Issue #40, opened 2014-07-30
add versioning to library


dominichamon:
from @mattyclarkson: [there is a guide](http://brianmilco.blogspot.co.uk/2012/11/cmake-automatically-use-git-tags-as.html) to pull the version number from the `git` tags. Then you can just tag the repo at [`v0.0.1`](http://semver.org/) and CMake would sort itself out.

PR #42, opened 2014-07-30
Implemented git versioning

mattyclarkson:
So this implements #40. It automatically versions the shared libraries from any annotated `git` tags:

```
git tag -a v1.0.0
```

It expects semver version tags such as `v1.0.0`. It would be trivial to support `1.0.0` but looking around it seems that most C/C++ projects follow `vX.X.X` rather that `X.X.X` like a lot of `Node.js` stuff.

This determines that the if the project has had a certain amount of commits since the last tag and also if the project is _dirty_ (has modified files), but does **nothing** with that information. In the future a more robust release could be implemented in the script.

This is pretty brittle and has little in the way of configuration. Ideally we should use `find_program` to work out where `git` is so that users can configure it. This implementation assumes that `git` will be available in `PATH`

Outputs the following on the command line:

```
-- git Version: v[MAJOR].[MINOR].[PATCH]-[COMMITS_SINCE_TAG]-[SHA1][-dirty]
-- Version: [MAJOR].[MINOR].[PATCH]
```

For example, dirty and ahead of a tag by one commit:

```
-- git Version: v1.4.0-1-gf492b08-dirty
-- Version: 1.4.0
```

On a tag and not dirty (this is what a release should be built from)

```
-- git Version: v0.0.1
-- Version: 0.0.1
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.so.?*
 *.dylib
 *.cmake
+!/cmake/*.cmake
 *~
 /test/benchmark_test
 /test/re_test
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required (VERSION 2.8)
 project (benchmark)
 
+# Make sure we can import out CMake functions
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+# We need threads in this project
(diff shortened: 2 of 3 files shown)


EricWF:
What if we don't use git for our version control system? Github supports SVN as well and CMake 2.8 only allows SVN to be used with ExternalProject. I don't see why we can't just define the version numbers in the CMakeLists.txt by hand.

I disagree with this approach entirely.

PR #64, opened 2014-11-03
MinGW  port

mattyclarkson:
So I've patched up the library to compile with `gcc` `4.9.2`:

```
cmake -g "MinGW Makefiles"
mingw32-make
```

I've just gone through and fixed up the compiler warnings/errors at the moment. I've tried to make the commits as fine grained as possible in case you would like to `git cherry-pick` them onto `master`.

The library doesn't seem to currently work due to hangs inside `WaitForMultipleObjects` in the Windows kernel. The backtrace looks like it is stalling inside the MinGW implementation of `pthread_cond_init` which is wrapped by `std::condition_variable`.  Thought I'd publish the branch in case anyone can provide any pointers?

I'll try to move forward with getting it compiling with VS2013 Express.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.a
 *.so
 *.so.?*
+*.exe
 *.dylib
 *.cmake
 !/cmake/*.cmake
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@
 *.cmake
 !/cmake/*.cmake
 *~
+*.pyc
+__pycache__
 
 # cmake files.
 /Testing
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.a
 *.so
 *.so.?*
+*.dll
 *.exe
 *.dylib
 *.cmake
--- a/mingw.py
+++ b/mingw.py
@@ -1,0 +1,320 @@
+#! /usr/bin/env python
+# encoding: utf-8
+
+import argparse
+import errno
+import logging
+import os
+import platform
+import re
+import sys
+import subprocess
+import tempfile
+
+try:
+    import winreg
+except ImportError:
+    import _winreg as winreg
+try:
+    import urllib.request as request
+except ImportError:
+    import urllib as request
+try:
+    import urllib.parse as parse
+except ImportError:
+    import urlparse as parse
+
+class EmptyLogger(object):
+    '''
+    Provides an implementation that performs no logging
+    '''
+    def debug(self, *k, **kw):
+        pass
+    def info(self, *k, **kw):
+        pass
+    def warn(self, *k, **kw):
+        pass
+    def error(self, *k, **kw):
+        pass
+    def critical(self, *k, **kw):
+        pass
+    def setLevel(self, *k, **kw):
+        pass
+
+urls = (
+    'http://downloads.sourceforge.net/project/mingw-w64/Toolchains%20'
+        'targetting%20Win32/Personal%20Builds/mingw-builds/installer/'
+        'repository.txt',
+    'http://downloads.sourceforge.net/project/mingwbuilds/host-windows/'
+        'repository.txt'
+)
+'''
+A list of mingw-build repositories
+'''
+
+def repository(urls = urls, log = EmptyLogger()):
+    '''
+    Downloads and parse mingw-build repository files and parses them
+    '''
+    log.info('getting mingw-builds repository')
+    versions = {}
+    re_sourceforge = re.compile(r'http://sourceforge.net/projects/([^/]+)/files')
+    re_sub = r'http://downloads.sourceforge.net/project/\1'
+    for url in urls:
+        log.debug(' - requesting: %s', url)
+        socket = request.urlopen(url)
+        repo = socket.read()
+        if not isinstance(repo, str):
+            repo = repo.decode();
+        socket.close()
+        for entry in repo.split('\n')[:-1]:
+            value = entry.split('|')
+            version = tuple([int(n) for n in value[0].strip().split('.')])
+            version = versions.setdefault(version, {})
+            arch = value[1].strip()
+            if arch == 'x32':
+                arch = 'i686'
+            elif arch == 'x64':
+                arch = 'x86_64'
+            arch = version.setdefault(arch, {})
+            threading = arch.setdefault(value[2].strip(), {})
+            exceptions = threading.setdefault(value[3].strip(), {})
+            revision = exceptions.setdefault(int(value[4].strip()[3:]),
+                re_sourceforge.sub(re_sub, value[5].strip()))
+    return versions
+
+def find_in_path(file, path=None):
+    '''
+    Attempts to find an executable in the path
+    '''
+    if platform.system() == 'Windows':
+        file += '.exe'
+    if path is None:
+        path = os.environ.get('PATH', '')
+    if type(path) is type(''):
+        path = path.split(os.pathsep)
+    return list(filter(os.path.exists,
+        map(lambda dir, file=file: os.path.join(dir, file), path)))
+
+def find_7zip(log = EmptyLogger()):
+    '''
+    Attempts to find 7zip for unpacking the mingw-build archives
+    '''
+    log.info('finding 7zip')
+    path = find_in_path('7z')
+    if not path:
+        key = winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, r'SOFTWARE\7-Zip')
+        path, _ = winreg.QueryValueEx(key, 'Path')
+        path = [os.path.join(path, '7z.exe')]
+    log.debug('found \'%s\'', path[0])
+    return path[0]
+
+find_7zip()
+
+def unpack(archive, location, log = EmptyLogger()):
+    '''
+    Unpacks a mingw-builds archive
+    '''
+    sevenzip = find_7zip(log)
+    log.info('unpacking %s', os.path.basename(archive))
+    cmd = [sevenzip, 'x', archive, '-o' + location, '-y']
+    log.debug(' - %r', cmd)
+    with open(os.devnull, 'w') as devnull:
+        subprocess.check_call(cmd, stdout = devnull)
+
+def download(url, location, log = EmptyLogger()):
+    '''
+    Downloads and unpacks a mingw-builds archive
+    '''
+    log.info('downloading MinGW')
+    log.debug(' - url: %s', url)
+    log.debug(' - location: %s', location)
+
+    re_content = re.compile(r'attachment;[ \t]*filename=(")?([^"]*)(")?[\r\n]*')
+
+    stream = request.urlopen(url)
+    try:
+        content = stream.getheader('Content-Disposition') or ''
+    except AttributeError:
+        content = stream.headers.getheader('Content-Disposition') or ''
+    matches = re_content.match(content)
+    if matches:
+        filename = matches.group(2)
+    else:
+        parsed = parse.urlparse(stream.geturl())
+        filename = os.path.basename(parsed.path)
+
+    try:
+        os.makedirs(location)
+    except OSError as e:
+        if e.errno == errno.EEXIST and os.path.isdir(location):
+            pass
+        else:
+            raise
+
+    archive = os.path.join(location, filename)
+    with open(archive, 'wb') as out:
+        while True:
+            buf = stream.read(1024)
+            if not buf:
+                break
+            out.write(buf)
+    unpack(archive, location, log = log)
+    os.remove(archive)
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -6,7 +6,9 @@
 !/cmake/*.cmake
 *~
 /test/benchmark_test
+/test/benchmark_test.exe
```
EricWF: Why do we need these? I don't think we should be building in-tree like this.
Either way the stuff already seems to be there for Unix so I guess this is fine.

CMakeLists.txt
```diff
@@ -20,15 +20,15 @@ link_directories(${binary_dir})
 
 # Enable the latest C++ standard possible
 include(CheckCXXCompilerFlag)
-check_cxx_compiler_flag(--std=c++14 HAVE_FLAG_CXX_14)
```
EricWF: Do we need these changes at all anymore?

src/benchmark.cc
```diff
@@ -22,8 +22,10 @@
 #include "sysinfo.h"
 #include "walltime.h"
 
+#ifndef COMPILER_MSVC
```
EricWF: Make sure to include the new file `"internal_macros.h"` to get `COMPILER_MSVC`.

src/colorprint.cc
```diff
@@ -16,6 +16,10 @@
 
 #include <stdarg.h>
 
+#ifdef OS_WINDOWS
```
EricWF: Make sure to include the new file `"internal_macros.h"` to get `OS_WINDOWS`.

src/cycleclock.h
```diff
@@ -37,7 +37,9 @@
 extern "C" uint64_t __rdtsc();
 #pragma intrinsic(__rdtsc)
 #endif
+#ifndef OS_WINDOWS
```
EricWF: Make sure to include the new file `"internal_macros.h"` to get `OS_WINDOWS`.

src/sleep.cc
```diff
@@ -17,10 +17,14 @@
 #include <time.h>
 #include <errno.h>
 
+#ifdef OS_WINDOWS
```
EricWF: Make sure to include the new file `"internal_macros.h"` to get `OS_WINDOWS`.

src/sysinfo.cc
```diff
@@ -19,11 +19,16 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifndef OS_WINDOWS
```
EricWF: Same as above.

src/sysinfo.cc
```diff
@@ -233,7 +237,7 @@ void InitializeSystemInfo() {
           SHGetValueA(HKEY_LOCAL_MACHINE,
                       "HARDWARE\\DESCRIPTION\\System\\CentralProcessor\\0",
                       "~MHz", NULL, &data, &data_size)))
-    cpuinfo_cycles_per_second = (int64)data * (int64)(1000 * 1000);  // was mhz
```
EricWF: Probably unneeded now?

src/sysinfo.cc
```diff
@@ -222,7 +227,6 @@ void InitializeSystemInfo() {
 // TODO: also figure out cpuinfo_num_cpus
 
 #elif defined OS_WINDOWS
-#pragma comment(lib, "shlwapi.lib")  // for SHGetValue()
```
EricWF: What did this do?

src/sysinfo.cc
```diff
@@ -233,7 +237,7 @@ void InitializeSystemInfo() {
           SHGetValueA(HKEY_LOCAL_MACHINE,
                       "HARDWARE\\DESCRIPTION\\System\\CentralProcessor\\0",
                       "~MHz", NULL, &data, &data_size)))
-    cpuinfo_cycles_per_second = (int64)data * (int64)(1000 * 1000);  // was mhz
+    cpuinfo_cycles_per_second = (int64_t)data * (int64_t)(1000 * 1000);  // was mhz
```
EricWF: Probably unneeded now?

src/walltime.cc
```diff
@@ -16,8 +16,10 @@
 
 #include <stdio.h>
 #include <string.h>
-#include <sys/time.h>
 #include <time.h>
+#ifndef COMPILER_MSVC
```
EricWF: Make sure to include the new file `"internal_macros.h"` to get `COMPILER_MSVC`.

src/sysinfo.cc
mattyclarkson: It automatically linked the `shwapi.lib` statically. I think I move it into the build files.

.gitignore
pphaneuf: Some people do build in-tree, though, even if it's not the best idea...

Me, I'd make a different comment: this should be a ```*.exe``` entry, the same way object files are handled.

mattyclarkson: I think we shouldn't build in tree. I tried to follow the current `.gitignore` style that was specifically adding the built executables. I think a catchall `*.exe` would be much more useful.

CMakeLists.txt
mattyclarkson: Looking at the `master` I don't think so. This branch has bit rotted for about 3 months. I'll pull it's socks up this week.

.gitignore
pphaneuf: Specifically adding the built executables makes sense for Unix platforms that don't have an executable extensions, but no need to duplicate this if there's an extension, just add a catchall.

mattyclarkson: Done. Thanks for the advice.

mattyclarkson:
That branch builds on Windows 8.1 and passes all tests with [`x86_64-4.9.2-release-posix-seh-rt_v4-rev2`](http://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/4.9.2/threads-posix/seh/) using the following commands:

```
cmake -G "MinGW Makefiles" -DCMAKE_BUILD_TYPE=DEBUG
mingw32-make
```

It would be cool to set up `appveyor` to build this on Windows continuously. Any thoughts on that?

mattyclarkson:
Just having a go at setting up appveyor on my personal repo. Will create a PR on top of this once it is merged.

dominichamon:
i don't know what appveyor is, but anything that builds windows continuously would be great.

src/sleep.cc
```diff
@@ -19,10 +19,14 @@
 
 #include "internal_macros.h"
 
+#ifdef OS_WINDOWS
+#include <Windows.h>
+#endif
+
 namespace benchmark {
 #ifdef OS_WINDOWS
 // Window's _sleep takes milliseconds argument.
```
dominichamon: nit: update the comment please

src/sysinfo.cc
```diff
@@ -13,13 +13,19 @@
 // limitations under the License.
 
 #include "sysinfo.h"
+#include "internal_macros.h"
 
+#ifndef OS_WINDOWS
```
dominichamon: consider flipping this so the positive is first:

#ifdef OS_WINDOWS
# include <Shlwapi.h>
# include <Windows.h>
#else
....
#endif  // OS_WINDOWS

src/sysinfo.cc
```diff
@@ -328,9 +335,26 @@ double MyCPUUsage() {
     }
   }
   return MyCPUUsageRUsage();
+#else
+  HANDLE proc = GetCurrentProcess();
```
dominichamon: it seems like changing MyCPUUsageRUsage to be this implementation for Windows is more accurate.

src/sysinfo.cc
```diff
@@ -340,8 +364,11 @@ double ChildrenCPUUsage() {
   } else {
     return 0.0;
   }
-}
+#else
+  // TODO(any.one): Not sure what this even means on Windows
```
dominichamon: don't put a name on here at all please .. just TODO:

mattyclarkson:
@dominichamon, it's a cloud based windows build service (it's free for open source projects). I'm most of the way with the integration - should be able to finish it up on Monday. Would be really good to get that in!

mattyclarkson:
The appveyor stuff has taken much longer than I thought it would - I needed to write a script that could download a specific version of MinGW depending on the version, threading model and exception model. The appveyor YAML file took a bit of cajoling but got there in the end. I'll push up the final changes today to the `nt` branch - I'm currently just working on the [`appveyor` branch](https://ci.appveyor.com/project/mattyclarkson/benchmark).

mattyclarkson:
Got the build working for [`mingw-4.9.2-posix-seh`](https://ci.appveyor.com/project/mattyclarkson/benchmark/build/40), Now just building up the matrix to cover some other common build situations.

mattyclarkson:
I edited the history to resolve your comments @dominichamon.

I've pushed up the `appveyor.yml` configuration file on top of the MinGW port commits. It currently is conservative with what it compiles for as the build times are lengthy due to downloading a specific version of MinGW. You can see a completed build [here](https://ci.appveyor.com/project/mattyclarkson/benchmark/build/44). I've told appveyor to cache the download of the MinGW compiler so that should decrease build times. I cannot set up the appveyor for `google/benchmark` as I don't have the necessary admin rights to give to appveyor to set up the web hooks. Would it be possible for one of the benchmark team to set this up?

dominichamon:
Nice work! I'll take a look at the patch.

I don't know if I have permission either - it may have to be a Google org admin. I'll poke around.

mingw.py
```diff
@@ -0,0 +1,320 @@
...
+    if args.quiet:
+        logger.setLevel(logging.WARN)
+    if args.verbose:
+        logger.setLevel(logging.DEBUG)
+
+    # Get MinGW
+    root_dir = root(location = args.location, arch = args.arch,
+        version = args.version, threading = args.threading,
+        exceptions = args.exceptions, revision = args.revision,
+        log = logger)
```
dominichamon: given you always pass this, you can kill the EmptyLogger above.

honestly, you can probably make the logger global and not even pass it through everywhere :)

mingw.py
```diff
@@ -0,0 +1,320 @@
...
+        default = os.path.join(tempfile.gettempdir(), 'mingw-builds'))
+    parser.add_argument('--arch', required = True, choices = ['i686', 'x86_64'],
+        help = 'the target MinGW architecture string')
+    parser.add_argument('--version', type = str2ver,
+        help = 'the version of GCC to download')
+    parser.add_argument('--threading', choices = ['posix', 'win32'],
+        help = 'the threading type of the compiler')
+    parser.add_argument('--exceptions', choices = ['sjlj', 'seh', 'dwarf'],
+        help = 'the method to throw exceptions')
+    parser.add_argument('--revision', type=int,
```
dominichamon: can you add some TODO (maybe) to limit the revision to versions we know we support?

mattyclarkson:
Cool - if the appveyor backend can get set up I can tune the configuration file to give speedy builds for MinGW.

mingw.py
mattyclarkson: The reasons it's there is so someone can use that python module as a library and turn logging on (or leave it off) when using the methods individually. For example I can then do:

```python
import mingw
versions = mingw.repository()
root = mingw.root(version = (4, 9, 2))
```

And that will give me all the mingw operations without logging anything or I could pass my own logger in.

That being said - completely happy to remove all that stuff if you don't want it in there! I tried to make the script as reusable as possible, if people want to use it in other projects for appveyor support or integrate it into python based tool chains.

mingw.py
mattyclarkson: Well we should always just use the latest revision, which is what the script does. The revision of the mingw-builds is just when niXman rebuilds the compiler because there was a bug in the build. It's actually really hard to limit the revision as each compiler in the matrix has a different number of revisions.

However, I may have misunderstood your question. The `appveyor.yml` limits the compilers we test against.

mingw.py
dominichamon: fair enough. i tend to err on the side of less code, minimal functionality, to avoid any subtle issues.

however, this isn't on the critical path so it seems reasonable.

dominichamon:
I also don't have access to the google-level appveyor. i'll ask around but in the meantime we should get this merged (assuming you're happy with it).

mattyclarkson:
:+1:

tamarlev:
Im currently trying to compile google benchmark for MingW and failing for c++11 features as mutex and threads, do you know how to resolve that?

dominichamon:
It's going to depend on your version of MinGW's gcc i imagine:
https://stackoverflow.com/questions/16136142/c11-functionality-with-mingw

Dominic Hamon | Google

PR #70, opened 2014-12-03
Adding extra libraries to integration, especially with googletest

dominichamon:


izaid:
This is useful -- is there a possibility to get it merged?

On a related issue, we've started using benchmark in our library and it works great. However, we already have Google Test included as a thirdparty library. Is it possible to make benchmark compile using that rather then download its own version of Google Test?

dominichamon:
rebased. can we get this in?

PR #76, opened 2015-02-19
One more 64-bit to 32-bit conversion

dominichamon:


dominichamon:
This should be the last one. Can someone take a look?

EricWF:
LGTM. Merged.

PR #78, opened 2015-02-21
Fix #77 build on freebsd 10

### LINKED ISSUE #77: Build error on freebsd 10
Hi, thanks for this awesome library! Unfortunatelly, I ran into a small issue with it. I tried to build it on freebsd 10.0 RELEASE and build failed. Compiler reported that header <sys/sysctl.h> has undefiened type u_int. The problem is that <sys/types.h> header should be included before it, cause this type is defined there. Can you please fix it, if you support freebsd?

pdavydov108:
I added a macro to detect freebsd in code (OS_FREEBSD), and put sys/types.h header before sys/sysctl.h. In fact it built itself fine even without freebsd macro. All tests passed for me, also checked it on linux.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -21,9 +21,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/resource.h>
+#include <sys/types.h>
 #include <sys/sysctl.h>
 #include <sys/time.h>
-#include <sys/types.h>
 #include <unistd.h>
 
 #include <iostream>
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -21,7 +21,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/resource.h>
-#include <sys/types.h>
+#include <sys/types.h> // this header must be included before 'sys/sysctl.h' to avoid compilation error on FreeBSD
 #include <sys/sysctl.h>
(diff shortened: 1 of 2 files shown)


src/sysinfo.cc
```diff
@@ -21,9 +21,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/resource.h>
+#include <sys/types.h>
```
dominichamon: huh. that's frustrating given the google style guide around include ordering. Maybe you can also add a comment to explain that this ordering is important (and why) so good samaritans don't put it back :)

EricWF:
This LGTM and I'll try and merge it by the end of the day. However CMake is currently segfaulting while configuring benchmark on FreeBSD 10.1 so I want to look into that first.

EricWF:
I fixed my CMake segfault issues. I'm happy to adopt your patch but I want to know if you have talked to the freeBSD folks about why sysctl.h doesn't include all the headers it uses? It seems to me this is a FreeBSD bug.

pdavydov108:
No, I haven't. It's written in man 3 sysctl that this headers should be included in this order, so I thought freebsd developers did it intentionally for some reason.

EricWF:
Bug filed under: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=198178

I'll merge the request. Thanks

PR #79, opened 2015-02-28
Update Copyright 2014 -> 2015

loverszhaokai:
Update Copyright 2014 -> 2015


EricWF:
Can we blanket update our copyright on files without actually making a change to that file?

loverszhaokai:
Maybe it's ok

dominichamon:
We can, but I'd rather not make these broad changes. Just updating when we touch the files is fine.

PR #80, opened 2015-03-03
Move to better version of benchmark library.

EricWF:
This is a HUGE patch that fixes problems in how the benchmark timings are performed.

The problem is the current version is that checking the time every iteration of the benchmark introduces a large amount of jitter and overhead into the timings.

This version fixes that problem by only checking the time at the beginning and end of every benchmark run.
It uses an iteration counter in state.KeepRunning() to track the number of times to run the benchmarks. 

I'll add more comments to explain this change later today.


CMakeLists.txt
```diff
@@ -1,89 +1,63 @@
 cmake_minimum_required (VERSION 2.8)
 project (benchmark)
 
+option(BENCHMARK_ENABLE_SHARED  "Enable building a shared library." OFF)
```
dominichamon: let's not lose this option please.

README.md
```diff
@@ -29,10 +29,10 @@ BENCHMARK(BM_StringCopy);
 
 // Augment the main() program to invoke benchmarks if specified
 // via the --benchmarks command line flag.  E.g.,
-//       my_unittest --benchmark_filter=all
-//       my_unittest --benchmark_filter=BM_StringCreation
-//       my_unittest --benchmark_filter=String
-//       my_unittest --benchmark_filter='Copy|Creation'
+//       my_unittest --benchmarks=all
```
dominichamon: why change this?

cmake/GetGitVersion.cmake
```diff
@@ -1,45 +0,0 @@
...
-#  get_git_version(<var>)
-#
-# - Example
-#
-# include(GetGitVersion)
-# get_git_version(GIT_VERSION)
-#
-# Requires CMake 2.6+
-
-if(__get_git_version)
```
dominichamon: this should come in too please

include/benchmark/benchmark.h
```diff
@@ -1,386 +1,45 @@
-// Support for registering benchmarks for functions.
-
-/* Example usage:
-// Define a function that executes the code to be measured a
-// specified number of times:
-static void BM_StringCreation(benchmark::State& state) {
-  while (state.KeepRunning())
-    std::string empty_string;
-}
+// Copyright 2014 Google Inc. All rights reserved.
```
dominichamon: 2015 :) (if we need a copyright notice at all, which i believe we no longer do.)

include/benchmark/benchmark.h
```diff
@@ -1,386 +1,45 @@
...
-  while (state.KeepRunning()) {
-    // Run the test as normal.
-  }
-  if (state.thread_index == 0) {
-    // Teardown code here.
-  }
-}
-BENCHMARK(BM_MultiThreaded)->Threads(4);
 */
+#ifndef TESTING_BASE_PUBLIC_BENCHMARK_H_
```
dominichamon: either follow the directory naming or use #pragma once.

include/benchmark/benchmark.h
```diff
@@ -1,386 +1,45 @@
...
-// Benchmark registration object.  The BENCHMARK() macro expands
-// into an internal::Benchmark* object.  Various methods can
-// be called on this object to change the properties of the benchmark.
-// Each method returns "this" so that multiple method calls can
-// chained into one expression.
 class Benchmark {
  public:
   // The Benchmark takes ownership of the Callback pointed to by f.
-  Benchmark(const char* name, BenchmarkFunction f);
+  Benchmark(const std::string& name, Function* f);
```
dominichamon: can we document ownership transfer by passing a std::unique_ptr<Function> instead?

CMakeLists.txt
EricWF: This option was added, not lost.

include/benchmark/benchmark.h
```diff
@@ -442,100 +101,122 @@ class Benchmark {
...
   // Following methods are not useful for clients
 
   // Used inside the benchmark implementation
   struct Instance;
 
-  // Measure the overhead of an empty benchmark to subtract later.
-  static void MeasureOverhead();
+  // Extract the list of benchmark instances that match the specified
+  // regular expression.
+  static void FindBenchmarks(const std::string& re,
```
dominichamon: as per the current version, i'm not sure that exposing this is the best API. take a look at how main looks with the current version to see what i think a good API is.

include/benchmark/benchmark.h
```diff
@@ -442,100 +101,122 @@ class Benchmark {
...
-  std::vector<Benchmark::Instance> CreateBenchmarkInstances(size_t rangeXindex,
-                                                            size_t rangeYindex);
-
   std::string name_;
-  BenchmarkFunction function_;
-  size_t registration_index_;
-  std::vector<int> rangeX_;
-  std::vector<int> rangeY_;
+  Function* function_;
+  int registration_index_;
```
dominichamon: make sure you test this with the 64 to 32 shortening warning.

include/benchmark/benchmark.h
```diff
@@ -442,100 +101,122 @@ class Benchmark {
...
-// Internal implementation details follow; please ignore
+// Benchmarks reporter interface + data containers.
 
-// Simple reporter that outputs benchmark data to the console. This is the
-// default reporter used by RunSpecifiedBenchmarks().
-class ConsoleReporter : public BenchmarkReporter {
- public:
-  virtual bool ReportContext(const Context& context) const;
-  virtual void ReportRuns(const std::vector<Run>& reports) const;
+struct BenchmarkContextData {
```
dominichamon: none of this needs to be in the public API, as you can see from the current version. please don't undo all this work :)

README.md
EricWF: It was unintentional. I meant to fix it.

cmake/GetGitVersion.cmake
EricWF: Noted.

include/benchmark/benchmark.h
EricWF: Noted.

include/benchmark/benchmark.h
EricWF: Noted.

include/benchmark/benchmark.h
EricWF: The pointer should be a static function pointer so we don't own it. Are there any use cases where somebody could (or would want to) pass a dynamically allocated function pointer?

dominichamon: I don't think so, given the requirements for registering a benchmark. but maybe then the comment is misleading :)

EricWF:
Address review comments.

include/benchmark/benchmark.h
EricWF: I have.

include/benchmark/benchmark.h
EricWF: It is in the public API to allow custom reporters. Should I remove it (I'm happy to).

CMakeLists.txt
```diff
@@ -1,89 +1,75 @@
...
 add_cxx_compiler_flag(-Werror)
 add_cxx_compiler_flag(-pedantic-errors)
+# TODO(ericwf): enable this for g++
 #add_cxx_compiler_flag(-Wzero-as-null-pointer-constant)
-
 # Release flags
 add_cxx_compiler_flag(-fno-strict-aliasing RELEASE)
 
-# Add a debug definition so we can make decisions in the compilation
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
```
dominichamon: we (in master) use ifdef DEBUG quite a bit. are you sure you want to remove this?

include/benchmark/benchmark.h
```diff
@@ -442,100 +101,122 @@ class Benchmark {
...
+void FindMatchingBenchmarkNames(const std::string& re,
+                                std::vector<std::string>* benchmark_names);
 
-#define BENCHMARK_TEMPLATE2(n, a, b)                         \
-  static ::benchmark::internal::Benchmark* BENCHMARK_CONCAT( \
-      __benchmark_, n, __LINE__) ATTRIBUTE_UNUSED =          \
-      (new ::benchmark::internal::Benchmark(#n "<" #a "," #b ">", n<a, b>))
+// Given a collection of reports, computes their mean and stddev.
+// REQUIRES: all runs in "reports" must be from the same benchmark.
+void ComputeStats(const std::vector<BenchmarkRunData>& reports,
```
dominichamon: does this need to be here?

CMakeLists.txt
EricWF: It's only used a couple of times in master and I don't see a reason why we can't use NDEBUG everywhere.

include/benchmark/benchmark.h
EricWF: These macros live in minimal_benchmark.h now.

EricWF:
Address more comments.

include/benchmark/minimal_benchmark.h
```diff
@@ -0,0 +1,487 @@
...
+  BENCHMARK_ALWAYS_INLINE
+  void SetLabel(const char* label) {
+    ::benchmark::SetLabel(label);
+  }
+
+  // Range arguments for this run. CHECKs if the argument has been set.
+  BENCHMARK_ALWAYS_INLINE
+  int range_x() const {
+    assert(has_range_x_);
+    ((void)has_range_x_); // Prevent unused warning.
```
dominichamon: __attribute__((unused)) ?

include/benchmark/minimal_benchmark.h
```diff
@@ -0,0 +1,487 @@
...
+#endif
+
+#define BENCHMARK_TEMPLATE2(n, a, b)                                      \
+  static ::benchmark::Benchmark*                                          \
+  BENCHMARK_CONCAT(__benchmark_, n, __LINE__) BENCHMARK_UNUSED =          \
+  (::benchmark::MinimalBenchmark(#n "<" #a "," #b ">",                    \
+                               &n<a, b>))
+
+// Helper macro to create a main routine in a test that runs the benchmarks
+#define BENCHMARK_MAIN()                             \
```
dominichamon: it might be best (as per the main.test branch) to have this as a separate library like googletest does it.

src/benchmark.cc
```diff
@@ -13,36 +13,66 @@
...
+DEFINE_string(benchmark_filter, "all",
               "A regular expression that specifies the set of benchmarks "
               "to execute.  If this flag is empty, no benchmarks are run.  "
               "If this flag is the string \"all\", all benchmarks linked "
               "into the process are run.");
 
-DEFINE_int32(benchmark_iterations, 0,
-             "Total number of iterations per benchmark. 0 means the benchmarks "
-             "are time-based.");
+DEFINE_int32(benchmark_min_iters, 1,
```
dominichamon: do you need the min and max?

src/benchmark.cc
```diff
@@ -53,700 +83,272 @@ DEFINE_double(benchmark_min_time, 0.5,
...
-DEFINE_bool(benchmark_memory_usage, false,
-            "Report memory usage for all benchmarks");
-
 DEFINE_int32(benchmark_repetitions, 1,
              "The number of runs of each benchmark. If greater than 1, the "
              "mean and standard deviation of the runs will be reported.");
 
-DEFINE_int32(v, 0, "The level of verbose logging to output");
-DEFINE_bool(color_print, true, "Enables colorized logging.");
+DEFINE_bool(benchmark_use_picoseconds, false,
```
dominichamon: should this be just wrapped in to how we report times? ie, report times with units.

src/benchmark.cc
```diff
@@ -53,700 +83,272 @@ DEFINE_double(benchmark_min_time, 0.5,
...
+  // Enter the barrier and wait until all other threads have also
+  // entered the barrier.  Returns iff this is the last thread to
+  // enter the barrier.
+  bool Barrier(MutexLock& ml) REQUIRES(lock_) {
+    CHECK_LT(entered_, num_threads_);
+    entered_++;
+    if (entered_ < num_threads_) {
+      // Wait for all threads to enter
+      int phase_number_cp = phase_number_;
+      auto cb = [this, phase_number_cp]() {
```
dominichamon: lambdas in C++03?

src/benchmark.cc
```diff
@@ -784,510 +383,569 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+  CHECK(running_benchmark);
+  timer_manager->StartTimer();
+}
+
+} // end namespace internal
+
+void PrintUsageAndExit() {
+  fprintf(stdout,
+          "benchmark"
+          " [--benchmarks=<regex>]\n"
```
dominichamon: benchmark_filter

src/sleep.h
```diff
@@ -1,7 +1,7 @@
 #ifndef BENCHMARK_SLEEP_H_
 #define BENCHMARK_SLEEP_H_
 
-#include <stdint.h>
+#include <cstdint>
```
dominichamon: maybe these header changes can be made on master in a separate patch.

test/benchmark_test.cc
```diff
@@ -63,13 +69,14 @@ static void BM_CalculatePiRange(benchmark::State& state) {
     pi = CalculatePi(state.range_x());
   std::stringstream ss;
   ss << pi;
-  state.SetLabel(ss.str());
+  state.SetLabel(ss.str().c_str());
```
dominichamon: there is a SetLabel that takes a std::string.

src/benchmark.cc
EricWF: Done.

src/sleep.h
EricWF: If the end goal is to merge this then does that make much sense?

include/benchmark/minimal_benchmark.h
EricWF: Getting that to work between clang and GCC and in debug mode vs release mode was a royal pain. This always works.

src/benchmark.cc
EricWF: A thought we didn't support building in C++03? This works with GCC 4.6 in c++0X.

src/benchmark.cc
EricWF: How do you mean? We do change the reported units if that is what your asking.

src/benchmark.cc
EricWF: I think they are useful constraints to have. Min changes the first amount of iterations attempted and max is useful for very fast benchmarks.

include/benchmark/minimal_benchmark.h
EricWF: I'll defer to you on this one. What do you think?

This macro works just fine for my uses.

src/benchmark.cc
EricWF: Yeah, I think they are useful to have. Why do you disagree?

include/benchmark/minimal_benchmark.h
EricWF: This works just fine for my cases. However I would like to build benchmark as a shared library and I don't think you can have a benchmark_main as a shared library.

We could provide both?

include/benchmark/benchmark.h
EricWF: I moved it into the internal namespace because that is how main does it. However I'm happy moving this out of the headers entirely.

include/benchmark/benchmark.h
EricWF: Didn't notice that. Removed.

dominichamon:
Can this be closed?

EricWF:
Closing this. The patch was broken up and committed in multiple parts.

PR #81, opened 2015-03-06
Step one towards merging timer changes.

EricWF:
This patch cleans up our use of generic macros and also merges changes in the
build system.

It adds the CMake options:
*-DBENCHMARK_ENABLE_TESTING: Turn on building googletest and the testsuite.
* -DBENCHMARK_ENABLE_SHARED: Build benchmark as a shared library.

This patch also removes a bunch of travis build matrix entries that are unneeded. Instead of having one entry to run the benchmark tests and another entry to run the regex tests there is now one that runs both.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,69 +1,10 @@
 cmake_minimum_required (VERSION 2.8)
 project (benchmark)
 
+option(BENCHMARK_ENABLE_SHARED  "Enable building a shared library." OFF)
+option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 # Make sure we can import out CMake functions
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
-
-# Import and build Google Test
-include(ExternalProject)
-set_directory_properties(properties EP_PREFIX "${CMAKE_BINARY_DIR}/third_party")
-ExternalProject_Add(googletest
-  URL "https://googletest.googlecode.com/files/gtest-1.7.0.zip"
-  URL_MD5 2d6ec8ccdf5c46b05ba54a9fd1d130d7
-  SOURCE_DIR "${CMAKE_BINARY_DIR}/third_party/gtest"
-  CMAKE_ARGS "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
-  INSTALL_COMMAND "")
-ExternalProject_Get_Property(googletest source_dir)
-include_directories(${source_dir}/include)
-ExternalProject_Get_Property(googletest binary_dir)
-link_directories(${binary_dir})
-
-# Try and enable C++11. Don't use C++14 because it doesn't work in some
-# configurations.
-include(CheckCXXCompilerFlag)
-check_cxx_compiler_flag(--std=c++11 HAVE_FLAG_CXX_11)
-check_cxx_compiler_flag(--std=c++0x HAVE_FLAG_CXX_0X)
-if (HAVE_FLAG_CXX_11)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --std=c++11")
-elseif (HAVE_FLAG_CXX_0X)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --std=c++0x")
-endif()
-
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -1,93 +1,62 @@
...
-  URL_MD5 2d6ec8ccdf5c46b05ba54a9fd1d130d7
-  SOURCE_DIR "${CMAKE_BINARY_DIR}/third_party/gtest"
-  CMAKE_ARGS "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
-  INSTALL_COMMAND "")
-ExternalProject_Get_Property(googletest source_dir)
-include_directories(${source_dir}/include)
-ExternalProject_Get_Property(googletest binary_dir)
-link_directories(${binary_dir})
+# Read the git tags to determine the project version
+include(GetGitVersion)
```
dominichamon: any reason why you moved this up?

src/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
-# Define the source files
-set(SOURCE_FILES "benchmark.cc" "colorprint.cc" "commandlineflags.cc" "sleep.cc" "sysinfo.cc" "walltime.cc")
+# Allow the source files to find headers in src/
+include_directories(${PROJECT_SOURCE_DIR}/src)
 
+# Define the source files
+set(SOURCE_FILES "benchmark.cc" "commandlineflags.cc"
```
dominichamon: nit - command should come after color .. alphabet and all :)

test/re_test.cc
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #include <gtest/gtest.h>
-#include "re.h"
+#include "../src/re.h"
```
dominichamon: this is unfortunate - should we add the src/ folder to the test/CMakeLists.txt include_directories instead?

CMakeLists.txt
EricWF: I felt like it didn't make sense that it lived between the `add_cxx_compiler_flag(...)` and `cxx_feature_check(...)` blocks. Those seem to be logically related to each other and setting the SO version there seemed out of place.

src/CMakeLists.txt
EricWF: Noted.

CMakeLists.txt
dominichamon: SGTM

test/CMakeLists.txt
```diff
@@ -1,18 +1,45 @@
...
 
+macro(compile_benchmark_test name)
+  add_executable(${name} "${name}.cc")
+  target_link_libraries(${name} benchmark gtest ${CMAKE_THREAD_LIBS_INIT})
+  add_dependencies(${name} googletest)
+endmacro(compile_benchmark_test)
+
+macro (add_benchmark_test name)
+  compile_benchmark_test(${name})
+  add_test(${name} ${name} --benchmarks=all)
```
dominichamon: s/benchmarks/benchmark_filter/

test/re_test.cc
EricWF: I made this change as a precaution that out tests didn't use any non-public headers accidentally. Instead forcing their inclusion to be somewhat more explicit.

I'll leave it up to you which you would rather have.

test/CMakeLists.txt
EricWF: Noted.

Yeah, it is currently unused. I'll remove it. I *had* a use for it but I didn't merge those tests yet.

test/re_test.cc
dominichamon: oh i get that. the re stuff has been a bit of a weird extra thing that we test -- i wonder if there's a better way of managing it.

Maybe we should add tests to the benchmark suite to explicitly test the 'find matching benchmarks' functionality instead, which is part of the public API.

EricWF: I agree we should add tests for that, but I would like to get through more of this patch before I add even more new code.

dominichamon: agreed - please add a TODO (or create an issue - or both!) to remove the re_tests and add the regex matching as a benchmark test against the public API.

EricWF: I'll open an issues. Also benchmark_test.cc already sort of tests the regex matching by checking how many benchmarks were run given a specific regex.

dominichamon:
LGTM modulo failing build

PR #84, opened 2015-03-09
Step two towards merging timer changes.

EricWF:
This part of the patch does three things:

1. Add logging class and macro to cleanup debug logging. Actually implement -v option.
2. Re-implement CHECK macros to provide better output and to allow extra error messages.
3. Re-implement static initialization of WallTime. Use a static variable in a function to initialize walltime instead of the atomic flags.

### DIFF
--- a/src/check.h
+++ b/src/check.h
@@ -1,12 +1,52 @@
 #ifndef CHECK_H_
 #define CHECK_H_
 
-#include <cassert>
+#include <cstdlib>
+#include <ostream>
 
-#define CHECK(b)             \
-  do {                       \
-    if (!(b)) assert(false); \
-  } while (0)
+#include "internal_macros.h"
+#include "log.h"
+
+namespace benchmark {
+namespace internal {
+
+// CheckHandler is the class constructed by failing CHECK macros. CheckHandler
+// will log information about the failures and abort when it is destructed.
+class CheckHandler {
+public:
+  CheckHandler(const char* check, const char* file, const char* func, int line)
+    : log_(GetErrorLogInstance())
+  {
+    log_ << file << ":" << line << ": " << func << ": Check `"
+          << check << "' failed. ";
+  }
+
+  std::ostream& GetLog() {
+    return log_;
+  }
+
+  BENCHMARK_NORETURN ~CheckHandler() {
+      log_ << std::endl;
+      std::abort();
+  }
+
+private:
+  std::ostream& log_;
+};
+
+} // end namespace internal
+} // end namespace benchmark
+
+// The CHECK macro returns a std::ostream object that can have extra information
+// written to it.
+#ifndef NDEBUG
+# define CHECK(b)  (b ? ::benchmark::internal::GetNullLogInstance()        \
+                      : ::benchmark::internal::CheckHandler(               \
+                          #b, __FILE__, __func__, __LINE__).GetLog())
+#else
+# define CHECK(b) ::benchmark::internal::GetNullLogInstance()
+#endif
+
 #define CHECK_EQ(a, b) CHECK((a) == (b))
 #define CHECK_NE(a, b) CHECK((a) != (b))
 #define CHECK_GE(a, b) CHECK((a) >= (b))
--- a/src/log.h
+++ b/src/log.h
@@ -1,0 +1,28 @@
+#ifndef BENCHMARK_LOG_H_
+#define BENCHMARK_LOG_H_
+
+#include <ostream>
+
+namespace benchmark {
+namespace internal {
+
+int GetLogLevel();
+void SetLogLevel(int level);
+
+std::ostream& GetNullLogInstance();
+std::ostream& GetErrorLogInstance();
+
+inline std::ostream& GetLogInstanceForLevel(int level) {
+  if (level >= GetLogLevel()) {
+    return GetErrorLogInstance();
+  }
+  return GetNullLogInstance();
+}
+
+} // end namespace internal
+} // end namespace benchmark
+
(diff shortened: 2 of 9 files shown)


src/check.h
```diff
@@ -1,12 +1,52 @@
...
+#include "log.h"
+
+namespace benchmark {
+namespace internal {
+
+// CheckHandler is the class constructed by failing CHECK macros. CheckHandler
+// will log information about the failures and abort when it is destructed.
+class CheckHandler {
+public:
+  CheckHandler(const char* check, const char* file, const char* func, int line)
```
dominichamon: strings instead?

src/check.h
```diff
@@ -1,12 +1,52 @@
...
+  std::ostream& log_;
+};
+
+} // end namespace internal
+} // end namespace benchmark
+
+// The CHECK macro returns a std::ostream object that can have extra information
+// written to it.
+#ifndef NDEBUG
+# define CHECK(b)  (b ? ::benchmark::internal::GetNullLogInstance()        \
```
dominichamon: should we just depend on glog? maybe create an issue to optionally use glog if it is available.

src/log.h
```diff
@@ -0,0 +1,28 @@
+#ifndef BENCHMARK_LOG_H_
+#define BENCHMARK_LOG_H_
+
+#include <ostream>
+
+namespace benchmark {
+namespace internal {
+
+int GetLogLevel();
```
dominichamon: use an enum please

src/walltime.cc
```diff
@@ -14,14 +14,14 @@
 
 #include "walltime.h"
 
-#include <stdio.h>
-#include <stdint.h>
-#include <string.h>
 #include <sys/time.h>
 #include <time.h>
-
+#include <cstdio>
```
dominichamon: newline between the C and C++ includes?

src/check.h
EricWF: What is the rational for switching to std::string if the constructor is only called from the macro?

src/walltime.cc
```diff
@@ -30,44 +30,9 @@
...
 
-WallTime Slow() {
-  struct timeval tv;
-  gettimeofday(&tv, NULL);
-  return tv.tv_sec + tv.tv_usec * 1e-6;
-}
-
-bool SplitTimezone(WallTime value, bool local, struct tm* t,
-                   double* subsecond) {
+inline bool SplitTimezone(WallTime value, bool local, struct tm* t,
```
dominichamon: no point in 'inline' here.

src/walltime.cc
```diff
@@ -81,58 +46,127 @@ bool SplitTimezone(WallTime value, bool local, struct tm* t,
...
-  // within kMaxErrorInterval of one another.
 
-  SetDrift(0.0);
-  last_adjust_time = static_cast<uint32_t>(uint64_t(base_cycletime) >> 32);
-  std::atomic_store(&initialized, true);
-}
+} // end anonymous namespace
+
+
+namespace internal {
```
dominichamon: maybe kill the internal namespace and continue the anonymous one instead?

src/walltime.cc
```diff
@@ -81,58 +46,127 @@ bool SplitTimezone(WallTime value, bool local, struct tm* t,
...
+
+class WallTimeImp
+{
+public:
+  WallTime Now();
+
+  static WallTimeImp & GetWallTimeImp() {
```
dominichamon: & with type please

src/walltime.cc
```diff
@@ -81,58 +46,127 @@ bool SplitTimezone(WallTime value, bool local, struct tm* t,
...
+  }
+
+  float GetDrift() const {
+    float f;
+    int32_t w = std::atomic_load(&drift_adjust_);
+    memcpy(&f, &w, sizeof(f));
+    return f;
+  }
+
+  WallTime Slow() {
```
dominichamon: const method?

src/check.h
EricWF: I don't think so. Then we introduce another dependency which some users don't want. If we do it optionally there is another configuration to support and provides little extra value.

But I'll defer to your judgement on this if you really want it.

src/walltime.cc
EricWF: Sure.

src/walltime.cc
EricWF: Noted.

src/log.h
EricWF: I'll make the change. Any input for what to call the levels?

FYI the reason I didn't use an enum is because the log level is set via the `-v=<int>` flag. So I figured there was no point obfuscating that input behind an enum.

src/walltime.cc
EricWF: Ok.

src/walltime.cc
EricWF: Ok.

src/walltime.cc
EricWF: Ok.

src/log.h
EricWF: Actually are you sure you want an enum here? Glog doesn't use an enum for it's verbose log either.

src/check.h
dominichamon: good point. it's a knee-jerk reaction to seeing const char*.

src/check.h
dominichamon: ack.

src/log.h
dominichamon: i did some work to get glog to use enums but it never landed. it's safer in that it's easy to check for expected levels. i suppose we don't really need that here as we're only emulating VLOG.

PR #85, opened 2015-03-09
Step three towards moving timer changes.

EricWF:
This patch is the third and final step before merging the timer changes.

This patch contains 3 main changes along with some small misc changes:
1. Create string_util.h and string_util.cc with string utility from benchmark.cc
2. Cleanup includes.
3. Update copyright.

### DIFF
--- a/src/string_util.h
+++ b/src/string_util.h
@@ -1,0 +1,48 @@
+#ifndef BENCHMARK_STRING_UTIL_H
+#define BENCHMARK_STRING_UTIL_H
+
+#include <string>
+#include <sstream>
+#include <utility>
+
+namespace benchmark {
+
+void ToExponentAndMantissa(double val, double thresh, int precision,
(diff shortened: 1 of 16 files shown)


src/string_util.h
```diff
@@ -0,0 +1,48 @@
+#ifndef BENCHMARK_STRING_UTIL_H_
+#define BENCHMARK_STRING_UTIL_H_
+
+#include <string>
+#include <sstream>
+#include <utility>
+
+namespace benchmark {
+
+void ToExponentAndMantissa(double val, double thresh, int precision,
```
dominichamon: which of these actually need to be in the header? i'm guessing not all.

would you mind checking before you merge?

EricWF: Done.

dominichamon:
LGTM modulo tests passing.

EricWF:
Tests are failing again due to an existing bug. pushing.

PR #86, opened 2015-03-09
Change timing internals

EricWF:
Finally upgrade the timing internals!

The problem is the current version is that checking the time every iteration of the benchmark introduces a large amount of jitter and overhead into the timings.

This version fixes that problem by only checking the time at the beginning and end of every benchmark run.
It uses an iteration counter in state.KeepRunning() to track the number of times to run the benchmarks.


src/CMakeLists.txt
```diff
@@ -2,9 +2,9 @@
 include_directories(${PROJECT_SOURCE_DIR}/src)
 
 # Define the source files
-set(SOURCE_FILES "benchmark.cc" "colorprint.cc" "commandlineflags.cc"
-                 "log.cc" "sleep.cc" "string_util.cc" "sysinfo.cc"
-                 "walltime.cc")
+set(SOURCE_FILES "benchmark.cc" "minimal_benchmark.cc" "commandlineflags.cc"
+                 "colorprint.cc" "log.cc" "sleep.cc" "string_util.cc"
```
dominichamon: colorprint before commandlineflags
minimal_benchmark after log

src/benchmark.cc
```diff
@@ -13,40 +13,42 @@
...
+#include "colorprint.h"
+#include "log.h"
+#include "mutex.h"
+#include "re.h"
+#include "stat.h"
+#include "string_util.h"
+#include "sysinfo.h"
+#include "walltime.h"
+
+DEFINE_string(benchmark_filter, "all",
```
dominichamon: 'all' is not a valid regex

src/benchmark.cc
```diff
@@ -13,40 +13,42 @@
...
               "A regular expression that specifies the set of benchmarks "
               "to execute.  If this flag is empty, no benchmarks are run.  "
               "If this flag is the string \"all\", all benchmarks linked "
               "into the process are run.");
 
-DEFINE_int32(benchmark_iterations, 0,
-             "Total number of iterations per benchmark. 0 means the benchmarks "
-             "are time-based.");
+DEFINE_int32(benchmark_min_iters, 1,
```
dominichamon: why not just one number for iterations?

src/benchmark.cc
```diff
@@ -57,596 +59,277 @@ DEFINE_double(benchmark_min_time, 0.5,
...
-        Stat1_d(it->real_accumulated_time / it->iterations, it->iterations);
-    cpu_accumulated_time_stat +=
-        Stat1_d(it->cpu_accumulated_time / it->iterations, it->iterations);
-    items_per_second_stat += Stat1_d(it->items_per_second, it->iterations);
-    bytes_per_second_stat += Stat1_d(it->bytes_per_second, it->iterations);
-    iterations_stat += Stat1_d(it->iterations, it->iterations);
-    max_heapbytes_used_stat +=
-        Stat1MinMax_d(it->max_heapbytes_used, it->iterations);
-  }
+namespace internal {
```
dominichamon: why not use the existing anonymous namespace?

src/benchmark.cc
```diff
@@ -57,596 +59,277 @@ DEFINE_double(benchmark_min_time, 0.5,
...
-  Benchmark* bm;
-  int threads;  // Number of concurrent threads to use
-
-  bool rangeXset;
-  int rangeX;
-  bool rangeYset;
-  int rangeY;
-
-  bool multithreaded() const { return !bm->thread_counts_.empty(); }
+  std::string   name;
```
dominichamon: no constructor?

src/benchmark.cc
```diff
@@ -57,596 +59,277 @@ DEFINE_double(benchmark_min_time, 0.5,
...
-
-namespace internal {
-
-Benchmark::Benchmark(const char* name, BenchmarkFunction f)
-    : name_(name), function_(f) {
-  registration_index_ = BenchmarkFamilies::GetInstance()->AddBenchmark(this);
+Benchmark::Benchmark(const char* name,
+                     Function* f) EXCLUDES(GetBenchmarkLock())
+                    : name_(name), function_(f), arg_count_(-1) {
+  MutexLock l(GetBenchmarkLock());
```
dominichamon: the BenchmarkFamilies class above is a helpful abstraction. I'm not sure why you aren't using it.

src/benchmark.cc
```diff
@@ -57,596 +59,277 @@ DEFINE_double(benchmark_min_time, 0.5,
...
   AddRange(&arglist2, lo2, hi2, kRangeMultiplier);
 
-  std::lock_guard<std::mutex> l(mutex_);
-  rangeX_.resize(arglist1.size());
-  std::copy(arglist1.begin(), arglist1.end(), rangeX_.begin());
-  rangeY_.resize(arglist2.size());
-  std::copy(arglist2.begin(), arglist2.end(), rangeY_.begin());
+  for (int i : arglist1) {
+    for (int j : arglist2) {
+      args_.emplace_back(i, j);
```
dominichamon: does this build on our target platforms?

src/benchmark.cc
```diff
@@ -684,505 +364,577 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-    report.report_label = state.label;
-    report.bytes_per_second = state.stats.bytes_processed / seconds;
-    report.items_per_second = state.stats.items_processed / seconds;
-    report.max_heapbytes_used = MeasurePeakHeapMemory(b);
-  }
 
-  br->ReportRuns(state.runs);
-}
+      Notification done;
+      timer_manager = new TimerManager(b.threads, &done);
```
dominichamon: unique_ptr?

src/benchmark.cc
```diff
@@ -684,505 +364,577 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
 
-void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter /*= nullptr*/) {
+void RunSpecifiedBenchmarks(BenchmarkReporter* reporter) {
   std::string spec = FLAGS_benchmark_filter;
-  if (spec.empty() || spec == "all")
-    spec = ".";  // Regexp that matches all benchmarks
-  internal::ConsoleReporter default_reporter;
-  internal::RunMatchingBenchmarks(
-      spec, reporter == nullptr ? &default_reporter : reporter);
+  if (spec.empty()) {
```
dominichamon: change in behaviour. why do you want an empty spec to do nothing?

src/benchmark.cc
```diff
@@ -684,505 +364,577 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+    // Nothing to do
+  } else {
+    if (spec == "all") {
+      spec = ".";         // Regexp that matches all benchmarks
+    }
+    internal::ConsoleReporter default_reporter;
+    internal::RunMatchingBenchmarks(spec, reporter ? reporter : &default_reporter);
+    std::exit(0);
```
dominichamon: why an explicit exit here? just let the function exit so people calling this aren't surprised.

src/benchmark.cc
```diff
@@ -684,505 +364,577 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+      --i;
+    } else if (IsFlag(argv[i], "help")) {
+      PrintUsageAndExit();
+    }
+  }
 }
 
 void Initialize(int* argc, const char** argv) {
-  internal::ParseCommandLineFlags(argc, argv);
+  ParseCommandLineFlags(argc, argv);
```
dominichamon: put this back under internal please

src/benchmark.cc
```diff
@@ -684,505 +364,577 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
   internal::SetLogLevel(FLAGS_v);
-  // Ensure walltime is initialized by a single thread by forcing the
-  // initialization.
+  // TODO remove this. It prints some output the first time it is called.
+  // We don't want to have this ouput printed during benchmarking.
+  MyCPUUsage();
```
dominichamon: why is it necessary? it wasn't before <---

src/CMakeLists.txt
EricWF: Done.

src/benchmark.cc
EricWF: Sure, but the documentation says "all" == "." in both the new and old versions.

src/benchmark.cc
EricWF: Different semantics. `benchmark_iterations` runs the test with exactly N iterations. `benchmark_min_iters` and `benchmark_max_iters` allow the benchmark to be run with any number of iterations in that range.

src/benchmark.cc
EricWF: I think this has to do with the thread annotation attributes not working with variables in anonymous namespaces.

include/benchmark/benchmark.h
```diff
@@ -1,386 +1,48 @@
-// Support for registering benchmarks for functions.
-
-/* Example usage:
-// Define a function that executes the code to be measured a
-// specified number of times:
-static void BM_StringCreation(benchmark::State& state) {
-  while (state.KeepRunning())
-    std::string empty_string;
-}
+// Copyright 2015 Google Inc. All rights reserved.
```
dominichamon: any chance you can move this change to another patch? it's making it harder to review the deltas between the benchmark API.

src/benchmark.cc
dominichamon: that's something we should change... it's a hangover from the first version that had 'all'.

src/benchmark.cc
EricWF: I could add one, but the Instance's are only constructed in one spot and it would be more unclear to use a constructor than it would be to assign values to each field.

src/benchmark.cc
dominichamon: i think the single number is a cleaner API that captures the same benefit. How the number of iterations and time interact is 'tricky' and i think deterministic behaviour is better.

src/benchmark.cc
EricWF: Because adding it would have resulted in even more churn. It seems like a good abstraction though. Do you want it in this patch?

src/benchmark.cc
dominichamon: in this case you're removing one. I'd rather the callsite was cleaner and didn't have to set things (if we refactor things later).

src/benchmark.cc
EricWF: It builds with GCC 4.6 in -std=c++0x mode. So yes?

src/benchmark.cc
dominichamon: i think there's more churn here because you're removing it.

src/benchmark.cc
EricWF: Reverted that change.

src/benchmark.cc
EricWF: Sure.

src/benchmark.cc
EricWF: My starting point was the internal version of this file, not the current one. So on my end it was more churn to add it.

src/benchmark.cc

src/benchmark.cc
EricWF: Because the first call to MyCPUUsage happened in MeasureOverhead().

include/benchmark/benchmark.h
```diff
@@ -180,17 +185,71 @@ void UseRealTime();
...
+      has_range_x_(has_x), range_x_(x),
+      has_range_y_(has_y), range_y_(y),
+      bytes_processed_(0), items_processed_(0),
+      thread_index(thread_i)
+  {}
+
   // Returns true iff the benchmark should continue through another iteration.
-  bool KeepRunning();
+  BENCHMARK_ALWAYS_INLINE
+  bool KeepRunning() {
```
dominichamon: can you please try this same change but with the implementation kept private? i'll be amazed if the overhead of the function call matches the inlined version, and the benefit to moving it is a simpler API with fewer internals exposed.

include/benchmark/benchmark.h
```diff
@@ -221,113 +296,89 @@ class State {
...
  public:
   struct Context {
     int num_cpus;
     double mhz_per_cpu;
-    // std::string cpu_info;
     bool cpu_scaling_enabled;
 
     // The number of chars in the longest benchmark name.
-    size_t name_field_width;
+    int name_field_width;
```
dominichamon: this was changed to fix a 64-32 shortening issue. be careful here :)

include/benchmark/benchmark.h
```diff
@@ -354,33 +405,16 @@ class BenchmarkReporter {
   // benchmark, thus have the same name.
   virtual void ReportRuns(const std::vector<Run>& report) const = 0;
 
-  virtual ~BenchmarkReporter() {}
+  virtual ~BenchmarkReporter();
```
dominichamon: does this do something now? if not, an empty implementation in the header should be fine.

include/benchmark/benchmark.h
```diff
@@ -354,33 +405,16 @@ class BenchmarkReporter {
...
-void RunMatchingBenchmarks(const std::string& spec,
-                           const BenchmarkReporter* reporter);
-
-// Extract the list of benchmark names that match the specified regular
-// expression.
-void FindMatchingBenchmarkNames(const std::string& re,
-                                std::vector<std::string>* benchmark_names);
+typedef void(Function)(State&);
 
-// ------------------------------------------------------
```
dominichamon: you don't think this comment is useful?

include/benchmark/benchmark.h
```diff
@@ -448,40 +482,25 @@ class Benchmark {
   // Used inside the benchmark implementation
   struct Instance;
 
-  // Measure the overhead of an empty benchmark to subtract later.
-  static void MeasureOverhead();
-
- private:
   friend class BenchmarkFamilies;
```
dominichamon: as per the previous, this friend declaration should be private

include/benchmark/benchmark.h
```diff
@@ -448,40 +482,25 @@ class Benchmark {
...
+ private:
   std::string name_;
-  BenchmarkFunction function_;
-  size_t registration_index_;
-  std::vector<int> rangeX_;
-  std::vector<int> rangeY_;
+  Function* function_;
+  std::size_t registration_index_;
+  int arg_count_;
+  std::vector< std::pair<int, int> > args_;  // Args for all benchmark runs
```
dominichamon: g++-4.6 should allow for >> so avoid the spaces.

include/benchmark/benchmark.h
```diff
@@ -491,16 +510,16 @@ class ConsoleReporter : public BenchmarkReporter {
  public:
   virtual bool ReportContext(const Context& context) const;
   virtual void ReportRuns(const std::vector<Run>& reports) const;
-
  private:
-  std::string PrintMemoryUsage(double bytes) const;
   virtual void PrintRunData(const Run& report) const;
-  mutable size_t name_field_width_;
+  // TODO(ericwf): Find a better way to share this information.
+  mutable int name_field_width_;
```
dominichamon: int -> size_t for 64-32 shortening.

src/benchmark.cc
```diff
@@ -684,505 +480,497 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+                                    ? benchmarks[i].name.size() + 7
+                                    : benchmarks[i].name.size());
   }
 
   // Print header here
   BenchmarkReporter::Context context;
-  context.num_cpus = NumCPUs();
-  context.mhz_per_cpu = CyclesPerSecond() / 1000000.0f;
-  //  context.cpu_info = base::CompactCPUIDInfoString();
+  context.num_cpus = benchmark::NumCPUs();
```
dominichamon: we should be in the benchmark namespace here... no?

src/benchmark.cc
```diff
@@ -684,505 +480,497 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter) {
+  std::string spec = FLAGS_benchmark_filter;
+  if (spec.empty()) {
+    // Nothing to do
+  } else {
+    if (spec.empty() || spec == "all")
+        spec = ".";  // Regexp that matches all benchmarks
+    internal::ConsoleReporter default_reporter;
+    internal::RunMatchingBenchmarks(spec, reporter ? reporter : &default_reporter);
+    std::exit(0);
```
dominichamon: calling exit from within a library is a bad idea.

src/benchmark.cc
```diff
@@ -684,505 +480,497 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+
+void ParseCommandLineFlags(int* argc, const char** argv) {
+  using namespace benchmark;
+  for (int i = 1; i < *argc; ++i) {
+    if (
+        ParseStringFlag(argv[i], "benchmark_filter",
+                        &FLAGS_benchmark_filter) ||
+        ParseInt32Flag(argv[i], "benchmark_iterations",
+                       &FLAGS_benchmark_iterations) ||
+        ParseInt32Flag(argv[i], "benchmark_min_iters",
```
dominichamon: did you mean to leave iterations and min_iters and max_iters? it's not clear how these interact.

src/benchmark.cc
```diff
@@ -684,505 +480,497 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+                       &FLAGS_benchmark_min_iters) ||
+        ParseInt32Flag(argv[i], "benchmark_max_iters",
+                       &FLAGS_benchmark_max_iters) ||
+        ParseDoubleFlag(argv[i], "benchmark_min_time",
+                        &FLAGS_benchmark_min_time) ||
+        ParseInt32Flag(argv[i], "benchmark_repetitions",
+                       &FLAGS_benchmark_repetitions) ||
+        ParseBoolFlag(argv[i], "benchmark_use_picoseconds",
+                       &FLAGS_benchmark_use_picoseconds) ||
```
dominichamon: do we need to check if this is even possible on the platform before allowing it?

src/benchmark.cc
```diff
@@ -684,505 +480,497 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-
-  std::lock_guard<std::mutex> l(that->shared_->mu);
 
-  that->shared_->exited++;
-  if (that->thread_index > 0 &&
-      that->shared_->exited == that->shared_->threads - 1) {
-    // All threads but thread 0 have exited the user-provided run function.
-    // Thread 0 can now wake up and exit.
-    that->shared_->cond.notify_one();
+  double const multiplier = FLAGS_benchmark_use_picoseconds ? 1e12 : 1e9;
```
dominichamon: can this be determined automatically? ie, if the time is < than 1e-9 then switch to pico.

src/benchmark.cc
```diff
@@ -684,505 +480,497 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-  std::vector<internal::Benchmark::Instance> benchmarks;
-  BenchmarkFamilies::GetInstance()->FindBenchmarks(spec, &benchmarks);
-  std::transform(benchmarks.begin(), benchmarks.end(), benchmark_names->begin(),
-                 [](const internal::Benchmark::Instance& b) { return b.name; });
+void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter) {
+  std::string spec = FLAGS_benchmark_filter;
+  if (spec.empty()) {
+    // Nothing to do
+  } else {
+    if (spec.empty() || spec == "all")
```
dominichamon: spec can't be empty here

include/benchmark/benchmark.h
EricWF: Won't this cause the v-table to be emitted in every translation unit?

include/benchmark/benchmark.h
EricWF: I think the headers should be usable in C++03. If this were in the library source I would make the change.

include/benchmark/benchmark.h
EricWF: Woopies. I do think it is useful but it somehow got lost.

include/benchmark/benchmark.h
EricWF: ack. Any particular reason besides style?

include/benchmark/benchmark.h
EricWF: ack.

include/benchmark/benchmark.h
EricWF: ack.

src/benchmark.cc
EricWF: Yep.

src/benchmark.cc
EricWF: I'm not sure I understand what you mean. Could you expand on this?

src/benchmark.cc
EricWF: It is indeed. Sorry about that.

src/benchmark.cc
EricWF: Reverting to old behavior where an empty spec matched all benchmarks.

src/benchmark.cc
EricWF: I wanted to put iterations back so that I didn't remove functionality. I think I will rename benchmark_min_iters and benchmark_max_iters so that their purpose is more clear.

src/benchmark.cc
EricWF: AFAIK this is just a formatting thing. The same unit is used internally but is converted to either nano seconds or pico seconds before outputting.

include/benchmark/benchmark.h
EricWF: I've done the legwork. Here are some of the things I noticed.

1. There was no difference when building in debug mode.
2. There was a difference when building in release. I saw a 5% change on some benchmarks in basic_test.

The one thing I like about moving `KeepRunning()` out of the header is that when the user compiles their tests with two different optimization levels the amount of overhead `KeepRunning()` introduces doesn't change.

However I also 

I think we can still expose fewer internals by moving some of the `State` methods out of the header. I'll make that change and not move `KeepRunning()` for now. I'll continue to look into this though.

include/benchmark/benchmark.h
```diff
@@ -221,94 +286,69 @@ class State {
...
 
-  // Each State object goes through a sequence of measurement intervals. By
-  // default each interval is approx. 100ms in length. The following stats are
-  // kept for each interval.
-  int64_t iterations_;
-  double start_cpu_;
-  double start_time_;
-  int64_t stop_time_micros_;
+  BENCHMARK_ALWAYS_INLINE
+  size_t iterations() const { return total_iterations_; }
```
dominichamon: should these just be public const member variables like thread_index?

include/benchmark/benchmark.h
```diff
@@ -221,94 +286,69 @@ class State {
...
-  friend class internal::Benchmark;
+private:
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(State);
 };
 
 // Interface for custom benchmark result printers.
 // By default, benchmark reports are printed to stdout. However an application
 // can control the destination of the reports by calling
-// RunSpecifiedBenchmarks and passing it a custom reporter object.
+// RunMatchingBenchmarks and passing it a custom reporter object.
```
dominichamon: why the rename of the method?

include/benchmark/benchmark.h
dominichamon: i don't see that it's an issue, but i'm not that particular about it.

src/benchmark.cc
```diff
@@ -45,8 +45,7 @@ DEFINE_string(benchmark_filter, ".",
               "into the process are run.");
 
 DEFINE_int32(benchmark_iterations, 0,
-             "Total number of iterations per benchmark. 0 means the benchmarks "
-             "are time-based.");
+             "The exact amount of iterations per benchmark");
```
dominichamon: does 0 still mean they're time based? also, period at the end of the sentence please :)

src/benchmark.cc
```diff
@@ -57,155 +56,203 @@ DEFINE_double(benchmark_min_time, 0.5,
...
+       ? str + strliterallen(prefix)                      \
        : NULL)
 
+
 namespace benchmark {
-namespace {
+
 // For non-dense Range, intermediate values are powers of kRangeMultiplier.
 static const int kRangeMultiplier = 8;
+static const int kMinIterations = 1;
```
dominichamon: still necessary?

src/benchmark.cc
```diff
@@ -269,384 +317,127 @@ void BenchmarkFamilies::FindBenchmarks(
...
-        ParseInt32Flag(argv[i], "v", &FLAGS_v)) {
-      for (int j = i; j != *argc; ++j) argv[j] = argv[j + 1];
-
-      --(*argc);
-      --i;
-    } else if (IsFlag(argv[i], "help"))
-      PrintUsageAndExit();
+  if (!re.Init(spec, &error_msg)) {
+    std::cerr << "Could not compile benchmark re: " << error_msg << std::endl;
+    std::exit(1);
```
dominichamon: std::exit in a library is unusual. we should really return a boolean (or raise an exception ;))

src/benchmark.cc
```diff
@@ -269,384 +317,127 @@ void BenchmarkFamilies::FindBenchmarks(
...
     }
-    return static_cast<int64_t>(t * kNumMicrosPerSecond);
-  }
+    for (auto const& args : family->args_) {
+      const std::vector<int>* thread_counts =
+        (family->thread_counts_.empty()
+         ? &one_thread
+         : &family->thread_counts_);
+      for (int num_threads : *thread_counts) {
+        if (num_threads == Benchmark::kNumCpuMarker) {
```
dominichamon: why not set num_threads to NumCPUs directly and do away with kNumCpuMarker? (see old implementation <---)

src/benchmark.cc
```diff
@@ -684,505 +472,480 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-      std::stringstream ss;
-      ss << "/threads:" << instance.threads;
-      instance.name += ss.str();
-    }
+bool running_benchmark = false;
 
-    instances.push_back(instance);
-  }
+// Global variable so that a benchmark can cause a little extra printing
+std::string report_label GUARDED_BY(GetBenchmarkLock());
```
dominichamon: std::string is not a POD. This shouldn't be public or shouldn't be a std::string. a std::string* would work.

src/benchmark.cc
```diff
@@ -684,505 +472,480 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
 
-  return instances;
+// Should this benchmark base decisions off of real time rather than
+// cpu time?
+bool use_real_time GUARDED_BY(GetBenchmarkLock());
+
+// Return prefix to print in front of each reported line
+static const char* Prefix() {
```
dominichamon: aside: this should probably be printed in the header info, not per run. maybe an issue is needed - what do you think?

include/benchmark/benchmark.h
EricWF: This one can't because it is modified in the keep running loop. The other one could.

src/benchmark.cc
EricWF: Yes. We now just use an iteration count to try and get the benchmarks to run for benchmark_min_time.

include/benchmark/benchmark.h
EricWF: Doc bug.

src/benchmark.cc
EricWF: kMinIterations isn't really needed.

src/benchmark.cc
```diff
@@ -684,505 +472,480 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-      break;
-    case STATE_STOPPING:
-      ret = MaybeStop();
-      break;
-    case STATE_STOPPED:
-      CHECK(false);
-      ret = true;
-      break;
+
+void ComputeStats(const std::vector<BenchmarkReporter::Run>& reports,
```
dominichamon: can you leave this where it was (in the anonymous namespace at top of file)?

src/benchmark.cc

src/string_util.cc
```diff
@@ -24,13 +24,13 @@ static_assert(arraysize(kBigSIUnits) == arraysize(kBigIECUnits),
 static_assert(arraysize(kSmallSIUnits) == arraysize(kBigSIUnits),
               "Small SI and Big SI unit arrays must be the same size");
 
-static const int kUnitsSize = arraysize(kBigSIUnits);
+static const int64_t kUnitsSize = arraysize(kBigSIUnits);
```
dominichamon: all these on the left were changed to cope with 64-32 shortening. please don't change them.

src/benchmark.cc

src/benchmark.cc

src/benchmark.cc
EricWF: I agree.

EricWF: Actually I'm not too sure it should be printed at all. Why are we telling them the library was built in debug mode?

src/benchmark.cc
EricWF: Sure. I think it makes more sense to put it at the bottom with the reporter stuff but I'll move this and some other stuff that used to be at the top of the file.

src/string_util.cc
EricWF: arraysize returns a size_t. These need to be 64 bits to prevent those exact warnings. Without these changes `-Wshorten-64-to-32` fires.

src/benchmark.cc
dominichamon: might be worth keeping the same help string as before then.

src/benchmark.cc
dominichamon: because the timings would be massively different, i suppose? hence having a warning in the header instead.

dominichamon: added issue

src/benchmark.cc
```diff
@@ -131,50 +142,41 @@ bool CpuScalingEnabled() {
...
+  Stat1_d items_per_second_stat;
+  int64_t total_iters = 0;
 
   // Populate the accumulators.
   for (std::vector<BenchmarkReporter::Run>::const_iterator it = reports.begin();
-       it != reports.end(); ++it) {
+       it != reports.end();
+       it++) {
     CHECK_EQ(reports[0].benchmark_name, it->benchmark_name);
+    total_iters += it->iterations;
```
dominichamon: when the runs are time based, this isn't necessarily correct. This was a fix made in the current version to report the mean iteration count correctly. i don't think you should change this.

src/benchmark.cc
```diff
@@ -183,29 +185,169 @@ void ComputeStats(const std::vector<BenchmarkReporter::Run>& reports,
...
-    stddev_data->cpu_accumulated_time = cpu_accumulated_time_stat.StdDev();
-  } else {
-    stddev_data->real_accumulated_time = real_accumulated_time_stat.StdDev() *
-                                         stddev_data->iterations;
-    stddev_data->cpu_accumulated_time = cpu_accumulated_time_stat.StdDev() *
-                                        stddev_data->iterations;
-  }
+  stddev_data->iterations = total_iters;
+  // We multiply by total_iters since PrintRunData expects a total time.
+  stddev_data->real_accumulated_time =
```
dominichamon: see the current implementation for an important bugfix when tracking iterations as a stat.

src/benchmark.cc
```diff
@@ -183,29 +185,169 @@ void ComputeStats(const std::vector<BenchmarkReporter::Run>& reports,
...
   stddev_data->bytes_per_second = bytes_per_second_stat.StdDev();
   stddev_data->items_per_second = items_per_second_stat.StdDev();
-  stddev_data->max_heapbytes_used = max_heapbytes_used_stat.StdDev();
 }
-}  // namespace
+
+} // end namespace
+
+
+struct ThreadStats {
```
dominichamon: this could be in the anonymous namespace, right? as could TimerManager?

src/benchmark.cc
```diff
@@ -259,394 +402,135 @@ size_t BenchmarkFamilies::AddBenchmark(Benchmark* family) {
...
-  std::lock_guard<std::mutex> l(mutex_);
-  for (internal::Benchmark* family : families_) {
-    if (family == nullptr) continue;  // Family was deleted
+  // Special list of thread counts to use when none are specified
+  std::vector<int> one_thread;
+  one_thread.push_back(1);
+
+  MutexLock l(mutex_);
+  for (Benchmark* family : families_) {
+    if (family == NULL) continue;  // Family was deleted
```
dominichamon: nullptr should be available in g++-4.6 - is there a reason to not use it?

src/benchmark.cc

src/benchmark.cc
EricWF: Could you explain the issues further? I don't understand.

src/benchmark.cc
```diff
@@ -684,505 +566,389 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+      // Note: Avoid division by zero with max(seconds, 1ns).
+      double multiplier = FLAGS_benchmark_min_time * 1.4 / std::max(seconds, 1e-9);
+      // If our last run was at least 10% of FLAGS_benchmark_min_time then we
+      // use the multiplier directly. Otherwise we use at most 10 times
+      // expansion.
+      // NOTE: When the last run was at least 10% of the min time the max
+      // expansion should be 14x.
+      bool is_significant = (seconds / FLAGS_benchmark_min_time) > 0.1;
+      multiplier = is_significant ? multiplier : std::min(10.0, multiplier);
+      // TODO(ericwf) I don't think this branch is reachable.
```
dominichamon: it is.

min_time = 0 -> multiplier = 0.0 and is_significant = true (because XX/0 is > 0.1)

src/benchmark.cc
```diff
@@ -684,505 +566,389 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-      name_field_width =
-          std::max<size_t>(name_field_width, benchmark.name.size() + 7);
-    } else {
-      name_field_width =
-          std::max<size_t>(name_field_width, benchmark.name.size());
-    }
+  std::size_t name_field_width = 10;
+  for (std::size_t i = 0; i < benchmarks.size(); i++) {
+    name_field_width = std::max<std::size_t>(name_field_width,
+                                    // Maybe add space for appending "_stddev"
```
dominichamon: the current implementation has bugfixes for printing that would be nice to not lose. it's not pretty, but it's the only way to get tabular data in all cases.

don't break this please :)

src/benchmark.cc
```diff
@@ -684,505 +566,389 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-  //  context.cpu_info = base::CompactCPUIDInfoString();
+
   context.cpu_scaling_enabled = CpuScalingEnabled();
   context.name_field_width = name_field_width;
 
-  if (reporter->ReportContext(context))
-    for (internal::Benchmark::Instance& benchmark : benchmarks)
-      Benchmark::RunInstance(benchmark, reporter);
+  if (reporter->ReportContext(context)) {
+    for (const auto& bench_instance : benchmarks) {
```
dominichamon: s/bench_instance/benchmark/

src/benchmark.cc
```diff
@@ -684,505 +566,389 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
-                 [](const internal::Benchmark::Instance& b) { return b.name; });
-}
+} // end namespace internal
 
-}  // end namespace internal
 
-void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter /*= nullptr*/) {
+void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter) {
   std::string spec = FLAGS_benchmark_filter;
   if (spec.empty() || spec == "all")
```
dominichamon: we should create an issue to change spec.empty() to mean none.

test/CMakeLists.txt
```diff
@@ -1,40 +1,25 @@
...
-include_directories(${source_dir}/include)
-link_directories(${binary_dir})
-
 find_package(Threads REQUIRED)
 
 macro(compile_benchmark_test name)
   add_executable(${name} "${name}.cc")
-  target_link_libraries(${name} benchmark gtest ${CMAKE_THREAD_LIBS_INIT})
-  add_dependencies(${name} googletest)
+  target_link_libraries(${name} benchmark ${CMAKE_THREAD_LIBS_INIT})
```
dominichamon: weird .. you're getting the filter test stuff in here but it's not showing up on the other side. maybe a rebase and push?

src/benchmark.cc
dominichamon: if you're not iteration bound then multiple runs (repetitions) may have different iteration counts. when it comes to printing the mean column, it's weird to print the total iterations in that column. the fixed version currently implemented tracks it as a stat so it is correctly reported.

src/benchmark.cc
EricWF: In this version iterations cannot differ between repetitions so I don't think you run into this bug because stddev_iterations = total_iterations.

src/benchmark.cc
EricWF: Yep. Done.

src/benchmark.cc
EricWF: I don't think so.

src/benchmark.cc
EricWF: Should min_time=0 be allowed?

src/benchmark.cc

src/benchmark.cc
EricWF: Ok.

src/benchmark.cc

test/CMakeLists.txt
EricWF: I can push, but I can't rebase because it won't rebase against master.

src/benchmark.cc
dominichamon: sure.. it means "i don't care how quickly this runs" :)

src/benchmark.cc
dominichamon: why not? if min_time is set then different runs might take different iterations to get there (assuming some variance in the benchmark time) which will lead to mean_iterations != total_iterations. unless i'm missing something?

src/benchmark.cc
EricWF: In this version repetitions always have the same iteration count. That iteration count is determined based on time but it isn't changed between runs.

dominichamon: oh.. that sounds like a bug. if i want the min_time and i don't care about iterations, i want each repetition to run to min_time.

src/benchmark.cc
EricWF: To duplicate my above comment:

The amount of iterations to run a benchmark for are computed so that the benchmark runs for ~min_time. This is done for the first run only. Once the amount of iterations every repetition of the benchmark uses the same amount of iterations.

src/benchmark.cc
EricWF: The iteration count is computed so that the benchmark runs in about 1.4x min_time. Computing the iteration count is a rather expensive operation because it requires repeatedly re-running the benchmark on sample iteration counts until you find one large enough. I don't think we want to repeat this operation for every repetition.

How many benchmarks differ by at least 40% between repetitions?

src/benchmark.cc
```diff
@@ -684,443 +564,291 @@ void Benchmark::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
...
+    fprintf(stdout, "***WARNING*** CPU scaling is enabled, the benchmark "
+                    "timings may be noisy\n");
+  }
+
+  int prefix_len = static_cast<int>(std::strlen(Prefix()));
+  CHECK(prefix_len <= static_cast<int>(name_field_width_));
+  int output_width =
+      fprintf(stdout,
+              "%s%-*s %10s %10s %10s\n",
+              Prefix(),
```
dominichamon: i thought you were going to remove this?

EricWF: Pushing the commit helps...

include/benchmark/benchmark.h
```diff
@@ -221,80 +287,54 @@ class State {
   //  BM_Compress   50         50   14115038  compress:27.3%
   //
   // REQUIRES: a benchmark has exited its KeepRunning loop.
-  void SetLabel(const std::string& label);
+  void SetLabel(const char* label);
```
dominichamon: so there's State::SetLabel and benchmark::SetLabel?

that's a little confusing.

src/benchmark.cc
dominichamon: I don't know. I do know that it's weird to report the total when the row is for the mean. That's the main thing that I fixed by making it a statistic.

ie, line 163 below outputs a total for a mean.

test/benchmark_test.cc
```diff
@@ -55,7 +53,7 @@ static void BM_Factorial(benchmark::State& state) {
   while (state.KeepRunning())
     fac_42 = Factorial(8);
   // Prevent compiler optimizations
-  EXPECT_NE(fac_42, std::numeric_limits<int>::max());
+  std::cout << fac_42;
```
dominichamon: while you're here, we should do something better ... maybe use setlabel instead of cout?

test/benchmark_test.cc
```diff
@@ -149,45 +147,12 @@ static void BM_LongTest(benchmark::State& state) {
...
-
-  // Make sure we ran all of the tests
-  const size_t count = test_reporter.GetCount();
-  const size_t expected = (argc == 2) ? std::stoul(argv[1]) : count;
-  if (count != expected) {
-    std::cerr << "ERROR: Expected " << expected << " tests to be ran but only "
-              << count << " completed" << std::endl;
-    return -1;
-  }
+  benchmark::RunSpecifiedBenchmarks();
```
dominichamon: BENCHMARK_MAIN?

include/benchmark/benchmark.h
```diff
@@ -153,14 +154,18 @@ void Initialize(int* argc, const char** argv);
 
 // Otherwise, run all benchmarks specified by the --benchmark_filter flag,
 // and exit after running the benchmarks.
-void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter = nullptr);
+void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter = NULL);
```
dominichamon: can you leave the nullptr stuff alone? or does it conflict with standard library benchmarking?

include/benchmark/benchmark.h
```diff
@@ -153,14 +154,18 @@ void Initialize(int* argc, const char** argv);
...
 
 // ------------------------------------------------------
 // Routines that can be called from within a benchmark
 
 //
 // REQUIRES: a benchmark is currently executing
-void SetLabel(const std::string& label);
+void SetLabel(const char* label);
```
dominichamon: please remove these in favour of setting it on 'state' for consistency with everything else.

EricWF: Woo! I would love to. The only reason I left these in was for backwards compat.

include/benchmark/benchmark.h
EricWF: I tried to leave it alone inside the library source. However I would like to keep the headers compiling in C++03 so I can benchmark stuff in C++03.

test/benchmark_test.cc
EricWF: Ack.

include/benchmark/benchmark.h
EricWF: Should we also do this for UseRealTime()?

dominichamon: probably. it's much more consistent.

test/benchmark_test.cc
EricWF: ack.

src/benchmark.cc
```diff
@@ -57,124 +57,121 @@ DEFINE_double(benchmark_min_time, 0.5,
...
+    bytes_per_second_stat += Stat1_d(run.bytes_per_second, run.iterations);
   }
 
-  // Get the data from the accumulator to BenchmarkRunData's.  In the
-  // computations below we must multiply by the number of iterations since
-  // PrintRunData will divide by it.
+  // Get the data from the accumulator to BenchmarkReporter::Run's.
   mean_data->benchmark_name = reports[0].benchmark_name + "_mean";
-  mean_data->iterations = iterations_stat.Mean();
+  mean_data->iterations = run_iterations;
```
dominichamon: this is still a regression over the current version. if there are multiple runs (when repetitions is set?) this will not reflect the mean as expected by the row label.

dominichamon: this is a regression - i just tested and when 'repetitions' is set > 1, this reports the "mean" iterations as the total. I understand that now the iterations per run will be identical, but then this should be reports[0].iterations, not the total.

just keeping the existing statistical code and allowing it to calculate the mean and stddev will be future proof.

src/benchmark.cc
```diff
@@ -57,124 +57,121 @@ DEFINE_double(benchmark_min_time, 0.5,
...
 
-DEFINE_bool(benchmark_memory_usage, false,
-            "Report memory usage for all benchmarks");
-
 DEFINE_int32(benchmark_repetitions, 1,
              "The number of runs of each benchmark. If greater than 1, the "
              "mean and standard deviation of the runs will be reported.");
 
-DEFINE_int32(v, 0, "The level of verbose logging to output");
 DEFINE_bool(color_print, true, "Enables colorized logging.");
```
dominichamon: i'm not seeing color in my output.

dominichamon:
Looks good

EricWF:
Merged.

Issue #82, opened 2015-03-06
Remove re_test and add the regex matching as a benchmark test against the public API.


EricWF:
The title says it all.

dominichamon:
if we do this, we only depend on googletest for a single EXPECT_NE in the factorial test to avoid an optimization. I wonder if we can do better ... ;)

PR #87, opened 2015-03-10
Add filter test, remove re test, and googletest deps

dominichamon:
Fixes issue #82.

Also removes googletest as a dependency!


dominichamon:
this also speeds up tests, even though more have been added.

- before: Total Test time (real) =  94.58 sec
- after: Total Test time (real) =  79.25 sec

EricWF:
LGTM.

PR #90, opened 2015-03-11
Enable UseRealTime and fix documentation for SetLabel.

### LINKED ISSUE #89: Undefined reference to UseRealTime
Hello,

I tried to use the function to use realtime instead of CPU time:

```c++
// If a particular benchmark is I/O bound, or if for some reason CPU
// timings are not representative, call this method from within the
// benchmark routine.  If called, the elapsed time will be used to
// control how many iterations are run, and in the printing of
// items/second or MB/seconds values.  If not called, the cpu time
// used by the benchmark will be used.
void UseRealTime();
```

like this: 

```c++
static void BM_foo(benchmark::State& state) {
  benchmark::UseRealTime();
  while (state.KeepRunning()) {
    // Stuff
  }
}
```

But this won't link.

The function is actually defined in the `benchmark::internal` namespace. Simply moving it to the `benchmark` namespace _seems_ to work (at least to compile).

Is this function supposed to be used like this ? 

Also the function `void SetLabel(const std::string& label);` (the free function, not the `State` member) is not defined at all.

Cheers,

dominichamon:
Fixes #89

UseRealTime was defined in the internal namespace by mistake.
Similarly, documentation suggested that benchmark::SetLabel should be
used to set a label, and a function was declared but not defined, while
actually the call should be 'state.SetLabel'.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -158,10 +158,6 @@ void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter = nullptr);
 // ------------------------------------------------------
 // Routines that can be called from within a benchmark
 
-//
-// REQUIRES: a benchmark is currently executing
-void SetLabel(const std::string& label);
-
 // If this routine is called, peak memory allocation past this point in the
 // benchmark is reported at the end of the benchmark report line. (It is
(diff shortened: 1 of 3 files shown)


EricWF:
LGTM.

PR #92, opened 2015-03-12
Pimpl the benchmark object to reduce dependencies on the STL.

EricWF:
This is a patch I need in order to use the library against two different STL versions.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -165,7 +165,7 @@ void RunSpecifiedBenchmarks(const BenchmarkReporter* reporter = NULL);
 
 namespace internal {
 class Benchmark;
-class BenchmarkFamilies;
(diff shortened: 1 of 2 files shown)


dominichamon:
Do you have a sense if this adds overhead to short running benchmarks?

You can't use unique_ptr for the imp pointer, right?

EricWF:
It might add overhead but nothing that would show up in the benchmark timings. I don't think an extra indirection would add any major overhoad. Also the compiler is free to inline the BenchmarkImp methods into the Benchmark calls. 

Unfortunately I can't use unique_ptr for the same reason I have to pimpl this but it would be really nice if I could.

dominichamon:
If you could check the benchmark numbers before and after the change, that'd be great. LGTM.

PR #97, opened 2015-03-13
Add output format to ConsoleReporter with support for CSV

dominichamon:


EricWF:
Silly Googlebot. Anyway what is the rational for not making this another class? If we add more formats (ie JSON) it seems like ConsoleReporter could become needlessly complicated to support them all.

dominichamon:
Well, we could. But I thought this was a quick way to extend it to support a new format, then we can look at the surface area of the changes (printing the header and run data) and then bump the ConsoleReporter to a base class with TableReporter and CSVReporter (and JSONReporter) as subclasses.

Do you think it's worth doing that jump now?

EricWF:
I don't think we need to make that jump now but I wanted to ask.

pphaneuf:
Heh, rather hilariously, @dominichamon's CLA doesn't appear to have been migrated to the new system. :smiley: 

I *know* you signed it in the old system, could you re-sign it in the new one? Silly, I know, sorry...

https://cla.developers.google.com/

dominichamon:
I could ... or we could wait until Monday when I'll be a Google employee again :P

pphaneuf:
Okay, now that is **properly** hilarious!

Welcome back, I guess... :smiley:

EricWF:
I don't know a lot about CSV, but I think templated benchmarks might cause problems with the output because the benchmark name may contain commas. Ex BM_myTest<Foo, Bar>.

Can you comment on this?

dominichamon:
closing as a CSVreporter along the lines of #100 will be better.

PR #98, opened 2015-03-17
move reporter internals in both headers and source

EricWF:
This patch is the last step to remove all required dependencies on the STL just to run the benchmarks.


include/benchmark/benchmark.h
```diff
@@ -1,541 +1,21 @@
...
+#define BENCHMARK_BENCHMARK_H_
 
-// Helper macro to create a main routine in a test that runs the benchmarks
-#define BENCHMARK_MAIN()                             \
-  int main(int argc, const char** argv) {            \
-    ::benchmark::Initialize(&argc, argv);            \
-    ::benchmark::RunSpecifiedBenchmarks();           \
-  }
+#include "macros.h"
+#include "minimal_benchmark.h"
```
dominichamon: i wonder if there's a better name here... benchmark_macros? benchmark_impl? benchmark_api?

include/benchmark/reporter.h
```diff
@@ -0,0 +1,99 @@
...
+  // cpu-time and heap memory usage during the benchmark run.
+  // Note that all the grouped benchmark runs should refer to the same
+  // benchmark, thus have the same name.
+  virtual void ReportRuns(const std::vector<Run>& report) const = 0;
+
+  virtual ~BenchmarkReporter();
+};
+
+
+namespace internal {
```
dominichamon: this doesn't need to be internal.

include/benchmark/benchmark.h
EricWF: - benchmark_macros.h: It contains more than just macros.
- benchmark_impl.h: That seems to imply it is private, but users should be able to us it.
- benchmark_api.h: I'm ok with this but I don't feel it convase the actual intention of the header.

How about `benchmark_test.h` because It gives you everything to create and run the tests?

include/benchmark/reporter.h
EricWF: ack.

include/benchmark/benchmark.h
dominichamon: that sounds like a test header :)

naming is hard... i think benchmark_api is the closest ...

EricWF: Or perhaps just `test.h`.

include/benchmark/benchmark_api.h
```diff
@@ -0,0 +1,465 @@
+// Support for registering benchmarks for functions.
```
dominichamon: no include guard or pragma once?

EricWF: Hmm? There is an include guard.

dominichamon:
LGTM with the include guard addition.

waiting for the build too before merging.

PR #99, opened 2015-03-17
Apply reporter interface changes. 

EricWF:
Make report methods non-const and add a Finalize method.

I don't think it makes sense to consider the reporter methods to be const. This patch changes this.

It also adds a Finalize method that is called after all tests have been reported. This is needed for JSON support.


dominichamon:
LGTM

EricWF:
Travis has passed on the last commit. Merging.

PR #100, opened 2015-03-17
Add JSON Reporter

EricWF:


dominichamon:
Woo #100!

reviewing...

src/benchmark.cc
```diff
@@ -807,19 +811,35 @@ void RunMatchingBenchmarks(const std::string& spec,
...
-void RunSpecifiedBenchmarks(BenchmarkReporter* provided_reporter) {
+void RunSpecifiedBenchmarks(BenchmarkReporter* reporter) {
   std::string spec = FLAGS_benchmark_filter;
   if (spec.empty() || spec == "all")
     spec = ".";  // Regexp that matches all benchmarks
-  ConsoleReporter default_reporter;
-  BenchmarkReporter* reporter = provided_reporter ? provided_reporter
-                                                  : &default_reporter;
+
+  std::unique_ptr<BenchmarkReporter> default_reporter;
```
dominichamon: this could be inside the if statement, right? and there's no point in it being a unique_ptr in this case as ownership is not passed.

if RunMatchingBenchmarks took a unique_ptr then this might be useful...

src/benchmark.cc
```diff
@@ -861,6 +884,10 @@ void ParseCommandLineFlags(int* argc, const char** argv) {
       PrintUsageAndExit();
     }
   }
+  if (FLAGS_benchmark_format != "tabular" &&
```
dominichamon: does this make the above check redundant (in GetDefaultReporter)?

src/reporter.cc
```diff
@@ -188,4 +189,145 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
+  return ss.str();
+}
+
+int64_t RoundDouble(double v) {
+    return static_cast<int64_t>(v + 0.5);
+}
+
+} // end namespace
+
+bool JSONReporter::ReportContext(const Context& context) {
```
dominichamon: consider a jsonreporter.cc file to minimize lines per file.

src/benchmark.cc
EricWF: It should.

src/reporter.cc
```diff
@@ -188,4 +189,145 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
+      << ",\n";
+  out << indent
+      << FormatKV("mhz_per_cpu", RoundDouble(context.mhz_per_cpu))
+      << ",\n";
+  out << indent
+      << FormatKV("cpu_scaling_enabled", context.cpu_scaling_enabled)
+      << ",\n";
+
+#if defined(NDEBUG)
+  const char* build_type = "release";
```
dominichamon: const char build_type[]

dominichamon:
This is great - can you add something to the README.md to document it, maybe with some example output?

src/benchmark.cc
EricWF: When a reporter is not provided `GetDefaultReporter()` dynamically allocates one and returns it via a unique_ptr, so there is ownership transfer in this case.

The unique_ptr instance needs to be declared outside of the if statement so that the unique_ptr is not destroyed until the end of the function (and not the end of the if statement).

src/reporter.cc
EricWF: ack.

src/reporter.cc
EricWF: Gladly.

src/benchmark.cc
dominichamon: ah true.. i was thinking that we might be spawning threads in RunMatchingBenchmarks still which would cause an issue with the lifetime of reporter.

You're absolutely right of course.

dominichamon:
pull out into a separate file and then LGTM.

src/reporter.cc
EricWF: Actually I'm going to hold off for now until I find a place to put ComputeStats

dominichamon: static protected method on the base?

but yeah. .. not obvious.

PR #101, opened 2015-03-18
Add variadic macro for BENCHMARK_TEMPLATE in c++11 and beyond.

EricWF:
This patch makes it easier to use the BENCHMARK_TEMPLATE macro when >= c++11 by using variadic macros to accept templates that have an arbitrary number of parameters (and not just 1).

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -445,7 +445,7 @@ class Benchmark {
 // BENCHMARK_TEMPLATE(BM_Foo, 1);
 //
 // will register BM_Foo<1> as a benchmark.
-#define BENCHMARK_TEMPLATE(n, a)                             \
+#define BENCHMARK_TEMPLATE1(n, a)                            \
   static ::benchmark::internal::Benchmark* BENCHMARK_CONCAT( \
       _benchmark_, n, __LINE__) BENCHMARK_UNUSED =           \
       (new ::benchmark::internal::Benchmark(#n "<" #a ">", n<a>))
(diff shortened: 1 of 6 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -455,6 +455,16 @@ class Benchmark {
       _benchmark_, n, __LINE__) BENCHMARK_UNUSED =           \
       (new ::benchmark::internal::Benchmark(#n "<" #a "," #b ">", n<a, b>))
 
+#if __cplusplus >= 201103L
+#define BENCHMARK_TEMPLATE(n, ...) \
```
dominichamon: can you add documentation to the comment above (and/or the README) regarding the use of this?

dominichamon:
LGTM

include/benchmark/benchmark_api.h
EricWF: Done.

include/benchmark/benchmark_api.h
```diff
@@ -445,7 +445,7 @@ class Benchmark {
 // BENCHMARK_TEMPLATE(BM_Foo, 1);
 //
 // will register BM_Foo<1> as a benchmark.
-#define BENCHMARK_TEMPLATE(n, a)                             \
+#define BENCHMARK_TEMPLATE1(n, a)                            \
```
dominichamon: these don't need to be defined for C++11 right? can they be in the #else below?

EricWF: They don't need to be but I see value in keeping them defined. If I have a `BM_foo<T, U>` in `foo.cpp` and I don't know if `foo.cpp` will be compiled in C++11 then I would want to use `BENCHMARK_TEMPLATE2` so that it always works.

PR #102, opened 2015-03-18
Fix and improve BENCHMARK macros.

EricWF:
This patch does a couple of things:

* Removes possibility of `BENCHMARK(...)` creating an reserved identifier. If the benchmark has a name beginning with an underscore (ex `_foo`) the `BENCHMARK(...)` macro will name the benchmark object `_benchmark__foo<line>`. Identifiers with two underscores in a row are reserved.
* Use `__COUNTER__` to generate unique names instead of `__LINE__` when possible. This allows users to define macros that create multiple benchmarks on the same line with the same name. For example:

```c++
#define CREATE_CONTAINER_BENCHMARKS(Cont) \
  BENCHMARK_TEMPLATE(BM_copy, Cont);               \
  BENCHMARK_TEMPLATE(BM_copy, Cont)->ThreadPerCPU()
```

* Make private macros begin with `BENCHMARK_PRIVATE_` .
* Use the utility macro `BENCHMARK_PRIVATE_DECLARE(n)` to declare benchmarks.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -421,14 +421,28 @@ class Benchmark {
 // ------------------------------------------------------
 // Macro to register benchmarks
 
-// Helpers for generating unique variable names
-#define BENCHMARK_CONCAT(a, b, c) BENCHMARK_CONCAT2(a, b, c)
-#define BENCHMARK_CONCAT2(a, b, c) a##b##c
+// Check that __COUNTER__ is defined and that __COUNTER__ increases by 1
+// every time it is expanded. X + 1 == X + 0 is used in case X is defined to be
+// empty. If X is empty the expression becomes (+1 == +0).
+#if defined(__COUNTER__) && (__COUNTER__ + 1 == __COUNTER__ + 0)
+#define BENCHMARK_PRIVATE_UNIQUE_ID __COUNTER__
+#else
+#define BENCHMARK_PRIVATE_UNIQUE_ID __LINE__
+#endif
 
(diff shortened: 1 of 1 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -421,14 +421,27 @@ class Benchmark {
...
+#define BENCHMARK_PRIVATE_UNIQUE_ID __COUNTER__
+#else
+#define BENCHMARK_PRIVATE_UNIQUE_ID __LINE__
+#endif
+
 // Helpers for generating unique variable names
-#define BENCHMARK_CONCAT(a, b, c) BENCHMARK_CONCAT2(a, b, c)
-#define BENCHMARK_CONCAT2(a, b, c) a##b##c
+#define BENCHMARK_PRIVATE_NAME(n) \
+    BENCHMARK_PRIVATE_NAME2(_benchmark_, BENCHMARK_PRIVATE_UNIQUE_ID, n)
```
dominichamon: BENCHMARK_PRIVATE_NAME2 could still be called BENCHMARK_CONCAT (or BENCHMARK_PRIVATE_CONCAT) which is a more useful name.

EricWF: Ok, but I think it is weird to call a function `CONCAT` when it only takes one input.

dominichamon: It takes 3...

EricWF: Oh, I see what you mean.

dominichamon:
this should go in if you still want to land it.

PR #103, opened 2015-03-19
Add more context information and print estimated runtime

EricWF:
It's useful to have an idea of how long it will take to run all of the benchmark. This patch adds enough information to the Context type to do that in some cases. ConsoleReporter now prints the estimated time.

NOTE: when --benchmark_iterations is given there is no way to estimate how long the tests will take to run.

### DIFF
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -14,21 +14,37 @@
 
 #include "benchmark/reporter.h"
 
-#include <cmath>
 #include <cstdio>
 #include <cstdlib>
 #include <iostream>
-#include <limits>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


src/reporter.cc
```diff
@@ -113,6 +115,14 @@ bool ConsoleReporter::ReportContext(const Context& context) {
 #ifndef NDEBUG
   fprintf(stdout, "Build Type: DEBUG\n");
 #endif
+  if (std::abs(context.benchmark_min_time)
```
dominichamon: fabs?

EricWF:
This doesn't work as well as I wanted. Closing.

PR #104, opened 2015-03-26
Fix #96: Prevent times from being negative.

### LINKED ISSUE #96: Negative CPU times
I'm pretty sure this is new: negative times for 'real' time.

Note, the below is release build.
```
 $ test/basic_test
Run on (4 X 775 MHz CPUs)
2015/03/12-21:22:09
***WARNING*** CPU scaling is enabled, the benchmark timings may be noisy
Benchmark                                                Time(ns)    CPU(ns) Iterations
---------------------------------------------------------------------------------------
BM_empty                                                        1          2  360765235                                 
BM_empty/threads:4                                              3          3  221043324                                 
BM_spin_empty/8                                                -4          4  194933402                                 
BM_spin_empty/512                                             552        172    4084371                                 
BM_spin_empty/8k                                            -7250       2660     265357                                 
BM_spin_empty/8/threads:4                                       7          9   79947008                                 
BM_spin_empty/512/threads:4                                   326        396    1797916                                 
BM_spin_empty/8k/threads:4                                   5087       6155     113608                                 
BM_spin_pause_before/8                                        -10          4  194293327                                 
BM_spin_pause_before/512                                      554        172    4085134                                 
BM_spin_pause_before/8k                                     -7281       2655     265210                                 
BM_spin_pause_before/8/threads:4                                7          9   80895864                                 
BM_spin_pause_before/512/threads:4                            334        388    1841124        
```

it seems to be related to ranges and i see it in a range of different benchmarks.

EricWF:
This patch uses std::chrono::*clock when CPU scaling is enabled at the start of the programs run.

It also removes the functionality for printing WallTime and replaces it with a function that returns the current date-time as a string.

NOTE: This breaks when CPU scaling is enabled during the program run but not at the start.


cmake/steady_clock.cpp
```diff
@@ -0,0 +1,7 @@
+#include <chrono>
+
+int main() {
+    typedef std::chrono::steady_clock Clock;
+    Clock::time_point tp = Clock::now();
+    ((void)tp);
+}
```
dominichamon: newline please

src/reporter.cc
```diff
@@ -98,16 +98,12 @@ bool ConsoleReporter::ReportContext(const Context& context) {
...
-                                walltime::Now(), "%Y/%m/%d-%H:%M:%S",
-                                true,  // use local timezone
-                                &remainder_us);
+  std::string walltime_str = LocalDateTimeString();
   fprintf(stdout, "%s\n", walltime_str.c_str());
 
   if (context.cpu_scaling_enabled) {
     fprintf(stdout, "***WARNING*** CPU scaling is enabled, the benchmark "
-                    "timings may be noisy\n");
+                    "timings may be noisy and will incure extra overhead.\n");
```
dominichamon: note this is just for real time measurements, right? CPU time shouldn't actually be affected.

src/walltime.cc
```diff
@@ -22,37 +22,40 @@
...
+    typedef std::chrono::steady_clock type;
+};
+#endif
 
-namespace {
+struct ChooseClockType {
+#if defined(HAVE_STEADY_CLOCK)
+  typedef typename ChooseSteadyClock<>::type type;
+#else
+  typedef std::chrono::high_resolution_clock type;
```
dominichamon: no need to check for high_resolution_clock?

src/walltime.h
```diff
@@ -8,15 +8,14 @@ typedef double WallTime;
...
-// GIVEN: walltime, generic format string (as understood by strftime),
-// a boolean flag specifying if the time is local or UTC (true=local).
-// RETURNS: the formatted string. ALSO RETURNS: the remaining number of
-// microseconds (never printed in the string since strftime does not understand
-// it)
-std::string Print(WallTime time, const char *format, bool local,
-                  int *remainder_us);
 }  // end namespace walltime
+
+std::string DateTimeString(bool local = false);
```
dominichamon: does this need to be exposed in the header?

dominichamon: does it need the default argument?

dominichamon:
Can you add some of your test results to the PR here so we have them for posterity?

src/walltime.cc
EricWF: Doesn't seem like it. GCC 4.6 implements system_clock and high_resolution_clock (likely as a typedef for system_clock).

cmake/steady_clock.cpp
EricWF: ack.

src/reporter.cc
EricWF: Noted.

src/walltime.h
EricWF: At this point? No. If a reporter wants to report a GMT time (as opposed to local) then it or something similar must be exposed.

dominichamon: hide this in the cc file and remove the default = false please.

PR #105, opened 2015-03-27
Change the available Benchmark options

EricWF:
This patch does the following:

* Move `State::UseRealTime()` to `Benchmark::UseRealTime()`.
* Add `Benchmark::MinTime(double t)` option to allow the min time to be set on a per-test basis.
* Remove `--benchmark_iterations=<n>` option. Only allow time based tests.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -802,26 +802,14 @@ void RunMatchingBenchmarks(const std::string& spec,
   auto families = benchmark::internal::BenchmarkFamilies::GetInstance();
   if (!families->FindBenchmarks(spec, &benchmarks)) return;
 
-
   // Determine the width of the name field using a minimum width of 10.
-  // Also determine max number of threads needed.
   size_t name_field_width = 10;
   for (const internal::Benchmark::Instance& benchmark : benchmarks) {
-    // Add width for _stddev and threads:XX
-    if (benchmark.threads > 1 && FLAGS_benchmark_repetitions > 1) {
-      name_field_width =
-          std::max<size_t>(name_field_width, benchmark.name.size() + 17);
-    } else if (benchmark.threads > 1) {
-      name_field_width =
-          std::max<size_t>(name_field_width, benchmark.name.size() + 10);
-    } else if (FLAGS_benchmark_repetitions > 1) {
-      name_field_width =
-          std::max<size_t>(name_field_width, benchmark.name.size() + 7);
-    } else {
-      name_field_width =
-          std::max<size_t>(name_field_width, benchmark.name.size());
-    }
+    name_field_width =
+        std::max<size_t>(name_field_width, benchmark.name.size());
   }
+  if (FLAGS_benchmark_repetitions > 1)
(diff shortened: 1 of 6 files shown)


src/benchmark.cc
```diff
@@ -101,6 +97,10 @@ GetBenchmarkLock()
 
 namespace {
 
+bool IsZero(double n) {
+    return std::abs(n) < std::numeric_limits<double>::epsilon();
```
dominichamon: std::fabs

src/benchmark.cc
```diff
@@ -424,6 +428,12 @@ bool BenchmarkFamilies::FindBenchmarks(
         if (family->arg_count_ >= 2) {
           AppendHumanReadable(instance.arg2, &instance.name);
         }
+        if (!IsZero(family->min_time_)) {
+          instance.name +=  StringPrintF("/min_time:%0.3f",  family->min_time_);
```
dominichamon: you'll need to take this into account in the width of the first column in tabular format.

src/benchmark.cc
```diff
@@ -493,6 +504,15 @@ void BenchmarkImp::RangePair(int lo1, int hi1, int lo2, int hi2) {
   }
 }
 
+void BenchmarkImp::MinTime(double t) {
+  CHECK(t > 0.0);
```
dominichamon: what if i set MinTime(0.0) to disable the minimum time for a given benchmark?

test/benchmark_test.cc
```diff
@@ -70,7 +68,7 @@ static void BM_FactorialRealTime(benchmark::State& state) {
   ss << fac_42;
   state.SetLabel(ss.str());
 }
-BENCHMARK(BM_FactorialRealTime);
+BENCHMARK(BM_FactorialRealTime)->UseRealTime();
```
dominichamon: no need for this .. just add this macro to the BM_Factorial benchmark above to show how it works.

src/benchmark.cc
EricWF: Why? I'm not opposed but std::abs(...) has overloads for floating point types.

dominichamon: oh that new standard library! It didn't used to.

src/benchmark.cc
EricWF: How long does the benchmark run for then? The maximum number of iterations? I think setting it to 0.0 is likely a bug and not a feature.

src/benchmark.cc
EricWF: The way we calculate the width seems weird. Shouldn't it just be the size of the longest name plus room for `_stddev`?

test/benchmark_test.cc
EricWF: ack.

src/benchmark.cc
EricWF: I'll fix the way the column width is calculated.

dominichamon: i didn't want to mess with it too much because it was doing voodoo.

PR #106, opened 2015-03-27
Add `benchmark::DoNotOptimize(...)` to help users prevent optimizations.

EricWF:
Calling `DoNotOptimize(Tp& value)` should be sufficient to prevent the compiler from optimizing away value and the expression that created it.


test/basic_test.cc
```diff
@@ -8,8 +8,7 @@
```
dominichamon: i think you can use this in the other benchmarks too with the CalculatePi benchmark.

test/benchmark_test.cc
```diff
@@ -84,9 +84,8 @@ BENCHMARK_RANGE(BM_CalculatePiRange, 1, 1024 * 1024);
 
 static void BM_CalculatePi(benchmark::State& state) {
   static const int depth = 1024;
-  double pi BENCHMARK_UNUSED = 0.0;
```
dominichamon: do we need BENCHMARK_UNUSED to be public any more?

EricWF: I think so. BENCHMARK_UNUSED is used in the `BENCHMARK()` macros.

test/basic_test.cc
EricWF: ack.

EricWF:
I buggered the revision history on this branch. Commited via patch.

PR #107, opened 2015-03-27
Introduce CSV reporter

dominichamon:


EricWF:
Both the names and the labels have to be quoted because they can contain commas. 
We also need to replace every occurrence of `"` in the label with `""`.

dominichamon:
How can the names have commas? We generate the names from the benchmark functions.

i'll try to fix up the labels.

EricWF:
The names can have commas in them if the benchmark function is a template.

```c++
template <class T, class U>
void BM_foo(benchmark::State& state);
// name: BM_foo<T, U>
```

EricWF:
It is possible for the name field to include nested quotes as well (in very rare cases). Example:
```c++
template <char MyChar>
void BM_foo(benchmark::State& state);
BENCHMARK_TEMPLATE(BM_foo, '"');
```

EricWF:
LGTM

PR #108, opened 2015-03-31
Add --benchmark_list_tests option and change filter to match generated name.

EricWF:


src/benchmark.cc
```diff
@@ -414,7 +418,8 @@ bool BenchmarkFamilies::FindBenchmarks(
           instance.name += StringPrintF("/threads:%d", instance.threads);
         }
 
-        benchmarks->push_back(instance);
+        if (re.Match(instance.name))
```
dominichamon: it's not required, but for consistency these should have {} even though they're single-line.

src/benchmark.cc
```diff
@@ -766,19 +771,29 @@ void State::SetLabel(const char* label) {
...
 
 namespace internal {
+namespace {
+
+void PrintBenchmarkList() {
+  std::vector<Benchmark::Instance> benchmarks;
+  auto families = BenchmarkFamilies::GetInstance();
+  if (!families->FindBenchmarks(".", &benchmarks)) return;
+
+  for (const internal::Benchmark::Instance& benchmark : benchmarks)
```
dominichamon: {} please

src/benchmark.cc

src/benchmark.cc

PR #113, opened 2015-04-01
Libraries are treated as runtime objects on Windows

mattyclarkson:
When the library is created as a *.dll on Windows it is treated like a
runtime object so we must proivde the destination for the runtime
objects in our install command


dominichamon:
this doesn't seem to change anything for  Linux (which is good) but i'm surprised that the dll has to be in the runtime directory. i trust your windows know-how :)

PR #115, opened 2015-04-01
Remove BENCHMARK_ENABLE_SHARED and prefer BUILD_SHARED_LIBS

EricWF:
Address comments from #114.

PR #116, opened 2015-04-01
Improve cmake dependencies

geoffromer:
Configure 'benchmark' cmake target so that when other targets depend on it, they get the appropriate include directories and link libraries automatically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 2.8.12)
 project (benchmark)
 
 option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 2.8.11)
 project (benchmark)
 
 option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,7 +20,7 @@ endif()
 add_library(benchmark ${SOURCE_FILES} ${RE_FILES})
 
 find_package(Threads REQUIRED)
-target_link_libraries(benchmark INTERFACE ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(benchmark ${CMAKE_THREAD_LIBS_INIT})
 
 set_target_properties(benchmark PROPERTIES
   OUTPUT_NAME "benchmark"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,10 +18,12 @@ else()
 endif()
 
 add_library(benchmark ${SOURCE_FILES} ${RE_FILES})
+find_package(Threads REQUIRED)
 
-# Link threading if building a shared library.
+# Link threading if building a shared library, otherwise let client do it.
 if (BUILD_SHARED_LIBS)
-  find_package(Threads REQUIRED)
+  target_link_libraries(benchmark PRIVATE ${CMAKE_THREAD_LIBS_INIT})
+else()
   target_link_libraries(benchmark ${CMAKE_THREAD_LIBS_INIT})
 endif()
 
@@ -35,6 +37,8 @@ set_target_properties(benchmark PROPERTIES
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   target_link_libraries(benchmark Shlwapi)
 endif()
+
+target_include_directories(benchmark PUBLIC ${PROJECT_SOURCE_DIR}/include)
 
 # Install target (will install the library to specified CMAKE_INSTALL_PREFIX variable)
 install(
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,7 +20,7 @@ endif()
 add_library(benchmark ${SOURCE_FILES} ${RE_FILES})
 
 find_package(Threads REQUIRED)
-target_link_libraries(benchmark ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(benchmark INTERFACE ${CMAKE_THREAD_LIBS_INIT})
 
 set_target_properties(benchmark PROPERTIES
   OUTPUT_NAME "benchmark"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,14 +18,9 @@ else()
 endif()
 
 add_library(benchmark ${SOURCE_FILES} ${RE_FILES})
-find_package(Threads REQUIRED)
 
-# Link threading if building a shared library, otherwise let client do it.
-if (BUILD_SHARED_LIBS)
-  target_link_libraries(benchmark PRIVATE ${CMAKE_THREAD_LIBS_INIT})
-else()
-  target_link_libraries(benchmark ${CMAKE_THREAD_LIBS_INIT})
-endif()
+find_package(Threads REQUIRED)
+target_link_libraries(benchmark ${CMAKE_THREAD_LIBS_INIT})
 
 set_target_properties(benchmark PROPERTIES
   OUTPUT_NAME "benchmark"


mattyclarkson:
@geoffromer, might want to have a look at #115 as that's changing the cmake code slightly.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 2.8.12)
```
dominichamon: looks like travis is stuck at 2.8.7. We're using that as a proxy for minspec as it is Ubuntu 12.04 LTS which seems like a good base system.

having said that, we could add something like https://gist.github.com/winterz/10cc2741d466cbd8ff21 to the travis build script to get the Trusty cmake (which is 2.8.12 iiuc).

geoffromer: It's also possible to leave the minimum at 2.8; it just means we'd have to drop `INTERFACE` from the `target_link_libraries` line. Would you prefer that?

geoffromer:
@mattyclarkson Thanks, I've re-synced to pick up that change.

CMakeLists.txt
dominichamon: that'll mean we'll only be linking them in the shared version again, right?

i think that's ok for now. having the find_package(Threads..) is an improvement even in the static case where we don't use them as not linking against it in your final executable causes runtime errors.

mattyclarkson: @geoffromer, can you detect what version of CMAKE is being used and put the INTERFACE in there if it is available?

Something like:

```
target_link_libraries(benchmark ${CMAKE_TARGET_LINK_INTERFACE} ${CMAKE_THREAD_LIBS_INIT})
```

`CMAKE_TARGET_LINK_INTERFACE` would be empty if cmake is too old. Just thinking out loud - then it would do the right thing if the user has a newer version of cmake but gracefully fallback if not.

mattyclarkson:
@geoffromer, consider running `git rebase master` on your branch to remove the unnecessary `Merge branch 'master' into cmake-patch` commit.

CMakeLists.txt
geoffromer: OK, dropped the `INTERFACE`.

@mattyclarkson I don't know cmake well enough to confidently pull that off, and it seems like unnecessary complexity.

geoffromer:
Squashed all my changes into a single commit. Sorry if that messed anything up; I'm new at Git collaboration.

src/CMakeLists.txt
```diff
@@ -18,10 +18,12 @@ else()
...
 add_library(benchmark ${SOURCE_FILES} ${RE_FILES})
+find_package(Threads REQUIRED)
 
-# Link threading if building a shared library.
+# Link threading if building a shared library, otherwise let client do it.
 if (BUILD_SHARED_LIBS)
-  find_package(Threads REQUIRED)
+  target_link_libraries(benchmark PRIVATE ${CMAKE_THREAD_LIBS_INIT})
+else()
   target_link_libraries(benchmark ${CMAKE_THREAD_LIBS_INIT})
```
dominichamon: i'm pretty sure this will statically link pthread into the static library, no? i'm not completely familiar with how target_link_libraries works under the different variants in cmake.

EricWF: Yes it will. The correct thing to do would be to use the `INTERFACE` specifier but since that isn't available I don't think we should statically link pthread.

src/CMakeLists.txt
```diff
@@ -18,10 +18,12 @@ else()
 endif()
 
 add_library(benchmark ${SOURCE_FILES} ${RE_FILES})
+find_package(Threads REQUIRED)
 
-# Link threading if building a shared library.
+# Link threading if building a shared library, otherwise let client do it.
 if (BUILD_SHARED_LIBS)
-  find_package(Threads REQUIRED)
+  target_link_libraries(benchmark PRIVATE ${CMAKE_THREAD_LIBS_INIT})
```
EricWF: Shouldn't this be PUBLIC when pthread is a shared library and PRIVATE when pthread is static? Also wouldn't we want to avoid linking pthread as a static library?

EricWF:
So I think we should find a way to upgrade CMake to 2.8.12. I checked the LLVM CMake and mailing lists because this is what they use. Here is the rational they have for choosing that version:

http://lists.cs.uiuc.edu/pipermail/llvmdev/2015-February/081970.html
http://lists.cs.uiuc.edu/pipermail/llvmdev/2015-March/083672.html

After doing this we should be able to use the target_link_directories command as well as specifying `INTERFACE` libraries.

src/CMakeLists.txt
geoffromer: Why not? IIUC, static linking only links the pieces that are needed, so if benchmark doesn't use pthread, this will have no effect.

src/CMakeLists.txt
geoffromer: Oh, OK, I thought linking pthread statically was the intent of this line. If the goal is just to add pthread to the link interface, why are the static and shared cases treated differently?

EricWF: If libbenchmark.so links to pthread.so then anything that links to benchmark must also link pthread in order to get the pthread symbols, otherwise it will fail to compile.

If libbenchmark.so links to libpthread.a then all of the symbols needed for benchmark are already provided and other targets do not need to link to pthread.

  -

geoffromer: So wait, does `BUILD_SHARED_LIBS` control whether pthread is shared, or whether benchmark is shared, or both? I'd been assuming it at least controlled benchmark, but your answer makes it sound like that's not the case.

EricWF: It controls whether benchmark is built as a shared library. I don't think it controls how pthread is linked.

geoffromer: OK, I remain confused, but I'm out of my depth here, so I'm happy to do what you tell me. But how can I "avoid linking pthread as a static library"?

geoffromer:
OK, I've updated this to require cmake 2.8.12.

src/CMakeLists.txt
EricWF: When we build benchmark as a static library we will copy some of the symbols from libpthread.a into libbenchmark.a. Now whenever libbenchmark.a is used in an executable that also links pthread we get two definitions of some symbols.

src/CMakeLists.txt
geoffromer: Actually, the cmake documentation seems to be saying that PUBLIC/PRIVATE/INTERFACE is only relevant for shared libraries (and dependencies are always propagated when linking statically), so I think this can be simplified a lot. What do you think of this?

EricWF: CMake is going to pass `-lpthread` to the compiler, so it is up to the compiler on how to link it. AFAIK the compiler will prefer to just about always link pthread dynamically. I think the only case where it won't do so is if your linking pthread into a static library but I'm not sure.

EricWF:
You'll actually need to change CMake to 2.8.11. That was the closest version I could find for travis. Sorry.

src/CMakeLists.txt
EricWF: > the cmake documentation seems to be saying that PUBLIC/PRIVATE/INTERFACE is only relevant for shared libraries.

Ah my mistake. CMake just ignores the pthread input when building a static library. My mistake.

EricWF:
Oh, I forgot to say... In order for the CMake upgrade to work with travis we will need to merge #117

geoffromer:
@EricWF the description of #117 indicates that you're upgrading to CMake 3.2. Is that right?

src/CMakeLists.txt
geoffromer: Judging from the [gcc documentation](https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-static-1225), gcc will link against shared libs when available, unless you pass `-static`. CMake doesn't seem to pass that flag, so I think this will not statically link pthread even when building static libs, and so it should work fine even without `INTERFACE`.

dominichamon:
I checked and this doesn't seem to make a difference to how pthread is linked in to static or shared libraries (checking that pthread symbols are weakly linked in both cases).

As such, i'm happy for this to land as soon as #117 does.

Thanks so much for the cleanup!

dominichamon:
Merged manually. travis should pass - watching.

PR #117, opened 2015-04-02
Rework travis configuration

EricWF:
This patch does the following:

1. Cleanup travis configuration.
2. Upgrade CMake to 3.2
3. Add Clang 3.6 testers.

I would prefer to use CMake 2.8.12 (the soon to be minimum version) when testing, but I couldn't find a repository that had it. Unfortunately cmake.org only provides x86_64 linux binaries for CMake 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,21 @@
 language: cpp
+
+# NOTE: The COMPILER variable is unused. It simply makes the display on
+# travis-ci.org more readable.
 matrix:
     include:
         - compiler: gcc
-          env: BUILD_TYPE=Debug STD=c++0x
+          env: COMPILER=g++-4.6     STD=c++0x BUILD_TYPE=Debug
         - compiler: gcc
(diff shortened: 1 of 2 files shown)


dominichamon:
LGTM

.travis.yml
```diff
@@ -1,32 +1,24 @@
...
-    - os: linux
-      env: BUILD_TYPE=Release STD=c++11
-    - os: osx
-      env: BUILD_TYPE=Debug   STD=c++11
-    - os: osx
-      env: BUILD_TYPE=Release STD=c++11
-
-language:
-    - cpp
+language: cpp
```
dominichamon: this doesn't seem to show up in the travis UI the same as the old version did.

EricWF: It didn't show up before. Here is the last build of master (with the old config) https://travis-ci.org/google/benchmark.

I was hoping a different syntax would make it show up, but it doesn't.

EricWF:
Pushed without commit history.

Issue #109, opened 2015-03-31
test fixtures


dominichamon:
The current design has an implicit setup/teardown in each benchmark function with the assumption that microbenchmarks shouldn't care about complicated setup. However, there are some cases where test fixtures would help coverage. Specifically benchmarking containers with non-integer value types and a variety of key values.

dominichamon:
This would have to fundamentally change the API I think. Currently, we just use macros to register user-defined static methods. With fixtures, we may have to change to the googletest model of defining classes from benchmarks.

Unless... Perhaps a 'fixture' can inherit from `benchmark::State` and be passed to the benchmark instead of the state. Ie, you'd have something like:

```
static void BM_Basic(benchmark::State& state) {
  while (state.KeepRunning()) {
    ...
  }
}
BENCHMARK(BM_basic);

class MyState : public benchmark::State {
 protected:
  MyState() : _count(0) {}
  ~MyState() final {
    std::stringstream ss;
    ss << _count;
    SetLabel(ss.str());
  }

  int _count;
};

static void BM_Fixture(MyState& state) {
  while (state.KeepRunning()) {
    ...
    ++_count;
  }
}

BENCHMARK_F(MyState, BM_Fixture);
```

We still need a second version of all the top-level registration macros that take a fixture, and we'd want to check that the given fixture is-a state. This will also require more templating on State...

What do you think, @EricWF?

EricWF:
That makes sense to me for the most part. I agree that we shouldn't need to break the current API. 

However I don't know if we should use the constructor/destructer for initialization/destruction. Since the fixture classes will likely have a static lifetime. I also don't know how your given example of `BM_Fixture` would have `_count` in scope.

dominichamon:
`_count` isn't in scope - that was a bug :P

Why would the fixture classes have static lifetime? `RunInThread` would need to create the state on the heap as a pointer using the correct fixture type (instead of on the stack as it is now) but it would have the same per-run lifetime. no?

EricWF:
Hmm, Perhaps I'm not understanding what your fully suggesting. Could you explain further how you see these fixture classes working?

dominichamon:
I've tooled around and can't find a syntax i like. I was hoping to have either `Stat`e be subclassed (but registering and instantiating that correctly is hard) so then I tried to have `State` contain an 'inner state' of some user-defined type, but that was equally disruptive.

[...]

PR #120, opened 2015-04-06
Add ability to use benchmark fixtures

EricWF:
This patch implements #109. It adds the ability to create and use benchmarks that derive from test fixtures.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -185,6 +185,8 @@ struct EnableIfString<T, typename Voider<typename T::basic_string>::type> {
 
 void UseCharPointer(char const volatile*);
 
+// Take ownership of the pointer and register the benchmark. Return the
+// registered benchmark.
 Benchmark* RegisterBenchmarkInternal(Benchmark*);
 
 } // end namespace internal
@@ -375,9 +377,7 @@ typedef void(Function)(State&);
 // Each method returns "this" so that multiple method calls can
 // chained into one expression.
 class Benchmark {
- public:
-  Benchmark(const char* name);
-
+public:
   virtual ~Benchmark();
 
   // Note: the following methods all return "this" so that multiple
@@ -455,8 +455,9 @@ class Benchmark {
   struct Instance;
 
 protected:
-   Benchmark(Benchmark const&);
-   void SetName(const char* name);
+  explicit Benchmark(const char* name);
+  Benchmark(Benchmark const&);
+  void SetName(const char* name);
 
 private:
   friend class BenchmarkFamilies;
@@ -480,23 +481,7 @@ private:
 
 }  // end namespace internal
 
-// The base class for all fixture tests. Fixture tests are created by
-// first defining a type that derives from ::benchmark::Fixture and then
-// creating/registering the tests using the following macros:
-//
-// * BENCHMARK_F(ClassName, Method)
-// * BENCHMARK_DEFINE_F(ClassName, Method)
-// * BENCHMARK_REGISTER_F(ClassName, Method)
-//
-// For Example:
-//
-// class MyFixture : public benchmark::Fixture {};
-//
-// BENCHMARK_F(MyFixture, FooTest)(benchmark::State& st) {
-//    while (st.KeepRunning()) {
-//      ...
-//    }
-// }
+// The base class for all fixture tests. 
 class Fixture: public internal::Benchmark {
 public:
     Fixture() : internal::Benchmark("") {}
@@ -580,44 +565,29 @@ protected:
 
 
 #define BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method) \
-class BaseClass##_##Method##_Test : public BaseClass { \
+class BaseClass##_##Method##_Benchmark : public BaseClass { \
 public:\
-    BaseClass##_##Method##_Test() : BaseClass() {this->SetName(#BaseClass "/" #Method);} \
+    BaseClass##_##Method##_Benchmark() : BaseClass() {this->SetName(#BaseClass "/" #Method);} \
 protected: \
     virtual void TestCase(::benchmark::State&); \
 };
 
-// The BENCHMARK_DEFINE_F(...) and BENCHMARK_REGISTER_F(...) macros are used
-// to define and register new fixture benchmarks in two steps.
-// Example:
-//
-// class MyFixture : public ::benchmark::Fixture {};
-//
-// BENCHMARK_DEFINE_F(MyFixture, Method)(benchmark::State& st) {
-//    while(st.KeepRunning()) {
-//        ...
-//    }
-// }
-// /* the test is not registered. */
-// BENCHMARK_REGISTER_F(MyFixture, Method)->Arg(42);
-// /* the test is now registered */
 #define BENCHMARK_DEFINE_F(BaseClass, Method) \
     BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method) \
-    void BaseClass##_##Method##_Test::TestCase
+    void BaseClass##_##Method##_Benchmark::TestCase
 
 #define BENCHMARK_REGISTER_F(BaseClass, Method) \
-    BENCHMARK_PRIVATE_REGISTER_F(BaseClass##_##Method##_Test)
+    BENCHMARK_PRIVATE_REGISTER_F(BaseClass##_##Method##_Benchmark)
 
 #define BENCHMARK_PRIVATE_REGISTER_F(TestName) \
     BENCHMARK_PRIVATE_DECLARE(TestName) = \
         (::benchmark::internal::RegisterBenchmarkInternal(new TestName()))
 
-// This function will define and register a benchmark within a fixture class.
-// See Fixture for more information.
+// This macro will define and register a benchmark within a fixture class.
 #define BENCHMARK_F(BaseClass, Method) \
     BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method) \
     BENCHMARK_REGISTER_F(BaseClass, Method); \
-    void BaseClass##_##Method##_Test::TestCase
+    void BaseClass##_##Method##_Benchmark::TestCase
 
 
 // Helper macro to create a main routine in a test that runs the benchmarks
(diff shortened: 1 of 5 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -372,9 +376,9 @@ typedef void(Function)(State&);
 // chained into one expression.
 class Benchmark {
  public:
-  Benchmark(const char* name, Function* f);
+  Benchmark(const char* name);
```
dominichamon: explicit on one-arg constructors please.

include/benchmark/benchmark_api.h
```diff
@@ -499,21 +560,66 @@ class Benchmark {
...
+protected: \
+    virtual void TestCase(::benchmark::State&); \
+};
+
+// The BENCHMARK_DEFINE_F(...) and BENCHMARK_REGISTER_F(...) macros are used
+// to define and register new fixture benchmarks in two steps.
+// Example:
+//
+// class MyFixture : public ::benchmark::Fixture {};
+//
```
dominichamon: we probably don't need this documentation twice. adding it to the README and then having some short note here is enough.

src/benchmark.cc
```diff
@@ -273,27 +273,23 @@ class BenchmarkFamilies {
...
-  std::vector<BenchmarkImp*> families_;
+  std::vector<std::unique_ptr<Benchmark>> families_;
   Mutex mutex_;
 };
 
 
 class BenchmarkImp {
 public:
-  BenchmarkImp(const char* name, Function* func);
+  BenchmarkImp(const char* name);
```
dominichamon: explicit on one-arg constructors

src/benchmark.cc
```diff
@@ -375,9 +348,10 @@ bool BenchmarkFamilies::FindBenchmarks(
   one_thread.push_back(1);
 
   MutexLock l(mutex_);
-  for (BenchmarkImp* family : families_) {
+  for (std::unique_ptr<Benchmark>& bench_family : families_) {
```
dominichamon: const ref? it doesn't look like bench_family is mutated.

src/benchmark.cc
```diff
@@ -906,6 +895,13 @@ void ParseCommandLineFlags(int* argc, const char** argv) {
   }
 }
 
+Benchmark* RegisterBenchmarkInternal(Benchmark* bench) {
```
dominichamon: to document the ownership passing, this could take a `std::unique_ptr<Benchmark>` directly.

include/benchmark/benchmark_api.h
```diff
@@ -445,15 +449,71 @@ class Benchmark {
...
+//    while (st.KeepRunning()) {
+//      ...
+//    }
+// }
+class Fixture: public internal::Benchmark {
+public:
+    Fixture() : internal::Benchmark("") {}
+
+    virtual void Run(State& st) {
+      this->SetUp();
```
dominichamon: it's a shame we can't rely on the constructor/destructor pair to be setup/teardown. however, i agree that that won't work well here.

include/benchmark/benchmark_api.h
```diff
@@ -499,21 +560,66 @@ class Benchmark {
...
+      (::benchmark::internal::RegisterBenchmarkInternal( \
+        new ::benchmark::internal::FunctionBenchmark( \
+        #n "<" #__VA_ARGS__ ">", n<__VA_ARGS__>)))
 #else
 #define BENCHMARK_TEMPLATE(n, a) BENCHMARK_TEMPLATE1(n, a)
 #endif
 
+
+#define BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method) \
+class BaseClass##_##Method##_Test : public BaseClass { \
```
dominichamon: s/_Test/_Benchmark/ :)

include/benchmark/benchmark_api.h
```diff
@@ -499,21 +560,66 @@ class Benchmark {
...
+    BENCHMARK_PRIVATE_REGISTER_F(BaseClass##_##Method##_Test)
+
+#define BENCHMARK_PRIVATE_REGISTER_F(TestName) \
+    BENCHMARK_PRIVATE_DECLARE(TestName) = \
+        (::benchmark::internal::RegisterBenchmarkInternal(new TestName()))
+
+// This function will define and register a benchmark within a fixture class.
+// See Fixture for more information.
+#define BENCHMARK_F(BaseClass, Method) \
+    BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method) \
```
dominichamon: it looks like `BaseClass` only has to be a `Benchmark`, rather than a `Fixture`.

Should the `RegisterBenchmarkInternal` method be overloaded on `Fixture` and `FunctionBenchmark` to ensure users don't create their own `Benchmark` specialization that defines `Run` and then something odd breaks?

include/benchmark/benchmark_api.h
```diff
@@ -445,15 +449,71 @@ class Benchmark {
...
+// For Example:
+//
+// class MyFixture : public benchmark::Fixture {};
+//
+// BENCHMARK_F(MyFixture, FooTest)(benchmark::State& st) {
+//    while (st.KeepRunning()) {
+//      ...
+//    }
+// }
+class Fixture: public internal::Benchmark {
```
dominichamon: nit: be consistent. `FixtureBenchmark` to match `FunctionBenchmark`

or flip them both: `BenchmarkFixture` and `BenchmarkFunction`.

include/benchmark/benchmark_api.h
EricWF: ack.

src/benchmark.cc

src/benchmark.cc
EricWF: ack. This can't take unique_ptr directly to satisfy my need that no c++ library types cross the library boundary.

include/benchmark/benchmark_api.h
EricWF: ack.

include/benchmark/benchmark_api.h
EricWF: I don't really want to name `Fixture` anything more complex than `Fixture`. `FunctionBenchmark` is an internal type so the name should matter a little less for now.

include/benchmark/benchmark_api.h
EricWF: It actually does have to be `Fixture` because the `Fixture` class declares `virtual void TestCase(State&) = 0;` that the resulting test case needs to override.

I want the `RegisterBenchmarkInternal` be the universal way to register benchmarks. I don't really understand why we would try and protect against silly users like that. Are user supposed to be able to manually write out benchmark specializations?

dominichamon: thinking about it more, if someone wants to hack around and find a way to create a benchmark specialization, more power to them. i'd love to hear their usecase.

include/benchmark/benchmark_api.h
```diff
@@ -480,7 +525,8 @@ class Benchmark {
   BENCHMARK_PRIVATE_NAME(n) BENCHMARK_UNUSED
 
 #define BENCHMARK(n) \
-    BENCHMARK_PRIVATE_DECLARE(n) = (new ::benchmark::internal::Benchmark(#n, n))
+    BENCHMARK_PRIVATE_DECLARE(n) = (::benchmark::internal::RegisterBenchmarkInternal( \
```
dominichamon: is this longer than 80 cols?

include/benchmark/benchmark_api.h
```diff
@@ -499,21 +545,51 @@ class Benchmark {
...
+class BaseClass##_##Method##_Benchmark : public BaseClass { \
+public:\
+    BaseClass##_##Method##_Benchmark() : BaseClass() {this->SetName(#BaseClass "/" #Method);} \
+protected: \
+    virtual void TestCase(::benchmark::State&); \
+};
+
+#define BENCHMARK_DEFINE_F(BaseClass, Method) \
+    BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method) \
+    void BaseClass##_##Method##_Benchmark::TestCase
```
dominichamon: s/Test/Benchmark/ throughout.

src/benchmark.cc
```diff
@@ -313,15 +310,13 @@ class BenchmarkImp {
...
-  Function* function_;
   int arg_count_;
   std::vector< std::pair<int, int> > args_;  // Args for all benchmark runs
   double min_time_;
   bool use_real_time_;
   std::vector<int> thread_counts_;
-  std::size_t registration_index_;
 
-  BENCHMARK_DISALLOW_COPY_AND_ASSIGN(BenchmarkImp);
+  BenchmarkImp& operator=(BenchmarkImp const&);
```
dominichamon: why not BENCHMARK_DISALLOW_COPY_AND_ASSIGN?

that should be fine given that we're not copying anywhere...

test/fixture_test.cc
```diff
@@ -0,0 +1,22 @@
+
+#include "benchmark/benchmark.h"
+
+class MyFixture : public ::benchmark::Fixture
+{
```
dominichamon: can you add SetUp and TearDown that get called so we see the usage?

EricWF: ack.

include/benchmark/benchmark_api.h
EricWF: Not anymore.

include/benchmark/benchmark_api.h
EricWF: ack.

src/benchmark.cc
EricWF: Unfortunately that would require

1. Passing std::unique_ptr across the library boundary.
2. C++11 in the headers.

dominichamon:
LGTM

Would the syntax for registering be any cleaner if we allowed a std::function as a third parameter?

```
BENCHMARK_F(BaseClass, BenchmarkName, [&state] {
  while(state.KeepRunning()) {
  }
})
```

and then we just pass a std::function around internally.

I know this might restrict the minimum compiler (though not by much) but does it block you from your work and do you think it's cleaner?

PR #121, opened 2015-04-07
Add Value-Parameterized Benchmarks

EricWF:
This patch adds value-parameterized benchmarks. These benchmarks accept any number of arguments after the first State argument. The value for these arguments are set when the benchmark is created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,25 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
 /* BarTest is now registered */
 ```
 
+Value Benchmarks
+-------------------
+Value benchmarks let you pass an arbitrary number of extra arguments to a
+benchmark test. The value of these extra arguments is specified when the
+benchmark is created. Use the `BENCHMARK_V(...)` macro to create value
+benchmarks.
+NOTE: Value benchmarks are only enabled in C++11 and beyond.
+
+```c++
+std::vector<int> test_vector1 = {1, 2, 3};
+
+void BM_find_in(benchmark::State&, std::vector<int>& cont, int value) {
+    while (st.KeepRunning()) {
+         DoNotOptimize(std::find(cont.begin(), cont.end(), value));
+    }
+}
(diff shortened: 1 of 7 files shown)


dominichamon:
Your example isn't very compelling as I can do the find_in benchmark without value parameterising the benchmark, I think. Create the vector on the stack and pass in 42 as an Arg.

Or did I miss something?

dominichamon:
I'm still not seeing it. Unlike googletest, you have a preamble to set up local values for the benchmark. Now we have fixtures as well, i'm not seeing why you'd need the values to be passed in.

I'm sorry, i'm clearly missing something.

EricWF:
Fixtures allow you to share some data between different tests.
Value benchmarks allow you to share different data with common tests.

Say I want to benchmark std::sort. I need to test std::sort against different inputs, different iterator types and different value types but I don't want to re-write the sort benchmark function for each input.

With value benchmarks I can re-use the same benchmark body for each of these tests and just pass in different inputs.

With fixtures I have to define a new fixture every time I want different input and I need to re-define the benchmark body for each test.

Does that make more sense?

dominichamon:
https://github.com/google/benchmark/blob/master/test/benchmark_test.cc#L97

That tests push_back for different container types and inputs. It would probably be cleaner with value benchmarks, but i'm not sure what you're not getting.

The only reason i'm pushing back is to try to keep the API simple and to avoid adding things that are C++11 only.

EricWF:
I'll come up with better and clearer motivating cases and post them here. I think there is great utility in supporting an interface similar to this that I have yet to demonstrate. Until I manage that I won't push to move this forward.

>The only reason i'm pushing back is to try to keep the API simple and to avoid adding things that are
> C++11 only.

Can you try and expand on what you think is not clean about this approach? I don't disagree but I want to focus in on what can be improved. Also I can implement this in C++03 for a limited function arity with a lot more code but we should have a discussion on if that is worth it. I think it might just be best to restrict these features to C++11.

PR #122, opened 2015-04-30
Coverage

mattyclarkson:
I've implemented a branch that allows for `lcov` coverage of the code base. To enable just install `lcov` and `gcov` with your package manager and do:

```
cmake -DCMAKE_BUILD_TYPE=Coverage .
make coverage
```

It'll create the coverage report in the `lcov` directory. Just run up a web server `python3 -m http.server` or `python2 -m SimpleHTTPServer` and go to `localhost:8000`.

Current coverage is:

| Directory         | Line Coverage  | Functions      |
|:------------------|:---------------|:---------------|
| include/benchmark | 84.1% 37/44    | 55.0 %  11/20  |
| src               | 62.0% 738/1190 | 73.4 % 146/199 |
| test              | 97.6% 242/248  | 95.2 %  60/63  |

Coverage isn't everything, as long as you're testing the critical stuff - but it's nice to have the metric!

As part of the branch I've have done some other improvements to the build:
 * The C++ flag checker has got better at detecting flags that are required by the linker
 * Added link time optimisation if supported
 * Added sanitization checkers to debug builds
 * Updated the `.gitignore` to ignore the new unit tests
 * Allow the user to determine the C++ standard to build with


CMakeLists.txt
```diff
@@ -17,38 +17,88 @@ message("-- Version: ${VERSION}")
...
+set_property(CACHE BENCHMARK_CXX_STANDARD PROPERTY STRINGS
+  auto c++14 c++11 c++98)
+if ("${BENCHMARK_CXX_STANDARD}" STREQUAL "auto")
+  # Try and enable C++11. Don't use C++14 because it doesn't work in some
+  # configurations.
+  #add_cxx_compiler_flag(-std=c++14)
+  if (NOT HAVE_CXX_FLAG_STD_CXX14)
+    add_cxx_compiler_flag(-std=c++11)
+    if (NOT HAVE_CXX_FLAG_STD_CXX11)
+      add_cxx_compiler_flag(-std=c++98)
```
EricWF: I don't think the code compiles as C++98.

CMakeLists.txt
```diff
@@ -17,38 +17,88 @@ message("-- Version: ${VERSION}")
...
+# The C and C++ standards to use
+set(BENCHMARK_CXX_STANDARD auto CACHE STRING
+  "The C++ standard to use; auto uses the latest supported by the compiler")
+set_property(CACHE BENCHMARK_CXX_STANDARD PROPERTY STRINGS
+  auto c++14 c++11 c++98)
+if ("${BENCHMARK_CXX_STANDARD}" STREQUAL "auto")
+  # Try and enable C++11. Don't use C++14 because it doesn't work in some
+  # configurations.
+  #add_cxx_compiler_flag(-std=c++14)
+  if (NOT HAVE_CXX_FLAG_STD_CXX14)
```
EricWF: What is the point of enabling C++14 for anything other than testing the code in the headers? I don't see what is limiting about only building the library in C++11. What problem are you trying to fix?

mattyclarkson: I wasn't actually trying to fix anything. Was allowing CMake to pick the latest standard but that's not really necessary. I can suck this commit out if you like?

CMakeLists.txt
```diff
@@ -17,38 +17,88 @@ message("-- Version: ${VERSION}")
...
-add_cxx_compiler_flag(-fno-strict-aliasing RELEASE)
-
 add_cxx_compiler_flag(-Wthread-safety)
 if (HAVE_WTHREAD_SAFETY)
   add_definitions(-DHAVE_WTHREAD_SAFETY)
   cxx_feature_check(THREAD_SAFETY_ATTRIBUTES)
 endif()
 
+# Debug flags
+add_cxx_compiler_flag(-fsanitize=address DEBUG)
```
EricWF: If you are using clang as your c++ compiler it will require a sanitized standard library. Not sure about GCC. Can you pull these changes out for now? They don't seem to fit.

CMakeLists.txt
mattyclarkson: ack, sorry.

CMakeLists.txt
EricWF: Yeah I think that makes sense. C++14 enables C11 which removes `gets(...)`. There are certain compiler/stdlib/libc combinations on linux where `gets(...)` is used but is not available in the headers. 

`check_cxx_compiler_flag(-std=c++14)` doesn't see this issue because the compile test does not include any system headers that use `gets(...)`.

Since we don't *need* any C++14 features it is easier to just limit the library to C++11 for now.

mattyclarkson: OK. Sorry about that! Will rebase now.

EricWF:
Thank you very much for the code coverage patch. It seems very cool. Sorry for being so negative off the bat.

CMakeLists.txt
mattyclarkson: This adds `-fsanitize=address` to the link line as well which results in linking the sanitization library correctly. I can add a check for `-static-asan` to link the address sanitizer statically if available.

Happy to remove though. Pretty handy in debug mode to get pointer bounds checking.

EricWF: Your right, it is super useful and it is a feature I want to have. I don't think the way you handle it is portable though. I'm sure it works with GCC and libstdc++, however I don't think this will work on any platform that ships Clang and libc++ as their C++ stack (OS X, FreeBSD).

To be clear, I would really really really really like to enable asan and ubsan for testing by default. This change just needs a little more work.

mattyclarkson: Ah, on board now. Will back out that change.

mattyclarkson:
@EricWF, no worries fella. Have got thick skin :wink:  Plus you own the repo so your decisions FTW.

CMakeLists.txt
```diff
@@ -17,38 +17,71 @@ message("-- Version: ${VERSION}")
...
 
 # Turn compiler warnings up to 11
 add_cxx_compiler_flag(-Wall)
 add_cxx_compiler_flag(-Wextra)
 add_cxx_compiler_flag(-Wshadow)
-add_cxx_compiler_flag(-Werror)
-add_cxx_compiler_flag(-pedantic-errors)
+add_cxx_compiler_flag(-Werror RELEASE)
+add_cxx_compiler_flag(-pedantic)
+add_cxx_compiler_flag(-pedantic-errors RELEASE)
```
EricWF: Why only use pedantic errors in release mode?

test/CMakeLists.txt
```diff
@@ -43,3 +43,32 @@ compile_benchmark_test(cxx03_test)
...
+  find_program(GCOV gcov)
+  find_program(LCOV lcov)
+  find_program(GENHTML genhtml)
+  # TODO: support more compilers and generators
+  find_program(MAKE make)
+  if (GCOV AND LCOV AND GENHTML AND MAKE AND "${CMAKE_GENERATOR}" STREQUAL "Unix Makefiles")
+    add_custom_command(
+      OUTPUT ${CMAKE_BINARY_DIR}/lcov/index.html
+      COMMAND ${LCOV} -q -z -d .
+      COMMAND ${LCOV} --no-external -q -c -b "${CMAKE_BINARY_DIR}" -d . -o initial.lcov  -i
```
dominichamon: this depends on the source being built without coverage, right? where is that dependency captured?

test/CMakeLists.txt
```diff
@@ -43,3 +43,32 @@ compile_benchmark_test(cxx03_test)
...
     PROPERTIES COMPILE_FLAGS "${CXX03_FLAGS}")
 add_test(cxx03 cxx03_test --benchmark_min_time=0.01)
+
+# Add the coverage command(s)
+if (${CMAKE_BUILD_TYPE} MATCHES "[Cc]overage")
+  find_program(GCOV gcov)
+  find_program(LCOV lcov)
+  find_program(GENHTML genhtml)
+  # TODO: support more compilers and generators
+  find_program(MAKE make)
```
dominichamon: i'm not sure you want to restrict this at all.. is there no way to get the 'make' command from cmake?

test/CMakeLists.txt
```diff
@@ -43,3 +43,32 @@ compile_benchmark_test(cxx03_test)
...
+      WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+      COMMENT "Running LCOV"
+    )
+    add_custom_target(coverage
+      DEPENDS ${CMAKE_BINARY_DIR}/lcov/index.html
+      COMMENT "LCOV report at lcov/index.html"
+    )
+    message(STATUS "Coverage command added")
+  else()
+    message(WARNING "Coverage command not available for ${CMAKE_C_COMPILER_ID}")
```
dominichamon: it's not the cmake compiler that's the problem .. it's the generator and/or make. can you make this error clearer?

dominichamon:
why is 'test' being checked for coverage? i thought the point was that 'test' is used to cover 'src' and 'include'...

EricWF:
I'm having no luck using your patch to generate coverage. The problem I ran into is that the `.gcno` and `.gcda` files are created in <build-root>/src/CMakeFiles/benchmark.dir/ and <build-root>/test/CMakeFiles/<test-target>.dir/ but those *specific* directories are not passed to lcov. This results in an empty info.lcov file.

What platform are you testing this on?

EricWF:
I figured out the problem. Are you building inside the top level of the repository? I build in a directory outside of the repo. For this reason `--no-external` excludes all coverage data because the actual sources files are not in `CMAKE_BINARY_DIR`. Adding `CMAKE_SOURCE_DIR` as a second directory fixed this problem.

CMakeLists.txt
mattyclarkson: Just so you can implement features in debug mode quickly without having to worry about the high warning levels. For example, I'm hacking on a new feature and I want to just block out some code with `#if 0` then I get unused variable errors but I'm just hacking so it's not an issue. However, once the feature stabilized in debug mode and the warnings have reduced (or gone) I can move to release mode and finalize the feature.

As always, if you don't agree happy to back it out.

test/CMakeLists.txt
mattyclarkson: It depends on the source being built _with_ coverage. In the `CMakeList.txt` file in the route it sets up the `COVERAGE` build variant. It adds the `--coverage` flag.

test/CMakeLists.txt
mattyclarkson: I can look into that.

test/CMakeLists.txt
mattyclarkson: Well it depends on three things:
 * coverage support in the compiler used (`--coverage` on gcc)
 * `lcov` available on the command line
 * Knowing the generator to run the unit tests

mattyclarkson:
@dominichamon, good point about `test` being included in the coverage. That's not very useful. Will ignore that directory.

@EricWF yeah I was doing an in place build. I can patch it up to support out of source builds. It's interesting the project allows in source builds - usually CMake projects disable that.

EricWF:
@mattyclarkson I've created a patch against your coverage branch that gets this working in out of tree builds and fixes some bugs.

https://gist.github.com/EricWF/eaf2dc4ce888cb659275

test/CMakeLists.txt
EricWF: Can we just use `ctest` instead?

CMakeLists.txt
EricWF: Seems reasonable. What about turning only turning on `-Werror` in `RELEASE` but always keeping `-pedantic-errors` on. I don't *think* that `-pedantic-errors` turns unused variables and other small warnings into errors.

cmake/AddCXXCompilerFlag.cmake
```diff
@@ -20,18 +20,18 @@ set(__add_cxx_compiler_flag INCLUDED)
...
   string(REGEX REPLACE "_+" "_" SANITIZED_FLAG ${SANITIZED_FLAG})
-  check_cxx_compiler_flag(${FLAG} ${SANITIZED_FLAG})
+  set(CMAKE_REQUIRED_FLAGS "${FLAG}")
+  check_cxx_compiler_flag("" ${SANITIZED_FLAG})
   if(${SANITIZED_FLAG})
-    string(REGEX REPLACE "[^A-Za-z_0-9]" "_" VARIANT "${VARIANT}")
-    string(TOUPPER "${VARIANT}" VARIANT)
+    set(VARIANT ${ARGV1})
+    if(ARGV1)
+      string(REGEX REPLACE "[^A-Za-z_0-9]" "_" VARIANT "${VARIANT}")
```
EricWF: I'm not sure we should need to sanitize `VARIANT` like this. The way it is used implies that it should always be a valid CMake identifier. I realize this is an existing problem but since your in the area :)

CMakeLists.txt
```diff
@@ -17,38 +17,71 @@ message("-- Version: ${VERSION}")
...
   add_definitions(-DHAVE_WTHREAD_SAFETY)
   cxx_feature_check(THREAD_SAFETY_ATTRIBUTES)
 endif()
 
+# Release flags
+add_cxx_compiler_flag(-fno-strict-aliasing RELEASE)
+add_cxx_compiler_flag(-flto RELEASE)
+
+# Configure the linker
+if ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
```
EricWF: I'm assuming these binaries are needed for GCC to do LTO? What happens if we add the -flto flag but we don't find the required ar/ranlib?

CMakeLists.txt
mattyclarkson: That's fine by me. Pedantic errors are usually checking for standards conformance.

CMakeLists.txt
mattyclarkson: Yeah, you're correct.  Without these wrappers the LTO won't work due to AR not loading the correct plugin. The `gcc-ar` translates the `-flto` into the correct plugin flag. Without the wrappers static linking of the library fails spectacularly on 4.9.2

cmake/AddCXXCompilerFlag.cmake
mattyclarkson: I can take that out. It doesnt do any harm being in there but _is_ redundant

test/CMakeLists.txt
mattyclarkson: Ack. `ctest` would make perfect sense. Thanks.

CMakeLists.txt
EricWF: I like the idea of using LTO. In particular because every optimization helps us remove  timing overhead from the benchmark measurements. However setting everything up to build with LTO seems fragile and non-portable. 

For that reason I don't think we should enable LTO by default just yet (or enable it without a way to turn it off). Could you guard all of this stuff within a cmake option (ex `BENCHMARK_ENABLE_LTO`)?

mattyclarkson:
Have pushed up a rebased branch addressing the comments. The error message for when coverage isn't supported is still a bit crappy:

```
    message(WARNING "Coverage not available. Currently only available when the compiler supports coverage")
```

Any suggestions on the best thing to put there? I've never actually used coverage on `clang` it seems that you can use `llvm-cov` to generate the `gcov` compatible data that `lcov` picks up. I don't currently have my Mac plugged in so can't test today. Might be able to look at that in another patch.

CMakeLists.txt
dominichamon: could you make this change please?

i'm not sure it's a great idea to have different compiler flags across debug/release boundaries because it's too easy to forget to test release before pushing. travis helps a little, but it's still a potential point of failure. however, i can see the benefit in flexibility.

CMakeLists.txt
```diff
@@ -17,38 +18,73 @@ message("-- Version: ${VERSION}")
...
-add_cxx_compiler_flag(-fno-strict-aliasing RELEASE)
-
 add_cxx_compiler_flag(-Wthread-safety)
 if (HAVE_WTHREAD_SAFETY)
   add_definitions(-DHAVE_WTHREAD_SAFETY)
   cxx_feature_check(THREAD_SAFETY_ATTRIBUTES)
 endif()
 
+# Release flags
+add_cxx_compiler_flag(-fno-strict-aliasing RELEASE)
```
dominichamon: why?

CMakeLists.txt
```diff
@@ -17,38 +18,73 @@ message("-- Version: ${VERSION}")
...
+  "Flags used by the shared libraries linker during coverage builds."
+  FORCE)
+mark_as_advanced(
+  CMAKE_CXX_FLAGS_COVERAGE
+  CMAKE_EXE_LINKER_FLAGS_COVERAGE
+  CMAKE_SHARED_LINKER_FLAGS_COVERAGE)
+set(CMAKE_BUILD_TYPE "${CMAKE_BUILD_TYPE}" CACHE STRING
+  "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel Coverage."
+  FORCE)
+add_cxx_compiler_flag(--coverage COVERAGE)
```
dominichamon: i may be reading this wrong, but doesn't this always add the --coverage flag?

test/CMakeLists.txt
```diff
@@ -43,3 +43,33 @@ compile_benchmark_test(cxx03_test)
...
+      WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+      COMMENT "Running LCOV"
+    )
+    add_custom_target(coverage
+      DEPENDS ${CMAKE_BINARY_DIR}/lcov/index.html
+      COMMENT "LCOV report at lcov/index.html"
+    )
+    message(STATUS "Coverage command added")
+  else()
+    message(WARNING "Coverage not available. Currently only available when the compiler supports coverage")
```
dominichamon: you could check which of gcov, lcov, genhtml, ctest, or the coverage flag are unset :)

CMakeLists.txt
EricWF: the presence of the second `COVERAGE` arg means that the flag gets added to CMAKE_CXX_FLAGS_COVERAGE and not directly to CMAKE_CXX_FLAGS.

CMakeLists.txt
EricWF: Agreed. I think we should always use `-fstrict-aliasing -Wstrict-aliasing`.

mattyclarkson: I haven't changed that - it was that way in the project on `master`. Look at line 44. I just moved it so that it was grouped with the release flags. Want me to make it into the following?

```
add_cxx_compiler_flag(-Wstrict-aliasing)
add_cxx_compiler_flag(-fstrict-aliasing
```

test/CMakeLists.txt
mattyclarkson: OK. Will do that.

CMakeLists.txt
EricWF: Sorry, I realize this already existed but since it ended up in the diff it got my attention. I discussed this @dominichamon last night and we agreed that there is no reason why benchmark should/would violate the strict aliasing rules. Please make the change.

mattyclarkson: Will do!

CMakeLists.txt
mattyclarkson: @EricWF, you're correct the `--coverage` flag will only be enabled if someone does `CMAKE_BUILD_TYPE=Coverage`

mattyclarkson:
@dominichamon, @EricWF addressed your comments and pushed up rebased commits.

EricWF:
@dominichamon Do you have anywhere you think we should host these results? I'm happy to put them somewhere one one of my domains (efcs.ca)

mattyclarkson:
@EricWF, @dominichamon want me to set up [coveralls support](https://coveralls.io/)? It's easy to do and uploads all the gcov data to there.

EricWF:
@mattyclarkson Oh, Cool resource. I'll have to let @dominichamon decide if that's what he wants. I don't work at Google anymore so I wouldn't feel comfortable saying yes to using that.

(however I'm going to try and use coveralls on other projects I work on. Thanks).

mattyclarkson:
@EricWF, I just pushed up the change to see if the coveralls works. Once the [build finishes](https://travis-ci.org/mattyclarkson/benchmark/jobs/62253448) the results should show up [here](https://coveralls.io/r/mattyclarkson/benchmark). I can drop the patch if @dominichamon  doesn't want to use coveralls.

CMakeLists.txt
```diff
@@ -17,38 +18,72 @@ message("-- Version: ${VERSION}")
...
+add_cxx_compiler_flag(-Werror RELEASE)
+add_cxx_compiler_flag(-pedantic)
 add_cxx_compiler_flag(-pedantic-errors)
 add_cxx_compiler_flag(-Wshorten-64-to-32)
 add_cxx_compiler_flag(-Wfloat-equal)
 add_cxx_compiler_flag(-Wzero-as-null-pointer-constant)
-# Release flags
-add_cxx_compiler_flag(-fno-strict-aliasing RELEASE)
-
+add_cxx_compiler_flag(-Wstrict-aliasing)
```
EricWF: Small nit. Could you reverse the order of these flags? `-Wstrict-aliasing` has no effect unless `-fstrict-aliasing` is given. For that reason I think the reverse order is better (although it probably has no effect).

CMakeLists.txt
```diff
@@ -17,38 +18,72 @@ message("-- Version: ${VERSION}")
...
+add_cxx_compiler_flag(-fstrict-aliasing)
 add_cxx_compiler_flag(-Wthread-safety)
 if (HAVE_WTHREAD_SAFETY)
   add_definitions(-DHAVE_WTHREAD_SAFETY)
   cxx_feature_check(THREAD_SAFETY_ATTRIBUTES)
 endif()
 
+# Link time optimisation
+if (BENCHMARK_ENABLE_LTO)
```
EricWF: If a user asks for LTO I think we should give it to them regardless of if it is a DEBUG or RELEASE build. Any thoughts?

mattyclarkson:
The coveralls support [works](https://coveralls.io/builds/2544981) Will just need to make it enabled for the repository after the PR is merged.

CMakeLists.txt
mattyclarkson: I'll add the warning conditionally depending on if the flag is supported.

CMakeLists.txt
mattyclarkson: Makes sense. Will push up the change.

dominichamon:
I'm fine with using coveralls and am looking into how to enable it now.

test/CMakeLists.txt
```diff
@@ -43,3 +43,44 @@ compile_benchmark_test(cxx03_test)
 set_target_properties(cxx03_test
     PROPERTIES COMPILE_FLAGS "${CXX03_FLAGS}")
 add_test(cxx03 cxx03_test --benchmark_min_time=0.01)
+
+# Add the coverage command(s)
+if (${CMAKE_BUILD_TYPE} MATCHES "[Cc]overage")
```
EricWF: I would really like `COVERAGE` to work as well. Could you convert `${CMAKE_BUILD_TYPE}` to upper-case (or lower-case) before comparing?

EricWF:
Other than the requested changes this LGTM. Thanks for all the work! I'll let @dominichamon merge it once he figures out coverall.

dominichamon:
Badge is ready and coveralls is waiting.

mattyclarkson:
image

test/CMakeLists.txt
mattyclarkson: Yeah. Will push up a patch.

mattyclarkson:
@dominichamon, looks like coveralls is all set up, even commented on the PR! Ready for merge I guess.

PR #125, opened 2015-05-22
Fixing cmake issue with empty cmake_build_type

dcoeurjo:
Checking if the cmake_build_type is not empty before trying to lower-case it.


dcoeurjo:
signed.

PR #126, opened 2015-06-01
Auto-detect whether Benchmark should produce colorized output

nickhutchinson:
Attached is a patch that makes Benchmark only produce colorized output
if stdout is a TTY. (This can be overridden with --color-print=true.)

The logic for deciding whether to use colorized output was lifted from
GTest: <https://code.google.com/p/googletest/source/browse/trunk/src/gtest.cc#2872>.

I haven't got a Windows system to test my changes on, but I see from #7
that there's more work to be done here anyway, so hopefully that's okay!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -24,6 +24,7 @@ Jussi Knuuttila <jussi.knuuttila@gmail.com>
 Kaito Udagawa <umireon@gmail.com>
 Lei Xu <eddyxu@gmail.com>
 Matt Clarkson <mattyclarkson@gmail.com>
+Nick Hutchinson <nshutchinson@gmail.com>
 Oleksandr Sochka <sasha.sochka@gmail.com>
 Paul Redmond <paul.redmond@gmail.com>
 Radoslav Yovchev <radoslav.tm@gmail.com>
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -34,6 +34,7 @@
 #include <thread>
 
 #include "check.h"
+#include "colorprint.h"
 #include "commandlineflags.h"
 #include "complexity.h"
 #include "log.h"
@@ -82,7 +83,12 @@ DEFINE_string(benchmark_out_format, "json",
 
 DEFINE_string(benchmark_out, "", "The file to write additonal output to");
 
-DEFINE_bool(color_print, true, "Enables colorized logging.");
+DEFINE_string(benchmark_color, "auto",
+              "Whether to use colors in the output.  Valid values: "
+              "'true'/'yes'/1, 'false'/'no'/0, and 'auto'. 'auto' means to use "
+              "colors if the output is being sent to a terminal and the TERM "
+              "environment variable is set to a terminal type that supports "
+              "colors.");
 
 DEFINE_int32(v, 0, "The level of verbose logging to output");
 
@@ -532,8 +538,14 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter,
   std::unique_ptr<BenchmarkReporter> default_console_reporter;
   std::unique_ptr<BenchmarkReporter> default_file_reporter;
   if (!console_reporter) {
-    auto output_opts = FLAGS_color_print ? ConsoleReporter::OO_Color
-                                          : ConsoleReporter::OO_None;
+    auto output_opts = ConsoleReporter::OO_None;
+    if (FLAGS_benchmark_color == "auto")
+      output_opts = IsColorTerminal() ? ConsoleReporter::OO_Color
+                                      : ConsoleReporter::OO_None;
+    else
+      output_opts = IsTruthyFlagValue(FLAGS_benchmark_color)
+                        ? ConsoleReporter::OO_Color
+                        : ConsoleReporter::OO_None;
     default_console_reporter = internal::CreateReporter(
           FLAGS_benchmark_format, output_opts);
     console_reporter = default_console_reporter.get();
@@ -588,7 +600,7 @@ void PrintUsageAndExit() {
           "          [--benchmark_format=<console|json|csv>]\n"
           "          [--benchmark_out=<filename>]\n"
           "          [--benchmark_out_format=<json|console|csv>]\n"
-          "          [--color_print={true|false}]\n"
+          "          [--benchmark_color={auto|true|false}]\n"
           "          [--v=<verbosity>]\n");
   exit(0);
 }
@@ -613,8 +625,12 @@ void ParseCommandLineFlags(int* argc, char** argv) {
                         &FLAGS_benchmark_out) ||
         ParseStringFlag(argv[i], "benchmark_out_format",
                         &FLAGS_benchmark_out_format) ||
-        ParseBoolFlag(argv[i], "color_print",
-                       &FLAGS_color_print) ||
+        ParseStringFlag(argv[i], "benchmark_color",
+                        &FLAGS_benchmark_color) ||
+        // "color_print" is the deprecated name for "benchmark_color".
+        // TODO: Remove this.
+        ParseStringFlag(argv[i], "color_print",
+                        &FLAGS_benchmark_color) ||
         ParseInt32Flag(argv[i], "v", &FLAGS_v)) {
       for (int j = i; j != *argc; ++j) argv[j] = argv[j + 1];
 
@@ -627,6 +643,9 @@ void ParseCommandLineFlags(int* argc, char** argv) {
   for (auto const* flag : {&FLAGS_benchmark_format,
                            &FLAGS_benchmark_out_format})
   if (*flag != "console" && *flag != "json" && *flag != "csv") {
+    PrintUsageAndExit();
+  }
+  if (FLAGS_benchmark_color.empty()) {
     PrintUsageAndExit();
   }
 }
--- a/src/colorprint.cc
+++ b/src/colorprint.cc
@@ -16,16 +16,20 @@
 
 #include <cstdarg>
 #include <cstdio>
-#include <cstdarg>
-#include <string>
+#include <cstdlib>
+#include <cstring>
 #include <memory>
+#include <string>
 
 #include "check.h"
 #include "internal_macros.h"
 
 #ifdef BENCHMARK_OS_WINDOWS
+#include <io.h>
 #include <Windows.h>
-#endif
+#else
+#include <unistd.h>
+#endif // BENCHMARK_OS_WINDOWS
 
 namespace benchmark {
 namespace {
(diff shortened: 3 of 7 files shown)


src/posix.h
```diff
@@ -0,0 +1,57 @@
...
+
+#ifdef OS_WINDOWS
+#include <io.h>  // _isatty
+#else
+#include <strings.h>  // strcasecmp
+#include <unistd.h>   // isatty
+#endif
+
+namespace benchmark {
+namespace posix {
```
EricWF: I know the top level namespace `posix` is reserved for the implementation in C++. I would prefer not using a nested one either. Could you come up with another name?

AUTHORS
```diff
@@ -19,6 +19,7 @@ Google Inc.
 JianXiong Zhou <zhoujianxiong2@gmail.com>
 Lei Xu <eddyxu@gmail.com>
 Matt Clarkson <mattyclarkson@gmail.com>
+Nick Hutchinson <nshutchinson@gmail.com>
```
EricWF: This doesn't really matter, but what is the difference between "authors" and "contributors"?

src/benchmark.cc
```diff
@@ -877,8 +881,8 @@ void ParseCommandLineFlags(int* argc, const char** argv) {
                        &FLAGS_benchmark_repetitions) ||
         ParseStringFlag(argv[i], "benchmark_format",
                         &FLAGS_benchmark_format) ||
-        ParseBoolFlag(argv[i], "color_print",
-                       &FLAGS_color_print) ||
+        ParseStringFlag(argv[i], "color_print",
```
EricWF: I think the value of this switch should be verified below.

src/colorprint.cc
```diff
@@ -73,13 +74,50 @@ PlatformColorCode GetPlatformColorCode(LogColor color) {
   };
 #endif
 }
+
+// Returns true iff Benchmark should use colors in the output.
+bool ShouldUseColor(bool stdout_is_tty) {
+  const char* const color_print = FLAG(color_print).c_str();
```
EricWF: It seems like it would be a lot easier to compare a std::string than a C string.

```c++
std::string const& color_print = FLAG(color_print);
const bool term_supports_color =
   term == "xterm" ||
   term == "foo" ||
  ...;
```

EricWF:
Seems like a reasonable patch. Is there a reason to perfer using C strings in the function interfaces as opposed to std::string? If there isn't then I would prefer to use std::string.

AUTHORS
dominichamon: AUTHORS are people who own the copyright, ie, if a person signs the CLA on behalf of a company, that company should be in here.

CONTRIBUTORS are people who contribute but do not have copyright necessarily.

Everyone who contributes should be in CONTRIBUTORS. Everyone who claims copyright should be in AUTHORS.

src/benchmark.cc
```diff
@@ -858,7 +862,7 @@ void PrintUsageAndExit() {
           "          [--benchmark_min_time=<min_time>]\n"
           "          [--benchmark_repetitions=<num_repetitions>]\n"
           "          [--benchmark_format=<tabular|json|csv>]\n"
-          "          [--color_print={true|false}]\n"
+          "          [--color_print={auto|true|false}]\n"
```
dominichamon: the usage says 'true' and 'false' while the help says 'yes' and 'no'.

src/benchmark.cc
```diff
@@ -67,7 +67,11 @@ DEFINE_string(benchmark_format, "tabular",
               "The format to use for console output. Valid values are "
               "'tabular', 'json', or 'csv'.");
 
-DEFINE_bool(color_print, true, "Enables colorized logging.");
+DEFINE_string(color_print, "auto",
```
dominichamon: a boolean might be better with default 'true' being overridden if the TERM doesn't support it.

src/posix.h
nickhutchinson: My model here was Google Test's [internal/gtest-port.h](https://code.google.com/p/googletest/source/browse/trunk/include/gtest/internal/gtest-port.h#2239). I'll have a ponder about an alternate namespace name. Would you be open to ditching the nested namespace altogether?

src/benchmark.cc
nickhutchinson: I think the three states are necessary: it's sometimes useful to pipe to some command that *does* support colorized output. In this case, the `isatty()` test would fail, so you do need a way to forcibly enable colors.

src/posix.h
EricWF: Not opposed towards dropping the namespace.

src/benchmark.cc
nickhutchinson: How lenient should the check be? ParseBoolFlag() is quite generous with what it accepts, as is Google Test's --gtest_color option. 

I've changed it to assert that the user passes a non-empty string. Is that enough?

src/benchmark.cc
nickhutchinson: Fixed

src/benchmark.cc
```diff
@@ -877,14 +882,17 @@ void ParseCommandLineFlags(int* argc, const char** argv) {
...
+        // Legacy flag for benchmark_color.
+        ParseStringFlag(argv[i], "color_print",
+                        &FLAGS_benchmark_color) ||
         ParseInt32Flag(argv[i], "v", &FLAGS_v)) {
       for (int j = i; j != *argc; ++j) argv[j] = argv[j + 1];
 
       --(*argc);
       --i;
-    } else if (IsFlag(argv[i], "help")) {
+    } else if (IsFlag(argv[i], "help") || HasBenchmarkFlagPrefix(argv[i])) {
```
dominichamon: !HasBenchmarkFlagPrefix?

also, redundant given how we parse the strings above.

src/commandlineflags.cc
```diff
@@ -216,4 +209,34 @@ bool ParseStringFlag(const char* str, const char* flag, std::string* value) {
...
+bool HasBenchmarkFlagPrefix(const char* str) {
+  assert(str);
+  return (SkipPrefix("--", &str) ||
+          SkipPrefix("-", &str) ||
+          SkipPrefix("/", &str)) &&
+         (SkipPrefix("benchmark_", &str) ||
+          SkipPrefix("benchmark-", &str));
+}
+
+bool IsTruthyFlagValue(const char* value)
```
dominichamon: this should be taking a string.

src/commandlineflags.cc
```diff
@@ -216,4 +209,34 @@ bool ParseStringFlag(const char* str, const char* flag, std::string* value) {
 bool IsFlag(const char* str, const char* flag) {
   return (ParseFlagValue(str, flag, true) != nullptr);
 }
+
+// If |*pstr| begins with |prefix|, advance *pstr so that it points
+// just past the prefix, and return true. Otherwise, return false.
+static bool SkipPrefix(const char* prefix, const char** pstr) {
```
dominichamon: neither of these functions are useful.

src/benchmark.cc
nickhutchinson: Not quite -- without this check, you could pass `--benchmark_format` or similar with no associated value, and no error will be reported.

src/commandlineflags.cc
nickhutchinson: Is there a reason this one should take a std::string when the others in the header take a c string?

src/commandlineflags.cc
nickhutchinson: Could you elaborate?

src/benchmark.cc
dominichamon: $ test/benchmark_test --benchmark_format=foo
benchmark [--benchmark_list_tests={true|false}]
          [--benchmark_filter=<regex>]
          [--benchmark_min_time=<min_time>]
          [--benchmark_repetitions=<num_repetitions>]
          [--benchmark_format=<tabular|json|csv>]
          [--color_print={true|false}]
          [--v=<verbosity>]

this is working as intended. the checks for specific args come next.

src/commandlineflags.cc
dominichamon: we don't need HasBenchmarkFlagPrefix (see earlier comment) and SkipPrefix is only used by it.

For colored output, this is not necessary.

src/commandlineflags.cc
dominichamon: I think they should all be taking std::string. @EricWF might have a preference though as some headers should be C-ish only.

EricWF: As long as the std::string doesn't cross the DSO boundary. Internally I agree we should use std::string.

dominichamon:
@nickhutchinson are you still working on this?

EricWF:
@nickhutchinson Are you interested in continuing with this?

nickhutchinson:
Sorry for the delay, I’ll try fix this up this week.

nickhutchinson:
Finally got around to rebasing this on master -- thanks for your patience. I believe I've addressed the review comments, in that the new patch:
- uses std::string instead of const char*
- removes error-checking of CLI arguments -- as pointed out, this isn't relevant to colorised output.

I've tested on Win10/MSVC2015 and OS X 10.11/Clang 3.8.

src/posix.h
```diff
@@ -0,0 +1,45 @@
...
+
+inline const char* GetEnv(const char* name) {
+#if defined(__BORLANDC__) || defined(__SunOS_5_8) || defined(__SunOS_5_9)
+  // Environment variables which we programmatically clear will be set to the
+  // empty string rather than unset (nullptr).  Handle that case.
+  const char* const env = getenv(name);
+  return (env != nullptr && env[0] != '\0') ? env : nullptr;
+#else
+  return getenv(name);
+#endif // BENCHMARK_OS_WINDOWS
```
EricWF: The comment is incorrect.

src/posix.h
```diff
@@ -0,0 +1,45 @@
...
+}
+
+inline int IsATTY(int fd) {
```
EricWF: Let's make this idiomatic C++ and change the return type to `bool`

src/posix.h
```diff
@@ -0,0 +1,45 @@
+#ifndef BENCHMARK_POSIX_H_
```
EricWF: I'm concerned about naming a header `posix.h` since it's a very generic name. Maybe `posix_support.h` or `benchmark_posix_support.h`

src/commandlineflags.cc
```diff
@@ -175,7 +167,7 @@ bool ParseBoolFlag(const char* str, const char* flag, bool* value) {
   if (value_str == nullptr) return false;
 
   // Converts the string value to a bool.
-  *value = !(*value_str == '0' || *value_str == 'f' || *value_str == 'F');
+  *value = value_str[0] == '\0' || IsTruthyFlagValue(value_str);
```
EricWF: This now returns `true` for the empty string. Simply calling `IsTruthyFlagValue` should give you the right answer.

src/benchmark.cc
```diff
@@ -1123,8 +1135,11 @@ void ParseCommandLineFlags(int* argc, char** argv) {
                         &FLAGS_benchmark_out) ||
         ParseStringFlag(argv[i], "benchmark_out_format",
                         &FLAGS_benchmark_out_format) ||
-        ParseBoolFlag(argv[i], "color_print",
-                       &FLAGS_color_print) ||
+        ParseStringFlag(argv[i], "benchmark_color",
+                        &FLAGS_benchmark_color) ||
+        // Legacy flag for benchmark_color.
```
EricWF: Add a `TODO` or `FIXME` tag to this comment so we can grep for it.

src/colorprint.h
```diff
@@ -23,6 +24,10 @@ std::string FormatString(const char* msg, ...);
 void ColorPrintf(std::ostream& out, LogColor color, const char* fmt, va_list args);
 void ColorPrintf(std::ostream& out, LogColor color, const char* fmt, ...);
 
+// Returns true if the stream identified by 'f' appears to be a terminal that
+// supports colored output, false otherwise.
+bool IsColorTerminal(FILE* f);
```
EricWF: We only really support color output to `stdout`, not arbitrary files. Therefore I think this should be `StdoutIsColorTerminal()` or similar.

src/posix.h
EricWF: Actually since `IsATTY` and `FileNo` seem like unneeded abstractions since they only have one caller, `IsColorTerminal`. I would remove these and just call `isatty` and `fileno` directly from there.

EricWF:
Thanks again for this patch!

PS. I really liked your command line argument checking. If you want to submit that as a separate patch it would be  very welcome.

nickhutchinson:
Addressed code review comments.

- rebase on master
- make IsTruthyFlagValue() return true if given the empty string.
- remove posix.h header, folded the functions into IsColorTerminal() (which no longer accepts any arguments)
- remove GetEnv() shim, update callers to use stdlib's getenv() directly.
- add FIXME to --color_print comment.

PR #130, opened 2015-06-16
Fix the compilation error against G++ on Mac OS X.

### LINKED ISSUE #129: Compilation error happens against G++ on Mac OS X.
The compilation with g++-5 on OS X does not success with emitting the following message
because the usage of 0 as null pointer is in OS X specific part of `sysinfo.cc`,
which is illegal when `-Wzero-as-null-pointer-constant` is given to the compiler.

```
$ cd /path/to/google/benchmark
$ mkdir build
$ cd build
$ CC=gcc-5 CXX=g++-5 cmake ..
$ make
--- snip ---
/path/to/google/benchmark/src/sysinfo.cc: In function 'void benchmark::{anonymous}::InitializeSystemInfo()':
/path/to/google/benchmark/src/sysinfo.cc:270:77: error: zero as null pointer constant [-Werror=zero-as-null-pointer-constant]
   if (::sysctl(numcpus_name, arraysize(numcpus_name), &num_cpus, &size, 0, 0) ==

$ g++-5 --version
g++-5 (Homebrew gcc 5.1.0) 5.1.0
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

The possible fix is use `nullptr` instead of `0` in the argument of `sysctl`, as shown in https://github.com/umireon/benchmark/commit/72ce39de3e428decfb1d29aa33ea81e0386c4441 .
This patch works on my Yosemite environment.

umireon:
Fixes #129

Using `0` as a null pointer is illegal when `-Wzero-as-null-pointer-constant` is given to G++. 
To avoid the warning `zero-as-null-pointer-constant`, `nullptr` (C++11 keyword) instead of `0` is used in the `sysctl` invocation.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,6 +17,7 @@ Evgeny Safronov <division494@gmail.com>
 Felix Homann <linuxaudio@showlabor.de>
 Google Inc.
 JianXiong Zhou <zhoujianxiong2@gmail.com>
+Kaito Udagawa <umireon@gmail.com>
 Lei Xu <eddyxu@gmail.com>
 Matt Clarkson <mattyclarkson@gmail.com>
 Oleksandr Sochka <sasha.sochka@gmail.com>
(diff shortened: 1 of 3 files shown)


EricWF:
LGTM

umireon:
THANKS!

PR #131, opened 2015-07-20
Fix the compiling error under XCode 6 in release mode

dulaccc:
The implicit cast throws a compiler error because of loosing integer precision, under Xcode 6 in Release mode.

```txt
benchmark.cc:700:57: Implicit conversion loses integer precision: 'long long' to 'size_t' (aka 'unsigned long')
```

Cheers

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -24,3 +24,4 @@ Oleksandr Sochka <sasha.sochka@gmail.com>
 Paul Redmond <paul.redmond@gmail.com>
 Shuo Chen <chenshuo@chenshuo.com>
 Yusuke Suzuki <utatane.tea@gmail.com>
+Pierre Dulac <dulacpier@gmail.com>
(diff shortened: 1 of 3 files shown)


dulaccc:
Thanks @googlebot :)

EricWF:
Why upcast it to `int64_t` before down casting it to `size_t` (on 32 bit systems)? Your not fixing the lossy conversion, simply silencing the warning.

Wouldn't it make more sense to just remove the `int64_t` cast all together?

pleroy:
@EricWF, A more interesting question is: why is the `iterations` field a `size_t` but the `iters` variable an `int`?  There is a lot of confusion between signed and unsigned for the number of iterations (I know, the Google style guide recommendations don't really help).  This has caused trouble for me on 32-bit machines (yes, that's still a thing).

EricWF:
@pleroy The answer isn't as interesting as the question. I think these oddities are leftovers from a big overhaul I made a couple of months ago. I'll try and cleanup the types used to make them consistent.

I think the number of iterations should be represented using an unsigned type because it's important to support a very large number of iterations and there is no need for that number to be negative.

dominichamon:
Agreed. Better to clean up the types than hide the warning behind casts.

see https://github.com/google/benchmark/issues/141

PR #135, opened 2015-09-07
Fix typo: iff -> if

myd7349:


EricWF:
That's not a typo. `iff` is commonly used to mean `if and only if`. If you want to change it to `if and only if` that's fine but changing it to 'if' is wrong.

myd7349:
@EricWF Thanks! Got it.

Issue #134, opened 2015-09-06
benchmark won't build on debian wheezy


mathieuk:
I've been on a mission to get https://github.com/facebook/osquery to build on Debian Wheezy. One of the issues I ran into was building benchmark, a dependency for osquery. Turns out that the libstd++ on Debian Wheezy doesn't have ```is_trivially_destructable``` and this situation is quite tricky to work around.  I found (like, literally :)) a workaround, which basically selects ```is_trivially_destructable``` or ```has_trivial_destructor``` based on availability.

OSQuery had a separate project to provide dependencies like benchmark and that fix has been applied there. But they've understandably moved to just using vanilla builds of benchmark to prevent unnecessary forking. 

The pull request I submitted for OSQuery is here https://github.com/osquery/third-party/pull/7 . Could that be added to benchmark?

EricWF:
Urg, std::is_trivially_destructible is the cause of a lot of pain. I think we can safely just remove the static_assert in walltime.cc though. I don't think that Walltime::now() is used at all during program termination. 

I'll submit a patch that removes the static_assert.

dominichamon:
We might. If StartTimer is called by a thread during termination we might try to access it.

A better fix would be to allocate the WallTimeImp on the heap as a pointer and never free it.

mathieuk:
So, what would be the best way to move this forward?

EricWF:
This has already been fixed. Please reopen if you disagree.

PR #136, opened 2015-09-08
Remove std::is_trivially_destructible assertion. Fix issue #134.

EricWF:
See https://github.com/google/benchmark/issues/134

### DIFF
--- a/src/walltime.cc
+++ b/src/walltime.cc
@@ -24,7 +24,6 @@
 #include <atomic>
 #include <chrono>
 #include <limits>
-#include <type_traits>
 
 #include "arraysize.h"
 #include "check.h"
@@ -64,11 +63,6 @@ public:
 
   static WallTimeImp& GetWallTimeImp() {
     static WallTimeImp imp;
-#if __cplusplus >= 201103L
-    static_assert(std::is_trivially_destructible<WallTimeImp>::value,
-                  "WallTimeImp must be trivially destructible to prevent "
-                  "issues with static destruction");
-#endif
     return imp;
   }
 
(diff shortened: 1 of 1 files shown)


theopolis:
Hi @EricWF, is there a problem with the CLA signing? It would be great to have this merged into benchmark's master ASAP!

dominichamon:
this is really weird... @EricWF has committed in the past without any problems. I wonder if it's because his email changed so he has to resign the CLA.

EricWF:
Sorry I've just been insanely busy with school, work and moving I haven't had a chance to work on this. I'll commit a fix tonight.

dominichamon:
@EricWF are you planning more changes to this as suggested in #134, or is this fine to go in?

i'm concerned that we might access Now during termination as it is called from `StartTimer()` which is called by each thread. The only case this might be a problem is if the user sigkills so maybe we need to catch that and wait for `phase_condition_`?

EricWF:
So I'll commit the change using a dynamic allocation because it clearly expresses intent. However because the type is trivially destructible I don't think it actually makes a semantic difference. Originally I thought avoiding dynamic allocation would prevent WallTimeImp from being wrapped in a static guard variable. However because WallTimeImp's constructor is not constexpr it will always have a static guard around it.

On another note I'm not convinced that the termination race condition is a real issue. You would have to call StartTimer from the destructor of a static object O, and where the construction of O was not sequenced before the first call to WallTimeImp. If your writing this kind of code your already screwed.

To be even more pedantic our fix still has undefined behavior in the case described above from the point of view of the standard. AFAIK accessing even a trivially destructible type after its destructors ran is UB.

PR #138, opened 2015-09-17
Fix sample documentation: using iterations()

eliben:


eliben:
@googlebot - this is certainly strange as I'm part of the https://github.com/google organization (Googler!) and contribute to other Google projects on Github (for example https://github.com/google/yapf)

dominichamon:
The author on the commit is set to eliben@users.noreply.github.com which presumably has not signed the CLA.

Can you fix the author and repush the PR?

eliben:
@dominichamon ah, this is what happens when I edit files from inside Github. This looks like something @goolebot should handle, but oh well.

Seems like I'd have to rewrite all history to change author - so I'll just create a new PR - there's another typo needs fixing there anyways.

eliben:
See #139

PR #139, opened 2015-09-18
Fix doc sample typos in header and README

eliben:
* iterations() is a method
* int64_t fix in a couple of places


eliben:
.... aaand @googlebot doesn't like my primary email account either :-/ I tried just submitting the CLA anew. Hope this helps.

PR #140, opened 2015-09-23
Fix documentation comment in benchmark_api.h w.r.t CustomArguments

eliben:
A custom arguments function returns void.


dominichamon:
nice catch. how do you feel about adding a test for this? :)

eliben:
Added.

Not sure about the Clang build failure on Travis. It complains about a libcstd++ header and was reported on the previous version of the PR which only had a comment change, so it looks suspicious.

dominichamon:
blech. that looks awful. i wonder if it's going to bust master if we merge this.

clearly not related to your change though.

dominichamon:
looks like travis updated clang but didn't include the libstdc++ 4.6 patches. 

http://llvm.org/bugs/show_bug.cgi?id=12893

Issue #141, opened 2015-09-30
fix the types of iteration counts throughout


dominichamon:
there's some confusion between size_t and int64_t in the codebase. These should be cleaned up.

iteration counts should be size_t (unsigned and large).

EricWF:
Closing this bug.

1. It's a year old.
2. It's a duplicate of #147 (albeit older).
3. int64_t still exists but not within iteration counts.

PR #146, opened 2015-10-01
Fix issue #141. Use size_t instead of int for the iteration count

EricWF:
This patch fixes the warnings but there is still a bug where the iteration count cannot be greater that what is represented in an integer. This is because kMaxIterationCount is an int. We should fix that at some point.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -596,7 +596,7 @@ namespace {
 // Execute one thread of benchmark b for the specified number of iterations.
 // Adds the stats collected for the thread into *total.
 void RunInThread(const benchmark::internal::Benchmark::Instance* b,
-                 int iters, int thread_id,
+                 size_t iters, int thread_id,
                  ThreadStats* total) EXCLUDES(GetBenchmarkLock()) {
   State st(iters, b->has_arg1, b->arg1, b->has_arg2, b->arg2, thread_id);
   b->benchmark->Run(st);
@@ -613,7 +613,7 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
 
 void RunBenchmark(const benchmark::internal::Benchmark::Instance& b,
                   BenchmarkReporter* br) EXCLUDES(GetBenchmarkLock()) {
-  int iters = 1;
+  size_t iters = 1;
 
   std::vector<BenchmarkReporter::Run> reports;
 
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -101,7 +101,7 @@ bool IsZero(double n) {
 
 // For non-dense Range, intermediate values are powers of kRangeMultiplier.
 static const int kRangeMultiplier = 8;
-static const int kMaxIterations = 1000000000;
+static const size_t kMaxIterations = 1000000000;
 
 bool running_benchmark = false;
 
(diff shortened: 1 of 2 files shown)


pleroy:
I believe that the member `iteration` in `reporter.h` (I seem to remember that it's much more pervasive).

dominichamon:
maybe this isn't the right solution. we have `max_iters` in `State` as a `size_t`, but maybe it should be `uint64_t` to support high iteration counts on 32-bit machines. consistency is great, and i usually prefer to use `size_t` for sizes because, y'know, it's in the name. But I wonder if this will cause surprises in cross-platform benchmarks.

what do you think?

dominichamon:
```
/home/travis/build/google/benchmark/src/benchmark.cc: In function ‘void benchmark::{anonymous}::RunBenchmark(const benchmark::internal::Benchmark::Instance&, benchmark::BenchmarkReporter*)’:
/home/travis/build/google/benchmark/src/benchmark.cc:683:21: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
```

EricWF:
Yeah I see the issues. The simple fix is to make `kMaxIterationsCount` unsigned. However a max iteration count of 1 billion is plenty most of the time. With the default settings the benchmark would be allowed to run for a total of 0.5 seconds. In order to do 1 billion iterations in 0.5 seconds each iteration would take 1ns.  In the default configuration we wouldn't benefit from allowing more iterations.

I'm hesitant to start using int64_t in the interface again. API usage starts to get cluttered up with
casts when users don't use the same integer types (and they probably don't). These conversions have already caused us tons of pain. I obviously don't want to pass that on to users.

Anyway, I'm going to defer to @dominichamon. He is much wiser than I am.

pleroy:
FWIW I'd vote for `size_t` because it's going to be less surprising for clients.  For better or for worse `int64_t` is not used very often in APIs.  And anyway I don't imagine a practical need to do more than 1B iterations.

I remember running into problems on 32-bit Windows with this line in `console_reporter.cc`:
```
ColorPrintf(COLOR_CYAN, "%10lld", result.iterations);
```
It would just print out garbage.  Changing to `%10u` fixed the problem for me.

ulvgard:
Dominic said:
> "... we have max_iters in State as a size_t, but maybe it should be uint64_t to support high iteration counts on 32-bit machines. consistency is great, and i usually prefer to use size_t for sizes because, y'know, it's in the name."

I agree, but most of all I wish this could be implemented for items_processed, and bytes_processed as these areused generally. I do realize that this is outside this discussion but I wanted to bring attention to this question as size_t is frequently insufficient for me.

dominichamon:
@EricWF wiser? you mean older, right? you have a better grasp on the constraints for the API and have the stricter limitations. `size_t` is fine if it's not causing issues.

@ulvgard for items and bytes processed i can see uint64_t being the right measure. What do you think, @EricWF?

EricWF:
I was concerned about the performance of 64 bit arithmetic on a 32 bit system. Obviously I decided to benchmark it. I posted the test here https://gist.github.com/EricWF/29868ce86eaa4a03132d.
It seems like using a 64 bit iteration count on a 32 bit system can dramatically affect the speed of a benchmark. For this reason I think we shouldn't use uint64_t to represent iterations.

I think it makes sense to support uint64_t for items and bytes though.

dominichamon:
Sounds great. Something needs to change in this PR, right?

EricWF:
I don't think we should put those changes in this PR. This change is still needed to fix some warnings

ulvgard:
Separate fix for bytes and items then?

dominichamon:
yes please

PR #149, opened 2015-10-06
Android fix

anton-dirac:
This patch makes it possible to build google-benchmark with the android-ndk.
A few of the fixes are questionable, like changing std::vsnprintf to just vsnprintf and std:stoul to std::atol but I have not found any better solution. The changes are needed due to limitations in Bionic (see: http://stackoverflow.com/questions/17950814/how-to-use-stdstoul-and-stdstoull-in-android).

### DIFF
--- a/src/commandlineflags.cc
+++ b/src/commandlineflags.cc
@@ -14,10 +14,10 @@
 
 #include "commandlineflags.h"
 
+#include <cstdlib>
 #include <cstring>
 #include <iostream>
 #include <limits>
-#include <cstdlib>
 
 namespace benchmark {
 // Parses 'str' for a 32-bit signed integer.  If successful, writes
--- a/src/commandlineflags.cc
+++ b/src/commandlineflags.cc
@@ -17,6 +17,7 @@
 #include <cstring>
 #include <iostream>
 #include <limits>
+#include <cstdlib>
 
 namespace benchmark {
 // Parses 'str' for a 32-bit signed integer.  If successful, writes
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -5,7 +5,6 @@
 #include <array>
 #include <memory>
 #include <sstream>
-
 #include <stdio.h>
 
 #include "arraysize.h"
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -128,7 +128,7 @@ std::string StringPrintFImp(const char *msg, va_list args)
   // allocation guess what the size might be
   std::array<char, 256> local_buff;
   std::size_t size = local_buff.size();
-  // 8/10/2015: vsnprintf is used instead of snd::vsnprintf due to a limitation in the android-ndk
+  // 2015-10-08: vsnprintf is used instead of snd::vsnprintf due to a limitation in the android-ndk
   auto ret = vsnprintf(local_buff.data(), size, msg, args_cp);
 
   va_end(args_cp);
@@ -143,7 +143,7 @@ std::string StringPrintFImp(const char *msg, va_list args)
   // add 1 to size to account for null-byte in size cast to prevent overflow
   size = static_cast<std::size_t>(ret) + 1;
   auto buff_ptr = std::unique_ptr<char[]>(new char[size]);
-  // 8/10/2015: vsnprintf is used instead of snd::vsnprintf due to a limitation in the android-ndk
+  // 2015-10-08: vsnprintf is used instead of snd::vsnprintf due to a limitation in the android-ndk
   ret = vsnprintf(buff_ptr.get(), size, msg, args);
   return std::string(buff_ptr.get());
 }
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -6,6 +6,8 @@
 #include <memory>
 #include <sstream>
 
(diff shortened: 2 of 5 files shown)


test/filter_test.cc
```diff
@@ -75,7 +76,7 @@ int main(int argc, char* argv[]) {
 
   // Make sure we ran all of the tests
   const size_t count = test_reporter.GetCount();
-  const size_t expected = (argc == 2) ? std::stoul(argv[1]) : count;
+  const size_t expected = (argc == 2) ? std::atol(argv[1]) : count;
```
pleroy: Mixing signed and unsigned is going to cause a mess on Windows.  I don't think we should do this at a time when other PRs are trying to fix signed/unsigned confusion.

dominichamon: i agree, however this is test code so we might be a little less strict if it allows them to run on a whole new platform.

i don't see any other workaround either for android. sadly their std implementation is broken in a few ways. perhaps we could define BENCHMARK_OS_ANDROID and have custom char -> unsigned long parsing just for them.

alternatively we could bring in stringstream here to parse the number instead, though that seems like overkill.

pleroy: If `stringstream` works on Android, I vote for using that.  It's a bit more code, but at least it will be clear what happens and we won't have a signed->unsigned cast hidden in the middle of a complicated expression.

anton-dirac: `<stringstream>`  works, for example:
```cpp
int main(int argc, char* argv[]) {

  TestReporter test_reporter;
  benchmark::RunSpecifiedBenchmarks(&test_reporter);

  if (argc == 2) {
      // Make sure we ran all of the tests
      std::stringstream ss(argv[1]);
      size_t expected;
      ss >> expected;

      const size_t count = test_reporter.GetCount();
      if (count != expected) {
          std::cerr << "ERROR: Expected " << expected << " tests to be ran but only "
              << count << " completed" << std::endl;
          return -1;
      }
  }
  return 0;
}
```

dominichamon: this is a better option i think. unless @EricWF has any reason to object?

src/string_util.cc
```diff
@@ -141,7 +143,7 @@ std::string StringPrintFImp(const char *msg, va_list args)
   // add 1 to size to account for null-byte in size cast to prevent overflow
   size = static_cast<std::size_t>(ret) + 1;
   auto buff_ptr = std::unique_ptr<char[]>(new char[size]);
-  ret = std::vsnprintf(buff_ptr.get(), size, msg, args);
+  ret = vsnprintf(buff_ptr.get(), size, msg, args);
```
EricWF: Leave a comment here with the date and why it can't be qualified with "std::". This way we prevent people from breaking it.

src/commandlineflags.cc
```diff
@@ -17,6 +17,7 @@
 #include <cstring>
 #include <iostream>
 #include <limits>
+#include <cstdlib>
```
EricWF: Lets keep the sorted order please.

test/benchmark_test.cc
```diff
@@ -13,6 +13,7 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include <cstdlib>
```
EricWF: sorted order.

test/filter_test.cc
EricWF: No reason. I agree that seems like a nice solution.

src/string_util.cc
```diff
@@ -6,6 +6,8 @@
 #include <memory>
 #include <sstream>
```
dominichamon: no reason for a newline here

dominichamon:
can you poke googlebot please? just reply 'I signed it!' when you have.

anton-dirac:
I signed it! 
(Corporation DiracResearch)

src/string_util.cc
```diff
@@ -127,7 +129,8 @@ std::string StringPrintFImp(const char *msg, va_list args)
   // allocation guess what the size might be
   std::array<char, 256> local_buff;
   std::size_t size = local_buff.size();
-  auto ret = std::vsnprintf(local_buff.data(), size, msg, args_cp);
+  // 8/10/2015: vsnprintf is used instead of snd::vsnprintf due to a limitation in the android-ndk
```
dominichamon: nit: please use yyyy/mm/dd for clarity (iso 8601)

src/string_util.cc
anton-dirac: OK

src/string_util.cc
anton-dirac: Sure thing.
https://xkcd.com/1179/

src/string_util.cc
```diff
@@ -5,6 +5,7 @@
 #include <array>
 #include <memory>
 #include <sstream>
+#include <stdio.h>
```
dominichamon: cstdio

anton-dirac: I added stdio.h because it's vsnprintf that is used, not std::vsnprintf.
Should it still be cstdio?

dominichamon: good point. sorry for the noise.

PR #150, opened 2015-10-06
Visual studio warnings fix

anton-dirac:
This patch fixes and removes warnings when compiling google-benchmark with Visual Studio.
The first commit changes the warning level from Wall to W4 when MSVC is used. Wall on MSVC is really strict and gives 10000+ warnings (most of them in windows and STL headers...). 

The largest change is to the OS macros to avoid a new clash in Shlwapi.h, see babe4e9.
While fixing the warnings I also found a bug, see: 3c16f40.


src/benchmark.cc
```diff
@@ -15,7 +15,7 @@
 #include "benchmark/benchmark.h"
 #include "internal_macros.h"
 
-#ifndef OS_WINDOWS
+#ifndef BENCHMARK_OS_WINDOWS
```
dominichamon: this is a nice cleanup. we should do the others to match.

hint.

src/benchmark.cc
```diff
@@ -623,7 +623,7 @@ void RunBenchmark(const benchmark::internal::Benchmark::Instance& b,
 
   for (int i = 0; i < FLAGS_benchmark_repetitions; i++) {
     std::string mem;
-    while (true) {
+    for (;;) {
```
dominichamon: this gives a warning in msvc? oy vey.

dominichamon:
change looks good. when the CLA is signed i'll pull it in.

thanks!

src/benchmark.cc
anton-dirac: Yes, with W4 it does... (warning C4127: conditional expression is constant)
An alternative would be to disable this warning in the CMake file.

anton-dirac:
I signed it! 
(Corporation DiracResearch)

PR #151, opened 2015-10-06
Fix for Ubuntu & Clang & C++14

izaid:
On modern Ubuntu, there is an issue with certain versions of Clang (notably, but not only, Clang 3.4) and how it uses the headers of GCC 4.9. This comes up if one wants to use -std=c++14, something we've done for the DyND library (https://github.com/libdynd/libdynd). We also include Google Benchmark as a third-party library in DyND.

For reference to this bug, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=744792%E2%80%8B or http://askubuntu.com/questions/523613/upgrade-to-gcc-4-9-broke-clang

There are a couple of ways to fix this, but the easiest workaround is to be careful with how one includes `<cstddef>` or `<stddef.h>`. I removed two unneeded includes of this from Google Benchmark, and was able to make it usable again (woohoo!) on Ubuntu with Clang and C++14. Without doing this, Google Benchmark fails to compile with exactly the error in those bug reports.

Can we merge this? It's such a trivial change to get back a solid use case.

### DIFF
--- a/src/arraysize.h
+++ b/src/arraysize.h
@@ -1,8 +1,6 @@
 #ifndef BENCHMARK_ARRAYSIZE_H_
 #define BENCHMARK_ARRAYSIZE_H_
 
-#include <cstddef>
-
 #include "internal_macros.h"
 
 namespace benchmark {
--- a/test/basic_test.cc
+++ b/test/basic_test.cc
@@ -1,6 +1,4 @@
 
-#include <cstddef>
-
 #include "benchmark/benchmark_api.h"
 
 #define BASIC_BENCHMARK_TEST(x) \


izaid:
I signed it!

dominichamon:
should we be linking against libc++ instead when compiling with clang? (as per the askubuntu site)

i'm happy for this change to land as nothing breaks, but i'm curious if there's a better long-term fix.

izaid:
Thanks @dominichamon! As to your question, I have two answers:

1) As far as I can tell (and I tried), linking to libc++ did not help here.

2) I think most users of clang on Linux will ultimately link to libstd anyway, as that is what is available. So we probably should make it work for that.

EricWF:
The change to `arraysize.h` is not OK. The header uses `size_t` but never includes a header that provides it. The code may compile now but it's a lot more brittle. It's also not reasonable to restrict the use of "<cstddef>" within a project to work around a compiler bug.

@izaid  This bug is caused because clang 3.4.0 does not provide which shipped a version of stddef.h that does not provide the required max_align_t typedef. I *think* this is fixed in the 3.4.2 release. The correct and only fix to your problem is to upgrade your compiler or at very least the stddef.h header. 

@dominic I think libc++ should work. We explicitly handle this problem in ToT. However it's wrong to force a non-default standard library on users and it might cause issues. If a user wants to compile with libc++ all they have to do is configure with `-DCMAKE_CXX_FLAGS="-stdlib=libc++"`.

izaid:
@EricWF Okay, so we should try and resolve this. If the change to `arraysize.h` is not okay, how can we make Google Benchmark work around this bug?

Yes, it is a compiler issue, but unfortunately it is a compiler issue that is widespread due to it being packaged this way in Ubuntu. I'd rather we fix it so that people can use Google Benchmark on Ubuntu with Clang and C++14. If we don't fix it here, it simply makes adoption of Google Benchmark harder.

EricWF:
In this case we can fix `arraysize.h` by including `stddef.h` instead of `cstddef`. I'm fine with trying to fix it here so long as it doesn't harm this code. 

We can work around the bug here, but your not going to get very far trying to use a broken compiler.

pleroy:
I don't think that we should go out of our way to support ancient compilers (3.4.0 dates back to December 2013 if I am not mistaken).  That way lies madness.  Every change, however innocuous, would need to be tested with all ten versions of Clang between 3.4.0 and 3.7.0.  And the same would have to be done with gcc, Visual Studio, your-favorite-compiler-here.

I don't buy the "adoption" argument, it's a straw man.  If someone is serious about benchmarking, they can easily `apt-get` a recent Clang, that will take all of 10 minutes.

izaid:
Actually, it's not that easy, because the bug affects all versions of Clang on Ubuntu -- 3.5 and 3.6 as well. It is literally any Clang packaged in the default Ubuntu repository.

EricWF:
I can't reproduce with Clang 3.4.2 nor clang 3.5 nor clang 3.6. Only clang 3.4.2. I did this on Ubuntu. I don't think your correct. I work on LLVM and I use clang on Ubuntu every day. It's super easy to get a working copy.

izaid:
@EricWF What I said is what I recall not working.

In any case, we are talking about removing one `include <cstddef>`. When I made this PR, Google Benchmark worked and it was great. If you feel very strongly about that, then put it back and I will test it again and I will tell you if it is broken.

I did not find it "super easy" to make things work otherwise I would have not made this PR.

Issue #155, opened 2015-10-18
get_git_version CMake function should use GIT_EXECUTABLE variable


Zbyl:
get_git_version CMake function uses 'git' command directly, instead of
GIT_EXECUTABLE variable. This causes CMake errors while generating
project files in environments, where 'git' command is not present
in PATH.

I have a patch ready and intend to create a pull request for this.

PR #156, opened 2015-10-18
Fix get_git_version to use GIT_EXECUTABLE.

Zbyl:
get_git_version CMake function uses 'git' command directly, instead of
GIT_EXECUTABLE variable. This causes CMake errors while generating
project files in environments, where 'git' command is not present
in PATH.

This is pull request for issue #155.


dominichamon:
Thanks!

PR #157, opened 2015-10-25
add static_cast to get rid of warnings for intel compiler on linux platform

guanqun:


dominichamon:
This looks reasonable, but what is the warning?

guanqun:
Actually it's an error:

```
/sandbox/benchmark/src/sysinfo.cc(128): error #2259: non-pointer conversion from "int64_t={long}" to "double" may lose significant bits
        cpuinfo_cycles_per_second = EstimateCyclesPerSecond();
                                  ^
```

dominichamon:
wait.. double isn't 64-bit on your machine? i think maybe the compiler is being overly conservative and not checking bit-lengths before warning, but ok. it should be safe on every sane platform to do this conversion and therefore the cast is fine.

pleroy:
@dominichamon: 64-bit doubles only have 53 bits of mantissa, so integers between 2^53 and 2^63 (roughly) can be represented as int64 but not as double.  Hence the warning.

PR #166, opened 2015-11-30
Fixes #165: CustomArguments ret type in README

### LINKED ISSUE #165: Wrong return type for `CustomArguments` in README's `Apply` example
Hey,

I have just started using google benchmark and found out that there's an error on the README page:

```cpp
static benchmark::internal::Benchmark* CustomArguments(
    benchmark::internal::Benchmark* b) {
  for (int i = 0; i <= 10; ++i)
    for (int j = 32; j <= 1024*1024; j *= 8)
      b = b->ArgPair(i, j);
  return b;
}
BENCHMARK(BM_SetInsert)->Apply(CustomArguments);
```

The `Apply` expects the argument to return `void`:

```cpp
Benchmark* Benchmark::Apply(void (*custom_arguments)(Benchmark* benchmark)) {
```

I have already forked repo and prepared commit with the fix.

disconnect3d:
* adds myself to AUTHORS/CONTRIBUTORS

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,3 +27,4 @@ Shuo Chen <chenshuo@chenshuo.com>
 Yusuke Suzuki <utatane.tea@gmail.com>
 Dirac Research 
 Zbigniew Skowron <zbychs@gmail.com>
+Dominik Czarnota <dominik.b.czarnota@gmail.com>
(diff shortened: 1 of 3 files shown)


dominichamon:
thank you!

PR #171, opened 2015-12-30
Update README to mention UseRealTime for wallclock time measurements.

### LINKED ISSUE #170: Benchmarking code that uses threads and UseRealTime
I just ran into a gotcha with the library and wanted to share - IMHO this issue should be made more discoverable.

Say you implement some algorithm, and then have another implementation that uses multiple threads. Currently `Benchmark` will incorrectly measure the two against each other because it uses CPU time by default - to figure out how many iterations to run, and to report "items/second".

The solution is to set `UseRealTime`. I wonder if this should just be made more prominent in the documentation, or perhaps added as command line option - currently the default is CPU time unless `UseRealTime` is attached on a `Benchmark*`. A command-line flag can reverse this default and affect the whole program.

eliben:
Also adding a use case in the API header.

Fixes #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,10 @@ Three macros are provided for adding benchmark templates.
 #define BENCHMARK_TEMPLATE2(func, arg1, arg2)
 ```
 
-In a multithreaded test, it is guaranteed that none of the threads will start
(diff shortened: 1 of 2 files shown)


dominichamon:
thanks!

PR #176, opened 2016-02-10
Add aarch64 support to src/cycleclock.h.

enh:
From google3's base/cycleclock.


dominichamon:
LG. I'll merge when travis gets around to finishing.

PR #182, opened 2016-02-15
Add number of threads to State.

buchgr:
Having access to the thread count from within a benchmark is useful,
for when one wants to distribute a workload dynamically among the
benchmarks running in parallel e.g when using ThreadRange() or
ThreadPerCpu().


dominichamon:
It sounds reasonable. Can you add a test that uses the new parameter, both to test it and serve as an example for others?

buchgr:
@dominichamon PTAL

dominichamon:
Thanks!

PR #183, opened 2016-02-16
PauseTiming and ResumeTiming are now private.

### LINKED ISSUE #179: `state.PauseTiming()` and state.ResumeTiming() take a long time
I'm trying to benchmark a custom map type. I basically want to see how long it takes to look up random keys.

To do so, I'm generating random keys in the `state.KeepRunning()` loop. I obviously don't want the random key generation to be added to the overall time of the benchmark. To prevent this, I tried to use `state.PauseTiming()` and `state.ResumeTiming()`, but I found using those functions added 100s of nanoseconds to my benchmark -- which is much more than the map lookup itself.

Is this a problem with these functions? Or should I be generating my keys a different way?

dominichamon:
Use fixtures if you need to do significant work to prepare for the
benchmark to run.

Fixes #179.


dominichamon:
@EricWF does this seem reasonable to you? See the linked issue #179 for the background.

pleroy:
FWIW this greatly annoys me.  It means that I cannot pull from head without doing major changes to my benchmarks.  As a matter of fact, I don't see that the fixtures cover all the possible use cases.  This seems like a drastic change for something which is only mildly broken.  These functions work just fine if what you are benchmarking takes long enough.

dominichamon:
@pleroy great feedback... can you explain the use-case you have that uses this? Or is it that you don't see the impact so you're happy with the simpler API?

pleroy:
Two examples:
* https://github.com/mockingbirdnest/Principia/blob/7cf932755c9c0546be3b5f16159a43b702ef4bd4/benchmarks/%D1%87%D0%B5%D0%B1%D1%8B%D1%88%D1%91%D0%B2_series.cpp has a benchmark BM_NewhallApproximation which generates a new set of (random) input data each time through the loop.  This makes sure that the benchmark samples a broad set of input values (the performance of the algorithm is quite sensitive to its input).
* https://github.com/mockingbirdnest/Principia/blob/807fd9b78eb586138e5dafcd43a9676f23b82b50/benchmarks/embedded_explicit_runge_kutta_nystr%C3%B6m_integrator.cpp has benchmarks that do some numerical processing (e.g. SolveHarmonicOscillatorAndComputeError3D) and compute the resulting error.  When trying to optimize the code under test it is useful to evaluate the impact on the accuracy (there are many performance/accuracy tradeoffs here).

It seems to me that these use cases would be awkward/impossible to cover with fixtures.

dominichamon:
And both are already taking long enough that you can discount the overhead, or you are less interested in the absolute numbers rather than the relative numbers over time so the overhead is ignorable, i assume.

OK, sounds good. Dropping the PR.

pleroy:
Exactly, these benchmarks take minutes to run.  In some cases I ran into some slowness due to `CHECK` but after fixing that it worked for me.  ([boatcat]: I should send a PR.)

PR #186, opened 2016-02-24
Pass const State to Fixture::TearDown. Fix memory leak in fixture_test

amin-jabri:
As in Fixture::SetUp, it is desirable to pass const State to Fixture::TearDown. For instance, this could be used for resources cleaning in a thread-safe manner (perform a test against the thread_index).
Also in TearDown method in test/fixture_test.cc, there is a memory leak due: data.release() call instead of data.reset().


amin-jabri:
I signed it!

dominichamon:
Thanks!

PR #192, opened 2016-03-24
Add optional ms time unit for console reporter

NewProggie:
Some benchmarks may run a few milliseconds which makes it kind of hard to visually compare, since the currently only available nanoseconds numbers can get very large in this case. Therefore this commit adds an optional command line flag `--benchmark_time_unit` which lets the user choose between ns and ms time units for displaying the mean execution time.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -216,6 +216,13 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 }
 #endif
 
+// TimeUnit is passed to a benchmark in order to specify the order of magnitude
+// for the measured time.
+enum TimeUnit {
+  kNanosecond,
+  kMicrosecond,
+  kMillisecond
+};
 
 // State is passed to a running Benchmark and contains state for the
 // benchmark to use.
@@ -390,6 +397,9 @@ public:
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Arg(int x);
 
+  // Run this benchmark with the given time unit for the generated output report
+  Benchmark* Unit(TimeUnit unit);
+
   // Run this benchmark once for a number of values picked from the
   // range [start..limit].  (start and limit are always picked.)
   // REQUIRES: The function passed to the constructor must accept an arg1.
@@ -534,6 +544,7 @@ protected:
 // Old-style macros
 #define BENCHMARK_WITH_ARG(n, a) BENCHMARK(n)->Arg((a))
 #define BENCHMARK_WITH_ARG2(n, a1, a2) BENCHMARK(n)->ArgPair((a1), (a2))
+#define BENCHMARK_WITH_UNIT(n, t) BENCHMARK(n)->Unit((t))
 #define BENCHMARK_RANGE(n, lo, hi) BENCHMARK(n)->Range((lo), (hi))
 #define BENCHMARK_RANGE2(n, l1, h1, l2, h2) \
   BENCHMARK(n)->RangePair((l1), (h1), (l2), (h2))
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -137,6 +137,13 @@ static void BM_MultiThreaded(benchmark::State& state) {
   }
 }
 BENCHMARK(BM_MultiThreaded)->Threads(4);
+
+
+If a benchmark runs a few milliseconds it may be hard to visually compare the
+measured times, since the output data is given in nanoseconds per default. In
+order to manually set the time unit, you can specify it manually:
+
+BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 */
 
 #ifndef BENCHMARK_BENCHMARK_API_H_
--- a/include/benchmark/reporter.h
+++ b/include/benchmark/reporter.h
@@ -36,13 +36,12 @@ class BenchmarkReporter {
 
     // The number of chars in the longest benchmark name.
     size_t name_field_width;
-    // The time unit for displayed execution time.
-    std::string time_unit;
   };
 
   struct Run {
     Run() :
       iterations(1),
+      time_unit(kNanosecond),
       real_accumulated_time(0),
       cpu_accumulated_time(0),
       bytes_per_second(0),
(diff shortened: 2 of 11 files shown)


NewProggie:
I signed it!

src/console_reporter.cc
```diff
@@ -92,7 +95,9 @@ void ConsoleReporter::PrintRunData(const Run& result) {
                    " items/s");
   }
 
-  double const multiplier = 1e9; // nano second multiplier
+  double const multiplier = time_unit_ == "ns" ? 1e9 : 1e3; // nano second or
```
dominichamon: we take pains to make other values human readable. could we do the same here?

ie, instead of a flag, can we add the unit suffix to the output and automatically detect if the value should be ns, us, ms, etc?

NewProggie: Ah, interesting. Say, for example the measured values are above a certain number the unit should switch to the next greater one (from ns -> us -> ms -> s)? The unit in the header row would be removed then. I would propose a time unit range from nanoseconds up to seconds (we can safely ditch the planck time here, for sure :-) )

dominichamon: right. that's essentially what we do with the values that go into the benchmark name.

it might make at-a-glance comparison a little more awkward if a single benchmark jumps between units. i'm not sure what's best.

NewProggie: Yes, indeed.
In order to stay backwards compatible, I'd suggest to keep the default behaviour the way it is (all values in ns). If there shouldn't be another commandline flag, we could just add more rows for ms and seconds besides the already existing ns row. In this case at-a-glance comparisons would still be possible for one unit. Otherwise I could change the commandline flag (as well as the implementation) to move units directly to the output.

dominichamon: i don't think anyone is parsing the console output so just changing it should be fine.

we have csv and json that is more readily parseable, and shouldn't be affected by this change.

pleroy:
Would it make sense to specify the unit in the benchmark registration itself?  Something like:
```
BENCHMARK(BM_NewhallApproximation)->Arg(4)->Arg(8)->Arg(16)->Unit(kMillisecond);
```
I'd say that when you write a benchmark you have a reasonable idea of the time it will take.  The notion of having the reporting magically decide what order of magnitude works best worries me: in the above example it would be very inconvenient if the run for `Arg(4)` used microseconds but the run for `Arg(16)` used milliseconds.

dominichamon:
That's an interesting compromise.

I have a strong preference for not requiring configuration, so i'd refactor things to run all the runs for a benchmark arg set and then decide which unit to present for the set.

However, if @NewProggie wants to add this as an option and see how it plays out, i'd love to see it.

NewProggie:
Gathering all times for all the runs before setting a time unit and reporting it would be more work I guess. Furthermore one would not see any measured times before the last run for a benchmark suite has finished (which might take a longer time). This is unfortunate.

src/console_reporter.cc
```diff
@@ -113,4 +121,16 @@ void ConsoleReporter::PrintRunData(const Run& result) {
               result.report_label.c_str());
 }
 
+TimeUnitMultiplier ConsoleReporter::getTimeUnitAndMultiplier(TimeUnit unit) {
```
dominichamon: If i've asked a benchmark to report in ms, it should be reported in ms no matter which reporter I use.

as such, this can be moved to `BenchmarkReporter` and called from each of csv, json, and console reporters.

include/benchmark/reporter.h
```diff
@@ -84,15 +86,21 @@ class BenchmarkReporter {
...
  public:
   virtual bool ReportContext(const Context& context);
   virtual void ReportRuns(const std::vector<Run>& reports);
-protected:
+
+ protected:
   virtual void PrintRunData(const Run& report);
 
+ private:
+  TimeUnitMultiplier getTimeUnitAndMultiplier(TimeUnit unit);
```
dominichamon: consistent naming style please: GetTimeUnitAndMultiplier.

Further, it doesn't need to be a member of any class as it doesn't use state from the class. It doesn't even need to be in `include/benchmark` as it'll only be called from implementations of reporters.

include/benchmark/benchmark_api.h
```diff
@@ -390,6 +397,9 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Arg(int x);
 
+  // Run this benchmark with the given time unit for the generated output report
+  Benchmark* Unit(TimeUnit unit);
```
dominichamon: Please add an example to README.md

src/console_reporter.cc
NewProggie: This comment is unclear to me, since csv and json reporter currently don't display the time unit. Nevertheless I moved the function to `BenchmarkReporter` as you suggested.

NewProggie:
By the way: I've noticed that the code formatting is inconsistent in several areas, but didn't want to pollute  this pull request with reformatting. I could provide an appropriate clang-format configuration (basically: use Style: Google) and open another pull request for this?

src/console_reporter.cc
dominichamon: right, but they should now scale appropriately given the user has requested a particular time scale.

they should maybe have a field for the time unit added too...

test/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 # Enable the tests
 
+# Allow the source files to find headers in src/
+include_directories(${PROJECT_SOURCE_DIR}/src)
```
dominichamon: you shouldn't need this. the only extra thing here is the time unit which is already in include/benchmark.

NewProggie: Actually, yes. I was using SleepForMilliseconds in my test in order to get appropriate timings for each time unit.

dominichamon: ah right. it's a shame to break the abstraction here and have tests that seem to rely on internals. could you add a comment to the CMakeLists.txt explaining why it's necessary?

NewProggie: Yes, this is unfortunate. Maybe we should use a different approach here, such as a bad implemented fibonacci function which generates the work load and ditch the `SleepForMilliseconds ` inside the test alltogether?

dominichamon: see https://github.com/google/benchmark/pull/199

the benchmark test there uses std::this_thread::sleep_for which doesn't require pulling from src.

dominichamon:
Thank you!

NewProggie:
Thanks as well. I've learned some new things along the way ;-)

Issue #190, opened 2016-03-21
HTML based chart reporter


LepelTsmok:
It would be nice, to have a reporter, which plots the results into a graph. This graph should be viewed with any browser, without the need of an Internet connection. 

I have a preliminary implementation using [HighCharts](http://www.highcharts.com/). I would be willing to work on getting it ready for a merge if there is interest.

dominichamon:
How would it work? output an HTML file (+JS/CSS) in some temporary folder that someone can use with `python -m SimpleHTTPServer`?

LepelTsmok:
I thought of something simpler, like in my [Result.zip](https://github.com/google/benchmark/files/184136/Result.zip). It's a single file embedding the data, javascript, and css. So there is no need for an HTTPServer.

dominichamon:
oh, neat.

if you'll allow me to bikeshed: having a chart per benchmark would be great... would the x/y ranges be shown along the x-axis? how about threaded benchmarks?

i think this is a really neat idea and i'd love to see more examples or thoughts about how the charts could be displayed.

LepelTsmok:
Another possible chart could be like this [Result.zip](https://github.com/google/benchmark/files/185711/Result.zip). The x-axis displays the 'range_x' value, which was passed to the benchmark. Sadly, there is no good way to get these value.
I can think of three possible ways:

1. You could change the structure `Run` to contain these values
2. You could extract the value from the name, but at some point the value changes to a human readable string instead of a number (e.g. 64k)
3. You could set the value via the label and extract it form there - this is what I did

What solution would you prefer?

dominichamon:
are you generating this from the existing output?

benchmark already supports multiple reporter types.. you could try reading the json version, for example.

alternatively, you could add an HTML reporter that has full access to the raw data as it is generated and can write out whatever you need.

LepelTsmok:
I used the json output as a basis for a new HTML reporter. But I wanted to generate the HTML output directly, without an additional program. 

I wrote an HTML reporter, which generated the examples I showed you. When the reporter runs, the range_x variable isn't available any more. So actually I don't have full access to the raw data.

dominichamon:
well... that can change. it's a matter of passing `b->arg1` and `b->arg2` to the `BenchmarkReporter::Run` instance created for a given run.

that might actually be a better way of managing the reporting anyway, instead of changing the benchmark name.

LepelTsmok:
[...]

PR #193, opened 2016-04-01
Add HTML based chart reporter (Issue #190)

LepelTsmok:
This implements issue #190. The work was mentored by @mattkretz.


LepelTsmok:
I just red the CLA and noticed, that the .js files are not my work. Can those be part of the pull request at all? If not, I would have to implement a download mechanism.

LepelTsmok:
I signed it!

dominichamon:
yes, please. i'd like to avoid having inline javascript. calls out to cdns from the html reporter should be fine.

LepelTsmok:
All fixes are done with this commit. I hope everything is okay.

CMakeLists.txt
```diff
@@ -121,6 +122,45 @@ find_package(Threads REQUIRED)
...
 include_directories(${PROJECT_SOURCE_DIR}/include)
 
+# Set up pathes to javascript
+
+if (BENCHMARK_DOWNLOAD_JS)
+    set(JQUERY_NAME "jquery-2.2.1.min.js")
+    set(HIGHSTOCK_NAME "highstock.js")
+    set(HIGHSTOCK_MORE_NAME "highcharts-more.js")
+
+    file(DOWNLOAD "https://code.jquery.com/jquery-2.2.2.min.js" "${PROJECT_BINARY_DIR}/include/benchmark/${JQUERY_NAME}" SHOW_PROGRESS STATUS state)
```
dominichamon: this isn't the approach i would take.

can the generated HTML reference the CDN from which these can be sourced? I understand there are benefits to having them locally, but this will do odd things to the installed package, which may be unexpected.

include/benchmark/reporter.h
```diff
@@ -90,33 +111,69 @@ class ConsoleReporter : public BenchmarkReporter {
...
+    int range_x;
+  };
+
+  struct BenchmarkData {
+    std::string name;
+    std::vector<RunData> runData;
+  };
+
+ private:
+  void writeFile(const char* file) const;
```
dominichamon: capital initial letters please.

include/benchmark/reporter.h
```diff
@@ -90,33 +111,69 @@ class ConsoleReporter : public BenchmarkReporter {
...
+  std::string replaceHTMLSpecialChars(const std::string& label) const;
+
+  void printHTML(std::ostream& out, const std::string& html) const;
+
+  void appendRunDataTo(std::vector<BenchmarkData> *container, const Run &data, bool isStddev) const;
+
+  std::vector<BenchmarkData> benchmarkTests_Line;
```
dominichamon: no underscores in variable names please. 

https://google.github.io/styleguide/cppguide.html

src/benchmark.cc
```diff
@@ -251,11 +250,40 @@ static std::unique_ptr<TimerManager> timer_manager = nullptr;
 
 } // end namespace
 
+std::string generateInstanceName(const std::string& name, int arg_count,
```
dominichamon: capital initial please

src/html_reporter.cc
```diff
@@ -0,0 +1,507 @@
...
+    "            });\n"
+    "       </script>\n"
+    "   </head>\n"
+    "   <body>\n"
+    "${DIV}"
+    "   </body>\n"
+    "</html>";
+
+namespace benchmark {
+std::string &replaceString(std::string &input, const std::string &from,
```
dominichamon: consistency with the rest of the code means keeping the & with the type, not the name.

src/html_reporter.cc
```diff
@@ -0,0 +1,507 @@
...
+                           const std::string &to) {
+  size_t position = input.find(from);
+
+  if (position != std::string::npos) {
+    input.replace(position, from.size(), to);
+  }
+
+  return input;
```
dominichamon: why return input and replace inline? either take input as const & and take a copy, or just do the inline replacement. however, if you do the inline replacement, input should be a *.

https://google.github.io/styleguide/cppguide.html#Reference_Arguments

src/html_reporter.cc
```diff
@@ -0,0 +1,507 @@
...
+
+    categories.clear();
+    series.clear();
+    chart.assign(highChart_Bar_Function);
+    div.assign(div_Element);
+  }
+}
+
+bool HTMLReporter::ReportContext(const Context &context) {
+  std::cerr << "Run on (" << context.num_cpus << " X " << context.mhz_per_cpu
```
dominichamon: maybe these should be in the HTML too...

include/benchmark/reporter.h
```diff
@@ -90,33 +111,69 @@ class ConsoleReporter : public BenchmarkReporter {
...
 
-} // end namespace benchmark
-#endif // BENCHMARK_REPORTER_H_
+class HTMLReporter : public BenchmarkReporter {
+ public:
+  virtual bool ReportContext(const Context& context);
+  virtual void ReportRuns(const std::vector<Run>& reports);
+  virtual void Finalize();
+
+  struct RunData {
```
dominichamon: do these structs need to be public?

src/html_reporter.cc
```diff
@@ -0,0 +1,507 @@
...
+  outputAllLineCharts(output, charts, benchmarkTests_Line,
+                      benchmarkTests_Line_stddev);
+  outputAllBarCharts(output, charts, benchmarkTests_Bar,
+                     benchmarkTests_Bar_stddev);
+
+  printHTML(std::cout,
+            replaceString(replaceString(output, "${CHART}", ""), "${DIV}", ""));
+}
+
+void HTMLReporter::writeFile(const char *file) const {
```
dominichamon: any reason why file isn't a std::string?

src/string_util.h
```diff
@@ -39,6 +39,11 @@ inline std::string StrCat(Args&&... args)
 void ReplaceAll(std::string* str, const std::string& from,
                 const std::string& to);
 
-} // end namespace benchmark
+std::string generateInstanceName(const std::string& name, int arg_count,
```
dominichamon: this isn't really a string util, it should be local to the benchmark code.

it could even be an instance method, i think...

CMakeLists.txt
LepelTsmok: [Line 14](https://github.com/LepelTsmok/benchmark/blob/master/CMakeLists.txt#L14) makes the download optional. The default is off, which should be what you want.

LepelTsmok:
Could you please restart Travis? It seems to fail in the STL.

dominichamon:
as you change the code, it should rerun anyway.

LepelTsmok:
Is everything okay now?

include/benchmark/benchmark_api.h
```diff
@@ -188,6 +189,11 @@ void UseCharPointer(char const volatile*);
 // registered benchmark.
 Benchmark* RegisterBenchmarkInternal(Benchmark*);
 
+
+std::string GenerateInstanceName(const std::string& name, int arg_count,
```
dominichamon: can you find somewhere to declare this that isn't in the public api header?

src/html_reporter.cc
```diff
@@ -0,0 +1,520 @@
...
+    "       </script>\n"
+    "   </head>\n"
+    "   <body>\n"
+    "${DIV}"
+    "${CONTEXT}"
+    "   </body>\n"
+    "</html>";
+
+namespace benchmark {
+void replaceString(std::string *input, const std::string &from,
```
dominichamon: consider using an unnamed inner namespace here to avoid these methods/structs being used by other translation units.

src/html_reporter.cc
```diff
@@ -0,0 +1,520 @@
...
+
+  if (position != std::string::npos) {
+    input->replace(position, from.size(), to);
+  }
+}
+
+template <typename T>
+std::function<std::string(const HTMLReporter::RunData &,
+                          const HTMLReporter::RunData &)>
+generateErrorbarCallable(T HTMLReporter::RunData::*member) {
```
dominichamon: update these methods to have initial capital letters please.

src/benchmark.cc
```diff
@@ -66,13 +66,12 @@ DEFINE_int32(benchmark_repetitions, 1,
 
 DEFINE_string(benchmark_format, "tabular",
               "The format to use for console output. Valid values are "
-              "'tabular', 'json', or 'csv'.");
+              "'tabular', 'json', 'csv', or 'html'.");
```
dominichamon: make sure this is reflected in the 'usage' printout.

dominichamon:
when i run this, there's no output for ages and then:

$ test/benchmark_test --benchmark_format=html
<waiting waiting...>
3D plotting is not available at the moment!

and that's it.

no idea where the HTML files are or how to see them. maybe some output would be useful?

LepelTsmok:
Now it is better! So I hope, that it is okay now.

include/benchmark/reporter.h
dominichamon: can you see if these can be made private? i think they can as they're not used externally to the class.

src/html_reporter.cc
```diff
@@ -0,0 +1,521 @@
...
+  };
+}
+
+struct ChartIt {
+  std::string (*value)(const HTMLReporter::RunData &);
+  std::function<std::string(const HTMLReporter::RunData &,
+                            const HTMLReporter::RunData &)> error;
+};
+
+void outputAllLineCharts(
```
dominichamon: method still not named to style requirements.

src/html_reporter.cc
```diff
@@ -0,0 +1,521 @@
...
+    ReplaceString(&output, "${CHART}", chart);
+    ReplaceString(&output, "${DIV}", div_element);
+    ReplaceString(&output, "${DIV_NAME}", benchmark_id[chart_num]);
+
+    series.clear();
+    chart.assign(high_chart_line_function);
+  }
+}
+
+void outputAllBarCharts(
```
dominichamon: method still not named to style requirements.

src/html_reporter.cc
```diff
@@ -0,0 +1,521 @@
...
+                     benchmark_tests_bar_stddev);
+
+  ReplaceString(&output, "${CONTEXT}", context_output);
+  ReplaceString(&output, "${CHART}", "");
+  ReplaceString(&output, "${DIV}", "");
+
+  printHTML(std::cout, output);
+}
+
+void HTMLReporter::writeFile(const std::string &file) const {
```
dominichamon: method still not named to style requirements.

src/html_reporter.cc
```diff
@@ -0,0 +1,521 @@
...
+    do {
+      fin.read(buffer, 256);
+      std::cout.write(buffer, fin.gcount());
+    } while (fin.gcount() > 0);
+
+    fin.close();
+  }
+}
+
+std::string HTMLReporter::replaceHTMLSpecialChars(
```
dominichamon: method still not named to style requirements.

src/html_reporter.cc
```diff
@@ -0,0 +1,521 @@
...
+        new_label.insert(pos, 1, '\\');
+        pos++;
+        break;
+    }
+  }
+
+  return new_label;
+}
+
+void HTMLReporter::printHTML(std::ostream &out, const std::string &html) const {
```
dominichamon: method still not named to style requirements.

src/html_reporter.cc
```diff
@@ -0,0 +1,521 @@
...
+  if (marker_pos != std::string::npos) {
+    out.write((html.c_str() + string_pos), (marker_pos - string_pos));
+    writeFile("@CMAKE_HIGHSTOCK_MORE_PATH@");
+    string_pos = (marker_pos + 17);
+  }
+
+  out << (html.c_str() + string_pos);
+}
+
+void HTMLReporter::appendRunDataTo(std::vector<BenchmarkData> *container,
```
dominichamon: method still not named to style requirements.

src/string_util.cc
```diff
@@ -12,6 +13,10 @@
 namespace benchmark {
 namespace {
 
+bool IsZero(double n) {
+    return std::abs(n) < std::numeric_limits<double>::epsilon();
```
dominichamon: indent 2 please

src/string_util.h
```diff
@@ -39,6 +39,10 @@ inline std::string StrCat(Args&&... args)
 void ReplaceAll(std::string* str, const std::string& from,
                 const std::string& to);
 
-} // end namespace benchmark
+std::string GenerateInstanceName(const std::string& name, int arg_count,
```
dominichamon: sorry, this doesn't fit here either. a new header would be reasonable in this case, i think. a private benchmark_util.h maybe?

LepelTsmok:
The 'IsZero' function is now in benchmark_util.cc.
I hope this is okay.

dominichamon:
it'd be nice to have some output other than warnings while the benchmarks are running (ie, when the reporter gets a request to add a report) so the user knows something is happening.

now i have output (yay!) i have some questions.

i'm surprised that each benchmark is on the same graph. i expected a section for each benchmark name, with a graph for each metric.

it doesn't really make sense to combine the graphs for all the benchmarks given the wide range of times and arguments, and the different meanings of the arguments, across benchmarks.

mattkretz:
> it'd be nice to have some output other than warnings while the benchmarks are running

I agree. However, if this is an issue, don't the JSON and CSV reporters have the same issue? (Assuming the use in all cases is via redirection to file.)

My opinion: do one of

1. Create a general progress report output on stderr
1. Have json, csv, and html write to a specified file and do progress reporting on stdout via the console reporter

Both solutions seem like a separate issue to solve.

> it doesn't really make sense to combine the graphs for all the benchmarks given the wide range of times and arguments, and the different meanings of the arguments, across benchmarks.

This seems hard / impossible to solve in a general way. The best presentation depends a lot on the benchmark itself. Consequently, one would need many configuration options to influence the way the data is presented. The motivation behind the current HTML reporter was to have a quick and easy path to *some* kind of charting. If you want fully flexibility use CSV or JSON and a dedicated charting program. This extra step is too much for the *quick and easy* charting use case.

What is really handy is to use the HTML reporter together with the regex filter to generate separate charts of the results you want to compare.

That said, there sure may be a better *one size fits all* default than the current one.

* Given e.g. three benchmarks without x/y ranges, should those really have their own graph? They would each have a single bar, no?
* So it seems the HTML reporter must combine multiple benchmarks with different names into the same chart. Can you think of any discriminator that the reporter could use to split benchmarks into different charts? Currently, it only splits via existence of an x range into line chart vs. bar chart.

dominichamon:
> > it'd be nice to have some output other than warnings while the benchmarks are running

> I agree. However, if this is an issue, don't the JSON and CSV reporters have the same issue? (Assuming the use in all cases is via redirection to file.)

They do not, precisely because they output to stdout.

> My opinion: do one of
> 
> Create a general progress report output on stderr
> Have json, csv, and html write to a specified file and do progress reporting on stdout via the console reporter
> Both solutions seem like a separate issue to solve.

HTML should probably output to stdout and then need to be redirected, if we're following the pattern of the JSON and CSV reporters.
 
> > it doesn't really make sense to combine the graphs for all the benchmarks given the wide range of times and arguments, and the different meanings of the arguments, across benchmarks.

> This seems hard / impossible to solve in a general way. The best presentation depends a lot on the benchmark itself. Consequently, one would need many configuration options to influence the way the data is presented. The motivation behind the current HTML reporter was to have a quick and easy path to some kind of charting. If you want fully flexibility use CSV or JSON and a dedicated charting program. This extra step is too much for the quick and easy charting use case.

is it a significant change to append a 'set' of charts for every benchmark name encountered? if you're building up the HTML as the runs are sent to the reporter it should be fairly easy to start a new set of charts whenever the name changes. Unless your templates assume one set of charts, which is unfortunate.

> What is really handy is to use the HTML reporter together with the regex filter to generate separate charts of the results you want to compare.

That's a good use-case, but out of the box not obvious. I'd like users who try it to get something less surprising when they see it.

> 
> That said, there sure may be a better one size fits all default than the current one.
> 
> Given e.g. three benchmarks without x/y ranges, should those really have their own graph? They would each have a single bar, no?

yes. and it might be that the HTML reporter isn't a good fit for the single benchmark case. having a single bar is fine as it still shows the magnitude. i think the real power of the HTML reporter is comparing across x/y ranges or threads.

> So it seems the HTML reporter must combine multiple benchmarks with different names into the same chart. Can you think of any discriminator that the reporter could use to split benchmarks into different charts? Currently, it only splits via existence of an x range into line chart vs. bar chart.

as above, per benchmark name.

WilliamTambellini:
Hi, Any plan to rebase and merge ?

mattkretz:
Hi @WilliamTambellini I'd love to get this merged. Currently, my priorities don't leave room for finalizing the PR. I'd be happy to lend a helping hand for anyone willing to see this through.

WilliamTambellini:
Hi @mattkretz I cannot help on that. 
Are we sure that it would be even accepted/merged if someone rebase/finish the PR ?

dominichamon:
I think this is much better solved by using JSON as an intermediate format and building charts around that (with some processing). This should then be a separate tool to the core library.

Issue #198, opened 2016-04-29
Support for benchmarking externally timed events, such as GPU code


jknuuttila:
Hello!

I would be interested to use Google Benchmark to benchmark GPU shaders (e.g. Direct3D or Vulkan draw calls or compute dispatches). Google Benchmark seems to provide a lot of extremely useful functionality that I would not want to roll on my own, such as definition of benchmarks, benchmarking with different arguments, handling iteration counts for statistical robustness, etc.

GPU shader executions are measured using timestamp queries, for which the results are provided asynchronously. The CPU side code sets up the shaders along with the timestamp queries, waits until the GPU is done, and then reads the timestamp results after. The CPU time spent doing all this can significantly overestimate the actual duration of interest (it is at least as much as the actual GPU execution, plus many forms of various overhead), and is usually of no interest. As such, Google Benchmark's default way of measuring time is not useful.

What I would require is a way for benchmark iterations to manually override the timing and be able to tell Google Benchmark that "this benchmark iteration took exactly 3.141 ms", regardless of how much CPU time was actually spent. A hypothetical API for this could look like the following:

``` c++
static void BM_GPUBenchmark(benchmark::State& state) {
  // Disable default time measurement
  state.ManualTiming();
  while (state.KeepRunning()) {
    // Run the actual benchmarked code, wait until it's done,
    // and measure the time spent.
    double seconds = ExecuteAndMeasureShader(state.range_x());
    // Report the actual time spent to Google Benchmark.
    state.SetIterationTiming(seconds);
  }
}
BENCHMARK(BM_GPUBenchmark)
    ->Arg(1)
    ->Arg(2)
    ->Arg(4)
    ->Arg(8);
```

Other possible uses for this could include e.g. benchmarking network latency or throughput using Google Benchmark.

It looks like one possible way to implement this could be to add a new `double manual_time_used_` member variable to `TimerManager`, add a flag to `State` or `TimerManager` to note when manual time is enabled, and then change `RunBenchmark` accordingly to use the `manual_time_used_` as the value of `seconds` when the manual timing mode is set.

If I was to implement this feature, would you be interested in accepting the pull request? If so, I would appreciate feedback on the proposed API.

Regards,
Jussi

romanlarionov:
I second this. It would be cool if I could test compute kernels for either OpenCL or CUDA with this.

dominichamon:
[...]

PR #199, opened 2016-04-30
Support for manual timing (see issue #198, https://github.com/google/benchmark/issues/198)

jknuuttila:
I have added support for manual timing, for benchmarking cases where CPU time and wallclock time are not useful. It is enabled by calling `UseManualTime` similar to how `UseRealTime` is used. When it is enabled, the benchmark loop must call `SetIterationTime(seconds)` once per each iteration. Manually timed benchmarks are decorated with "manual_time" when outputting results, and byte and item throughput are calculated with the manual time instead of the wallclock time.

I have added a test for manual timing in the `benchmark_test.cc` file. The test simulates a workload with `std::this_thread::sleep`, and measures the duration of the sleep manually using `std::chrono::high_resolution_clock::now`. The manually measured time is reported back with `SetIterationTime`. The benchmark is run with both `UseRealTime` and `UseManualTime` to see that the timings roughly match, which is expected in this case.

I have also added a short piece of documentation for the feature in the main `README.md` file.

I have also added one `static_cast` to replace an implicit cast which triggered a warning with Visual Studio 2015.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -283,6 +283,17 @@ public:
   // within each benchmark iteration, if possible.
   void ResumeTiming();
 
+  // REQUIRES: called exactly once per iteration of the KeepRunning loop.
+  // Set the manually measured time for this benchmark iteration, which
+  // is used instead of automatically measured time if UseManualTime() was
+  // specified.
+  //
+  // For threaded benchmarks the SetIterationTime() function acts
+  // like a barrier.  I.e., the ith call by a particular thread to this
+  // function will block until all threads have made their ith call.
+  // The time will be set by the last thread to call this function.
+  void SetIterationTime(double seconds);
+
   // Set the number of bytes processed by the current benchmark
   // execution.  This routine is typically called once at the end of a
   // throughput oriented benchmark.  If this routine is called with a
@@ -443,6 +454,13 @@ public:
   // run, and in the printing of items/second or MB/seconds values.  If not
   // called, the cpu time used by the benchmark will be used.
   Benchmark* UseRealTime();
+
+  // If a benchmark must measure time manually (e.g. if GPU execution time is being
+  // measured), call this method. If called, each benchmark iteration should call
+  // SetIterationTime(seconds) to report the measured time, which will be used
+  // to control how many iterations are run, and in the printing of items/second
+  // or MB/second values.
+  Benchmark* UseManualTime();
 
   // Support for running multiple copies of the same benchmark concurrently
   // in multiple threads.  This may be useful when measuring the scaling
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -660,7 +660,7 @@ void RunBenchmark(const benchmark::internal::Benchmark::Instance& b,
             thread.join();
         }
         for (std::size_t ti = 0; ti < pool.size(); ++ti) {
-            pool[ti] = std::thread(&RunInThread, &b, iters, ti, &total);
+            pool[ti] = std::thread(&RunInThread, &b, iters, static_cast<int>(ti), &total);
         }
       } else {
         // Run directly in this thread
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -130,6 +130,7 @@ class TimerManager {
         running_(false),
         real_time_used_(0),
         cpu_time_used_(0),
+        manual_time_used_(0),
         num_finalized_(0),
         phase_number_(0),
         entered_(0) {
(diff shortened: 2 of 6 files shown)


src/benchmark.cc
```diff
@@ -647,7 +688,7 @@ void RunBenchmark(const benchmark::internal::Benchmark::Instance& b,
             thread.join();
         }
         for (std::size_t ti = 0; ti < pool.size(); ++ti) {
-            pool[ti] = std::thread(&RunInThread, &b, iters, ti, &total);
+            pool[ti] = std::thread(&RunInThread, &b, iters, static_cast<int>(ti), &total);
```
dominichamon: nice catch. a shame we can't support more than 2 billion threads though :P

src/benchmark.cc
```diff
@@ -699,7 +743,11 @@ void RunBenchmark(const benchmark::internal::Benchmark::Instance& b,
         report.report_label = label;
         // Report the total iterations across all threads.
         report.iterations = static_cast<int64_t>(iters) * b.threads;
-        report.real_accumulated_time = real_accumulated_time;
+        if (b.use_manual_time) {
```
dominichamon: do you think it's worth separating out manual accumulated time from real accumulated time? both could be set on a benchmark, and we already have both CPU and real accumulated time.

jknuuttila: I'm not sure. Usually, for my use cases (GPU workloads), the real time doesn't make sense when manual time is required, and I would be interested in having the throughput measured against the manual time.

However, it's possible that there are cases where it could be useful to have both. The console output might get a little cluttered in tabular mode, but I guess that's not a big problem.

dominichamon: I like to minimise surprises, and i think that if someone calls `->UseRealTime()->UseManualTime()` on a benchmark it should do the right thing (show real time and manual time).

yes, it might be cluttered in console mode but it's already fairly heavy-weight and the user has requested the extra data.

jknuuttila: I can make a new version of the pull request that supports this.

jknuuttila:
I have implemented support for separate tracking of manual and real times.

However, this patch turned out to be very invasive and hairy, because it affects output formatting and stat computation significantly, and as a result it is very possible that bugs may have been introduced. Indeed, it seems that some automatic check failed (which I noticed when writing this comment), so there is certainly some bug.

After implementing this and seeing how many changes were required, I would now personally strongly prefer the original simpler approach, which is to just use the manual time instead of real-time (requiring no modifications to output and stat computation) when `UseManualTime` has been specified. Enabling both real time and manual time simultaneously seems like a very niche use case to begin with, and I don't believe it is worth the additional code complexity.

I would suggest handling the combination of `UseManualTime` and `UseRealTime` either with an error message (i.e. the combination is forbidden), or by executing the benchmark once in both modes and reporting them completely separately. If you agree, I can implement this functionality and submit it as a separate pull request. If you would still prefer tracking both separately, I can fix the bugs remaining in this pull request.

jknuuttila:
It seems that I also made some mistake while rebasing the pull request, and as a result commits of other people incorrectly show up as parts of this pull request. Perhaps it would be easiest if I just made a new pull request after it is decided which approach is the best one.

dominichamon:
alright.. start with the simpler approach (maybe you can just revert your local branch to that version, and avoid creating a new PR?) and then i may take a crack at extending it.

jknuuttila:
Sure! I'll try to revert the branch.

jknuuttila:
I updated the PR again with the simpler version, and this time I hopefully did the rebasing better, too.

I added a couple of CHECKs to warn against using real time and manual time simultaneously.

include/benchmark/benchmark_api.h
```diff
@@ -444,6 +455,14 @@ class Benchmark {
   // called, the cpu time used by the benchmark will be used.
   Benchmark* UseRealTime();
 
+  // If a benchmark must measure time manually (e.g. if GPU execution time is being
+  // measured), call this method. If called, each benchmark iteration should call
+  // SetIterationTime(seconds) to report the measured time, which will be used
+  // to control how many iterations are run, and in the printing of items/second
+  // or MB/second values. If not called, either the CPU time or the elapsed time
+  // will be used (if UseRealTime() was called).
```
dominichamon: this would be an error now.

test/benchmark_test.cc
```diff
@@ -174,5 +176,30 @@ static void BM_ParallelMemset(benchmark::State& state) {
...
+    auto end     = std::chrono::high_resolution_clock::now();
+
+    auto elapsed =
+      std::chrono::duration_cast<std::chrono::duration<double>>(
+        end - start);
+
+    state.SetIterationTime(elapsed.count());
+    slept_for += microseconds;
+  }
+  state.SetItemsProcessed(slept_for);
```
dominichamon: SetLabel might be a better method to use to report this. or just skip it .. it doesn't add anything to the example that i can see.

jknuuttila: You mean the `SetItemsProcessed`? I put it there to test that the item throughput is calculated according to the manual time, as expected. I can remove it if you think it's a problem.

dominichamon: ah, i see. it's fine :)

include/benchmark/benchmark_api.h
dominichamon: update this comment, and i think we're good to go.

jknuuttila:
The comment should now be fixed.

PR #203, opened 2016-05-05
Benchmarking of mpi codes and exports package cmake style

mdavezac:
For the most part, the api is unchanged.

- if BENCHMARKING_ENABLE_MPI is True (defaults to FALSE) then the benchmarks are compiled with MPI
- BENCHMARK_MAIN calls mpi init and mpi finalize
- whenever a time is needed, the max across all processes is used
- as far as the benchmark is concerned, mpi calls only occur before or after calling the function to benchmark (e.g. not in KeepRunning)
- reporters on non-root nodes do not print
- benchmarks currently works with MPI_COMM_WORLD only. This is about benchmarking parallel codes not parallelizing benchmarks.

This pull-request also includes a cmake export mechanism:

- to avoid name collision, the benchmark target is called google.benchmark
- google.benchmark exposes its dependencies on MPI_LIBRARIES and threading
- google.benchmark keeps *private* its dependency on Shlwapi. I'm not a Windows user, so this may not be correct. Just following the previous code here.
- google.benchmark exposes the benchmark include directory
- BENCHMARK_ENABLE_MPI is exported


include/benchmark/benchmark_api.h
```diff
@@ -154,6 +154,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <stdint.h>
 
 #include "macros.h"
+#include "benchmark_mpi.h"
```
pleroy: This include is making `benchmark_mpi.h` part of the benchmark API.  That's a bad idea, it's leaking implementation details.  You should call `initialize_mpi` at the beginning of `Initialize` and `finalize_mpi` at the end of `RunSpecifiedBenchmarks`.

src/benchmark_mpi.in.h
```diff
@@ -0,0 +1,55 @@
+#ifndef BENCHMARK_MPI_H
+#define BENCHMARK_MPI_H
+
+#cmakedefine BENCHMARK_ENABLE_MPI
```
pleroy: This file in general and this line in particular are making it impossible to compile the project without using cmake.  That's not acceptable: the code should be self-standing and compilable with any modern C++ compiler, at least in the case where mpi is not being used.

src/console_reporter.cc
```diff
@@ -25,10 +25,13 @@
 #include "colorprint.h"
 #include "string_util.h"
 #include "walltime.h"
+#include "benchmark_mpi.h"
 
 namespace benchmark {
 
 bool ConsoleReporter::ReportContext(const Context& context) {
+  if(not mpi_is_world_root())
```
pleroy: This PR has many style issues, I'll let @dominichamon comment on them, but this statement is missing a space before `(` and the code in this project uses `!` not `not`.

include/benchmark/benchmark_api.h
mdavezac: That would make it difficult to create a custom main where MPI_Initialize (MPI_Finalize) are called by outside libraries, whenever those libraries do not separate mpi initialization from other initializations. For instance, it would be difficult to benchmark [Trilinos](https://trilinos.org/).

dominichamon:
this is a very intrusive change. it's the sort of change i'd like to support without any changes to the implementation. for example, can we define a custom main in whatever binary you link the benchmarks in? similarly, can you use a fixture or a custom reporter to calculate the world max time for reporting?

mdavezac:
for example, can we define a custom main in whatever binary you link the benchmarks in

You mean MPI users would have to define their own main calling `::benchmark::initialize` and `::benchmark::RunspecifiedBenchmarks`? That would be easy to do. The con is that it changes the API for MPI users with respect to non-MPI users.

    Can you use a fixture or a custom reporter to calculate the world max time for reporting

Not so sure about this. The times are used to compute the next number of iterations. This has to match across all processes (more explicitly, `state.max_iterations` has to be the same across all processes) if the processes are doing any sort of work in concert. Even adding a barrier before calling `KeepRunning` does not necessarily ensure all the times are the same and hence that the number of iterations is rigorously the same. One could add just sync the number of iterations, but that would lead to less accurate timings (not sure max and mean are commutative). Another option is to use #199, if that is going anywhere. But, frankly, not having to understand the intricacies of timing is one reason I'm using a benchmarking library in the first place.

src/benchmark_mpi.in.h
mdavezac: Okay. I'll make changes in that direction.

src/console_reporter.cc
mdavezac: Okay

mdavezac:
Since #199 has been merged, it is now possible to support benchmarking of mpi code without adding any mpi to benchmark itself.
All that would be required is the ability to use a null reporter.

Would that be a preferred route?

dominichamon:
that would be fantastic. if not a null reporter, being able to register your own reporter would be even better :)

mdavezac:
It looks like it is already possible to use your own reporter by passing it to RunSpecifiedBenchmarks in a custom main. So everything is already there!

dominichamon:
wonderful!

if you could write up something somewhere about how to benchmark MPI, i'd be happy to include it in the documentation here, or point to it if it's hosted elsewhere.

mdavezac:
@dominichamon, here is a [gist](https://gist.github.com/mdavezac/eb16de7e8fc08e522ff0d420516094f5) with a noddy MPI benchmark. I'm happy to reformat it into the README and do a pull request, but it seems a bit of a long example.

dominichamon:
I think it'd be a fine addition as README.mpi.

PR #212, opened 2016-05-14
added option to change range multiplier

ismaelJimenez:
Added option to change range multiplier to a number bigger than two, because for some benchmarks a range multiplier different than 8 is more suitable.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -287,6 +287,7 @@ struct Benchmark::Instance {
   bool           has_arg2;
   int            arg2;
   TimeUnit       time_unit;
+  int            range_multiplier;
   bool           use_real_time;
   bool           use_manual_time;
   double         min_time;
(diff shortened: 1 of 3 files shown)


src/benchmark.cc
```diff
@@ -500,6 +504,11 @@ void BenchmarkImp::RangePair(int lo1, int hi1, int lo2, int hi2) {
   }
 }
 
+void BenchmarkImp::RangeMultiplier(int multiplier) {
+  CHECK_GE(multiplier, 2);
```
dominichamon: it might make sense to add this check to `AddRange` instead where we already check the limits.

ismaelJimenez: yes, that's right. I have already moved the assert to the AddRange function

dominichamon:
wonderful! thank you!

PR #214, opened 2016-05-17
Use a separate template for DoNotOptimize with pointer arguments (gcc…

zabereer:
… only)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@
 #
 # Please keep the list sorted.
 
+Albert Pretorius <pretoalb@gmail.com>
 Arne Beer <arne@twobeer.de>
 Christopher Seymour <chris.j.seymour@hotmail.com>
 David Coeurjolly <david.coeurjolly@liris.cnrs.fr>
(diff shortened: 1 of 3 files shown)


zabereer:
I have a CLA for pretoalb@gmail.com.
Many thanks.

EricWF:
Just using "+m" massively changes the codegen, which is not what we want for micro benchmarks.
I outlined the differences is codegen here: https://gist.github.com/EricWF/90e14dacab3b28d1b811#donotoptimize

Also `template <class Tp*>` should be `template <class Tp>`.

EricWF:
@zabereer So I've looked into this more. I think we should go ahead with your change, or a change similar to it. I'm happy to do it myself but I wanted to ask you first.

zabereer:
There was a rookie error in the pull request I submitted as you pointed out, and the pointer should not be const either (sorry about that). I will fix it tomorrow morning and re-submit if ok with you. I guess this is an interim solution until the compiler issue is resolved?

EricWF:
Yeah. Also you don't need to provide the additional overload for pointers. When you make the change could you please merge the clang and GCC implementations since this will make them the same.

PR #215, opened 2016-05-21
Add reporting of asymptotic complexity and NRMSE

ismaelJimenez:
I added the option to report asymptotic complexity and normalized root-mean square error. One of the most important things when implementing a family of benchmarks is not only to see the actual time, but how it relates to the scalability curve. Up until now I had to parse the output of the benchmark with a python script, and calculate the coefficient for the high order term using scipy.optimize.leastsq. With the changes on this check-in, the capability of calculating the high order coefficient and RMS, are built inside the benchmark.

The check-in consists basically on the integration of a minimal least squares fitter that I developed as an standalone project and adapted it to work smoothly with the architecture of the benchmark.


include/benchmark/reporter.h
```diff
@@ -78,15 +92,23 @@ class BenchmarkReporter {
...
+  virtual void ReportComplexity(const std::vector<Run>& complexity_reports) = 0;
 
   // Called once and only once after ever group of benchmarks is run and
   // reported.
   virtual void Finalize();
 
   virtual ~BenchmarkReporter();
 protected:
-  static void ComputeStats(std::vector<Run> const& reports, Run* mean, Run* stddev);
+  static void ComputeStats(const std::vector<Run> & reports, Run& mean, Run& stddev);
```
dominichamon: as per google c-style, please don't use non-const reference arguments. https://google.github.io/styleguide/cppguide.html

src/benchmark.cc
```diff
@@ -411,6 +415,8 @@ bool BenchmarkFamilies::FindBenchmarks(
         instance.min_time = family->min_time_;
         instance.use_real_time = family->use_real_time_;
         instance.use_manual_time = family->use_manual_time_;
+        instance.last_benchmark_instance = (args == family->args_.back());
```
dominichamon: there's a possible bug here: you'll set this to true, but then it can be filtered out by the regex match below (l444). you only want to set it if the args is the last one and if it's going to be added to the benchmarks set.

src/console_reporter.cc
```diff
@@ -100,7 +115,22 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
+  if(result.report_bigO) {
+    std::string big_o = result.report_bigO ? GetBigO(result.complexity) : "";
+    ColorPrintf(COLOR_YELLOW, "%10.4f %s %10.4f %s ",
+                result.real_accumulated_time * multiplier,
+                big_o.c_str(),
+                result.cpu_accumulated_time * multiplier,
+                big_o.c_str());
+  }  
+  else if(result.report_rms) {
+    ColorPrintf(COLOR_YELLOW, "%10.0f %s %10.0f %s ",
```
dominichamon: why %s for a %? just use %% in the format string.

src/minimal_leastsq.cc
```diff
@@ -0,0 +1,113 @@
...
+
+// Source project : https://github.com/ismaelJimenez/cpp.leastsq
+// Adapted to be used with google benchmark
+
+#include "minimal_leastsq.h"
+
+#include <math.h>
+
+// Internal function to calculate the different scalability forms
+double fittingCurve(double N, benchmark::BigO Complexity) {
```
dominichamon: please use the same style guide in regards to naming and indent as the rest of the code.

src/minimal_leastsq.cc
```diff
@@ -0,0 +1,113 @@
...
+	if (Complexity == benchmark::O_N)
```
dominichamon: switch?

src/minimal_leastsq.cc
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Internal function to find the coefficient for the high-order term in the running time, by minimizing the sum of squares of relative error.
+//   - N          : Vector containing the size of the benchmark tests.
+//   - Time       : Vector containing the times for the benchmark tests.
+//   - Complexity : Fitting curve.
+// For a deeper explanation on the algorithm logic, look the README file at http://github.com/ismaelJimenez/Minimal-Cpp-Least-Squared-Fit
+
+LeastSq leastSq(const std::vector<int>& N, const std::vector<double>& Time, const benchmark::BigO Complexity) {
+	assert(N.size() == Time.size() && N.size() >= 2);
```
dominichamon: src/check.h has some useful macros for logging these errors.

include/benchmark/reporter.h
ismaelJimenez: I wasn't aware about that guideline. It is already fixed.

src/benchmark.cc
ismaelJimenez: Yes, good catch. I have moved it after the filter to avoid that problem.

src/console_reporter.cc
ismaelJimenez: fixed

src/minimal_leastsq.cc
ismaelJimenez: I changed the function names to camel case and the variables to small case separated by "_". If you find something else please let me know.

src/benchmark.cc
```diff
@@ -781,7 +798,14 @@ void RunBenchmark(const benchmark::internal::Benchmark::Instance& b,
         report.cpu_accumulated_time = cpu_accumulated_time;
         report.bytes_per_second = bytes_per_second;
         report.items_per_second = items_per_second;
+        report.arg1 = b.arg1;
```
dominichamon: are you using arg1 and arg2 anywhere in the reporters? i can't see them at first glance.

src/reporter.cc
```diff
@@ -77,6 +79,74 @@ void BenchmarkReporter::ComputeStats(
...
+  rms->report_label = big_o->report_label;
+  rms->iterations = 0;
+  rms->real_accumulated_time = result_real.rms / multiplier;
+  rms->cpu_accumulated_time = result_cpu.rms / multiplier;
+  rms->report_rms = true;
+  rms->complexity = result_cpu.complexity;
+}
+
+std::string BenchmarkReporter::GetBigO(BigO complexity) {
+  switch (complexity) {
```
dominichamon: the code /might/ be simpler if the definition of the BigO enum, this method, and FittingCurve, are moved into their own files. That keeps all of the complexity methods together and allows this file to focus on things specific to the reporter, which this isn't.

src/console_reporter.cc
```diff
@@ -100,7 +115,22 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   ColorPrintf(COLOR_GREEN, "%-*s ",
```
dominichamon: consider different colors for the complexity report. they get lost in the output otherwise.

also, i don't think there's any need to output for both time and cpu, but you may think different.

you can suppress the iteration count too.

src/json_reporter.cc
```diff
@@ -115,6 +115,31 @@ void JSONReporter::ReportRuns(std::vector<Run> const& reports) {
...
+    out << ",\n";
+  }
+  
+  Run big_o_data;
+  Run rms_data;
+  BenchmarkReporter::ComputeBigO(complexity_reports, &big_o_data, &rms_data);
+  
+  // Output using PrintRun.
+  out << indent << "{\n";
+  PrintRunData(big_o_data);
```
dominichamon: consider suppressing 'iterations' for this (and time unit).

src/csv_reporter.cc
```diff
@@ -66,7 +66,22 @@ void CSVReporter::ReportRuns(std::vector<Run> const& reports) {
...
+    // We don't report asymptotic complexity data if there was a single run.
+    return;
+  }
+  
+  Run big_o_data;
+  Run rms_data;
+  BenchmarkReporter::ComputeBigO(complexity_reports, &big_o_data, &rms_data);
+  
+  // Output using PrintRun.
+  PrintRunData(big_o_data);
```
dominichamon: same as elsewhere: iterations are unimporntant, and time unit should be suppressed.

src/benchmark.cc
ismaelJimenez: Yes, on the function ComputeBigO() on the file reporter.cc to populate the vector "n". I need arg1 and arg2 to know the length of the test. currently I am only using arg1, because I only use one variable, but in the near future I will use also arg2. This will let me calculate, for example, the time of string concatenation as a function of the length of both strings (N+M or in this case arg1 and arg2), but currently the bigO can only be calculated as a function of arg1.

dominichamon: ah, so you're assuming that arg1 and arg2 are used as loop indices, and only used in a single loop.

we have a couple of tests in basic_test that use the range loop twice (so it's already 2N) and examples in benchmark_test (BM_SetInsert) that use range_x to construct test data, then range_y to do the work. anything there based on assumptions of N and M will fail.

it might be worth making this more like the state methods SetItemsProcessed and SetBytesProcessed. SetComplexityN and SetComplexityM?

src/console_reporter.cc
ismaelJimenez: I have changed the color to blue, but I do not have any particular preference. So if you have any preference I can change it to any other color. With respect to the output, I actually only use the CPU, but as on the reports the real time is given I supposed that somebody might be interested on it, maybe for some test where CPU value is not reliable, I don´t know... I think it does not heart to have it there but you know better than me the users of the tool, so if you think is not useful, I could just remove it.

src/json_reporter.cc
ismaelJimenez: I have suppressed  iteration and time unit for RMS report but on BigO report i kept the time unit, because it is important. The multiplication factor given for the high order term is with respect to the time unit. So, if we change from nano to micro, this factor shall be scaled.

src/csv_reporter.cc
ismaelJimenez: Also fixed. Same comment as for json reporter.

src/reporter.cc
ismaelJimenez: I have extracted BigO enum and GetBigO in a file. But I am not sure about extracting FittingCurve because this is a local method of the leastSq calculation that nobody else should know about.

src/benchmark.cc
ismaelJimenez: Hello dominic. I have already added the field SetComplexityN. If there is something else please let me know.

PR #216, opened 2016-05-23
Add checks that <Resume|Pause>Timing functions are not called outside of the KeepRunning() loop. Fixes #204

### LINKED ISSUE #204: warn/error on PauseTiming and ResumeTiming outside the timing loop
there are a number of places that we mark, in comments, a requirement for a function call. we should enforce these with errors if they occur.

EricWF:


src/benchmark.cc
```diff
@@ -830,11 +830,13 @@ State::State(size_t max_iters, bool has_x, int x, bool has_y, int y,
 void State::PauseTiming() {
   // Add in time accumulated so far
   CHECK(running_benchmark);
```
dominichamon: is this check necessary any more?

test/diagnostics_test.cc
```diff
@@ -0,0 +1,43 @@
+
+#include "benchmark/benchmark_api.h"
+#include "../src/check.h"
```
dominichamon: please add a comment that this shouldn't be copied by users of the library or something :)

src/check.h
```diff
@@ -10,6 +10,18 @@
 namespace benchmark {
 namespace internal {
 
+typedef void(AbortHandlerT)();
+
+inline AbortHandlerT*& get_abort_handler() {
```
dominichamon: stick to google-style naming please. GetAbortHandler, etc.

src/benchmark.cc
EricWF: Sort of. 'running_benchmark' checks that the benchmark was actually started by the library. So that check would still fire if a user constructed their own 'State' object and tried to use it.

PR #217, opened 2016-05-24
Add SkipWithError(...) function.

EricWF:
This patch depends on pull request #216.

This patch adds the "SkipWithError(...)" function that allows a benchmark to be skipped and an error reported. When a benchmark is skipped no statistics are reported, only the error message.

Once "SkipWithError(...)" is called "KeepRunning()" will return false. If the user needs to exit scope immediatly it is there responsibility to add a "return;".

In multi-threaded benchmarks only the error threads exit early. All other threads continue as normal. It would be nice if other threads could detect the error and cease execution but currently that is not possible. 

If multiple threads report errors the first reported error string is used and the others discarded.

I'll update the doc and examples shortly.


README.md
```diff
@@ -276,6 +276,37 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+```c++
+static void BM_test(benchmark::State& state) {
+  auto resource = GetResource();
+  if (!resource.good()) {
+      state.SkipWithError("Resource is not good!");
+      // KeepRunning() loop will not be entered.
+  }
+  while (state.KeepRunning()) {
+      auto data = resource.read_data();
+      if (!resource.good()) {
```
dominichamon: `!data.good`?

src/benchmark.cc
```diff
@@ -830,14 +862,26 @@ State::State(size_t max_iters, bool has_x, int x, bool has_y, int y,
...
   CHECK(running_benchmark);
+  CHECK(started_ && !finished_);
   timer_manager->StartTimer();
 }
 
+void State::SkipWithError(const char* msg) {
+  CHECK(msg);
+  error_occurred_ = true;
+  const char* expected_no_error_msg = nullptr;
+  error_message.compare_exchange_weak(expected_no_error_msg, msg);
```
dominichamon: so just one error per benchmark. if every iteration errors, we'll only report the.. first?

src/console_reporter.cc
```diff
@@ -80,6 +85,14 @@ void ConsoleReporter::ReportRuns(const std::vector<Run>& reports) {
 }
 
 void ConsoleReporter::PrintRunData(const Run& result) {
+  ColorPrintf(COLOR_GREEN, "%-*s ",
+              name_field_width_, result.benchmark_name.c_str());
+
+  if (result.error_occurred) {
+    ColorPrintf(COLOR_RED, "ERROR OCCURRED: \'%s\'", result.error_message.c_str());
```
dominichamon: 'on iteration x' might be nice, if it's not too hard.

README.md
EricWF: I was imagining the "std::basic_istream" semantics where you read from the stream and then ask the stream, not the data, if it's still valid.

If you want this change made I'm happy to make it.

src/benchmark.cc
EricWF: I think we differ on the definition of "iteration".  Once a thread reports an error it performs no future iterations. I'll add code to enforce that each thread only calls "SkipWithError(...)" once.

This code selects the first error reported by a multi-threaded benchmark. Different threads may report an error later but IMO the first error message is likely the most relevant.

src/console_reporter.cc
EricWF: Ack.

dominichamon:
is this ready to merge or were you going to add the 'on iteration x' to the error output?

EricWF:
> is this ready to merge or were you going to add the 'on iteration x' to the error output?

I'm going to look into improving the diagnostic but that could be done as a followup. Other than this is ready to land. Note that pull request #216 should land before this.

dominichamon:
resolve yo conflicts

src/benchmark.cc
```diff
@@ -197,6 +203,16 @@ class TimerManager {
     }
   }
 
+  void RemoveErroredThread() EXCLUDES(lock_) {
+    MutexLock ml(lock_);
+    int last_thread = --running_threads_ == 0;
+    if (last_thread && running_) {
+        InternalStop();
```
dominichamon: nit: extra indent here

src/benchmark.cc
```diff
@@ -253,22 +270,27 @@ class TimerManager {
...
-    } else {
-      // Last thread has reached the barrier
-      phase_number_++;
-      entered_ = 0;
-      return true;
+      while (true) {
+        phase_condition_.wait(ml.native_handle(), cb);
+        if (phase_number_ > phase_number_cp)
+          return false;
+        else if (running_threads_ == entered_)
```
dominichamon: you don't need the 'else' here

src/benchmark.cc
EricWF: Ack.

src/benchmark.cc
EricWF: Ack. It seems like I can kill the whole loop.

EricWF:
I seem to have buggered up the console reporter as well. Working to fix it.

EricWF:
Is this OK to land?

dominichamon:
merging now

dominichamon:
running the skip_with_error_test doesn't allow for benchmark_format override:

```
$ test/skip_with_error_test --benchmark_format=json
Run on (12 X 3501 MHz CPU s)
2016-05-25 09:04:15
Benchmark                                    Time           CPU Iterations
--------------------------------------------------------------------------
BM_error_before_running             ERROR OCCURRED: 'error message'
BM_error_during_running/1/threads:1 ERROR OCCURRED: 'error message'
```

EricWF:
Huh, no Idea how I managed to break the json reporter.

dominichamon:
i think the issue is the test. other tests allow override.

maybe something in your custom main is funky? ¯\_(ツ)_/¯

EricWF:
OK, so that specific test provides a custom reporter which the `--benchark_format` option does not override. It is not the only test to do this. I think it behaves as intended.

PR #218, opened 2016-05-24
Force DoNotOptimize operand to memory for both gcc and clang

zabereer:
This is kind of unfortunate. Initially I thought we could use templates to determine register and memory based operands, however structs up to 4 eightbyte sizes (according to ABI) can be passed in registers.
(So one can impact the codegen negatively by forcing to memory, and also by forcing to register thus evicting other data.)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@
 #
 # Please keep the list sorted.
 
+Albert Pretorius <pretoalb@gmail.com>
 Arne Beer <arne@twobeer.de>
 Christopher Seymour <chris.j.seymour@hotmail.com>
 David Coeurjolly <david.coeurjolly@liris.cnrs.fr>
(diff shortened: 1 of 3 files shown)


EricWF:
Super! Thanks. LGTM. Do you need me to commit this?

zabereer:
Yes please and thanks very much!

EricWF:
Now that I think about it... In order to prevent regression would you be able to write a new tests that checks 'DoNotOptimize(...)'. Said test doesn't even have to contain any benchmarks. A main function in a new test file that simply calls 'DoNotOptimize(...)' with a variety of types would be great. Let me know if you need any help with the CMake magic.

zabereer:
Sure, that would be good.

dominichamon:
can you make sure your commit authors are set correctly so the cla bot stops complaining?

zabereer:
Sure I am doing this now.

PR #219, opened 2016-05-25
Change reporters to use a specified output and error stream. Add tests for output.

EricWF:
This patch attempts to make it easier to test the output of the reporters. Bugs in the output seems to pop up frequently and it would be nice to write unit tests for it.

This patch solves the test-ability problem by making all output use user-provided streams. Once this patch lands we can slowly add tests for the output under all the various benchmark options.

This branch is cloned from the 'SkipWithError(...)' branch so please ignore those changes here until it langs.


include/benchmark/reporter.h
```diff
@@ -105,12 +110,45 @@ class BenchmarkReporter {
   // reported.
   virtual void Finalize();
 
+  // REQUIRES: 'out' shall be valid for the lifetime of the reporter.
+  // Set the stream used for non-error output to the value specified by 'out'.
+  void SetOutputStream(std::ostream& out) {
```
dominichamon: google style: no non-const references. please take a pointer directly (and CHECK for nullptr. sorry).

include/benchmark/reporter.h
```diff
@@ -105,12 +110,45 @@ class BenchmarkReporter {
...
+    output_stream_ = &out;
+  }
+
+  // REQUIRES: 'out' shall be valid for the lifetime of the reporter.
+  // Set the stream used for error output to the value specified by 'out'.
+  void SetErrorStream(std::ostream& out) {
```
dominichamon: rename `out` to `err`?

include/benchmark/reporter.h
```diff
@@ -105,12 +110,45 @@ class BenchmarkReporter {
...
+    error_stream_ = &out;
+  }
+
+  // REQUIRES: 'out' shall be valid for the lifetime of the reporter.
+  // Set the stream used for both error and non-error output to the value
+  // specified by 'out'.
+  void SetStreams(std::ostream& out) {
```
dominichamon: not sure this is useful given you have the above...

src/console_reporter.cc
```diff
@@ -32,26 +34,36 @@ namespace benchmark {
 bool ConsoleReporter::ReportContext(const Context& context) {
   name_field_width_ = context.name_field_width;
 
-  std::cerr << "Run on (" << context.num_cpus << " X " << context.mhz_per_cpu
+  auto& Out = GetOutputStream();
+  auto& Err = GetErrorStream();
+
+#ifdef BENCHMARK_OS_WINDOWS
+  if (FLAGS_color_print && &Out != &std::cout) {
```
dominichamon: what if Out is std::cerr?

include/benchmark/reporter.h
EricWF: Ack. Does this apply to the getters as well?

include/benchmark/reporter.h
EricWF: Yeah. I'll remove this.

include/benchmark/reporter.h
dominichamon: doesn't seem to, just arguments.

ftr, i don't like this rule, but there it is.

src/colorprint.cc
```diff
@@ -74,14 +78,53 @@ PlatformColorCode GetPlatformColorCode(LogColor color) {
...
+
 }  // end namespace
 
-void ColorPrintf(LogColor color, const char* fmt, ...) {
+std::string FormatString(const char *msg, va_list args) {
+  // we might need a second shot at this, so pre-emptivly make a copy
+  va_list args_cp;
+  va_copy(args_cp, args);
+
+  //TODO, use std::array for first attempt to avoid one memory allocation
```
EricWF: Note to self: remove this comment since it's been implemented.

src/console_reporter.cc
EricWF: It could likely be made to work but I have no way of testing on Windows so I avoided changing it.

dominichamon:
can you resolve conflicts so i can merge please?

PR #221, opened 2016-05-25
Allow benchmarks to take arbitrary arguments.

EricWF:
This patch adds a `BENCHMARK_CAPTURE(func, description, args...)` macro which creates a benchmark that invokes `func` as `func(state, args...)`.

I've wanted this for a while but I was unsure of how to implement it without a large amount of complexity. Thankfully this solution is quite simple. `BENCHMARK_CAPTURE` works by creating a lambda of the form `[=](State& st) { func(state, args...); }` where both `func` and `args...` are captured by value. The lambda is then stored in a Benchmark object and it behaves exactly the same as all other benchmarks.

Obviously there is still some testing to be done, but I want to get feedback before spending too much time on it.

@dominichamon you've expressed concerns about similar features in the past. What do you think of this?


EricWF:
One example of where this functionality is useful is testing container templates. For example say I want to test the performance of `emplace(...Args)` on the Unordered containers in the STL. Writing a good benchmark is hard and I only want to do it once. Therefore I need to write this benchmark in such a way that it works with:

1. 4 different container types. `unordered_map`, `unordered_set`, `unordered_multimap` and`unordered_multiset`. 
2. 2 different value_type's between the 4 containers. `std::pair<Key, Value>` for maps and `Value` for sets.
3. Multiple key types with different hashing behaviors and costs.
4. Pathological sets of keys in an attempt to trigger hash collisions.
5. Multiple different types of arguments to emplace. In particular ones that don't match the containers value_type. For example `std::pair<const Key, Value>` and `std::pair<Key, Value>`.

The biggest issue is that inputs to the benchmark need to be constructed within the benchmark itself. This essentially requires writing a different benchmark whenever you want to test with sufficiently different inputs (be it types or values).

Fixtures don't solve this problem. They allow different benchmarks to share the same the same `SetUp()` code but they don't allow the same benchmark to use different `SetUp` functions.

The solution to these problems is allowing the inputs to the benchmark to be passed in as a function argument. The benchmarks I'm hoping to write look like this:

```c++
template <class Container, class Inputs>
void BM_emplace_test(State& state, Container C, Inputs const& inputs) {
  auto It = Inputs.begin();
  const auto ItEnd = Inputs.end();
  while (state.KeepRunning()) {
      DoNotOptimize(C.emplace(*It++));
     if (It == ItEnd) It = Inputs.begin();
  } 
}
BENCHMARK_CAPTURE(BM_emplace_test, unordered_set_of_strings, 
    std::unordered_set<std::string>{}, std::vector<std::string>{/* inputs */});
BENCHMARK_CAPTURE(BM_emplace_test, unordered_set_of_strings_with_collisions, 
    std::unordered_set<std::string>{}, std::vector<std::string>{/* inputs */});
BENCHMARK_CAPTURE(BM_emplace_test, unordered_map_of_ints, 
   std::unordered_map<int, int>{}, std::vector<std::pair<int, int>>{/* inputs */});
```

Going one step further this change makes it possible to write incredibly generic benchmarks by passing in a lambda to perform the operation under test. For example:

```c++
template <class Container, class Inputs>
void BM_container_op(State& state, Container C, Inputs const& inputs, Functor F) {
  auto It = Inputs.begin();
  const auto ItEnd = Inputs.end();
  while (state.KeepRunning()) {
     F(C, *It++);
     if (It == ItEnd) It = Inputs.begin();
  } 
}
BENCHMARK_CAPTURE(BM_container_op, insert_a_set_of_strings, 
    std::unordered_set<std::string>{}, std::vector<std::string>{/* inputs */},
   [](auto& C, auto& Input) { DoNotOptimize(C.insert(Input)) });
BENCHMARK_CAPTURE(BM_container_op, emplace_a_set_of_strings, 
    std::unordered_set<std::string>{}, std::vector<std::string>{/* inputs */},
   [](auto& C, auto& Input) { DoNotOptimize(C.emplace(Input)) });
```

Unlike calls through opaque function pointers the lambdas call operator will always be inlinable. within the benchmark. This would allow benchmarks to be written with almost no boilerplate. We have previously discussed deprecating the `KeepRunning()` loop and this seems like one possible way to do it.

test/basic_test.cc
```diff
@@ -99,4 +99,12 @@ void BM_empty_stop_start(benchmark::State& state) {
 BENCHMARK(BM_empty_stop_start);
 BENCHMARK(BM_empty_stop_start)->ThreadPerCpu();
 
+template <class ...Args>
+void BM_with_args(benchmark::State& state, Args&&...) {
```
dominichamon: guard this with a `__cplusplus >= 201103L` check please.

include/benchmark/benchmark_api.h
```diff
@@ -557,6 +557,24 @@ class FunctionBenchmark : public Benchmark {
...
+      : Benchmark(name), func_(std::forward<L>(l)) {}
+  virtual void Run(State& st) { func_(st); }
+private:
+  Lambda func_;
+};
+
+template <class Invoker>
+Benchmark* CreateLambdaBenchmark(const char* name, Invoker&& fn) {
+    return RegisterBenchmarkInternal(
+        new LambdaBenchmark<typename std::decay<Invoker>::type>(name, std::move(fn)));
```
dominichamon: that reminds me .. RegisterBenchmarkInternal should take a std::unique_ptr.

dominichamon:
I like it.

test/basic_test.cc
EricWF: C++11 language and library features are used elsewhere in the tests.  Only `cxx03_test.cc` compiles in C++03.

Is there a reason this particular feature needs guarding?

EricWF: Nevermind this change is needed. GCC 4.6 is a C++0x compiler and so BENCHMARK_CAPTURE isn't defined for it.

EricWF:
PS. I'll want to squash the commit history on this patch before merging. It's way too noisy.

include/benchmark/benchmark_api.h
dominichamon: not for you, by the way. note for me.

dominichamon:
it's showing me diffs including the error reporting change. shouldn't those be present on the left side too?

are you all rebased?

EricWF:
It has now been rebased and I don't see any error reporting changes.

PR #222, opened 2016-05-25
Upgraded complexity functionality

ismaelJimenez:
- Upgrade of the interface for automatic calculation of complexity. 
        From "-> Complexity(benchmark::oAuto)" to "-> Complexity()"
- Rework of Least Square functionality to use lambdas internally (and prepare for possible future upgrade
  of complexity interface moving from enums to lambdas)
- Integrated leastsq and complexity files, to provide a cleaner interface


include/benchmark/complexity.h
```diff
@@ -19,24 +37,34 @@ enum BigO {
...
+//                   interpolated from data.
+//   - rms         : Normalized Root Mean Squared Error.
+//   - complexity  : Scalability form (e.g. oN, oNLogN). In case a scalability
+//                   form has been provided to MinimalLeastSq this will return
+//                   the same value. In case BigO::oAuto has been selected, this
+//                   parameter will return the best fitting curve detected.
+
+struct LeastSq {
+  LeastSq() :
+    coef(0),
```
dominichamon: nit: initialize doubles with 0.0 please

src/reporter.cc
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "benchmark/reporter.h"
-#include "minimal_leastsq.h"
```
dominichamon: you need complexity.h, right?

include/benchmark/complexity.h
ismaelJimenez: fixed

src/reporter.cc
ismaelJimenez: Yes, that's right.  "complexity.h" is included via "benchmark_api.h" (because of the BigO enum definition) and then via "reporter.h". As I found that the header was already included via the previous path, I just removed if from the reporter.cc. But, if you think that it will increase readability, I could add it again.

dominichamon: please do. i don't want refactorings to benchmark_api to lead to missing symbols. everything should include what it uses.

ismaelJimenez: Yes, makes perfect sense. I have also added the include to console_reporter.cc, because function "GetBigOString" comes also from complexity.h.
On a different topic, I am not sure why after changing the doubles to 0.0, coveralls started to complain about coverage decreased.

include/benchmark/complexity.h
```diff
@@ -19,24 +37,34 @@ enum BigO {
...
+// This data structure will contain the result returned by MinimalLeastSq
+//   - coef        : Estimated coeficient for the high-order term as
+//                   interpolated from data.
+//   - rms         : Normalized Root Mean Squared Error.
+//   - complexity  : Scalability form (e.g. oN, oNLogN). In case a scalability
+//                   form has been provided to MinimalLeastSq this will return
+//                   the same value. In case BigO::oAuto has been selected, this
+//                   parameter will return the best fitting curve detected.
+
+struct LeastSq {
```
dominichamon: thinking about it more, these shouldn't be in this header. we want to keep the public api (under include/benchmark) as clean as possible and users would never need these.

as such, can you move this into src/ and add the enum definition into benchmark_api.h where the Complexity method is defined?

ismaelJimenez: Ok. Please have a look at the last check-in

src/complexity.cc
```diff
@@ -0,0 +1,162 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Source project : https://github.com/ismaelJimenez/cpp.leastsq
+// Adapted to be used with google benchmark
+
+#include "complexity.h"
```
dominichamon: needs benchmark_api.h for BigO

src/reporter.cc
dominichamon: it's been complaining all along about test coverage reduction.

src/complexity.cc
ismaelJimenez: Right. Added it with last check-in.

dominichamon:
attempting manual merge...

PR #226, opened 2016-05-27
Cleanup reporters

EricWF:
This patch attempts to remove duplicate code in the reports. The main changes are:

1. Additional run reports representing the mean, stddev, complexity and RMS are computed outside of the reporters and are passed as additional  elements to "ReportRuns(...)".

2. The "ReportComplexity(...)" interface has been removed since they are now reported using "ReportRuns(...)". 

3. Add "GetAdjusted<CPU|Real>Time()" helpers were added to "BenchmarkReporter::Run" which compute the per-iteration time in the correct unit. This prevent the reporter from having to do it themselves. 

3. Add static `BenchmarkReporter::PrintBasicContext(...)` helper which provides a default implementation used to report the context. This allows less duplication in the reporters.

I would like @ismaelJimenez's input on my changes to how complexity is reported.


ismaelJimenez:
I have just check it (regarding the complexity report) and everything looks fine. I have also taken a look at the refactoring and I really like having ComputeBigO in complexity.cc so it is easier to maintain, because all the logic is at the same place. 
Also I have just thought, that a future update could be moving ComputeStats into complexity.cc, because the logic of computing stats is an small subset of what ComputeBigO can do (maybe I can also merge both of them). It is like calculating the benchmark executions time with O(1), just getting the mean and stddev. What do you think?

EricWF:
@ismaelJimenez I moved `ComputeStats` to complexity.cc as suggested. It seems like a better place than inside `benchmark.cc`.

include/benchmark/benchmark_api.h
```diff
@@ -236,6 +236,30 @@ enum TimeUnit {
   kMillisecond
 };
 
+inline const char* GetTimeUnitString(TimeUnit unit) {
```
dominichamon: why are these in the public API?

dominichamon:
LGTM .. feel free to merge if you're happy with it.

include/benchmark/benchmark_api.h
EricWF: To support custom reporters.

dominichamon: it should be in include/benchmark/reporter.h then i think.

EricWF: Done in a follow up commit.

PR #231, opened 2016-06-02
Added option to specify complexity function with a lambda function

ismaelJimenez:
- Complexity function can now be specified with a lambda function, allowing for custom complexity functions (e.g. N/2, 2**N, ...)
- Json and Csv complexity reports have been improved. Now not only the coefficient, but the actual complexity calculated is also printed.
- Type of "complexity_n" has been changed from int to size_t
- Complexity_test.cc has been changed using reporter_output_test.cc as a base. Now the results of complexity tests are checked for correctness in console, csv and json reporters.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -152,7 +152,6 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <assert.h>
 #include <stddef.h>
 #include <stdint.h>
-#include <functional>
 
 #include "macros.h"
 
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -152,6 +152,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <assert.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <functional>
 
 #include "macros.h"
 
@@ -247,15 +248,20 @@ enum BigO {
   oNCubed,
   oLogN,
   oNLogN,
-  oAuto
+  oAuto,
+  oLambda
 };
+
+// BigOFunc is passed to a benchmark in order to specify the asymptotic 
+// computational complexity for the benchmark.
+typedef double(BigOFunc)(size_t);
 
 // State is passed to a running Benchmark and contains state for the
 // benchmark to use.
 class State {
 public:
   State(size_t max_iters, bool has_x, int x, bool has_y, int y,
-        int thread_i, int n_threads);
+    int thread_i, int n_threads);
 
   // Returns true iff the benchmark should continue through another iteration.
   // NOTE: A benchmark may not return from the test until KeepRunning() has
@@ -268,13 +274,13 @@ public:
     }
     bool const res = total_iterations_++ < max_iterations;
     if (BENCHMARK_BUILTIN_EXPECT(!res, false)) {
-        assert(started_ && (!finished_ || error_occurred_));
-        if (!error_occurred_) {
-            PauseTiming();
-        }
-        // Total iterations now is one greater than max iterations. Fix this.
-        total_iterations_ = max_iterations;
-        finished_ = true;
+      assert(started_ && (!finished_ || error_occurred_));
+      if (!error_occurred_) {
+        PauseTiming();
+      }
+      // Total iterations now is one greater than max iterations. Fix this.
+      total_iterations_ = max_iterations;
+      finished_ = true;
     }
     return res;
   }
@@ -359,7 +365,7 @@ public:
   // represent the length of N.
   BENCHMARK_ALWAYS_INLINE
   void SetComplexityN(size_t complexity_n) {
-    complexity_n_ = complexity_n;
+	  complexity_n_ = complexity_n;
   }
 
   BENCHMARK_ALWAYS_INLINE
@@ -533,10 +539,14 @@ public:
   // to control how many iterations are run, and in the printing of items/second
   // or MB/second values.
   Benchmark* UseManualTime();
-
+  
   // Set the asymptotic computational complexity for the benchmark. If called
   // the asymptotic computational complexity will be shown on the output. 
   Benchmark* Complexity(BigO complexity = benchmark::oAuto);
+  
+  // Set the asymptotic computational complexity for the benchmark. If called
+  // the asymptotic computational complexity will be shown on the output.
+  Benchmark* Complexity(BigOFunc* complexity);
 
   // Support for running multiple copies of the same benchmark concurrently
   // in multiple threads.  This may be useful when measuring the scaling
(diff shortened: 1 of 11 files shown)


include/benchmark/reporter.h
```diff
@@ -155,7 +156,7 @@ class BenchmarkReporter {
 // Simple reporter that outputs benchmark data to the console. This is the
 // default reporter used by RunSpecifiedBenchmarks().
 class ConsoleReporter : public BenchmarkReporter {
- public:
+public:
```
dominichamon: one space is correct for google style.

include/benchmark/benchmark_api.h
```diff
@@ -247,34 +247,39 @@ enum BigO {
...
   oLogN,
   oNLogN,
-  oAuto
+  oAuto,
+  oLambda
 };
 
+// BigOFunc is passed to a benchmark in order to specify the asymptotic 
+// computational complexity for the benchmark.
+typedef double(BigOFunc)(size_t);
```
dominichamon: `std::function<double(size_t)>` doesn't compile on the target platforms?

also, there's no need for n to be size_t. it isn't the length of a container, so int is preferable.

src/complexity.cc
```diff
@@ -144,34 +132,34 @@ LeastSq MinimalLeastSq(const std::vector<int>& n,
...
+    best_fit = MinimalLeastSq(n, time, FittingCurve(complexity));
     best_fit.complexity = complexity;
   }
 
   return best_fit;
 }
 
 std::vector<BenchmarkReporter::Run> ComputeStats(
-    const std::vector<BenchmarkReporter::Run>& reports)
+  const std::vector<BenchmarkReporter::Run>& reports)
```
dominichamon: please run this through clang-format with google style to ensure these indents are correct and not changing for the sake of it.

src/complexity.cc
```diff
@@ -256,14 +244,16 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
...
-
-  // result_cpu.complexity is passed as parameter to result_real because in case
-  // reports[0].complexity is oAuto, the noise on the measured data could make
-  // the best fit function of Cpu and Real differ. In order to solve this, we
-  // take the best fitting function for the Cpu, and apply it to Real data.
-  LeastSq result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
+  LeastSq result_cpu;
+  LeastSq result_real;
 
+  if (reports[0].complexity != oLambda) {
```
dominichamon: can you switch this to being the positive case instead? it's easier to read.

test/complexity_test.cc
```diff
@@ -1,37 +1,189 @@
...
   }
-  return v;
-}
+};
 
-std::map<int, int> ConstructRandomMap(int size) {
-  std::map<int, int> m;
-  for (int i = 0; i < size; ++i) {
-    m.insert(std::make_pair(rand() % size, rand() % size));
+std::vector<TestCase> ConsoleOutputTests;
```
dominichamon: no global non-PODs please, even in tests. These should be lazily instantiated pointers.

test/complexity_test.cc
```diff
@@ -1,37 +1,189 @@
...
 
+template <class First>
+std::string join(First f) { return f; }
+
+template <class First, class ...Args>
+std::string join(First f, Args&&... args) {
+    return std::string(std::move(f)) + "[ ]+" + join(std::forward<Args>(args)...);
+}
+
+std::string dec_re = "[0-9]+\\.[0-9]+";
```
dominichamon: no static non-PODs please.

const char[] instead for this.

test/complexity_test.cc
```diff
@@ -1,37 +1,189 @@
...
 void BM_Complexity_O1(benchmark::State& state) {
   while (state.KeepRunning()) {
   }
   state.SetComplexityN(state.range_x());
 }
 BENCHMARK(BM_Complexity_O1) -> Range(1, 1<<18) -> Complexity(benchmark::o1);
+BENCHMARK(BM_Complexity_O1) -> Range(1, 1<<18) -> Complexity([](size_t){return 1.0; });
+BENCHMARK(BM_Complexity_O1) -> Range(1, 1<<18) -> Complexity();
+
+std::string big_o_1_test_name = "BM_Complexity_O1_BigO";
```
dominichamon: as above

include/benchmark/reporter.h
ismaelJimenez: Hello dominic. I removed the space because JSONReporter and CSVReporter classes had no space before "public:". So I though that we either add an space on those other to classes or remove the space in this one, to keep consistency. Is there a preferred style?

dominichamon: yes. clang-format with google style setting. whatever that says is what we go with :)

include/benchmark/benchmark_api.h
ismaelJimenez: No, it doesn't compile on the c++0x compiler because it generates the error _"function’ in namespace ‘std’ does not name a template type_ I investigated the problem, and I found out that this happens with versions of g++ less that 4.7 and we have target compilers that are g++4.6. 
In fact, the first implementation I did was with std::function<double(size_t)> but I had to change it by the function definition to make it compile on target compilers.

test/complexity_test.cc
ismaelJimenez: This is just a copy paste from reporter_output_test.cc. Then we have the same problem over there.

dominichamon: then two comments:

1. never mind, we'll fix it afterwards.
2. can we abstract that so we don't copy-paste?

ismaelJimenez: Yes, I think it would be a good idea to abstract the implementation of the logic of reporter_output_test.cc, because it is very useful for testing the output of any benchmark on console, json and csv and I think future developments will benefit from it.
At first it seems to me that some things could be moved to a .h file (like a couple of classes and macros), but probably there is a better way to reuse it. I think @EricWF did the reporter_output_test.cc, so may be he can help us on the proper way to restructure it to be reused more efficiently.
@dominichamon : All the other comments are already fixed. Please check if I missed something.

include/benchmark/reporter.h
```diff
@@ -85,8 +85,9 @@ class BenchmarkReporter {
     double max_heapbytes_used;
 
     // Keep track of arguments to compute asymptotic complexity
-    BigO   complexity;
-    int complexity_n;
+    BigO complexity;
+    BigOFunc* complexity_lambda;
+    size_t complexity_n;
```
dominichamon: this should be int. google style (https://google.github.io/styleguide/cppguide.html) prefers int.

include/benchmark/reporter.h
```diff
@@ -132,22 +133,22 @@ class BenchmarkReporter {
     error_stream_ = err;
   }
 
-  std::ostream& GetOutputStream() const {
+   std::ostream& GetOutputStream() const {
```
dominichamon: there's an extra space here, and extra whitespace below.

ismaelJimenez: fixed

include/benchmark/reporter.h
ismaelJimenez: fixed

dominichamon:
starting the manual merge

PR #253, opened 2016-07-12
Add missing `volatile`s to 32-bit ARM cycleclock assembler.

enh:
Without these, clang reorders these instructions as if they were
regular loads/stores which causes SIGILL from the kernel because
it performs all the loads before it starts testing the values.


enh:
(if you're interested, this fixes internal bug http://b/30076265.)

PR #257, opened 2016-07-19
Support multiple ranges in the benchmark

loganek:
google-benchmark library allows to provide up to two ranges to the
benchmark method (range_x and range_y). However, in many cases it's not
sufficient. The patch introduces multi-range features, so user can easily
define multiple ranges by passing a vector of integers, and access values
through the method range(i).

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -153,6 +153,8 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <stddef.h>
 #include <stdint.h>
 
+#include <vector>
+
 #include "macros.h"
 
 namespace benchmark {
@@ -258,7 +260,7 @@ typedef double(BigOFunc)(int);
 // benchmark to use.
 class State {
 public:
-  State(size_t max_iters, bool has_x, int x, bool has_y, int y,
+  State(size_t max_iters, const std::vector<int>& ranges,
         int thread_i, int n_threads);
 
   // Returns true if the benchmark should continue through another iteration.
@@ -414,16 +416,18 @@ public:
   // Range arguments for this run. CHECKs if the argument has been set.
   BENCHMARK_ALWAYS_INLINE
   int range_x() const {
-    assert(has_range_x_);
-    ((void)has_range_x_); // Prevent unused warning.
-    return range_x_;
+    return range(0);
   }
 
   BENCHMARK_ALWAYS_INLINE
   int range_y() const {
-    assert(has_range_y_);
-    ((void)has_range_y_); // Prevent unused warning.
-    return range_y_;
+      return range(1);
+  }
+
+  BENCHMARK_ALWAYS_INLINE
+  int range(std::size_t pos) const {
+      assert(range_.size() > pos);
+      return range_[pos];
   }
 
   BENCHMARK_ALWAYS_INLINE
@@ -434,11 +438,7 @@ private:
   bool finished_;
   size_t total_iterations_;
 
-  bool has_range_x_;
-  int range_x_;
-
-  bool has_range_y_;
-  int range_y_;
+  std::vector<int> range_;
 
   size_t bytes_processed_;
   size_t items_processed_;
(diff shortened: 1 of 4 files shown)


test/multiple_ranges_test.cc
```diff
@@ -0,0 +1,14 @@
...
+void BM_empty(benchmark::State& state) {
+  while (state.KeepRunning()) {
+    int product = state.range(0) * state.range(1) * state.range(2);
+    for (int x = 0; x < product; x++) {
+      benchmark::DoNotOptimize(x);
+    }
+  }
+}
+
+BENCHMARK(BM_empty)->RangeMultiplier(2)->Ranges({1, 3, 5}, {2, 7, 15})->Args({7, 6, 3});
```
dominichamon: why this and not `Ranges({{1,2},{3,7},{5,15}})` ? ie, a vector of pairs instead of a pair of vectors. 

i find this more intuitive, and it's slightly easier to process too.

dominichamon:
i really like this, thank you. however, there may be a limitation on using std::vector in the api header. @EricWF for more details on this.

test/multiple_ranges_test.cc
loganek: @dominichamon I agree, think your API is more intuitive. I've amended my patch

include/benchmark/benchmark_api.h
```diff
@@ -414,16 +416,18 @@ class State {
   // Range arguments for this run. CHECKs if the argument has been set.
   BENCHMARK_ALWAYS_INLINE
   int range_x() const {
```
dominichamon: we should mark these as deprecated as they're largely useless with your change. at some point  we should remove them.

src/benchmark.cc
```diff
@@ -371,12 +370,13 @@ class BenchmarkImp {
 
   static void AddRange(std::vector<int>* dst, int lo, int hi, int mult);
 
+  int ArgsCnt() const { return args_.size() ? static_cast<int>(args_.front().size()) : -1; }
```
dominichamon: args_.empty() is a more accurate check.

test/multiple_ranges_test.cc
```diff
@@ -0,0 +1,14 @@
+#include "benchmark/benchmark.h"
+
+void BM_empty(benchmark::State& state) {
+  while (state.KeepRunning()) {
+    int product = state.range(0) * state.range(1) * state.range(2);
+    for (int x = 0; x < product; x++) {
```
dominichamon: you should probably test that the product is what you expect.. or even better, that each range(x) is what you expect.

include/benchmark/benchmark_api.h
loganek: @dominichamon Do we have any tag for documentation, like doxygen's https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmddeprecated ?

test/multiple_ranges_test.cc
loganek: Added assertions

src/benchmark.cc
loganek: Yeah, I agree. Fixed it.

include/benchmark/benchmark_api.h
dominichamon: we do not. a comment will do for now, or you could just remove them. anyone building this from head should be open to breaking changes, and we'll make sure to add this to any release notes for future releases.

loganek: I've removed range_x() and range_y().
Also, ArgPair() and RangePair() have been removed.

include/benchmark/benchmark_api.h
```diff
@@ -153,6 +153,8 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <stddef.h>
 #include <stdint.h>
 
+#include <vector>
```
dominichamon: @EricWF is this an issue for the public API and your use in libc++?

loganek:
@dominichamon @EricWF ping

include/benchmark/benchmark_api.h
EricWF: No issue. So long as STL stuff doesn't affect the timing results then I think we should encourage it's use.

dominichamon:
i'm going to have a crack at resolving conflicts and do some local testing before merging.

loganek:
@dominichamon I've just rebased to master and resolved conflicts. Please test my changes and let me know if I should do some improvements.
Thanks a lot

PR #258, opened 2016-07-21
Add --benchmark_output=<filename> option.

EricWF:
This patch adds support for writing benchmark results to a file. The file is specified using `--benchmark_output=<filename>`. The output format used is specified by `--benchmark_output_format={json:csv:console}`.

This patch is intended to help support additional tooling which needs to be able to access the data in a less human friendly manner than the colored console reporter.

Specifying `--benchmark_output` does not change the console output in any way. This is useful when you want to both (A) watch the output and (B) save it for later.

In order to support the `console` output format when writing to files I had to clean up how `ConsoleReporter` handles color printing.

I have no idea how to test this change since the tests can't (shouldn't?) be creating files. Suggestions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,9 +495,9 @@ name,iterations,real_time,cpu_time,bytes_per_second,items_per_second,label
 
 ## Output Files
 The library supports writing the output of the benchmark to a file specified
-by `--benchmark_output=<filename>`. The format of the output can be specified
-using `--benchmark_output_format={json|console|csv}`. Specifying
-`--benchmark_output` does not suppress the console output.
+by `--benchmark_out=<filename>`. The format of the output can be specified
+using `--benchmark_out_format={json|console|csv}`. Specifying
+`--benchmark_out` does not suppress the console output.
 
 ## Debug vs Release
 By default, benchmark builds as a debug library. You will see a warning in the output when this is the case. To build it as a release library instead, use:
--- a/README.md
+++ b/README.md
@@ -427,10 +427,10 @@ static void BM_test(benchmark::State& state) {
 
 ## Output Formats
 The library supports multiple output formats. Use the
-`--benchmark_format=<tabular|json|csv>` flag to set the format type. `tabular` is
-the default format.
+`--benchmark_format=<console|json|csv>` flag to set the format type. `console`
+is the default format.
 
-The Tabular format is intended to be a human readable format. By default
+The Console format is intended to be a human readable format. By default
 the format generates color output. Context is output on stderr and the 
 tabular data on stdout. Example tabular output looks like:
 ```
@@ -492,6 +492,12 @@ name,iterations,real_time,cpu_time,bytes_per_second,items_per_second,label
 "BM_SetInsert/1024/8",116606,18810.1,9766.64,3.27646e+06,819115,
 "BM_SetInsert/1024/10",106365,17238.4,8421.53,4.74973e+06,1.18743e+06,
 ```
+
+## Output Files
+The library supports writing the output of the benchmark to a file specified
+by `--benchmark_output=<filename>`. The format of the output can be specified
+using `--benchmark_output_format={json|console|csv}`. Specifying
+`--benchmark_output` does not suppress the console output.
 
 ## Debug vs Release
 By default, benchmark builds as a debug library. You will see a warning in the output when this is the case. To build it as a release library instead, use:
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -165,11 +165,16 @@ void Initialize(int* argc, char** argv);
 // of each matching benchmark. Otherwise run each matching benchmark and
 // report the results.
 //
-// The second overload reports the results using the specified 'reporter'.
+// The second and third overload use the specified 'console_reporter' and
+//  'file_reporter' respectively. 'file_reporter' will write to the file specified
+//   by '--benchmark_output'. If '--benchmark_output' is not given the
+//  'file_reporter' is ignored.
 //
 // RETURNS: The number of matching benchmarks.
 size_t RunSpecifiedBenchmarks();
-size_t RunSpecifiedBenchmarks(BenchmarkReporter* reporter);
+size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter);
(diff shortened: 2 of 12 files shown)


EricWF:
I honestly have no clue what's causing AppVeyorBot to fail :-S

dominichamon:
appveyor failure is some weirdness with ctest. i'm working on a fix (slowly) but not making much progress.

README.md
```diff
@@ -493,6 +493,12 @@ name,iterations,real_time,cpu_time,bytes_per_second,items_per_second,label
 "BM_SetInsert/1024/10",106365,17238.4,8421.53,4.74973e+06,1.18743e+06,
 ```
 
+## Output Files
+The library supports writing the output of the benchmark to a file specified
+by `--benchmark_output=<filename>`. The format of the output can be specified
+using `--benchmark_output_format={json|console|csv}`. Specifying
+`--benchmark_output` does not suppress the console output.
```
dominichamon: why not just invoke and pipe to tee then? i'm not sure that this library should be writing files at all, given that every OS has a way of redirecting output to a file.

EricWF: There's a big difference between output for humans and output for tooling. Simply put I can't grok JSON and tooling can't grok colored console output, so I don't want to choose between the two. I want the both.

I agree the library shouldn't write to files in general, but `-o output` is a pretty universal feature.

dominichamon: how often do you think someone would want to push the output to tooling and have it human readable? I'd expect any tooling requirement to override humans reading the output. ie, i think if you're going to push the output to a tool that tool should take care of the human readable bit too. whether that's a summary, or reformatting the raw results.

ie, i'd expect `$ benchmark_my_stuff --benchmark_format=json | analyse_my_bench`

to provide me with human readable analysis results. if i need the output for debugging, then `tee` or some equivalent would at least give me something to look at after the fact.

```
$ benchmark_my_stuff --benchmark_output=out.json --benchmark_output_format=json --benchmark_format=csv
```

it's very easy to get wrong. i mean, output is also to stdout, so output_format vs format is subtle.

this may just be bikeshedding the arguments, i guess. i'm wary of redundancy and confusion.

EricWF:
@dominichamon Do you have an issue with the coverage drop here? I don't want to write tests that create output files since it's messy to clean them up.

dominichamon:
no, i have no issue with the coverage drop or trying to test code that writes files. i'm still wary of the command line flag confusion.

README.md
EricWF: > how often do you think someone would want to push the output to tooling and have it human readable? 

I do it all the time. I want to see the benchmark results as the come in, and I also want to do something with the results later.

> , i think if you're going to push the output to a tool that tool should take care of the human readable bit too. 

I think that tool should take care of the human readable input for the information it presents, I don't expect the tool to repeat the benchmark output. 

> ie, i'd expect $ benchmark_my_stuff --benchmark_format=json | analyse_my_bench
> to provide me with human readable analysis results. if i need the output for debugging, then > tee or some equivalent would at least give me something to look at after the fact.

I wouldn't want to require "analyse_my_bench" to know how to pretty print basic benchmark output better than the library can already do.

> ```$ benchmark_my_stuff --benchmark_output=out.json --benchmark_output_format=json --benchmark_format=csv```
> it's very easy to get wrong. i mean, output is also to stdout, so output_format vs format is subtle.
> i'm wary of redundancy and confusion.

I'm sympathetic to that. A better way would be nice, or better names. However I don't think offering two output methods is too confusing. This behavior mirrors how LLVM's test runner LIT works. For example `$ lit -o out.json test.cpp` will give human readable output to stdout, and will print a copy of the JSON results to out.json.

In summary I agree with the general concern over confusion, but I think the feature is still useful to have.

dominichamon: ok, i'll buy that. bikeshedding then:

--benchmark_out: the file to which to write results
--benchmark_out_format: the format in which to write the results to --benchmark_out, if set
--benchmark_format: the format in which to write the results to stdout

?

src/benchmark.cc
```diff
@@ -1025,10 +1040,17 @@ std::unique_ptr<BenchmarkReporter> GetDefaultReporter() {
 } // end namespace internal
 
 size_t RunSpecifiedBenchmarks() {
```
dominichamon: why overload instead of using default arguments?

include/benchmark/reporter.h
```diff
@@ -157,13 +157,18 @@ class BenchmarkReporter {
 // default reporter used by RunSpecifiedBenchmarks().
 class ConsoleReporter : public BenchmarkReporter {
  public:
+  explicit ConsoleReporter(bool color_output = true)
```
dominichamon: one rather nice trick for avoiding boolean parameters (which can be confusing at the callsite) is to define an enum with two values.

src/benchmark.cc
EricWF: Good question: It's to avoid the "zero as null pointer constant" warning GCC emits in C++98 mode. 
Optimally we would use "= nullptr" as the default argument, but that's C++11 only, and using either "NULL" or "0" as the default value will trigger the warning.

I would prefer to use default arguments though.

README.md
EricWF: Sounds good to me. Changes incoming.

include/benchmark/reporter.h
EricWF: Ack. Changed in the newest revision.

EricWF:
@googlebot confirmed.

include/benchmark/benchmark_api.h
```diff
@@ -489,9 +494,9 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Range(int start, int limit);
 
-  // Run this benchmark once for every value in the range [start..limit]
+  // Run this benchmark once for all values in the range [start..limit] with specific step
```
dominichamon: why is this part of this PR?

dominichamon:
LGTM

include/benchmark/benchmark_api.h
EricWF: It's just git messing up history. This change is already in master, but I merged and caught the commit. I'll commit this patch without history.

EricWF:
Sounds good. I'm going to commit directly to involve getting the history.

PR #259, opened 2016-07-21
Add `RegisterBenchmark(name, func, args...)`  for creating/registering benchmarks.

EricWF:
The current method of registering benchmarks can be quite limiting. Currently registration must be done at the global scope during program startup, and each benchmark must be registered separately (on a separate line). For the simple cases this works OK, but quickly gets unwieldy as the number of benchmarks grows.

The solution is simple. We should allow benchmarks to be created and registered at a local (non-global) scope.  This patch adds `RegisterBenchmark(name, func, args...)` , which can be used to create and register benchmarks anywhere at any time and with any arguments. This is infinity more flexible than the current API.

The signature of `RegisterBenchmarks` is:

```c++
template <class Functor, class ...Args>
internal::Benchmark* RegisterBenchmark(const char* name, Functor&& fn, Args&&... args);
```
The function creates a new benchmark with the specified `name`, registers it, and returns a non-owning pointer to it. When the benchmark is run it will invoke `fn(st, args...)` where `st` is the `State` object. 

Example Usage:

```c++
auto BM_test = [](benchmark::State& st, auto Inputs) { /* ... */ };

  for (auto& test_input : { /* ... */ })
      benchmark::RegisterBenchmark(test_input.name(), BM_test, test_input);
  benchmark::RunSpecifiedBenchmarks();
}
```


EricWF:
This patch expects fully conformant C++11 compilers, which excludes most of our bots unfortunately. However it's been half a decade since C++11 and I think we should start using it.

include/benchmark/benchmark_api.h
```diff
@@ -587,6 +592,20 @@ class Benchmark {
...
+
+// Create and register a benchmark with the specified 'name' that invokes
+// the specified functor 'fn'.
+//
+// RETURNS: A pointer to the registered benchmark.
+internal::Benchmark* RegisterBenchmark(const char* name, internal::Function* fn);
+
+#if defined(BENCHMARK_HAS_CXX11)
+template <class Lambda>
+internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn);
```
dominichamon: why return the pointer?

dominichamon:
no matter how long it has been, we have users who do not have C++11 support. dropping them would be unfortunate.

having said that, if we state that as of version X we are only supporting C++11, that just leaves them on an old (working) version. What do you think?

EricWF:
> no matter how long it has been, we have users who do not have C++11 support. dropping them would be unfortunate.

I agree. I use the public API in C++03 within libc++. I don't want to drop support for that. Note that we already require C++11 to build/test the library, so we've already "dropped" C++03 support in that regard. Users who do not have "C++11 support" when benchmarking their project often do not have it because their project doesn't support C++11.  The same users can often find a pre-compiled version of the library or can use a newer compiler to build the library.

> having said that, if we state that as of version X we are only supporting C++11, that just leaves them on an old (working) version. What do you think?

I don't think we should kill C++03 compatibility in the headers and instead we should maintain the core API we have now so that it stays C++03 compatible. Therefore all old C++03 tests will keep working *and* they can continue upgrading to newer versions of the library. However I also want to allow new API additions to use C++11.  So I think the rule should be "Our core API is C++03 compatible, but additional features may use C++11 where appropriate so long as they don't conflict with the core API."

However I think we should be more aggressive in the compilers we require to build/test the library.
If we want to allow C++11 features we need to test against quality C++11 compilers, otherwise the tests will fail. Currently Jenkins uses GCC 4.6, and this is *NOT* a C++11 compiler. (GCC 4.7 isn't much better)

My suggestion is that we upgrade Jenkins to GCC 4.9. This way we can test new C++11 features without running into compiler bugs. However it would be nice if the library continued to build with older GCC versions even if they can't run all of the tests.

dominichamon:
oh, poor gcc 4.6. only around due to the long term ubuntu support, iirc.

i completely agree with everything you've written.

dominichamon:
trying to find a way to kick travis into building a new version with the latest travis...

dominichamon:
ah, it passed the 4.8 anyway.

EricWF:
I just pushed a merge commit. Hopefully that will get things going.

EricWF:
Ah nevermind...

PR #260, opened 2016-07-22
Add support for step in DenseRange()

drozdvadym:
This simple pull-request adds support for step in DenseRange function.


drozdvadym:
I signed it!

PR #261, opened 2016-07-22
Upgrade travis GCC version to 4.8.4

EricWF:
This patch attempts to upgrade the Travis CI to use GCC 4.8.4 instead of GCC 4.6. It does this by upgrading the Travis CI to run on Ubuntu Trusty which provides the CMake 2.8.12, GCC 4.8.4 and Clang 3.5. 

This patch is part of dropping support for building/testing with C++0x compilers.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -489,9 +489,9 @@ public:
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Range(int start, int limit);
 
-  // Run this benchmark once for every value in the range [start..limit]
+  // Run this benchmark once for all values in the range [start..limit] with specific step
   // REQUIRES: The function passed to the constructor must accept an arg1.
-  Benchmark* DenseRange(int start, int limit);
+  Benchmark* DenseRange(int start, int limit, int step = 1);
 
   // Run this benchmark once with "x,y" as the extra arguments passed
   // to the function.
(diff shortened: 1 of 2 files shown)


EricWF:
Urg... The patch looks like it works but I think "complexity_test.cc" is flaky.

.travis.yml
```diff
@@ -1,26 +1,23 @@
...
         - compiler: gcc
-          env: COMPILER=g++-4.6     STD=c++0x BUILD_TYPE=Coverage
-        - compiler: gcc
-          env: COMPILER=g++-4.6     STD=c++0x BUILD_TYPE=Debug
-        - compiler: gcc
-          env: COMPILER=g++-4.6     STD=c++0x BUILD_TYPE=Release
+          env: COMPILER=g++         STD=c++11 BUILD_TYPE=Debug
         - compiler: gcc
-          env: COMPILER=g++-4.8     STD=c++11 BUILD_TYPE=Debug
+          env: COMPILER=g++         STD=c++11 BUILD_TYPE=Debug
```
dominichamon: what's the difference between this and the one above?

EricWF: This line was lying. "COMPILER=4.8" is ignored entirely. If you look at the output you'll see this was invoking 4.6

EricWF: The COMPILER bit is only helpful in naming the compiler in diagnostic output. Otherwise it has no effect.

dominichamon: sorry, by "above" i meant within this file.

lines 9-10 and 11-12 are identical. right? can we remove one stanza?

EricWF: Ah good catch. I changed the top line to read "BUILD_TYPE=COVERAGE" since that's what it should have been.

Issue #240, opened 2016-06-15
support for additional counters (eg HW counters via PAPI)


biojppm:
Hi all,

First of all, thanks for making this library available.

I'm interested in adding additional counters. Time/manual time would still be the main counter, but additionally, I would like to see more counters such as cache or branch prediction misses (in some cases) and shader performance parameters (in other cases). If using PAPI, for example, the results would be something like this:

```
Benchmark              Time(ns)  CPU(ns) Iterations  PAPI_L1_DCM  PAPI_L2_DCM  PAPI_L3_TCM
------------------------------------------------------------------------------------------
BM_SetInsert/1024/1       28928    29349      23853            2            0            0
BM_SetInsert/1024/8       32065    32913      21375            4            2            1
BM_SetInsert/1024/10      33157    33648      21431           30           26           10
```

Above, I've put only cache miss counters. There are more varieties of counters, and not all are interesting to each microbenchmark. So ideally, the actual counters to use would be specifiable in a per-benchmark basis. If that would be too disruptive, then specifying them per-fixture or even per-session (ie once per executable) would also work.

I'm using PAPI here as a common example; this would be a client dependency and the client would be responsible for getting those counters and reporting them to googlebenchmark.

So in essence, this boils down to adding (string) named counters and getting the printers to show them. On the client side I am thinking of doing something like

```
static void BM_SetInsert(benchmark::State& state) {
  while (state.KeepRunning()) {
    //...
  }

  // get counter values
  double cL1, cL2, cL3;
  // ... call PAPI functions
  // report additional counters
  state.AddCounter("PAPI_L1_DCM", cL1);
  state.AddCounter("PAPI_L2_DCM", cL2);
  state.AddCounter("PAPI_L3_TCM", cL3);
}
```
Using c++11 initializer lists, the three AddCounter() calls could be reduced to:
```
  state.AddCounters({{"PAPI_L1_DCM", cL1}, {"PAPI_L2_DCM", cL2}, {"PAPI_L3_TCM", cL3}});
```
I've put double as the type of the counter as it is general enough to accomodate both integer and float-type counters. Also I'm using strings above but if this is too slow, then the client could pre-define the names in a fixture (while getting an integer handle for the name) and offer also an interface working strictly with those integer handles.

So, is this something you're interested in? Would it be too disruptive?

Thanks,
João

dominichamon:
[...]

PR #262, opened 2016-07-23
Add user-defined counters.

biojppm:
This PR adds user-defined counters, as discussed in issue #240 . I've added usage notes to the README.md file.

This was implemented mainly by adding two classes: `BenchmarkCounters` which is responsible for managing the collection of counters and `Counter` which contains logic for each individual counter. `BenchmarkCounters` objects can be summed up to aggregate multithread benchmark results. Then I added a `BenchmarkCounters` member to `State` and `Run`, which are accessible as needed.

Though the changes in this PR generally adhere to the gist of the discussion in issue #240 , I've opted to leave the bytes_processed and items_processed as they previously were. In an early commit I did move these quantities to the `BenchmarkCounters` class, but otherwise noted that these changes were maybe a bit too far-reaching for what I wanted to do here. As an illustration of the resulting approach, I left some of this code commented out (eg, see src/benchmark.cc:470).

I added two methods to the `Fixture` class: `InitState()` and `TerminateState()`. These were needed to change the state before and after the benchmark. They receive a non-const reference to benchmark::State so that they start and stop hardware counters as needed. I did this instead of changing the State constness in `Setup()` and `TearDown()` to retain compatibility. See src/benchmark.cc:758.

The JSON reporter was the easiest to adapt. The Console and CSV reporters had the part printing the header moved out of `Reporter::PrintContext()`. In the console reporter, I store the previously-printed header, and check on each Run whether the header is different; if it is then a new header is printed. The printing of header inside the table was not immediately clear, so I added another separator line filled with '-' before the names; this makes the in-table headers more visible by separating them more clearly from the results above. In the CSV reporter, I gather first all the fields in a set, and then print counters (or nothing) based on this set. This set is local so that part is done outside of the `PrintRunData()` method which I promptly recognize as awkward. See src/csv_reporter.cc:80 below.

Note that I needed to add `#includes of <utility>,<vector>,<string> and <initializer_list>` (this last only when compiling in C++11) to the benchmark_api.h file. If you do not want this, let me know, or just feel free to suggest alternative approaches that would obviate these includes.

All the existing unit tests are passed successfully. I added also another test showing use with the PAPI library, but will do a separate PR for that, after this one is closed. Also, I did not add new specific tests for the user counters. If this PR is approved, then we'll discuss what to test for.

Please let me know if something in the code of this PR should be changed. For example, maybe `BenchmarkCounters` could be called `UserCounters`. Also, I tried to be faithful to the style of this project, but if for any reason something was overlooked feel free to point it out; I'll gladly correct it.


biojppm:
gcc 4.6 is failing in the use of override. That is easy to fix. 
(It's also failing in previously existing tests that call `std::this_thread::sleep_for()`.) 
OTOH, there's also PR #261 which apparently will stop requiring compilation with gcc 4.6.

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +234,113 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+class BenchmarkCounters {
+public:
+
+  typedef Counter                         value_type;
+  typedef std::vector< Counter >          container_type;
+  typedef container_type::const_iterator  const_iterator;
+
+public:
+
+  BenchmarkCounters(size_t initial_capacity = 16);
```
dominichamon: don't use default arguments please. two separate constructors would be best.

and don't forget to mark the one-argument constructor as explicit!

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +234,113 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+
+  // Add/Set a counter. Will overwrite any additional counter properties.
+  size_t  Set(Counter const& c);
+  // Add/Set a counter. When the counter already exists, f is ignored.
+  size_t  Set(const char*        n, double v, uint32_t f = Counter::kDefaults);
+  // Add/Set a counter. When the counter already exists, f is ignored.
+  size_t  Set(std::string const& n, double v, uint32_t f = Counter::kDefaults);
```
dominichamon: if we could avoid std::string for the benchmark_api.h, that'd be great

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +234,113 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+
+#ifdef BENCHMARK_INITLIST
```
dominichamon: in #259, @EricWF adds a BENCHMARK_HAS_CXX11 macro which you should consider using.

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +234,113 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  void    Set(std::initializer_list< Counter > il) { for(auto &c : il) { Set(c); } }
+#endif
+
+  // Returns the index where name is located, or size() if the name was not found.
+  size_t  Find(const char*        name) const;
```
dominichamon: please don't add these spaces for alignment. go with the style throughout the rest of the code.

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +234,113 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  Counter const& GetItemsPerSecond() const;
+  size_t  BytesPerSecond() const;
+  size_t  ItemsPerSecond() const;
+  void    BytesPerSecond(size_t b);
+  void    ItemsPerSecond(size_t i);
+*/
+
+private:
+
+  std::vector< Counter > counters_;
```
dominichamon: i'm not sure having this private is necessary. you already have methods above for adding/setting/getting. why not just have this public?

include/benchmark/benchmark_api.h
```diff
@@ -429,6 +543,35 @@ class State {
...
+  size_t SetCounter(Counter const& c) { return counters_.Set(c); }
+  // Add/set a user counter. Return the counter handle.
+  BENCHMARK_ALWAYS_INLINE
+  size_t SetCounter(const char* n, double v = 0., uint32_t f = Counter::kDefaults) { return counters_.Set(n, v, f); }
+  // Add/set a user counter. Return the counter handle.
+  BENCHMARK_ALWAYS_INLINE
+  size_t SetCounter(std::string const& n, double v = 0., uint32_t f = Counter::kDefaults) { return counters_.Set(n, v, f); }
+  // Given a counter handle, set a previously existing user counter.
+  BENCHMARK_ALWAYS_INLINE
+  void   SetCounter(size_t counter_id, double v) { counters_.Get(counter_id).Set(v); }
```
dominichamon: consider reducing this API surface to the essential. If we want users to define a Counter type, then don't supply overrides to do it for them. it just adds complexity that's unnecessary.

include/benchmark/benchmark_api.h
```diff
@@ -609,11 +756,15 @@ class Fixture: public internal::Benchmark {
...
     virtual void Run(State& st) {
       this->SetUp(st);
+      this->InitState(st);
       this->BenchmarkCase(st);
+      this->TerminateState(st);
       this->TearDown(st);
     }
 
     virtual void SetUp(const State&) {}
+    virtual void InitState(State &) {}
```
dominichamon: i need more details why these are necessary, sorry. if the above API can add counters as well as set them, why would you need to set anything up in the fixture?

include/benchmark/benchmark_api.h
dominichamon: std::map<const char*, Counter> might be better. You can avoid the 'find' cost on every API access.

src/benchmark.cc
```diff
@@ -753,6 +963,7 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
     total->bytes_processed += st.bytes_processed();
     total->items_processed += st.items_processed();
     total->complexity_n += st.complexity_length_n();
+    counters_total->Sum(st.Counters());
```
dominichamon: why aren't counters part of ThreadStats?

include/benchmark/benchmark_api.h
biojppm: ok, will do.

include/benchmark/benchmark_api.h
biojppm: I used std::string as per your suggestion in #240 . I acknowledge that does not mean that it should be exposed in the public header. Would you be OK with a PIMPL idiom here? That would hide the need for including both <vector> and <string> .

include/benchmark/benchmark_api.h
biojppm: I went with vector first precisely to minimize API lookup cost: when the user wants, he can store the handle and get O(1) lookup with the handle. Also, when the user does not bother doing that, for small numbers of counters (low dozens) as is likely here, even a linear lookup through the vector will probably be better here because of cache behavior. For example, data here: eg http://scottmeyers.blogspot.pt/2015/09/should-you-be-using-something-instead.html .

Not that lookup cost is paramount here. But anyway, if despite this you still want a map, say so and I'll change this.

include/benchmark/benchmark_api.h
biojppm: ok.

include/benchmark/benchmark_api.h
biojppm: These were needed to be able to change the state before and after the benchmark, AND outside of the benchmark. It wouldn't be possible to start or stop the counters in SetUp() and TearDown() because they receive a const reference to State, which is where the counters are. The solutions could be:

- Remove const from the State argument to the SetUp() and TearDown() of Fixture. I chose not to do this because it was more disruptive to how the public API was. But if this was my own project, I would have gone for this.
- Make the counters interface mutable in the State class, ie add const to the methods adding counters. I chose not to do this because it would change the const semantics of a significant part of the class.
- Finally, adding these two methods. This has the advantage of being merely additive to the existing API and does not have the constness problems above.

If you prefer a different solution let me know!

src/benchmark.cc
```diff
@@ -302,6 +302,214 @@ static std::unique_ptr<TimerManager> timer_manager = nullptr;
 
 } // end namespace
 
+
+
+Counter::Counter(std::string const& n, double v, uint32_t f)
```
biojppm: cool. ok.

src/benchmark.cc
biojppm: Good suggestion. I'll move them there. (This happened because in my initial implementation there was no ThreadStats, then when I rolled back it was just quicker).

include/benchmark/benchmark_api.h
biojppm: As a usage example, here's a PAPI fixture class. Note that the counters are started and stopped in these new Fixture methods, and not inside the actual benchmark.

```
#include "benchmark/benchmark_api.h"                                                                                     
                                                                                                                         
#include "papi.h"                                                                                                        
                                                                                                                         
#include <random>                                                                                                        
#include <algorithm>                                                                                                     
#include <iostream>                                                                                                      
                                                                                                                         
                                                                                                                         
class PAPIFixture : public ::benchmark::Fixture {                                                                        
public:                                                                                                                  
                                                                                                                         
  void SetUp(const benchmark::State& st) {                                                                               
    values.resize(st.range_x());                                                                                         
    sequence.resize(st.range_x());                                                                                       
    int count = 0;                                                                                                       
    for(int &i : sequence) i = count++;                                                                                  
    std::random_device rng;                                                                                              
    std::mt19937 urng(rng());                                                                                            
    std::shuffle(sequence.begin(), sequence.end(), urng);
  }

  void InitState(benchmark::State &) {
    if(PAPI_start_counters(&event_ids[0], (int)event_ids.size()) != PAPI_OK) {
        std::cerr << "could not start PAPI counters\n";
    }
  }
  void TerminateState(benchmark::State &st) {
    if(PAPI_stop_counters(&events[0], (int)events.size()) != PAPI_OK) {
        std::cerr << "could not get PAPI counters\n";
    }
    for(int i = 0, s = (int)events.size(); i < s; i++) {
      st.SetCounter(event_names[i], (double)events[i]);
    }
  }

  std::vector< int          > event_ids   = { PAPI_L1_DCM ,  PAPI_L2_DCM ,/*  PAPI_L3_DCM ,  PAPI_MEM_SCY ,*/  PAPI_FP_INS };
  std::vector< const char * > event_names = {"PAPI_L1_DCM", "PAPI_L2_DCM",/* "PAPI_L3_DCM", "PAPI_MEM_SCY",*/ "PAPI_FP_INS"};
  std::vector< long_long    > events      = { 0           ,  0           ,/*  0           ,  0            ,*/  0           };


  //--------------------------------------------------------------

  float reduce_linear() const {
    float sum = 0.;
    for(size_t i = 0, s = values.size(); i < s; ++i) {
      sum += values[i];
    }
    return sum;
  }
  float reduce_jumps() const {
    float sum = 0.;
    for(size_t i = 0, s = values.size(); i < s; ++i) {
        sum += values[sequence[i]];
    }
    return sum;
  }

  std::vector< int > values;
  std::vector< int > sequence;

};



BENCHMARK_DEFINE_F(PAPIFixture, ReduceLinear)(benchmark::State& st) {
  float sum = 0.;
  while (st.KeepRunning()) {
    benchmark::DoNotOptimize(sum);
    sum = reduce_linear();
  }
}
BENCHMARK_DEFINE_F(PAPIFixture, ReduceJumps)(benchmark::State& st) {
  float sum = 0.;
  while (st.KeepRunning()) {
    benchmark::DoNotOptimize(sum);
    sum = reduce_jumps();
  }
}
BENCHMARK_REGISTER_F(PAPIFixture, ReduceLinear)->Arg(1<<6 );
BENCHMARK_REGISTER_F(PAPIFixture, ReduceJumps )->Arg(1<<6 );
BENCHMARK_REGISTER_F(PAPIFixture, ReduceLinear)->Arg(1<<8 );
BENCHMARK_REGISTER_F(PAPIFixture, ReduceJumps )->Arg(1<<8 );
BENCHMARK_REGISTER_F(PAPIFixture, ReduceLinear)->Arg(1<<10);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceJumps )->Arg(1<<10);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceLinear)->Arg(1<<12);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceJumps )->Arg(1<<12);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceLinear)->Arg(1<<20);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceJumps )->Arg(1<<20);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceLinear)->Arg(1<<24);
BENCHMARK_REGISTER_F(PAPIFixture, ReduceJumps )->Arg(1<<24);

BENCHMARK_MAIN()
```

include/benchmark/benchmark_api.h
biojppm: cool.

include/benchmark/benchmark_api.h
biojppm: ok, sorry for that.

include/benchmark/benchmark_api.h
biojppm: FWIW, in README.md I added some lines explaining this too.

include/benchmark/benchmark_api.h
biojppm: Please confirm :-).

dominichamon: pimpl would be fine, i think. @EricWF can confirm, but anything in the _api header should avoid as much of the standard library as possible (as it's used to benchmark libc++).

include/benchmark/benchmark_api.h
dominichamon: True, the number of counters will be small. If we think that 'id' is the best way then that's what we should expose to users. ie, they can only look up counters by id and they have to store them. what do you think?

include/benchmark/benchmark_api.h
biojppm: ok pimpl it is.

include/benchmark/benchmark_api.h
biojppm: I wanted to give choice to clients: id maybe the best way in some cases, but OTOH, it does require the client to store the handle. This is cumbersome for vanilla cases where lookup cost is negligible. So to make things agreeable in both cases, I always return the handle and the user can use or not via the handle overloads, `Get(size_t)`. 

So I think the benefit of efficiently dealing with both these usage scenarios outweighs the API cost of exposing the different `Get()` overloads. But it's fine if you disagree and prefer one approach only. In that case, I would remove the handle overload instead of forcing it on the user.

include/benchmark/benchmark_api.h
biojppm: Just to confirm: are you fine with this?

src/csv_reporter.cc
```diff
@@ -46,20 +48,51 @@ std::vector<std::string> elements = {
...
+    // would also require the include.
+    // So I'll defer judgment here.
+    for (const auto &name : user_counter_names) {
+      Out << ",";
+      if(run.counters.Exists(name)) {
+        Out << run.counters.Get(name).Value();
+      }
+    }
+    Out << '\n';
+  }
```
biojppm: Are you OK with this solution here?

biojppm: It's a bit of a hack.

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +231,126 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+
+  const_iterator begin() const { return counters_; }
+  const_iterator end () const { return counters_ + num_counters_; }
+
+private:
+
+  // using raw pointers here to avoid inclusion of std::vector in the public api
+  Counter * counters_;
+  size_t num_counters_;
+  size_t capacity_;
```
biojppm: NOTE: this fixes the <vector> include. I ended up using raw pointers instead of PIMPL. This was validated in valgrind and passed all tests so far.

README.md
```diff
@@ -277,6 +277,8 @@ static void BM_ManualTiming(benchmark::State& state) {
 BENCHMARK(BM_ManualTiming)->Range(1, 1<<17)->UseManualTime();
 ```
 
+
+
```
EricWF: Unnecessary white space change.

README.md
```diff
@@ -393,6 +395,76 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+  // Set the counter as a thread-average quantity. It will
+  // be presented divided by the number of threads. */
+  state.SetCounter("FooRate", numFoos, benchmark::Counter::kAvgThreads);
+
+  // the third argument is a bitwise mask, so you can combine.
+  // This will mark as a thread-average rate:
+  state.SetCounter("FooRate", numFoos, benchmark::Counter::kIsRate|benchmark::Counter::kAvgThreads);
+```
+
+Note that setting the counter flags only works on the first call to `state.SetCounter()`; any subsequent values will be silently ignored.
```
EricWF: We should absolutely diagnose this in debug mode, but probably in release mode as well.

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +231,126 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 // FIXME Add ClobberMemory() for non-gnu compilers
 #endif
 
+
+
+class Counter {
+public:
+
+  typedef enum {
```
EricWF: What's wrong with
```c++
enum Flags_e {
  kDefaults = 0,
  kIsRate = 1,
  kisRate = 2
};
```

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +231,126 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  bool SameName(const char *n) const;
+
+  Counter& operator+= (double v) { value_ += v; return *this; }
+  Counter& operator-= (double v) { value_ -= v; return *this; }
+  Counter& operator*= (double v) { value_ *= v; return *this; }
+  Counter& operator/= (double v) { value_ /= v; return *this; }
+
+private:
+
+  // using raw C strings here to avoid inclusion of std::string in the public api
```
EricWF: std::string is fine now. Please use that instead.

include/benchmark/benchmark_api.h
```diff
@@ -227,6 +231,126 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  char name_buf_[16]; // strings smaller than 15 characters are stored here
+  char* name_mem_;    // whereas longer strings are stored here
+  size_t mem_size_;   // current size of name_mem_
+  char* name_;        // this points to the right one
+
+  double   value_;
+  uint32_t flags_;
+
+  void _SetName(const char *name);
```
EricWF: `_SetName` is a reserved identifier, please choose a different name.

For reference the following names are reserved and cannot be used:
* identifiers with a double underscore anywhere are reserved;
* identifiers that begin with an underscore followed by an uppercase letter are reserved;
* identifiers that begin with an underscore are reserved in the global namespace.

EricWF:
> The solutions could be:
> 
> * Remove const from the State argument to the SetUp() and TearDown() of Fixture. I chose not to do this because it was more disruptive to how the public API was. But if this was my own project, I would have gone for this.

Let's do this,  but we don't have to break the API right away. Instead modify `Fixture` as follows:

```c++
class Fixture : public internal::Benchmark {
/* [...] */
public:
  // Already exist. We should deprecate them soon.
  virtual void SetUp(const State&) {}
  // Add these!
  virtual void SetUp(State& st) { SetUp(static_cast<State const&>(st)); }
  virtual void TearDown(State& st) { TearDown(static_cast<State const&>(st)); }
};
```

The library will call the new non-const functions which can be overridden by the user, otherwise the default implementation will forward to the old const API.

EricWF:
I think this patch would be a lot smaller if we used the standard library more freely in the public API, especially std::string.I would like to see this patch update so that it uses the C++03 STL as needed within the public API.

`BenchmarkCounters` should be replaced with `std::map<std::string, Counter>` or even `std::map<std::string, double>` if we can get rid of the `Counter` class all together. The current `BenchmarkCounters` implementation is doing all kinds of sketchy memory management. This would get rid of that entirely, and it would provided an standard container interface which seems to be what we want.

Additionally the `State` class should just expose the `counters_` (renamed `counters`) object directly as a public member instead of the `SetCounter`/`GetCounter/Counters` interface. 

Inline comments regarding `Counter` coming soon...

EricWF:
Addendum (On general policy, not necessarily this patch):

Although STL usage should be encouraged were appropriate within the public API,
it *must not* be used to implement features called within the `KeepRunning()` loop since things
like STL containers/std::string are going to introduce jitter and differ between platforms.

README.md
biojppm: Good point.

This needs clarification: `SetCounter()` _always_ changes the value of the counter, whether it is new or existing. The only thing it does not change is the value of its flags (kRate, etc) _after_ the first call for that particular counter. This is intended as changing the type of counter through the flags would not be typical use. 

I'll clarify this in the docs. Maybe this:

> Note that setting the counter flags only works on the first call to `state.SetCounter()`; any subsequent _flag_ values will be silently ignored.

What do you think?

include/benchmark/benchmark_api.h
biojppm: Nothing really, just what I'm used to. Changing in comming commit.

include/benchmark/benchmark_api.h
biojppm: D'oh, of course - sorry. Changing in a comming commit.

README.md
```diff
@@ -393,6 +395,76 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+to the counter. You can use this handle to get the counter with the method
+`state.GetCounter()`. You can also get the counter by name, either with
+`const char*` or `std::string const&`; this will do a string lookup, linear
+in the number of counters. The counter object has compound
+assignment overloads `+=`, `-=`, `*=`, `/=`. You can then change the counter
+as needed. For example:
+
+```c++
+static void BM_UserCountersExample(benchmark::State& state) {
+  state.SetCounter("LighterCounter"); // counter value defaults to 0
```
dominichamon: CreateCounter?

README.md
```diff
@@ -393,6 +395,76 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+  size_t handle = state.SetCounter("HeavyCounter"); // also 0
+  while (state.KeepRunning()) {
+    state.GetCounter("LighterCounter") += numLighterEvents; // will do a string lookup, linear in number of counters
```
dominichamon: GetCounter is going to incur overhead which, if it's in the context of a benchmark keeprunning loop is not a great idea.

the examples here should be best in class, ie:

```
while (state.KeepRunning()) {
  numLighterEvents++;
  bigNumEvents++;
}
state.GetCounter("LighterCounter") += numLighterEvents;
state.GetCounter(handle) += bigNumEvents;
```

include/benchmark/benchmark_api.h
dominichamon: s/Flags_e/Flags/

include/benchmark/benchmark_api.h
```diff
@@ -429,6 +553,33 @@ class State {
   BENCHMARK_ALWAYS_INLINE
   size_t iterations() const { return total_iterations_; }
 
+
+  // Add a user counter. Return the counter handle.
+  BENCHMARK_ALWAYS_INLINE
+  size_t SetCounter(const char* n) { return counters_.Set(n); }
```
dominichamon: given these always return a handle, i see no reason to allow counter lookup by name. it's just going to cause developers pain in the future when their benchmarks take longer than they expected.

Please simplify this such that SetCounter (actually, CreateCounter) takes a string and everything else takes a handle.

include/benchmark/benchmark_api.h
dominichamon: i'm not a fan of the two-stage initialization and shutdown just to support counters in fixtures, no. I suggest that for now we don't support fixtures and get the basic counters in. then we can figure out how to support fixtures better.

it may be ok to pass a mutable State, but i need to think about it.

src/benchmark.cc
```diff
@@ -302,6 +303,7 @@ static std::unique_ptr<TimerManager> timer_manager = nullptr;
 
 } // end namespace
 
+
```
dominichamon: lots of unnecessary extra whitespace. please remove it.

src/console_reporter.cc
```diff
@@ -46,17 +48,35 @@ bool ConsoleReporter::ReportContext(const Context& context) {
...
-                             static_cast<int>(name_field_width_), "Benchmark",
-                             "Time", "CPU", "Iterations");
-  GetOutputStream() << str << std::string(str.length() - 1, '-') << "\n";
 
   return true;
 }
 
+void ConsoleReporter::PrintHeader(const Run& run) {
+  std::string str = FormatString("%-*s %13s %13s %10s",
+                             static_cast<int>(name_field_width_), "Benchmark",
```
dominichamon: it doesn't look like this was formatted with clang-format. due you mind doing so with Google style?

src/counter.cc
```diff
@@ -0,0 +1,327 @@
...
+#include "benchmark/benchmark.h"
+#include "check.h"
+#include "internal_macros.h"
+
+#include <cstring>
+
+namespace benchmark {
+
+Counter::Counter()
+  : name_buf_(), name_mem_(nullptr), mem_size_(0), name_(nullptr), value_(0.), flags_(kDefaults) {
```
dominichamon: lines too long. please run clang-format on the file with Google style.

src/counter.cc
```diff
@@ -0,0 +1,327 @@
...
+
+#ifdef BENCHMARK_HAS_CXX11
+Counter::Counter(Counter && that)
+  : name_buf_{0}, name_mem_(that.name_mem_), mem_size_(that.mem_size_), name_(that.name_), value_(that.value_), flags_(that.flags_) {
+  memcpy(name_buf_, that.name_buf_, sizeof(name_buf_));
+  that.name_mem_ = nullptr;
+  that.mem_size_ = 0;
+  that.name_ = nullptr;
+}
+Counter& Counter::operator= (Counter && that) {
```
dominichamon: the default operator isn't good enough? if you switch to string it should be, right?

biojppm: yes, but as we've talked before (at least before these commits), std::string was not to be included in the public header. Even with PIMPL, these operators could not be defaulted.

src/counter.cc
biojppm: ok

src/console_reporter.cc
biojppm: ok

src/benchmark.cc
biojppm: ok

README.md
biojppm: I put this example here precisely to alert to the cost...!

include/benchmark/benchmark_api.h
biojppm: Fair point. As you were writing this, I was finishing up a commit changing the interface to something map-like. Maybe this way the users will be more alerted to potential costs. Please inspect that and if you still consider the access by name confusing I'll remove it.

FWIW, I still think that offering access by name is practical for the user, provided he is alert to the cost. For instance, accessing by name inside the KeepRunning loop is not a problem if the code being benchmarked is heavy in comparison and will result in few iterations.

README.md
biojppm: Yesss :-).  I changed it back and forth a few times already, but ended up removing it to make for a smaller API. Anyway, check the new map-like interface which maybe makes this question moot.

include/benchmark/benchmark_api.h
biojppm: Oh well. I just changed it from std::string... And what about the array in BenchmarkCounters - do I change that too and then `#include <vector>`?

include/benchmark/benchmark_api.h
biojppm: ok

README.md
biojppm: Actually, I went and changed the API to map-like. With that change, this question no longer stands.

README.md
biojppm: ok.

dominichamon:
Thank you for sticking with us on this. I understand you've gone back and forth on the API, but it's a big thing to introduce and I want to make sure it's going in the right way.

biojppm:
> 

> Thank you for sticking with us on this. I understand you've gone back and forth on the API, but it's a big thing to introduce and I want to make sure it's going in the right way.

No problem. It's a big addition in a mature project, so I guess it's natural for this back and forth to occur.

biojppm:
Eric,

Regarding your changes in the fixture interface:

> The library will call the new non-const functions which can be overridden by the user, otherwise the default implementation will forward to the old const API.

Great! I would have gone for this too.

biojppm:
Please see whether you agree with the new map-like interface. I will change the C arrays to std after we narrow this API down.

Meanwhile, I added some benchmarks for accessing counters inside the KeepRunning() loop by id and by name, with the latter being done both through the current API and accelerated with several alternative data structures. These tests lookup two counters out of n:

```c++
  while (st.KeepRunning()) {
    st.counters[do_a_lookup("Foo")] += 1.;
    st.counters[do_a_lookup("Bar")] += 1.;
  }
```

I've put counter sizes from 2 to 256, although in general we will always have small counter numbers (say, less than 16). The tests are the following:

- (ById) Accessing by id with the current data structure. O(1) in number of counters.
- (ByName) Accessing by name with the current data structure. O(n).
- (VsMap) Simulates the counters being stored in an std::map< std::string, Counter >. Note that this prevents access by id. O(log(n)).
- (NameIdMS) Accessing by name with the current data structure, but the lookup is accelerated with an auxiliary std::map< std::string, size_t >. O(log(n)).
- (NameIdFS) Accelerated as above, but the accelerator is a sorted std::vector< std::pair< std::string, size_t > > (ie, a flat map) which should provide better cache behaviour than the map. O(log(n)).
- (NameIdFC) Accelerated with a flat map as above, but the vector uses char[] instead of std::string which should provide even better cache behaviour. O(log(n)).

The results are expressive:

```
Run on (4 X 3728.84 MHz CPU s)
2016-08-05 14:47:19
-------------------------------------------------------
Benchmark                      Time      CPU Iterations
-------------------------------------------------------
ById____/2        3ns     4 ns  196629213   538.851M items/s
ByName__/2       17ns    18 ns   38043478   104.859M items/s
VsMap___/2       39ns    34 ns   20348837   55.7648M items/s
NameIdFC/2       19ns    19 ns   37234043   102.038M items/s
NameIdFS/2       42ns    40 ns   17500000   47.4128M items/s
NameIdMS/2       32ns    35 ns   19662921   54.1966M items/s
-------------------------------------------------------
ById____/8        3ns     4 ns  198863636    541.86M items/s
ByName__/8       36ns    40 ns   17676768   48.1654M items/s
VsMap___/8       55ns    50 ns   13461538   38.2081M items/s
NameIdFC/8       29ns    28 ns   25362319   68.7142M items/s
NameIdFS/8       60ns    66 ns   10294118   29.0451M items/s
NameIdMS/8       45ns    50 ns   14112903   38.2361M items/s
-------------------------------------------------------
ById____/16       4ns     3 ns  201149425   548.089M items/s
ByName__/16      63ns    62 ns   11666667   30.5665M items/s
VsMap___/16      60ns    65 ns   10937500   29.1363M items/s
NameIdFC/16      33ns    37 ns   19230769    52.102M items/s
NameIdFS/16      91ns    83 ns    8333333    22.969M items/s
NameIdMS/16      61ns    67 ns   10937500   28.4995M items/s
-------------------------------------------------------
ById____/32       3ns     3 ns  198863636   548.125M items/s
ByName__/32     131ns   119 ns    5833333   16.0783M items/s
VsMap___/32      76ns    76 ns    9210526   25.2409M items/s
NameIdFC/32      36ns    40 ns   17500000    48.235M items/s
NameIdFS/32      91ns    99 ns    7000000   19.1831M items/s
NameIdMS/32      84ns    76 ns    9210526   24.9541M items/s
-------------------------------------------------------
ById____/64       4ns     3 ns  203488372   548.197M items/s
ByName__/64     208ns   228 ns    3070175   8.36556M items/s
VsMap___/64      97ns    86 ns    7954545   22.1814M items/s
NameIdFC/64      50ns    49 ns   13461538   38.6684M items/s
NameIdFS/64     107ns   117 ns    5833333   16.3621M items/s
NameIdMS/64      97ns    87 ns    7954545    21.799M items/s
-------------------------------------------------------
ById____/128      4ns     3 ns  203488372   551.311M items/s
ByName__/128    491ns   463 ns    1521739   4.12285M items/s
VsMap___/128     98ns   107 ns    6481481   17.7621M items/s
NameIdFC/128     60ns    54 ns   12500000    35.269M items/s
NameIdFS/128    137ns   134 ns    5147059   14.1868M items/s
NameIdMS/128    101ns   110 ns    6481481    17.363M items/s
-------------------------------------------------------
ById____/256      3ns     4 ns  196629213   538.851M items/s
ByName__/256   1038ns   944 ns     729167   2.02148M items/s
VsMap___/256    105ns   114 ns    6034483   16.7295M items/s
NameIdFC/256     72ns    65 ns   10937500      29.3M items/s
NameIdFS/256    169ns   165 ns    4268293   11.5641M items/s
NameIdMS/256    106ns   115 ns    6250000   16.5568M items/s

```

Some conclusions:

- (ById) is always fastest, standing head and shoulders above all the other ones. This is expected.

Compared to (ById) the performance of access by name is always very bad, irrespective of which approach is used. But there are some notable differences among them.

There's a change in regime precisely at 16 counters. Up until then (ByName) is best for very small numbers of counters, but is then topped by the flat char array accelerator (NameIdFC). This is always best from here on.

For more than 16 counters, surprisingly, the map accelerator (NameIdMS) performs better than the flat std::string accelerator (NameIdFS). But the flat char array NameIdFC) always takes half the time of the other name accelerators; this is an impressive speedup. The map accelerator (NameIdMS) has the same performance as directly storing the counters in a map (VsMap).

So all in all, given these results, I would essentially keep the current structure, with both access by id and by name, as it provides greater freedom. Of course, it should be clearly stated that access by name is slow and should be made outside of the KeepRunning() loop. Even this may be possible and be of low impact when the benchmarked code is heavy in comparison.

biojppm:
@dominichamon @EricWF 
ping...
Just let me know if you are fine with the current iteration of the interface, or what I should change in it. I'll change back to std after we narrow this down.

README.md
```diff
@@ -421,6 +421,91 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+  // Set the counter as a thread-average quantity. It will
+  // be presented divided by the number of threads. */
+  state.counters.Insert("FooAvg", numFoos, benchmark::Counter::kAvgThreads);
+
+  // the third argument is a bitwise mask, so you can combine.
+  // This will mark as a thread-average rate:
+  state.counters.Insert("FooAvgRate", numFoos, benchmark::Counter::kIsRate|benchmark::Counter::kAvgThreads);
+```
+
+A further use of `state.counters.Insert()` is to minimize lookup time.
```
dominichamon: given we know this can be an issue, and it's hard for users to know which to use (or they use the simplest, which is expensive) can we do anything to ensure they do the right thing? ie, should Insert be the only API and should we restrict it to only being called outside KeepRunning?

include/benchmark/benchmark_api.h
```diff
@@ -239,6 +240,149 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+    /** Mark the counter as a rate. It will be presented divided by the duration of the benchmark. */
+    kIsRate         = 0x01 << 0,
+    /** Mark the counter as a thread-average quantity. It will be presented divided by the number of threads. */
+    kAvgThreads     = 0x01 << 1
+  } Flags;
+
+public:
+
+  Counter();
+  Counter(const char *name);
```
dominichamon: if you keep this, make sure it's explicit to avoid accidental conversion from string to Counter.

include/benchmark/benchmark_api.h
dominichamon: if @EricWF says it's ok to use <vector> and <string>, we should.

src/complexity.cc
```diff
@@ -196,6 +213,10 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
       cpu_accumulated_time_stat.Mean() * run_iterations;
   mean_data.bytes_per_second = bytes_per_second_stat.Mean();
   mean_data.items_per_second = items_per_second_stat.Mean();
+  // user counters
+  for(auto const& kv : counter_stats) {
+    mean_data.counters[kv.first.c_str()] = kv.second.Mean(); /**@todo add the rest of the settings (fmt,flags)*/
```
dominichamon: are you going to do this in this PR?

if not, please watch the line length: move the todo comment to the line before.

src/counter.cc
```diff
@@ -0,0 +1,432 @@
...
+#include "benchmark/benchmark.h"
+#include "check.h"
+#include "internal_macros.h"
+
+#include <cstring>
+
+namespace benchmark {
+
+Counter::Counter()
+    : name_(nullptr),
```
dominichamon: i think all the supported compilers allow delegating constructors now so feel free to use them.

include/benchmark/benchmark_api.h
```diff
@@ -239,6 +240,149 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  // Get a counter by id. Complexity: O(1)
+  Counter & operator[] (size_t id);
+  // Get a counter by id. Complexity: O(1)
+  Counter const& operator[] (size_t id) const;
+
+
+  // Insert a counter. Return its id. Complexity: O(nameLength*size())
+  size_t Insert(const char* name);
+  // Insert or update a counter. Return its id. Complexity: O(nameLength*size())
+  size_t Insert(const char *name, double value);
```
pleroy: It would be nice if the style was consistent, eg with the way that `*` and `&` are glued to the type or the variable (compare the two previous declarations).  I would suggest to run clang-format with style Google.

EricWF:
FYI I plan to look at this tonight.

src/counter.cc
```diff
@@ -0,0 +1,432 @@
...
+      s += name;
+    } else {
+      char tmp[256];
+      snprintf(tmp, sizeof(tmp), format.c_str(), value);
+      s = tmp;
+    }
+  }
+  return s;
+}
+*/
```
pleroy: Please don't leave commented out code around.

include/benchmark/benchmark_api.h
```diff
@@ -239,6 +240,149 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
 #endif
 
+
+
+class Counter {
+public:
+
+  typedef enum {
+    kDefaults       = 0x00,
+    /** Mark the counter as a rate. It will be presented divided by the duration of the benchmark. */
```
pleroy: Please use `//` comments whenever possible.

include/benchmark/benchmark_api.h
```diff
@@ -239,6 +240,149 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  char name_buf_[16]; // names smaller than 15 characters are stored here
+  char* name_mem_;    // whereas longer strings are stored here
+  size_t mem_size_;   // current size of name_mem_
+
+  void SetName_(const char *name);
+};
+
+
+
+class BenchmarkCounters {
```
pleroy: I have seen the benchmarks, but I remain unconvinced by the wisdom of having a half-baked map-like class that's not quite like a map.  In particular (1) the iterator exposes internal details (such as the fact that the element is not a pair) and (2) the method names are misleading wrt their complexity (Find is O(N)).

I don't quite buy the argument that this class has to be super-efficient and, like @EricWF, I would prefer to use STL containers whenever possible.  Your benchmark is a bit of a red-herring: anyone wanting to update the counters in the tight loop would be well-advised to grab an iterator to the counter before the loop and use it in the loop.  A `std::map` would work just fine for that purpose.

As it stands this PR increases considerably the surface of the API.  While I think that user-defined counters are a great idea, there has to be a way to add them with a much more compact API, and without an implementation that does its own management of storage and containers.

EricWF:
>Your benchmark is a bit of a red-herring: anyone wanting to update the counters in the tight loop would be well-advised to grab an iterator to the counter before the loop and use it in the loop. A std::map would work just fine for that purpose.

Exactly right. Instead of `size_t ID = counters.Insert("foo", value)` you would just write `iterator IT = counters.emplace("foo", value).first`.Admittedly it would be faster to use `unordered_map` but that requires C++11 and `benchmark::State` is limited to C++03 only.

> I don't quite buy the argument that this class has to be super-efficient 

I violently agree. The `counters` container *shouldn't* be used in the `KeepRunning()` loop, since that loop should contain *only* the code under test. Trying to optimize the performance of the container is akin to optimizing for poorly written benchmarks because a benchmark that modifies `counters` within the loop is a bad benchmark.

> As it stands this PR increases considerably the surface of the API. While I think that user-defined counters are a great idea, there has to be a way to add them with a much more compact API, and without an implementation that does its own management of storage and containers.

Again I violently agree! I'm opposed to accepting this patch with `BenchmarkCounters` and anything that resembles hand-rolled containers or manual memory management.

EricWF:
@biojppm @pleroy I implemented a different version of this as PR #276 . Please take a look at the updated interface and tell me what you think.

I don't mean to hijack @biojppm pull request, I'm simply created PR #276  in order to show an alternative interface.

biojppm:
@EricWF @dominichamon @pleroy 

I finally changed to std::map as you requested. Please tell me now whether this implementation is ok. To reduce API surface I placed some helper functions (previously members of BenchmarkCounters) inside the internal namespace; these are all still defined in counters.cc.

For the record, and to make this clear (sometimes it seemed as it wasn't), rolling out custom built containers was not my first choice - at the time, policy apparently was to have no std:: includes in the main benchmark_api.h file - and so that's why I rolled out these containers. Neither was I proposing to keep them in the last iteration - I just wanted to establish whether std::map or std::vector should be used before I proceed to rewrite the code to use std:: containers.

As for the array vs map lookup, you made good points. Yes, iterators can be used just the same as size_t handles, and the mental overhead is smaller. So map it is!

@EricWF I saw your PR #276 as I was finishing this rewrite. When you mention you do not want to hijack my PR, I gather you still want this to go ahead. So that's why I'm still pushing this. If you want me to merge from yours or something, let me know.

README.md
EricWF: These docs are out of date.

include/benchmark/benchmark_api.h
EricWF: Still needs to be changed.

include/benchmark/benchmark_api.h
```diff
@@ -239,6 +242,45 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  } Flags;
+
+public:
+
+  double   value;
+  uint32_t flags;
+
+public:
+
+  BENCHMARK_ALWAYS_INLINE Counter(double v, uint32_t flags_) : value(v), flags(flags_) {}
```
EricWF: This can be a single constructor `Counter(double v = 0, Flags flags = kDefaults)`

include/benchmark/benchmark_api.h
EricWF: And let's make this a non-bitmask enum so we can store `Flags` instead of `uint32_t`.  I would add a `kAvgThreadRate` to handle that case.

include/benchmark/benchmark_api.h
EricWF: This class should have an assignment operator `Counter& operator=(double)` to support things like `c = 42.0;`

include/benchmark/benchmark_api.h
```diff
@@ -674,8 +721,12 @@ class Fixture: public internal::Benchmark {
       this->TearDown(st);
     }
 
+    // These will be deprecated ...
```
EricWF: This change should be made as a separate PR, since it's unrelated to this patch AFAIK.

include/benchmark/macros.h
```diff
@@ -26,6 +26,7 @@
 # define BENCHMARK_DISALLOW_COPY_AND_ASSIGN(TypeName)  \
     TypeName(const TypeName&) = delete;                \
     TypeName& operator=(const TypeName&) = delete
+# define BENCHMARK_HAS_CXX11
```
EricWF: This change seems unneeded since the macro is already defined.

src/CMakeLists.txt
```diff
@@ -5,7 +5,18 @@ include_directories(${PROJECT_SOURCE_DIR}/src)
 set(SOURCE_FILES "benchmark.cc" "colorprint.cc" "commandlineflags.cc"
                  "console_reporter.cc" "csv_reporter.cc" "json_reporter.cc"
                  "log.cc" "reporter.cc" "sleep.cc" "string_util.cc"
-                 "sysinfo.cc" "walltime.cc" "complexity.cc")
+                 "sysinfo.cc" "walltime.cc" "complexity.cc" "counter.cc")
+# Add headers to the list of source files. cmake does not require this,
```
EricWF: Please submit this change as a separate PR.

include/benchmark/benchmark_api.h
```diff
@@ -239,6 +242,45 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+
+  BENCHMARK_ALWAYS_INLINE operator double const& () const { return value; }
+  BENCHMARK_ALWAYS_INLINE operator double      & ()       { return value; }
+
+};
+
+typedef std::map< std::string, Counter > BenchmarkCounters;
+
+// these counter-related functions are hidden to reduce API surface.
+namespace internal {
```
EricWF: If the functions are internal then they shouldn't be declared in the public headers. Please move the declarations elsewhere.

EricWF:
Alright this is looking a lot better. Thanks for all the hard work. I have a couple of comments on the new ouput format though.

I would prefer that additional counters simply get printed at the end of the test line, instead of printing a new header for every test. Parsing "<CounterName>=<CounterValue> <Unit>" should be as easy as parsing "<CounterName>\n----------------------\n<CounterValue>`.

EricWF:
Also see closed issue #240.

src/CMakeLists.txt
biojppm: D'oh! Indeed, this was meant to go as a separate PR. It got accidentally included as a blanket git add.

include/benchmark/benchmark_api.h
biojppm: Would it be ok if I added a counter.h to be included where needed?

include/benchmark/benchmark_api.h
biojppm: ok, good point. Indeed you had it like that in your example PR #276 .

biojppm: @EricWF It's currently Flags instead of Flags_e. Do you mean something else?

include/benchmark/benchmark_api.h
biojppm: ok.

include/benchmark/benchmark_api.h
biojppm: Actually the fixture I wrote (which was removed in the meantime) to benchmark the counter api depended on the non-const state. And counter setup is the main driver behind this change. Do you still want it as a separate PR?

include/benchmark/macros.h
biojppm: ok

test/CMakeLists.txt
```diff
@@ -72,6 +72,20 @@ endif()
...
+    if(PAPI_INCLUDE_DIRS AND PAPI_LIBRARIES)
+        find_library(LIBPFM pfm) # needed in Ubuntu 16.04
+        include_directories(${PAPI_INCLUDE_DIRS})
+        compile_benchmark_test(papi_test)
+        add_test(papi_test papi_test --benchmark_min_time=0.01)
+        target_link_libraries(papi_test ${PAPI_LIBRARIES} ${LIBPFM})
+    endif()
+endif(UNIX)
+
+
```
biojppm: D'oh, this was also accidentally put in. BTW, I take this opportunity to ask: would you be interested in an user-defined counter example with PAPI counters? If so, I'll make a separate PR with this. Otherwise I'll just remove this.

README.md
biojppm: Indeed.

biojppm:
Glad you're happier with the current iteration. I'm on vacation away from the computer so it'll take me a couple of days to get down to finish this.

As for the output format, I don't see what you mean. I assume you're talking about the tabular reporter. I changed this reporter so that the header is printed once, or when the counter set changes from one test to another. So the header should be printed only when needed, and if it isn't then that's a bug.
Do you mean you want the counter names out of the header? Ie, printed like bytes processed? I did do this in the first implementation, but the counters become unaligned when their set changes and the output is confusing to read (eg, where do the names go? printed on every counter value? at the top they can't go because the set may be different).

include/benchmark/benchmark_api.h
biojppm: Actually, it already has an implicit `operator=(double)`: via the implicit conversion to `double&` below.

include/benchmark/benchmark_api.h
```diff
@@ -243,6 +246,41 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 // FIXME Add ClobberMemory() for non-gnu compilers
 #endif
 
+
+
+// This class is used for user-defined counters.
+class Counter {
+public:
+
+  typedef enum {
```
dominichamon: just

```
enum Flags {
 ...
};
```

include/benchmark/benchmark_api.h
```diff
@@ -243,6 +246,41 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+    // Mark the counter as a rate. It will be presented divided
+    // by the duration of the benchmark.
+    kIsRate     = 1,
+    // Mark the counter as a thread-average quantity. It will be
+    // presented divided by the number of threads.
+    kAvgThreads = 2,
+    kAvgThreadsRate = kIsRate|kAvgThreads
+  } Flags;
+
+public:
```
dominichamon: no need to restate public

include/benchmark/benchmark_api.h
```diff
@@ -243,6 +246,41 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+
+  double value;
+  Flags  flags;
+
+public:
+
+  BENCHMARK_ALWAYS_INLINE Counter(double v = 0., Flags f = kDefaults) : value(v), flags(f) {}
```
dominichamon: watch your line-length please. you can always use clang-format with the Google style.

src/console_reporter.cc
```diff
@@ -91,30 +112,35 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   // Format items per second
   std::string items;
   if (result.items_per_second > 0) {
-    items = StrCat(" ", HumanReadableNumber(result.items_per_second),
-                   " items/s");
- }
+    items =
+        StrCat(" ", HumanReadableNumber(result.items_per_second), " items/s");
```
dominichamon: these formatting changes introduce noise into the PR. can you revert them?

src/counter.h
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "benchmark/benchmark_api.h"
+
+namespace benchmark {
+
+// these counter-related functions are hidden to reduce API surface.
+namespace internal {
+void Finish(BenchmarkCounters *l, double time, double num_threads);
```
dominichamon: it's _almost_ worth making BenchmarkCounters a struct containing the map and these methods as members.

src/csv_reporter.cc
```diff
@@ -46,20 +48,51 @@ std::vector<std::string> elements = {
 
 bool CSVReporter::ReportContext(const Context& context) {
   PrintBasicContext(&GetErrorStream(), context);
+  return true;
+}
+
+void CSVReporter::ReportRuns(const std::vector<Run> & reports) {
+
```
dominichamon: nit: extra blank line.

src/console_reporter.cc
biojppm: sure.

src/counter.h
biojppm: quite. I admit I was on the fence there.

src/complexity.cc
dominichamon: ping: are you going to do this TODO in this PR?

biojppm: Woops, missed this. Commit coming.

EricWF:
> Do you mean you want the counter names out of the header? Ie, printed like bytes processed?

Yes. I was thinking consistency with "bytes processed" would be preferred. In particular because "bytes processed" should really just become a "custom counter" in the future. 

>I did do this in the first implementation, but the counters become unaligned when their set changes and the output is confusing to read (eg, where do the names go? printed on every counter value? at the top they can't go because the set may be different).

I was thinking the name, unit and value for each counter would be printed for every counter value. The output would look something link:

```
counters["foo"] = 42;
counters["bar"] = {100, benchmark::Counter::kIsRate};

BM_Foo 4 4 100 foo=42 bar=100/ns
```

Arguably your format is nicer when you have a number of tests all with the same custom counters, since the header will only be printed once.

However when you have (A) one or two counters or (B) different counters for every test then it starts to get noisy. In the case of (B) each benchmark will be 4 lines of output instead of one.

I'm not opposed to your output plan either. I think we should consider both.

@dominichamon @pleroy Do you have any preference between the two output formats? (see below)

```
Run on (4 X 4227.95 MHz CPU s)
2016-09-01 18:53:44
------------------------------------------------------------------------------------------------------------------
Benchmark                    Time           CPU Iterations           Bar           Bat           Baz           Foo
------------------------------------------------------------------------------------------------------------------
BM_UserCounter               2 ns          2 ns  307017544             2             5             3         1024m
----------------------------------------------------------------------------------------------------
Benchmark                    Time           CPU Iterations          ABar          AFoo           Bat
----------------------------------------------------------------------------------------------------
BM_UserCounter2/0/1          2 ns          2 ns  296610169             2         1024m             5
Benchmark                    Time           CPU Iterations
BM_NoCounters                2 ns          2 ns  291666667
------------------------------------------------------------------------------------------------------------------
Benchmark                    Time           CPU Iterations          ABar          AFoo           Bat           Baz
------------------------------------------------------------------------------------------------------------------
BM_UserCounter2/1/2          2 ns          2 ns  307017544             2         1024m             5             3
```

```
Run on (4 X 4227.95 MHz CPU s)
2016-09-01 18:53:44
Benchmark                    Time           CPU Iterations          Additional Counters
------------------------------------------------------------------------------------
BM_UserCounter               2 ns          2 ns  307017544 Bar=2 Bat=5 Baz=3 Foo=1024m
BM_UserCounter2/0/1          2 ns          2 ns  296610169 ABar=2 AFoo=1024m Bat=5
BM_NoCounters                2 ns          2 ns  291666667
BM_UserCounter2/1/2          2 ns          2 ns  307017544 ABar=2 AFoo=1024m Bat=5 Baz=3
```

dominichamon:
A strong preference for the latter with 'User counters' as the header. For CSV it's a little trickier, but we can start with user counters containing the list of user counters as tabular. JSON is easy.

README.md
```diff
@@ -430,6 +430,60 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+
+  // Set the counter as a thread-average quantity. It will
+  // be presented divided by the number of threads. */
+  state.counters["FooAvg"] = Counter(numFoos, benchmark::Counter::kAvgThreads);
+
+  // There's also a combined flag:
+  state.counters["FooAvgRate"] = Counter(numFoos,benchmark::Counter::kAvgThreadsRate);
+```
+
+When you're compiling in C++11 mode or later you can use the `insert()` and `emplace()`
```
EricWF: 'emplace()' doesn't have initializer list overloads.

test/cxx03_test.cc
```diff
@@ -39,4 +39,13 @@ void BM_template1(benchmark::State& state) {
 BENCHMARK_TEMPLATE(BM_template1, long);
 BENCHMARK_TEMPLATE1(BM_template1, int);
 
+void BM_counters(benchmark::State& state) {
+    while (state.KeepRunning()) {
+        volatile std::size_t x = state.iterations();
```
EricWF: An empty benchmark body is fine, no need to waste cycles.

EricWF:
@biojppm OK I think it's been decided that the single-line output format is prefered. I'm just waiting on those changes before finishing this review.

EricWF:
@biojppm I just created some merge conflicts for you (sorry). They are pretty tame but let me know if you need help.

src/console_reporter.cc
```diff
@@ -44,22 +47,39 @@ bool ConsoleReporter::ReportContext(const Context& context) {
...
 
   return true;
 }
 
+void ConsoleReporter::PrintHeader(const Run& run) {
+  std::string str = FormatString("%-*s %13s %13s %10s",
+                                 static_cast<int>(name_field_width_),
+                                 "Benchmark", "Time", "CPU", "Iterations");
+  if(!run.counters.empty()) {
+    str += " UserCounters...";
```
biojppm: @EricWF @dominichamon note the use of the ellipsis, as an indication that the output below is not vertically aligned. Example results:

```
-----------------------------------------------------------------------------------------------
Benchmark                                         Time           CPU Iterations UserCounters...
-----------------------------------------------------------------------------------------------
BM_UserCounter/threads:8                       2266 ns      10283 ns      68080 Bar=16 Bat=40 Baz=24 Foo=8
BM_UserCounter/threads:1                       9791 ns       9785 ns      71503 Bar=2 Bat=5 Baz=3 Foo=1024m
BM_UserCounter/threads:2                       4915 ns       9829 ns      71076 Bar=4 Bat=10 Baz=6 Foo=2
BM_UserCounter/threads:4                       2603 ns      10283 ns      68256 Bar=8 Bat=20 Baz=12 Foo=4
BM_UserCounter/threads:8                       2152 ns      10263 ns      68440 Bar=16 Bat=40 Baz=24 Foo=8
BM_UserCounter/threads:16                      1484 ns      10283 ns      68224 Bar=32 Bat=80 Baz=48 Foo=16
BM_UserCounter/threads:32                       906 ns      10285 ns      68096 Bar=64 Bat=160 Baz=96 Foo=32
BM_UserCounter/threads:4                       2605 ns      10279 ns      68140 Bar=8 Bat=20 Baz=12 Foo=4
```

biojppm:
@EricWF @dominichamon @pleroy I finally got around to make the changes you requested. Output is now all at the end, _before_ `items_processed` and `bytes_processed`.

I would like to further discuss the tabular printing issue, as I'm considering a future PR for this.

First, you certainly have a point in that the tabular format will make runs with few lines more complicated. But the situation tends to reverse as the number of lines increases. So I still think that there's a place for tabular output of user counters, specially when a run will contain more than a few lines. Here's an example comparison which I hope shows how much easier it is in the eye. (Note that I removed some whitespace from both examples). 

First the current, named-at-the-end version:
```
------------------------------------------------------------------------------
Benchmark                        Time           CPU Iterations UserCounters...
------------------------------------------------------------------------------
BM_UserCounter/threads:8      2248 ns      10277 ns      68808 Bar=16 Bat=40 Baz=24 Foo=8
BM_UserCounter/threads:1      9797 ns       9788 ns      71523 Bar=2 Bat=5 Baz=3 Foo=1024m
BM_UserCounter/threads:2      4924 ns       9842 ns      71036 Bar=4 Bat=10 Baz=6 Foo=2
BM_UserCounter/threads:4      2589 ns      10284 ns      68012 Bar=8 Bat=20 Baz=12 Foo=4
BM_UserCounter/threads:8      2212 ns      10287 ns      68040 Bar=16 Bat=40 Baz=24 Foo=8
BM_UserCounter/threads:16     1782 ns      10278 ns      68144 Bar=32 Bat=80 Baz=48 Foo=16
BM_UserCounter/threads:32     1291 ns      10296 ns      68256 Bar=64 Bat=160 Baz=96 Foo=32
BM_UserCounter/threads:4      2615 ns      10307 ns      68040 Bar=8 Bat=20 Baz=12 Foo=4
BM_Factorial                    26 ns         26 ns   26608979 40320
BM_Factorial/real_time          26 ns         26 ns   26587936 40320
BM_CalculatePiRange/1           16 ns         16 ns   45704255 0
BM_CalculatePiRange/8           73 ns         73 ns    9520927 3.28374
BM_CalculatePiRange/64         609 ns        609 ns    1140647 3.15746
BM_CalculatePiRange/512       4900 ns       4901 ns     142696 3.14355
```
... compared to the tabular version:
```
---------------------------------------------------------------------------------------
Benchmark                        Time           CPU Iterations    Bar   Bat   Baz   Foo
---------------------------------------------------------------------------------------
BM_UserCounter/threads:8      2198 ns       9953 ns      70688     16    40    24     8
BM_UserCounter/threads:1      9504 ns       9504 ns      73787      2     5     3 1024m
BM_UserCounter/threads:2      4775 ns       9550 ns      72606      4    10     6     2
BM_UserCounter/threads:4      2508 ns       9951 ns      70332      8    20    12     4
BM_UserCounter/threads:8      2055 ns       9933 ns      70344     16    40    24     8
BM_UserCounter/threads:16     1610 ns       9946 ns      70720     32    80    48    16
BM_UserCounter/threads:32     1192 ns       9948 ns      70496     64   160    96    32
BM_UserCounter/threads:4      2506 ns       9949 ns      70332      8    20    12     4
--------------------------------------------------------------
Benchmark                        Time           CPU Iterations
--------------------------------------------------------------
BM_Factorial                    26 ns         26 ns   26392245 40320
BM_Factorial/real_time          26 ns         26 ns   26494107 40320
BM_CalculatePiRange/1           15 ns         15 ns   45571597 0
BM_CalculatePiRange/8           74 ns         74 ns    9450212 3.28374
BM_CalculatePiRange/64         595 ns        595 ns    1173901 3.15746
BM_CalculatePiRange/512       4752 ns       4752 ns     147380 3.14355
BM_CalculatePiRange/4k       37970 ns      37972 ns      18453 3.14184
BM_CalculatePiRange/32k     303733 ns     303744 ns       2305 3.14162
BM_CalculatePiRange/256k   2434095 ns    2434186 ns        288 3.1416
BM_CalculatePiRange/1024k  9721140 ns    9721413 ns         71 3.14159
BM_CalculatePi/threads:8      2255 ns       9943 ns      70936
```

So I'd like to ask here if you would consider a future PR giving an option to the user. Default behaviour would still be to print named-at-the-end as you want, but providing eg --benchmark-tab-counters or somesuch named option would give us the tabular printing. What do you say to this?

README.md
```diff
@@ -593,4 +649,4 @@ required to build the library.
 ### Windows
 
 * Users must manually link `shlwapi.lib`. Failure to do so may result
-in resolved symbols.
\ No newline at end of file
+in resolved symbols.
```
dominichamon: should this say 'unresolved'?

include/benchmark/benchmark_api.h
```diff
@@ -244,6 +246,39 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+  BENCHMARK_ALWAYS_INLINE
+  Counter(double v = 0., Flags f = kDefaults) : value(v), flags(f) {}
+
+  BENCHMARK_ALWAYS_INLINE operator double const& () const { return value; }
+  BENCHMARK_ALWAYS_INLINE operator double      & ()       { return value; }
+
+};
+
+// This is the container for the user-defined counters.
+typedef std::map< std::string, Counter > BenchmarkCounters;
```
dominichamon: nit: you don't need spaces between < > any more.

include/benchmark/reporter.h
```diff
@@ -92,6 +93,9 @@ class BenchmarkReporter {
     // Inform print function whether the current run is a complexity report
     bool report_big_o;
     bool report_rms;
+
+    BenchmarkCounters counters;
+
```
dominichamon: nit: remove this newline

src/benchmark_api_internal.h
```diff
@@ -14,20 +14,21 @@ namespace internal {
...
-  int              range_multiplier;
-  bool             use_real_time;
-  bool             use_manual_time;
-  BigO             complexity;
-  BigOFunc*        complexity_lambda;
-  bool             last_benchmark_instance;
-  int              repetitions;
-  double           min_time;
-  int              threads;    // Number of concurrent threads to us
+  std::string       name;
```
dominichamon: was this a whitespace change? can you revert it (unless it comes from clang-format)?

src/console_reporter.cc
```diff
@@ -74,11 +90,11 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   auto name_color =
       (result.report_big_o || result.report_rms) ? COLOR_BLUE : COLOR_GREEN;
   printer(Out, name_color, "%-*s ", name_field_width_,
-              result.benchmark_name.c_str());
+               result.benchmark_name.c_str());
```
dominichamon: nit: weird indent here

README.md
biojppm: Yes, I guess so. Anyway, apparently this diff here is my editor's fault - it adds a newline at the last character if there isn't one.

src/benchmark_api_internal.h
biojppm: Yes and no. I added the counters (line 27), it sticked out like a sore thumb, and to make it aligned I added one space to the second column.

src/console_reporter.cc
biojppm: To be clear, what I wanted to say here is that if you find the use of the ellipsis awkward, say so and I'll remove it.

biojppm:
@dominichamon @EricWF @pleroy 
Let's finish this.
I didn't notice till yesterday that there was a conflict. I've merged upstream master into this PR. In the process, I discovered a bug of mine in CsvReporter whereby the header was being printed on every call to ReportRuns(). I fixed it, so be sure to check that out too.

dominichamon:
paging @EricWF and @pleroy .. do you have any thoughts?

src/csv_reporter.cc
```diff
@@ -38,21 +39,51 @@ std::vector<std::string> elements = {
...
+      Out << ",\"" << *B++ << "\"";
+    }
+    Out << "\n";
+
+    printed_header_ = true;
+  } else {
+    // check that all the current counters are saved in the name set
+    for (const auto& run : reports) {
+      for (const auto& cnt : run.counters) {
+        CHECK(user_counter_names_.find(cnt.first) != user_counter_names_.end())
```
EricWF: This assertion fires when running `benchmark_test`

README.md
```diff
@@ -430,6 +430,60 @@ BENCHMARK_REGISTER_F(MyFixture, BarTest)->Threads(2);
...
+  state.counters["Bar"] = numBars;
+  state.counters["Baz"] = numBazs;
+}
+```
+
+The `state.counters` object is a `std::map` with `std::string` keys
+and `Counter` values. The latter is a `double`-like class, via an implicit
+conversion to `double&`. Thus you can use all of the standard arithmetic
+assignment operators (`=,+=,-=,*=,/=`) to change the value of each counter.
+
```
EricWF: You should mention that the `counters` map is per-thread and describe how the final value is calculated with multiple threads.

src/csv_reporter.cc
```diff
@@ -38,21 +39,51 @@ std::vector<std::string> elements = {
 
 bool CSVReporter::ReportContext(const Context& context) {
   PrintBasicContext(&GetErrorStream(), context);
+  return true;
+}
 
+void CSVReporter::ReportRuns(const std::vector<Run> & reports) {
```
EricWF: This still seems all wrong.  I think we need to make the CSV reporter store every reported run and only report them inside `Finalize`. That would mean the CSV reporter wouldn't be usable as a console reporter, instead it would be limited to outputting to files. @dominichamon Does this seem reasonable to you? If so I'm willing to fix this post-commit.

EricWF:
@biojppm I'm happy to merge this once all of the tests pass in debug mode, even if the `CSV` stuff isn't fully correct yet.

@biojppm Other than that this LGTM. Please ping me once you've updated this patch so we can merge it.

dominichamon:
@biojppm are you still planning on getting the tests passing in debug?

biojppm:
@dominichamon yes, I'll do it ASAP

src/csv_reporter.cc
biojppm: @EricWF I can't get this to fire. Can you give more info?

biojppm:
@dominichamon @EricWF I was not able to reproduce the failed tests. currently in Linux. tried in Debug with g++5.4, g++6.2, clang3.8, clang3.9

EricWF:
@biojppm I can't reproduce the failures I was seeing either. So I must have done something wrong on my end. 

@biojppm This LGTM. Thanks for all your time working on this!

@dominichamon Any last words or can I merge this?

dominichamon:
dooo eeett

PR #265, opened 2016-08-02
Upgrade travis GCC version to 4.8.4

EricWF:
Creating a new PR for this. After I force-squashed the git history Github closed the old PR.


dominichamon:
to confirm: we don't need to install cmake or clang any more?

EricWF:
Correct, This patch upgrades travis to use Ubuntu Trusty, and Clang 3.5 is preinstalled on those machines.

PR #266, opened 2016-08-03
Add a "compare_bench.py" tooling script.

EricWF:
This patch adds the `compare_bench.py` utility which can be used to compare the result of benchmarks.
The program is invoked like:

```bash
$ compare_bench.py <old-benchmark> <new-benchmark> [benchmark options]...
```

Where `<old-benchmark>` and `<new-benchmark>` either specify a benchmark executable file, or a JSON output file. The type of the input file is automatically detected. If a benchmark executable is specified then the benchmark is run to obtain the results. Otherwise the results are simply loaded from the output file.

The sample output using the JSON test files under `Inputs/` gives:
```bash
$ ./compare_bench.py ./gbench/Inputs/test1_run1.json ./gbench/Inputs/test1_run2.json
Comparing ./gbench/Inputs/test1_run1.json to ./gbench/Inputs/test1_run2.json
Benchmark                   Time           CPU
----------------------------------------------
BM_SameTimes               +0.00         +0.00
BM_2xFaster                -0.50         -0.50
BM_2xSlower                +1.00         +1.00
BM_10PercentFaster         -0.10         -0.10
BM_10PercentSlower         +0.10         +0.10
```

When a benchmark executable is run, the raw output from the benchmark is printed in real time to stdout. The sample output using `benchmark/basic_test` for both arguments looks like:

```
./compare_bench.py  test/basic_test test/basic_test  --benchmark_filter=BM_empty.*
RUNNING: test/basic_test --benchmark_filter=BM_empty.*
Run on (4 X 4228.32 MHz CPU s)
2016-08-02 19:21:33
Benchmark                              Time           CPU Iterations
--------------------------------------------------------------------
BM_empty                               9 ns          9 ns   79545455
BM_empty/threads:4                     4 ns          9 ns   75268816
BM_empty_stop_start                    8 ns          8 ns   83333333
BM_empty_stop_start/threads:4          3 ns          8 ns   83333332
RUNNING: test/basic_test --benchmark_filter=BM_empty.*
Run on (4 X 4228.32 MHz CPU s)
2016-08-02 19:21:35
Benchmark                              Time           CPU Iterations
--------------------------------------------------------------------
BM_empty                               9 ns          9 ns   76086957
BM_empty/threads:4                     4 ns          9 ns   76086956
BM_empty_stop_start                    8 ns          8 ns   87500000
BM_empty_stop_start/threads:4          3 ns          8 ns   88607596
Comparing test/basic_test to test/basic_test
Benchmark                              Time           CPU
---------------------------------------------------------
BM_empty                              +0.00         +0.00
BM_empty/threads:4                    +0.00         +0.00
BM_empty_stop_start                   +0.00         +0.00
BM_empty_stop_start/threads:4         +0.00         +0.00
```

Obviously this example doesn't give any useful output, but it's intended to show the output format when 'compare_bench.py' needs to run benchmarks.


dominichamon:
i'm not sure about a script deciding to run executables to get the output. it seems to me that one could do that directly from a shell and this script is much simpler when it can ignore being passed executable files.

and yes, my unix is showing.

EricWF:
> i'm not sure about a script deciding to run executables to get the output. it seems to me that one could do that directly from a shell and this script is much simpler when it can ignore being passed executable files.

It's absolutely possible, but it's a lot more typing for the user. For example a common invocation I use is:

```
./compare_bench.py a.out b.out --benchmark_repetitions=5 --benchmark_min_time=2.0
```

The equivalent invocation without supporting executables is:

```
./a.out --benchmark_repetitions=5 --benchmark_min_time=2.0 --benchmark_out=/tmp/a.json && ./b.out --benchmark_repetitions=5 --benchmark_min_time=2.0 --benchmark_out=/tmp/b.json && ./compare_bench.py /tmp/a.json /tmp/b.json && rm /tmp/a.json /tmp/b.json
```

Which is a bunch longer and makes manually invoking the script a bit of a pain. The unix-y thing to do would be to pipe the results into the script, but since we have two sets of results and only one stdin we can't do this.

EricWF:
>  this script is much simpler when it can ignore being passed executable files.

Yes, most of the code in this PR is silly utility code to determine the input type, but I don't intend for this to be the *only* script. My basic design idea was to create a python module containing everything you need to easily write new/custom tooling. For this reason I think providing utility functions to determine/diagnose/load input files is helpful.

dominichamon:
LFTM

(Looks fine to me)

Every cell in my unix body is screaming but it is more useful in its current form.

EricWF:
> Every cell in my unix body is screaming but it is more useful in its current form.

Better command line invocations are welcome, but this is a starting point :-)

PR #267, opened 2016-08-03
Add --benchmark_report_aggregates_only={true|false} flag for better summary output.

EricWF:
When running a benchmark multiple times the result of each run is reported along with the final mean/stddev. This is often more verbose than desired when only the mean/stddev is wanted by the user, and the additional reports make the output harder to read.

This patch adds the `--benchmark_report_repetitions={true|false}` flag and `Benchmark::ReportRepetitions(bool)` method. When either of these is specified as `false` only the mean/stddev are reported from repeated benchmarks. Otherwise all runs are reported, which is the current default.


dominichamon:
<bikeshed>can this be more explicit: --benchmark_aggregates_only

EricWF:
Yes it can be "--benchmark_aggregates_only". Additionally I'll change `ReportRepetitions` to `ReportAggregates`

EricWF:
The new names are:

* `--benchmark_report_aggregates_only`
* `->ReportAggregatesOnly(bool = true)`

src/benchmark.cc
```diff
@@ -66,6 +66,11 @@ DEFINE_int32(benchmark_repetitions, 1,
              "The number of runs of each benchmark. If greater than 1, the "
              "mean and standard deviation of the runs will be reported.");
 
+DEFINE_bool(benchmark_report_aggregates_only, false,
+            "Report the result of each benchmark repetitions. When 'false' is "
```
dominichamon: when true :)

src/benchmark.cc
```diff
@@ -575,6 +621,13 @@ void BenchmarkImp::Repetitions(int n) {
   repetitions_ = n;
 }
 
+void BenchmarkImp::ReportAggregatesOnly(bool value) {
+  if (report_options_ == RO_Unspecified)
+      report_options_ = RO_Default;
+  if (value) report_options_ &= ~RO_ReportAggregatesOnly;
```
dominichamon: this is very short and neat, but it requires a lot of work above. can you just be more explicit here about the state changes and lose all the operator overload cleverness please?

src/benchmark.cc
EricWF: Fixed.

src/benchmark.cc
EricWF: Done. The operator overloads are removed and  the report options are no longer treated as a bit mask.

dominichamon:
appveyor now failing because http://stackoverflow.com/questions/14191566/c-mutex-in-namespace-std-does-not-name-a-type

LGTM

EricWF:
> appveyor now failing because http://stackoverflow.com/questions/14191566/c-mutex-in-namespace-std-does-not-name-a-type

Apparently only MinGW-64 provides POSIX threading. I'm juts going to remove the MinGW32 builder. I think one bot is enough coverage.

PR #269, opened 2016-08-07
Set and document supported compilers

EricWF:
This patch adds documentation for the compilers the library supports.

@dominichamon Do these minimum versions sound reasonable to you?

Side note: The reason I chose GCC 4.8 instead of 4.7 is (A) we have CI that tests 4.8 and (B) 4.7.x currently fails to compile the library on my machine. There is no easy workaround for this issue.


dominichamon:
and GCC 4.7 has fundamentally broken C++11 support. IIRC it mishandled the feature macro in some key cases.

dominichamon:
LGTM

Issue #274, opened 2016-08-16
Google Benchmark doesn't have benchmark_main library (gtest_main equivalent)


AlexanderSidorenko:
Hello,

Google Test has a nice feature: gtest_main library that supplies default main.
Since Google Test has default recommended BENCHMARK_MAIN, I thought it would be great to have a library that supplies this default implementation.
It would make Google Benchmark consistent with Google Test.

Thank you,
Alex

AlexanderSidorenko:
I have created pull request:
https://github.com/google/benchmark/pull/275
Please let me know if any additional changes are required.

AlexanderSidorenko:
As discussed in pull request (see above), team doesn't see the need to retrofit functionality in question at this stage due to added complexity.

PR #275, opened 2016-08-16
Adding benchmark_main library that is baseline benchmark + BENCHMARK_MAIN()

AlexanderSidorenko:
Please see issue:
https://github.com/google/benchmark/issues/274

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ BENCHMARK_MAIN();
 ```
 
 You only need to have `BENCHMARK_MAIN` in your program if you are linking against `benchmark` library.
-Alternatively, you can sue `benchmark_main` library that has `BENCHMARK_MAIN` compiled in.
+Alternatively, you can use `benchmark_main` library that has `BENCHMARK_MAIN` compiled in.
 
 ### Passing arguments
 Sometimes a family of benchmarks can be implemented with just one routine that
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ BENCHMARK(BM_StringCopy);
 BENCHMARK_MAIN();
 ```
 
+You only need to have `BENCHMARK_MAIN` in your program if you are linking against `benchmark` library.
+Alternatively, you can sue `benchmark_main` library that has `BENCHMARK_MAIN` compiled in.
+
 ### Passing arguments
 Sometimes a family of benchmarks can be implemented with just one routine that
 takes an extra argument to specify which one of the family of benchmarks to
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,7 @@
 # Allow the source files to find headers in src/
 include_directories(${PROJECT_SOURCE_DIR}/src)
+# Allow the source files to find headers in include/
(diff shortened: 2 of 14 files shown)


AlexanderSidorenko:
I signed CLA. Should go through the system shortly.

README.md
```diff
@@ -32,6 +32,9 @@ BENCHMARK(BM_StringCopy);
 BENCHMARK_MAIN();
 ```
 
+You only need to have `BENCHMARK_MAIN` in your program if you are linking against `benchmark` library.
+Alternatively, you can sue `benchmark_main` library that has `BENCHMARK_MAIN` compiled in.
```
dominichamon: s/sue/use/

src/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 # Allow the source files to find headers in src/
 include_directories(${PROJECT_SOURCE_DIR}/src)
+# Allow the source files to find headers in include/
```
dominichamon: this wasn't needed before, why is it needed now?

README.md
AlexanderSidorenko: Fixed, thanks.

src/CMakeLists.txt
AlexanderSidorenko: BENCHMARK_MAIN() is currently defined in publicly exposed interface (include/ folder), and it is now required in src/ folder.
Unfortunately, there is no common place that is shared between src/ and include/: they are decoupled. Otherwise, I would move BENCHMARK_MAIN() there.
I added comment explaining why it's now needed.

AlexanderSidorenko:
Will look into build failure due to test failure in the evening when I get my hands on Windows machine.

dominichamon:
i don't think it's your change. it looks like complexity test has been broken for a while (https://ci.appveyor.com/project/google/benchmark/build/361/job/jq6e4wp072gtu90i)

AlexanderSidorenko:
Thanks for clarification.
Is there anything else that I could help with to make this go through?

EricWF:
I would love to know what convenience this offers over `BENCHMARK_MAIN()`
because it's not immediately apparent to me.

AlexanderSidorenko:
The difference is subtle - no need to have `BENCHMARK_MAIN()` in your code, but it has added benefit of being consistent with Google Test. Google test supplies gtest and gtest_main libraries, with latter taking care of main for you.

pleroy:
+1 to @EricWF's comment.  Each time we add something to the API we are making it a little bit harder to use.  Now you have two ways to create a main and the naive user has no way to decide which one they should pick.

The analogy with gtest is flawed: AFAICT gtest doesn't provide a "main" macro, it has `RUN_ALL_TESTS` but that's a different thing entirely.  If we were starting from scratch it might be nicer to have a "main" library than a "main" macro, but we cannot break compatibility so that's not an option.

I'd say that the whole thing is insufficiently broken and that fixing it is not worth the added complexity.

AlexanderSidorenko:
Pascal,
The reasoning you provided makes sense to me. Agree let's not introduce it.
I will close pull request, then.

Thanks!

PR #276, opened 2016-08-18
Alternative to user-defined counters

EricWF:
@biojppm I took the liberty of implementing a version of your patch with the interface I would like to see. Please take a look at this PR, in particular the removal of `BenchmarkCounters` and  the changes to `Counter`. 

@biojppm I don't want to hijack your pull request, I just wanted you to see the interface I had in mind.


pleroy:
FWIW I like this API much better.  No time to look at #262 tonight though.

EricWF:
Abandoning since this was only ever meant as an example.

PR #277, opened 2016-08-18
Refactor output test runner into standalone module.

EricWF:
I plan to commit this without review. I just want to make sure the bots handle it first.

### DIFF
--- a/test/output_test.h
+++ b/test/output_test.h
@@ -1,0 +1,83 @@
+#ifndef TEST_OUTPUT_TEST_H
+#define TEST_OUTPUT_TEST_H
+
+#undef NDEBUG
+#include "benchmark/benchmark.h"
+#include "../src/re.h"
+#include <vector>
+#include <string>
+#include <initializer_list>
+#include <memory>
+#include <utility>
+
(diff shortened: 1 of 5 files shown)


test/output_test.h
```diff
@@ -0,0 +1,76 @@
...
+// Run all output tests.
+void RunOutputTests(int argc, char* argv[]);
+
+// ========================================================================= //
+// --------------------------- Misc Utilities ------------------------------ //
+// ========================================================================= //
+
+namespace {
+
+std::string dec_re = "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?";
```
dominichamon: a static non-pod defined in a header? say it ain't so!

this should probably be

extern const char* const dec_re;

and then in the .cc

const char* const dec_re = "...";

dominichamon:
so much cleaner :)

test/output_test.h
EricWF: The non-pod part is definitely an issue, but static initialization is safer if the initializer is in the header so the compiler can see it.

PR #278, opened 2016-08-28
Supply old `RangePair` and `ArgPair` API for C++03 compatibility.

EricWF:
The new `Ranges` and `Args` interface requires C++11 initializer lists to be useful. In C++03 the new interface is almost impossible to use. This patch re-adds the old C++03 interface for continued compatibility.

Additionally this patch re-adds the `range_x()` and `range_y()` methods but marks them as deprecated. This should allow older benchmarks to continue to compile while making it clear that this interface is deprecated.

The final change in this patch gives `ranges(pos)` the default argument `0` allowing `ranges()` to work. This seems like a convenience to me but I"m happy to remove it.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -425,10 +425,16 @@ public:
 
   // Range arguments for this run. CHECKs if the argument has been set.
   BENCHMARK_ALWAYS_INLINE
-  int range(std::size_t pos) const {
+  int range(std::size_t pos = 0) const {
       assert(range_.size() > pos);
       return range_[pos];
   }
+
+  BENCHMARK_DEPRECATED_MSG("use 'range(0)' instead")
+  int range_x() const { return range(0); }
+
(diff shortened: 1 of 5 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -502,11 +508,31 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept arg1, arg2 ...
   Benchmark* Args(const std::vector<int>& args);
 
+  // Equivalent to Args({x, y})
+  // NOTE: This is a legacy C++03 interface provided for compatibility only.
+  //   New code should use 'Args'.
+  Benchmark* ArgPair(int x, int y) {
```
dominichamon: should we also mark these as deprecated?

EricWF: I don't think so. Using the replacements requires C++11. So if you want C++03 benchmarks you still need this interface.

dominichamon:
LGTM

PR #279, opened 2016-08-28
Add warning for large benchmark families

EricWF:
`->Ranges(...)` generates a list of arguments which is the Cartesian product of  all of the supplied ranges. 
If the list contains 100 items then the benchmark will be run 100 times, which can be very slow and possibly unexpected. This patch adds a warning message when the generated list is "large".

Additionally this patch re-works the error output to use the reporters error stream instead of `std::cerr`.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -115,6 +115,9 @@ bool IsZero(double n) {
 
 // For non-dense Range, intermediate values are powers of kRangeMultiplier.
 static const int kRangeMultiplier = 8;
+// The size of a benchmark family determines is the number of inputs to repeat
+// the benchmark on. If this is "large" then warn the user during configuration.
+static const size_t kMaxFamilySize = 100;
 static const size_t kMaxIterations = 1000000000;
 
 bool running_benchmark = false;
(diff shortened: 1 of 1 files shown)


src/benchmark.cc
```diff
@@ -446,12 +452,23 @@ bool BenchmarkFamilies::FindBenchmarks(
...
-
-    for (auto const& args : family->args_) {
-      const std::vector<int>* thread_counts =
+    const std::vector<int>* thread_counts =
         (family->thread_counts_.empty()
          ? &one_thread
-         : &family->thread_counts_);
+         : &static_cast<const std::vector<int>&>(family->thread_counts_));
+    const size_t family_size = family->args_.size() * thread_counts->size();
+    // The benchmark will be run at least 'total' different inputs. If 'total'
```
dominichamon: change total to family_size in the comment

src/benchmark.cc
```diff
@@ -1093,47 +1110,52 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter,
...
-  } else {
-    // Setup the reporters
-    std::ofstream output_file;
-    std::unique_ptr<BenchmarkReporter> default_console_reporter;
-    std::unique_ptr<BenchmarkReporter> default_file_reporter;
-    if (!console_reporter) {
-      auto output_opts = FLAGS_color_print ? ConsoleReporter::OO_Color
-                                           : ConsoleReporter::OO_None;
-      default_console_reporter = internal::CreateReporter(
+  // Setup the reporters
```
dominichamon: we do a lot of work now just to list the benchmarks.. why?

EricWF: I was to replace usage of `std::cout`/`std::cerr` with the reporters output/error streams, even when just listing the tests.

dominichamon:
LGTM

PR #280, opened 2016-08-29
Add -DBENCHMARK_USE_LIBCXX:BOOL=<value> option.

EricWF:
This patch is a bit unfortunate. Currently it's impossible to configure/build the library using libc++ and GCC. This patch adds the option `-DBENCHMARK_USE_LIBCXX=<value>` which can be used to build against libc++, even with GCC.

This option is needed because GCC doesn't support the -stdlib=libc++ flag and because additional C++ libraries cannot be passed during CMake configuration (only compile/link flags can be). The reason `-lc++` cannot be passed in `CMAKE_EXE_LINKER_FLAGS` is because the linker flags are added *before* the input files but `-lc++` needs to appear after them. Unfortunately CMake doesn't provide command line options for adding additional libraries.

Sorry about this PR. I would prefer fixing CMake to support this case but that's unlikely to happen soon.
I need this patch to support ongoing libc++ benchmark work.


CMakeLists.txt
```diff
@@ -78,7 +79,6 @@ else()
   add_cxx_compiler_flag(-pedantic-errors)
   add_cxx_compiler_flag(-Wshorten-64-to-32)
   add_cxx_compiler_flag(-Wfloat-equal)
-  add_cxx_compiler_flag(-Wzero-as-null-pointer-constant)
```
EricWF: I removed this warning because GCC generates it when compiling w/ libc++ even though the actual warning originates from a system header.

dominichamon: maybe comment it out with this as documentation? unless you think there's never a time we'll be able to put it back.

dominichamon:
thank you so much for bothering with this, even though it's a pain.

there is another alternative: should we replace cmake with bazel or something else?

CMakeLists.txt
EricWF: I'll just disable it when `BENCHMARK_USE_LIBCXX` is `ON`.

As I side note I dislike this warning in general. The warning seems to take into account constant propagation when it should really only warn on the literal token '0'.

EricWF:
> thank you so much for bothering with this, even though it's a pain.

No worries. I kinda need it :-P

> there is another alternative: should we replace cmake with bazel or something else?

I don't think this issues is motivation enough to switch from CMake. Hopefully the situation for using libc++ will improve over time.

dominichamon:
LGTM assuming CI passes

Issue #282, opened 2016-08-30
Add Benchmark::ThreadRange() version with increment instead of multiply


hydroo:
Hi,

for my work I want to use a range of threads.
So far the incrementing strategy seems to be limited to just multiplying with a factor.
I.e. you can only use thread counts like 1, 2, 4, 8, 16 or 1, 3, 9, 27.

So I made this small patch.
https://github.com/hydroo/google-benchmark/commit/345d2bac3c132e523b7a7a94160d4dba13c23ead
https://github.com/hydroo/google-benchmark/commit/11aeb3198c4bb8436e9e8892071832ba40cf1546

Should I create a pull request?

Best regards,
Ronny

P.S.: Is ThreadRange() not documented? I only came across it while reading the source.

EricWF:
> Should I create a pull request?

If you have an idea and a patch you should always make a PR. Issues are for people without patches :-P

> for my work I want to use a range of threads

You must have a lot of threads... I'm initially skeptical of the need for this feature. Typically you don't want to exceed the number of threads supported by the hardware (at least by much). Please try and add a motivating example when you open the PR.

I'm closing this issue to make way for the PR.

PR #283, opened 2016-08-31
Add Benchmark::ThreadRange() version with increment instead of multiply

hydroo:
Hi,

referring to #282.

>> for my work I want to use a range of threads.
>> So far the incrementing strategy seems to be limited to just multiplying with a factor.
>> I.e. you can only use thread counts like 1, 2, 4, 8, 16 or 1, 3, 9, 27.

>> So I made this small patch.

>> Should I create a pull request?

> If you have an idea and a patch you should always make a PR. Issues are for people without patches :-P

>> for my work I want to use a range of threads

> You must have a lot of threads... I'm initially skeptical of the need for this feature. Typically you don't want to exceed the number of threads supported by the hardware (at least by much). Please try and add a motivating example when you open the PR.

> I'm closing this issue to make way for the PR.

Yes we do have up to 24 or 48 threads per board at the moment.

I use it for benchmarking synchronization algorithms. (Not a very common thing to do,)
I will probably need a few more things in the future like thread pinning and more accurate version of PauseTiming so I can make sure they leave at an exact time.
I'm not even sure google/benchmark is the right tool for the job. But it seems nice. The output, functionality and code is pretty simple and clear.

Motivating example:

```c++
template<typename barrier>
static void BM_barrier(benchmark::State& state) {

    while (state.KeepRunning()) {

        static barrier b;
        if (state.thread_index == 0) { b.init(state.threads); }

        state.PauseTiming();

        auto start = std::chrono::high_resolution_clock::now();
        b.wait(state.thread_index);
        auto end   = std::chrono::high_resolution_clock::now();

        auto elapsed_seconds = std::chrono::duration_cast<std::chrono::duration<double>>(end - start);

        if (elapsed_seconds < std::chrono::nanoseconds(0)) {
            state.SkipWithError("elapsed seconds < 0");
            break;
        }
                                                                                                                                                                   
        state.SetIterationTime(elapsed_seconds.count());       
                                                                                                                                                                            
        if (state.thread_index == 0) { b.destroy(); }
    }                                                                                                                                                                       
}

BENCHMARK_TEMPLATE(BM_barrier, baseline_barrier)->ThreadRange2(1, 4)->UseManualTime();
BENCHMARK_TEMPLATE(BM_barrier, central_counter_barrier)->ThreadRange2(1, 4)->UseManualTime();
BENCHMARK_TEMPLATE(BM_barrier, sense_reversing_centralized_barrier)->ThreadRange2(1, 4)->UseManualTime();

BENCHMARK_MAIN();
```

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -595,7 +595,13 @@ public:
   //    Foo in 4 threads
   //    Foo in 8 threads
   //    Foo in 16 threads
-  Benchmark* ThreadRange(int min_threads, int max_threads);
+  Benchmark *ThreadRange(int min_threads, int max_threads);
+
+  // For each value n in the range, run this benchmark once using n threads.
+  // min_threads and max_threads are always included in the range.
+  // stride specifies the increment. E.g. DenseThreadRange(1, 8, 3) starts
+  // a benchmark with 1, 4, 7 and 8 threads.
+  Benchmark *DenseThreadRange(int min_threads, int max_threads, int stride = 1);
 
   // Equivalent to ThreadRange(NumCPUs(), NumCPUs())
   Benchmark* ThreadPerCpu();
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -293,6 +293,7 @@ public:
   void ComplexityLambda(BigOFunc* complexity);
   void Threads(int t);
   void ThreadRange(int min_threads, int max_threads);
+  void DenseThreadRange(int min_threads, int max_threads, int stride);
   void ThreadPerCpu();
   void SetName(const char* name);
 
@@ -548,6 +549,18 @@ void BenchmarkImp::ThreadRange(int min_threads, int max_threads) {
   AddRange(&thread_counts_, min_threads, max_threads, 2);
 }
 
+void BenchmarkImp::DenseThreadRange(int min_threads, int max_threads,
+                                    int stride) {
+  CHECK_GT(min_threads, 0);
+  CHECK_GE(max_threads, min_threads);
+  CHECK_GE(stride, 1);
+
+  for (auto i = min_threads; i < max_threads; i += stride) {
+    thread_counts_.push_back(i);
+  }
+  thread_counts_.push_back(max_threads);
+}
+
 void BenchmarkImp::ThreadPerCpu() {
   static int num_cpus = NumCPUs();
   thread_counts_.push_back(num_cpus);
@@ -579,6 +592,7 @@ void BenchmarkImp::AddRange(std::vector<int>* dst, int lo, int hi, int mult) {
     dst->push_back(hi);
   }
 }
+
 
 Benchmark::Benchmark(const char* name)
     : imp_(new BenchmarkImp(name))
@@ -684,6 +698,12 @@ Benchmark* Benchmark::Threads(int t) {
 
 Benchmark* Benchmark::ThreadRange(int min_threads, int max_threads) {
   imp_->ThreadRange(min_threads, max_threads);
+  return this;
+}
+
+Benchmark *Benchmark::DenseThreadRange(int min_threads, int max_threads,
+                                       int stride) {
+  imp_->DenseThreadRange(min_threads, max_threads, stride);
   return this;
 }
 
--- a/test/benchmark_test.cc
+++ b/test/benchmark_test.cc
@@ -220,5 +220,26 @@ BENCHMARK_CAPTURE(BM_non_template_args, basic_test, 0, 0);
 
 #endif // __cplusplus >= 201103L
 
-BENCHMARK_MAIN()
+static void BM_DenseThreadRanges(benchmark::State &st) {
+  switch (st.range(0)) {
+  case 1:
+    assert(st.threads == 1 || st.threads == 2 || st.threads == 3);
+    break;
+  case 2:
+    assert(st.threads == 1 || st.threads == 3 || st.threads == 4);
+    break;
+  case 3:
+    assert(st.threads == 5 || st.threads == 8 || st.threads == 11 ||
+           st.threads == 14);
+    break;
+  default:
+    assert(false && "Invalid test case number");
+  }
+  while (st.KeepRunning()) {
+  }
+}
+BENCHMARK(BM_DenseThreadRanges)->Arg(1)->DenseThreadRange(1, 3);
+BENCHMARK(BM_DenseThreadRanges)->Arg(2)->DenseThreadRange(1, 4, 2);
+BENCHMARK(BM_DenseThreadRanges)->Arg(3)->DenseThreadRange(5, 14, 3);
 
+BENCHMARK_MAIN()


hydroo:
I signed it!

EricWF:
> I will probably need a few more things in the future like thread pinning and more accurate version of PauseTiming so I can make sure they leave at an exact time.

FYI I PR #256  makes PauseTiming() per thread instead of per-process so it doesn't perform synchronization and block the threads.

hydroo:
Thanks for the info.

include/benchmark/benchmark_api.h
```diff
@@ -574,6 +574,8 @@ class Benchmark {
   //    Foo in 16 threads
   Benchmark* ThreadRange(int min_threads, int max_threads);
 
+  Benchmark* ThreadRange2(int min_threads, int max_threads);
```
EricWF: I would suggest `DenseThreadRange(int min_threads, int max_threads, int stride=1)` since it matches the `DenseRange` function we already have.

src/benchmark.cc
```diff
@@ -625,6 +627,13 @@ void BenchmarkImp::ThreadRange(int min_threads, int max_threads) {
   AddRange(&thread_counts_, min_threads, max_threads, 2);
 }
 
+void BenchmarkImp::ThreadRange2(int min_threads, int max_threads) {
+  CHECK_GT(min_threads, 0);
+  CHECK_GE(max_threads, min_threads);
+
+  AddRange2(&thread_counts_, min_threads, max_threads, 1);
```
EricWF: Just implement `AddRange2` directly in this function body. No need to dispatch.

EricWF:
Alright so there are a couple things left to do.

1. The `DenseThreadRange(...)` declaration in `benchmark_api.h` needs to be documented like the other functions around it. Take a look at `DenseRange` for an example.

2. This needs a test. Somethink like this should work:

```c++

static void BM_ThreadRanges(benchmark::State& st) {
  switch (st.range(0)) {
  case 1:
    assert(st.threads == 1 || st.threads == 2 || st.threads == 3);
    break;
  case 2:
    assert(st.threads == 1 || st.threads == 3);
    break;
  case 3:
    assert(st.threads == 5 || st.threads == 8 || st.threads == 11 || st.threads == 14);
    break;
  default:
    assert(false && "Invalid test case number");
  }
  while (st.KeepRunning()) {}
}
BENCHMARK(BM_ThreadRanges)->Arg(1)->DenseThreadRange(1, 3);
BENCHMARK(BM_ThreadRanges)->Arg(2)->DenseThreadRange(1, 3, 2);
BENCHMARK(BM_ThreadRanges)->Arg(3)->DenseThreadRange(5, 14, 3);
```

hydroo:
Is there a clang-format file for this project? I just tried clang-format -style=Google on benchmark_api.h and it changed a lot lot. So that's probably not what you do.

EricWF:
There is a config file now. It was just checked in today.
To format patches use [clang-format-diff](http://clang.llvm.org/docs/ClangFormat.html#script-for-patch-reformatting)

hydroo:
When I make test, benchark_test.cc is not executed. Is there a non-manual way to run it? Or how is this intended?

I add the test to the end of benchark_test.cc, ok?

hydroo:
This clang format file seems to be for a clang version newer than 3.8 (Newest Ubuntu).
Error: error: unknown key 'IncludeIsMainRegex'

EricWF:
Stick the test wherever makes sense and use --style=Google if the config doesn't work.

hydroo:
I force pushed my master, since I had troubles with git. I used this opportunity to clean up my commits as well. It's now only two.

EricWF:
Looks good to me! Thanks for the patch.

EricWF:
To discuss some of your other comments:

> more accurate version of PauseTiming so I can make sure they leave at an exact time.

PauseTiming now act's a lot differently. In particular it's thread-local and doesn't act as a barrier so hopefully this solves your problem. Also `SetIterationTime` is also per-thread now. The final value is the average of each of the threads values. 

I'm looking into "thread pinning" as a feature, but I'm not sure what the correct way to pin the threads is.
Could you clarify what you need here?

hydroo:
Thanks!

Last Saturday I had big troubles getting my code to work in the benchmark harness, again.
I liked having the barrier there.
I am not sure what was going on.
It would help, if I could acquire shared resources before spawning any threads and pass them.
If all threads start the benchmark at the same time, I have to make sure everyone has the correct data.
Starting the benchmarks via my own main seems to not help with that, since I can only start all benchmarks at the same time, and not individually (I can only acquire all resources before running any benchmark, and release them afterwards).

Can threads overtake each other in KeepRunning() or is there a barrier between iterations?

As for SetIterationTime, I would also like to have the maximum of all threads and the time from the earliest `before` to the latest `after` (kind of time maximum time collectively spent in the code).

Since I benchmark synchronization primitives I want to make sure (1) threads are not migrated during benchmark execution, (2) threads are placed in a sensible way (not 1, 3 on 2 cpus with 2 threads each, e.g.), perhaps I want to even place them individually myself because distances between CPUs affect performance.

Last time I used `pthread_getaffinity_np`, `pthread_setaffinity_np`. Haven't looked into it yet. Maybe there is a newer, nicer way. Also `sched_setaffinity` exists on linux.

EricWF:
> Last Saturday I had big troubles getting my code to work in the benchmark harness, again.
> I liked having the barrier there.

Sorry but the barrier in `PauseTiming()` essentially made multithreaded benchmarks synchronous , which is obviously not what we want. Removing the barrier drastically improved the timing accuracy of multithreaded benchmarks.

I plan to add barrier functionality to `State` so that users can block when need be, but I don't think `PauseTiming()` should be it.

> I am not sure what was going on.
> It would help, if I could acquire shared resources before spawning any threads and pass them.
> If all threads start the benchmark at the same time, I have to make sure everyone has the correct data.
> Starting the benchmarks via my own main seems to not help with that, since I can only start all benchmarks at the same time, and not individually (I can only acquire all resources before running any benchmark, and release them afterwards).

That shouldn't be a problem with the existing API. See the example below.

> Can threads overtake each other in KeepRunning() or is there a barrier between iterations?

`KeepRunning()` only acts as a barrier on the first and last iteration. This ensures that the initialization code has been run before any thread enters the loop and that the teardown code only runs after all threads are done. You should be able to setup/teardown shared state between multiple threads like this:

```c++
void SetupGlobalState();
void TeardownGlobalState();
std::vector<int> global_state;
static BM_Foo(benchmark::State& st) {
  if (st.thread_index == 0) { // A single thread executes this setup.
    SetupGlobalState();
  }
  PinCurrentThread(); // All thread execute this
  while (st.KeepRunning()) { // No thread starts the loop until all threads are here.
     /* ... */
  } // No thread passes here until all threads are complete.
  if (st.thread_index == 0) {
    TeardownGlobalState();
  }
}
```

> As for SetIterationTime, I would also like to have the maximum of all threads and the time from the earliest before to the latest after (kind of time maximum time collectively spent in the code).

Yeah I think `SetIterationTime()` could offer a bunch more functionality. I'm looking into better support for custom timers.

> 
> Since I benchmark synchronization primitives I want to make sure (1) threads are not migrated during benchmark execution, (2) threads are placed in a sensible way (not 1, 3 on 2 cpus with 2 threads each, e.g.), perhaps I want to even place them individually myself because distances between CPUs affect performance.

I think users will generally want to place the threads themselves. That's why IDK if this can be a library feature. Also it's super easy for users to pin threads today (see above).

> Last time I used pthread_getaffinity_np, pthread_setaffinity_np. Haven't looked into it yet. Maybe there is a newer, nicer way. Also sched_setaffinity exists on linux.

No that's the way to do it. But as mentioned above it's not clear what to set the affinity to.

PR #285, opened 2016-08-31
Fixture: add non const Setup() and TearDown().

biojppm:
This allows write-access to the State argument, which is important in upcoming user-defined counter functionality.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -713,8 +713,8 @@ public:
     virtual void SetUp(const State&) {}
     virtual void TearDown(const State&) {}
     // ... In favor of these.
-    virtual void SetUp(State& st) { SetUp(static_cast<State const&>(st)); }
-    virtual void TearDown(State& st) { TearDown(static_cast<State const&>(st)); }
+    virtual void SetUp(State& st) { SetUp(static_cast<const State&>(st)); }
+    virtual void TearDown(State& st) { TearDown(static_cast<const State&>(st)); }
 
 protected:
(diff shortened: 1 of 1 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -709,8 +709,12 @@ class Fixture: public internal::Benchmark {
       this->TearDown(st);
     }
 
+    // These will be deprecated ...
     virtual void SetUp(const State&) {}
     virtual void TearDown(const State&) {}
+    // ... In favor of these.
+    virtual void SetUp(State& st) { SetUp(static_cast<State const&>(st)); }
```
pleroy: Please be consistent wrt the placement of `const`.  This library generally puts it before the type name.  (I am in the "after the type name" camp, but consistency trumps.)

include/benchmark/benchmark_api.h
```diff
@@ -709,8 +709,12 @@ class Fixture: public internal::Benchmark {
       this->TearDown(st);
     }
 
+    // These will be deprecated ...
     virtual void SetUp(const State&) {}
     virtual void TearDown(const State&) {}
+    // ... In favor of these.
+    virtual void SetUp(State& st) { SetUp(static_cast<const State&>(st)); }
```
dominichamon: const_cast, probably.

EricWF: Nah this should be static_cast. const_cast is for removing const.

dominichamon: gah, completely read it wrong. i thought it was the other way around (with the old ones forwarding to the new ones)

PR #287, opened 2016-09-03
Refactor benchmark.cc into benchmark_register.cc and benchmark.cc

EricWF:
The benchmark.cc file is getting really big and it contains a bunch of
unrelated components. This patch separates the files into two separate
parts. The "runtime" parts (benchmark.cc) and the "registration" parts (benchmark_register.cc).

Optimally benchmark.cc should become benchmark_run.cc but I don't want to ruin the commit history.

This patch also removes the PIMPL used by Benchmark. Previously we couldn't
have STL types in the interface but now we can. Therefore there is no reason
to keep BenchmarkImp.


EricWF:
A couple of other changes in this patch:

* Change CMake to glob for source/header files so we don't have to enumerate them.
* Merge `re_std.cc` and `re_posix.cc` into `re.h`.

EricWF:
I was originally waiting on PR #262 to land before committing this (in order to avoid merge conflicts). But it's getting to be a big patch so I'm committing it before it gets any bigger.

include/benchmark/benchmark_api.h
```diff
@@ -616,9 +628,25 @@ class Benchmark {
   Benchmark(Benchmark const&);
   void SetName(const char* name);
 
+  int ArgsCnt() const;
+
+  static void AddRange(std::vector<int>* dst, int lo, int hi, int mult);
+
 private:
   friend class BenchmarkFamilies;
-  BenchmarkImp* imp_;
```
dominichamon: you popped my pimpl!

i'm actually fine with this. originally this was here to avoid having too much detail in the public header.

EricWF: Sorry! I figured since we already use STL types in `State` there was no need to keep PIMPL'ing benchmark. Mostly it was just the cost of the double dispatch that I wanted to remove.

PR #288, opened 2016-09-04
Flush output stream after writing a benchmark run

nickhutchinson:
If a reporter's output stream isn't line-buffered (e.g. it's not writing
to a terminal) then it can be some time before a write to it becomes
visible.

This is problematic if, say, you're wanting to use `tail -f` to view the
file written to via --benchmark_out. Or if the application crashes,
leaving you with no results.

Addressed by flushing the reporters' output streams at the end of
ReportRuns() and ReportContext().

Tested with:
```
./test/basic_test --benchmark_format=[console|json|csv] | cat
```

Before, nothing would be printed until all benchmarks had run.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -480,17 +480,31 @@ void RunMatchingBenchmarks(const std::vector<Benchmark::Instance>& benchmarks,
   // Keep track of runing times of all instances of current benchmark
   std::vector<BenchmarkReporter::Run> complexity_reports;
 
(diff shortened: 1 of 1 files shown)


EricWF:
I think it would be better to flush the streams in `RunMatchingBenchmarks` rather than in each reporter.

nickhutchinson:
Addressed review comments

EricWF:
Thanks!

PR #292, opened 2016-09-26
Fix working on mac clock_gettime

PSIAlt:
The problem was mac os x has CLOCK_PROCESS_CPUTIME_ID definition, but does not have clock_gettime()

### DIFF
--- a/src/timers.cc
+++ b/src/timers.cc
@@ -101,6 +101,7 @@ BENCHMARK_NORETURN static void  DiagnoseAndExit(const char* msg) {
 
 
 double ProcessCPUUsage() {
+// FIXME We want to use clock_gettime, but its not available in MacOS 10.11. See https://github.com/google/benchmark/pull/292
 #if defined(CLOCK_PROCESS_CPUTIME_ID) && !defined(BENCHMARK_OS_MACOSX)
   struct timespec spec;
   if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &spec) == 0)
@@ -124,6 +125,7 @@ double ProcessCPUUsage() {
 }
 
 double ThreadCPUUsage() {
+// FIXME We want to use clock_gettime, but its not available in MacOS 10.11. See https://github.com/google/benchmark/pull/292
 #if defined(CLOCK_THREAD_CPUTIME_ID) && !defined(BENCHMARK_OS_MACOSX)
   struct timespec ts;
   if (clock_gettime(CLOCK_THREAD_CPUTIME_ID, &ts) == 0)
--- a/src/timers.cc
+++ b/src/timers.cc
@@ -101,7 +101,7 @@ BENCHMARK_NORETURN static void  DiagnoseAndExit(const char* msg) {
 
 
 double ProcessCPUUsage() {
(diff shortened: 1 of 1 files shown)


PSIAlt:
I signed it!

EricWF:
That's weird... IDK why it would define that macro if it doesn't provide `clock_gettime`.

What version of OS X do you have? I can't reproduce on 10.11.6.

PSIAlt:
Same, 10.11.6. It compiles, but does not pass tests. If I pick one test, it does not work also:

```
The following tests FAILED:
	  1 - benchmark (OTHER_FAULT)
	  2 - filter_simple (OTHER_FAULT)
	  4 - filter_suffix (OTHER_FAULT)
	  6 - filter_regex_all (OTHER_FAULT)
	  8 - filter_regex_blank (OTHER_FAULT)
	 12 - filter_regex_wildcard (OTHER_FAULT)
	 14 - filter_regex_begin (OTHER_FAULT)
	 16 - filter_regex_begin2 (OTHER_FAULT)
	 18 - filter_regex_end (OTHER_FAULT)
	 20 - options_benchmarks (OTHER_FAULT)
	 21 - basic_benchmark (OTHER_FAULT)
	 22 - diagnostics_test (OTHER_FAULT)
	 23 - skip_with_error_test (OTHER_FAULT)
	 25 - fixture_test (OTHER_FAULT)
	 26 - register_benchmark_test (OTHER_FAULT)
	 27 - map_test (OTHER_FAULT)
	 28 - multiple_ranges_test (OTHER_FAULT)
	 29 - reporter_output_test (OTHER_FAULT)
	 30 - cxx03 (OTHER_FAULT)
	 31 - complexity_benchmark (OTHER_FAULT)
Errors while running CTest
make: *** [test] Error 8
Alt ~/repos/google-benchmark $ ./test/benchmark_test 
Run on (4 X 1000 MHz CPU s)
2016-09-26 23:45:01
Benchmark                                         Time           CPU Iterations
-------------------------------------------------------------------------------
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /Users/Alt/repos/google-benchmark/./test/benchmark_test
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /Users/Alt/repos/google-benchmark/./test/benchmark_test
  Expected in: /usr/lib/libSystem.B.dylib

Trace/BPT trap: 5

```

EricWF:
OK, so this is interesting. My "Command line" headers don't provide the `CLOCK_PROCESS_CPUTIME_ID` but the XCode SDK ones do. In order to use those headers I have to define `MACOSX_DEPLOYMENT_TARGET=10.12`, which is the only SDK provided by XCode. 

I'm guessing that `/usr/lib/libSystem.B.dylib` will contain the required `clock_gettime` on OS X 10.12, but not 10.11 which is what we are both using.

I would prefer to use `clock_gettime` once OS X actually provides it, but I'm OK with this fix for now. Please add a `// FIXME` note above the change so we know to remove it in the future.

EricWF:
LGTM. Thanks for the fix.

PR #298, opened 2016-10-05
Pass flag string to `check_cxx_compiler_flag`

ktnyt:
The commit provides a patch for [Issue #297 ](https://github.com/google/benchmark/issues/297) and related compiler flag problems.


dominichamon:
i don't know how we didn't catch that earlier. thank you!

PR #299, opened 2016-10-05
Fixing invalid json format

jjones646:
When exporting json, the _rms_ key does not wrap its value in quotes. This fixes that.

**Before:**
```
{
    "name": "my_benchmark",
    "rms": 12%
}
```

**After:**
```
{
    "name": "my_benchmark",
    "rms": "12%"
}
```


jjones646:
signed it

PR #302, opened 2016-10-07
Simplify clang-format and apply to tests

dominichamon:
@EricWF I think you have more clang-format experience.. this format file should work everywhere, i believe (no version-specific issues).


EricWF:
@dominichamon This LGTM. I imagine the new formulation shouldn't run into any versioning issues.

PR #304, opened 2016-10-21
Fix displayed time unit in mean and standard deviation.

mkurdej:
`Unit` passed to benchmark was taken into account only for individual runs, not for aggregated info.


dominichamon:
Thanks!

PR #305, opened 2016-10-24
Add ArgName() and ArgNames() methods to name arguments/ranges.

mkurdej:
This PR allows the user to name arguments/ranges.

Instead of showing only *"benchmark/1"* (when called with `Arg(1)`), one can call the benchmark with `Arg(1)->ArgName("arg")` and the displayed name will be *"benchmark/arg:1"*.
Readability purposes only.

### DIFF
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -153,12 +153,16 @@ bool BenchmarkFamilies::FindBenchmarks(
         // Add arguments to instance name
         size_t arg_i = 0;
         for (auto const& arg : args) {
+          instance.name += "/";
+
           if (arg_i < family->arg_names_.size()) {
-            instance.name +=
-                StringPrintF("/%s:", family->arg_names_[arg_i].c_str());
-          } else {
-            instance.name += "/";
+            const auto& arg_name = family->arg_names_[arg_i];
+            if (!arg_name.empty()) {
+              instance.name +=
+                  StringPrintF("%s:", family->arg_names_[arg_i].c_str());
+            }
           }
+
           AppendHumanReadable(arg, &instance.name);
           ++arg_i;
         }
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -151,8 +151,16 @@ bool BenchmarkFamilies::FindBenchmarks(
         instance.threads = num_threads;
 
         // Add arguments to instance name
+        size_t arg_i = 0;
         for (auto const& arg : args) {
+          if (arg_i < family->arg_names_.size()) {
+            instance.name +=
+                StringPrintF("/%s:", family->arg_names_[arg_i].c_str());
+          } else {
+            instance.name += "/";
+          }
           AppendHumanReadable(arg, &instance.name);
+          ++arg_i;
         }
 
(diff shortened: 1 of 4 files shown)


test/reporter_output_test.cc
```diff
@@ -52,6 +52,46 @@ ADD_CASES(TC_JSONOut, {{"\"name\": \"BM_error\",$"},
...
+
+// ========================================================================= //
+// ------------------------ Testing Arg Name Output ----------------------- //
+// ========================================================================= //
+
+void BM_arg_name(benchmark::State& state) {
+  while (state.KeepRunning()) {
+  }
+}
+BENCHMARK(BM_arg_name)->Arg(3)->ArgName("first");
```
dominichamon: what happens if i call `Args({1, 3})->ArgName("first")` or if I call `ArgName` directly on the benchmark without `Args` being set?

Would this be a cleaner api if `Args` took a `map<string, int>`

mkurdej: Well, to preserve current behaviour, the output will be exactly as in the current version if you do not specify `ArgName`. If you have more arguments than names, than only the first will be named.
One can find it a possible source of incoherence, but that's similar to the behaviour of `Args`, imagine calling it as `Args({1, 2, 3})->Args({1, 2})->Args({1,2,3,4,5})` - that's user's responsibility to define all the arguments its benchmark will need.

For your suggestion, I find it a bit "heavy" to set argument names each time, I would certainly not want to have to write this: `Args{{"first", 1}, {"second", 3}}->Args{{"first", 2}, {"second", 5}}` and so on. And it is disturbing that the same argument can be named differently each time (or isn't it?)

Anyway, your remarks made me think of a case that I didn't anticipate: the user might want to name only few arguments, e.g. `ArgNames({"first", "", "third"})`, my current PR will write `BM_benchmark/first:1/:2/third:3` (note the semicolon before 2). I'll correct it ASAP.

src/benchmark_register.cc
```diff
@@ -293,6 +305,16 @@ Benchmark* Benchmark::Ranges(const std::vector<std::pair<int, int>>& ranges) {
   return this;
 }
 
+Benchmark* Benchmark::ArgName(const std::string& name) {
+  arg_names_ = {name};
+  return this;
+}
+
+Benchmark* Benchmark::ArgNames(const std::vector<std::string>& names) {
+  arg_names_ = names;
```
dominichamon: should this be called after Args? If we can add a constraint on that, and then check that names.size() == args.size(), i think it would be much clearer to users when they do something unexpected. what do you think?

mkurdej: Oh, I see now that there are checks in `Arg/Args/Range/Ranges/DenseRange` that use `ArgsCnt`. Previously I thought that you could call `Args` with different number of elements each time.
I'll add these checks then, it seems to be the way to go.

mkurdej: I've added the constraint for the size. I've also modified ArgsCnt, so that it could be possible to call `Args` and `ArgNames` in any order. If you see a simpler/more elegant way to do this, I'm all ears.

dominichamon:
Thanks!

PR #328, opened 2016-12-20
Issue 327

BRevzin:
Arguments shouldn't be AppendHumanReadable()-ed, they should just be shown as-is. Otherwise, when I try to run a benchmark with an argument of 10000 it gets printed as 9.76562k, which is very confusing.


BRevzin:
I signed it!

dominichamon:
Thank you!

Issue #191, opened 2016-03-22
add a bazel BUILD file


dominichamon:
cmake is great, but it can't generate bazel (yet?). Create a bazel BUILD file alongside the cmake stuff.

EricWF:
Any particular reason for bazel? Alternatively we have a pretty simple build process and we could simply provide a buildit script or makefile.

dominichamon:
company allegiance, and ease of reintegration into google's source. ie, having a parallel build setup is overhead that i was trying to avoid by using cmake, but there are benefits to having a bazel build, but they're entirely selfish :)

the benefit of cmake over a script or makefile are two-fold: generation of native "projects" for MSVC users, XCode, etc, and dependency enforcement. as such, i don't think we can have just non-cmake.

i don't actually know if bazel has something similar or relies on all dependencies being integrated into the project so it's hermetic. ie, if we'd need to have submodules/subtrees for any libraries on which we depend. if it does, then maybe a simpler script is the way forward?

jupp0r:
As a starting point, this is what I'm using for [prometheus-cpp](https://github.com/jupp0r/prometheus-cpp) (hard coded to C++11s regex engine, because I require c++11 anyway):

```
cc_library(
    name = "googlebenchmark",
    srcs = glob(["src/*.cc"],
                exclude = ["src/re_posix.cc", "src/gnuregex.cc"]),
    hdrs = glob(["src/*.h", "include/benchmark/*.h"],
                exclude = ["src/re_posix.h", "src/gnuregex.h"]),
    includes = [
         "include",
    ],
    visibility = ["//visibility:public"],
    copts = [
          "-DHAVE_STD_REGEX"
    ],
)
```

ChrisCummins:
Thanks @jupp0r !

ffreling:
I think this issue has been fixed in version v1.4.0
https://github.com/google/benchmark/releases/tag/v1.4.0

PR #329, opened 2016-12-25
Create bazel builds

dominichamon:
As per #191 adding some basic bazel BUILD files. This will require some extra maintenance if files are removed/added but it might become the default some day. For now it's useful to have for interacting with other projects that already use bazel.


test/BUILD
```diff
@@ -0,0 +1,105 @@
+cc_library(
+    name = "output_test_helper",
+    srcs = [
+        "output_test.h",
+        "output_test_helper.cc",
+    ],
+    deps = ["//:benchmark"],
+)
+
+cc_binary(
```
nlkm: Is it possible to use cc_test instead?

Shouldn't users be able to do "bazel test :all" to validate that all tests are passing?

dominichamon: I wish they could, but we don't run googletest tests (which cc_test expects, as i understand it).

BUILD
```diff
@@ -0,0 +1,44 @@
...
+    hdrs = [
+        "include/benchmark/benchmark.h",
+        "include/benchmark/benchmark_api.h",
+        "include/benchmark/macros.h",
+        "include/benchmark/reporter.h",
+    ],
+    visibility = ["//test:__pkg__"],
+    defines = ["HAVE_POSIX_REGEX"],
+    includes = ["include"],
+    linkopts = ["-pthread"],
```
EricWF: Shouldn't we be adding -std=c++11 somewhere?

test/BUILD
```diff
@@ -0,0 +1,105 @@
...
+    srcs = ["complexity_test.cc"],
+    deps = [
+        ":output_test_helper",
+        "//:benchmark",
+    ],
+    linkopts = ["-lm", "-pthread"],
+)
+
+# cc_binary(
+#     name = "cxx03_test",
```
EricWF: This test needs an explicit `-std=c++03`.

EricWF:
@dominichamon What's the plan with this?

dominichamon:
on its one year anniversary i plan on reverting it.

actually i should probably just do it now. noone is screaming out for it, and this closed PR will be in the history for anyone that wants to do it.

shakeel:
Abseil.io mandates using Bazel for building code, it would be so nice if benchmark would have Bazel BUILD files, it would make the task of using Benchmark http_archive so much easier.

dominichamon:
i speak to the abseil folks quite regularly, and they're considering alternatives to bazel to make integration with existing projects easier. Ie, this issue isn't as high priority as i originally expected.

g-easy:
cctz inlines a BUILD file using new_http_archive: https://github.com/google/cctz/blob/master/WORKSPACE

keir:
Just wanted to add to this bug: Having a supported BUILD would be great, so that you can transitively include Google Benchmark easily in other projects. It seems strictly worse to have a multitude of projects copying and pasting their own forked version of a Google Benchmark BUILD file.

I am working to integrate Google Benchmark with Ceres Solver, and will end up copying and pasting a BUILD into the WORKSPACE file. This feels unsatisfying, especially since Bazel makes it natural and easy to automatically download and build dependencies.

Would a patch to add a BUILD file back be accepted?

keir:
Now landed: https://github.com/ceres-solver/ceres-solver/commit/22fa21c6f824ebefae12a9a2fa459714907acc57

dominichamon:
There are a few patches in progress adding BUILD files, but none of them
are complete. Feel free to work with the authors to help them finish the
job :)

Dominic Hamon | Google

Issue #320, opened 2016-12-05
Report/error unrecognized command line flags


NiklasRosenstein:
Are there any use cases where additional flags that are not recognized by `ParseCommandLineFlags()` are handled from another context? I would prefer if, by default, benchmark would report unrecognized command line flags.

Something like

```cpp
// benchmark.cc : ParseCommandLineFlags()
    if (...) {
      // ...
    } else if (IsFlag(argv[i], "help")) {
      PrintUsageAndExit();
    }
    else {
      fprintf(stdout, "unrecognized command line flag: %s\n", argv[i]);
      exit(1);
    }
```

pleroy:
The library being benchmarked might very well have its own flags and want to parse `argc`/`argv`.

NiklasRosenstein:
Very well. Would you agree with me when I'm saying that `BENCHMARK_MAIN()` should report unrecognized flags? Since it is expanded to a full main function, there's no way you're going to handle `argc`/`argv` any further.

pleroy:
I don't care that much about `BENCHMARK_MAIN`, but I don't think we should deviate too much from the Google flags library.  Otherwise it makes it very hard to use different Google libraries together (protocol buffers, gtest, gmock come to mind) just because the handling of flags was "improved" in slightly different manners in different libraries.

dominichamon:
should we start using gflags directly?

pleroy:
I think it would be nicer.  The sad thing though is that gflags is based on the Google flag library as it existed in 2008, so it lags a bit.

EricWF:
We could probably be safe diagnosing unrecognized flags starting with `--benchmark_*`, but like others have said some tests might have to parse their own flags. (For example people frequently use gtest and gbenchmark in the same executable).

NiklasRosenstein:
It does seem like using gflags for both libraries would make things a bit simpler, especially in the case of using the libraries in a single executable. As far as I can tell they both use their own derivative implementation/interface of gflags, `GTEST_DEFINE_...` in googletest, and `DEFINE_...` in benchmark.

@EricWF Unless these flags would then also be consumed by `ParseCommandLineFlags()`, it could lead to problems when you implement your own reporting of unrecognised flags. My [initial comment](https://github.com/google/benchmark/issues/320#issue-193394298) to this issue has the same problem.

[...]

PR #332, opened 2017-01-15
Report unrecognized arguments from BENCHMARK_MAIN() macro

NiklasRosenstein:
- Add `benchmark::ReportUnrecognizedArguments()`
- Update `BENCHMARK_MAIN()` macro

See #320 for reference

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -168,6 +168,9 @@ class BenchmarkReporter;
 
 void Initialize(int* argc, char** argv);
 
+// Report to stdout all arguments in 'argv' as unrecognized except the first.
+void ReportUnrecognizedArguments(int argc, char** argv);
+
 // Generate a list of benchmarks matching the specified --benchmark_filter flag
 // and if --benchmark_list_tests is specified return after printing the name
 // of each matching benchmark. Otherwise run each matching benchmark and
@@ -858,11 +861,8 @@ class Fixture : public internal::Benchmark {
 #define BENCHMARK_MAIN()                   \
   int main(int argc, char** argv) {        \
     ::benchmark::Initialize(&argc, argv);  \
-    for (int i = 1; i < argc; ++i) {       \
-      fprintf(stdout, "%s: error: unrecognised command-line flag: %s\n", \
-              argv[0], argv[i]);           \
-    }                                      \
     if (argc > 1) {                        \
+      ::benchmark::ReportUnrecognizedArguments(argc, argv); \
       return 1;                            \
     }                                      \
     ::benchmark::RunSpecifiedBenchmarks(); \
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -169,7 +169,8 @@ class BenchmarkReporter;
 void Initialize(int* argc, char** argv);
 
 // Report to stdout all arguments in 'argv' as unrecognized except the first.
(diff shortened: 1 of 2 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -858,6 +861,10 @@ class Fixture : public internal::Benchmark {
 #define BENCHMARK_MAIN()                   \
   int main(int argc, char** argv) {        \
     ::benchmark::Initialize(&argc, argv);  \
+    if (argc > 1) {                        \
```
EricWF: Why have this check at all? `ReportUnrecognizedArguments` handles that case already, doesn't it?

NiklasRosenstein: It's necessary if we want to treat unrecognized arguments as fatal errors and exit the application (see `return 1;` two lines below). If the benchmark should still be executed (which IMHO should not be the case, but other people might have a different opinion about that), we can remove the `if` and `return 1;`.

I also considered adding `exit(1);` to `ReportUnrecognizedArguments()` when `argc > 1`, but I figured it would not be obvious that the function can actually cause the application to terminate.

EricWF: You could have `ReportUnrecognizedArguments(...)` return `true` if there are unrecognized arguments and false otherwise.  Then change this to `if (ReportUnrecognizedArguments(...)) exit(1);`

EricWF:
LGTM. I'll merge this tomorrow night unless @dominichamon has any comments.

include/benchmark/benchmark_api.h
dominichamon: I don't really understand why this was made into a separate function in the first place as i think it was much clearer inline (albeit in a macro).

i have no massive concerns with the current approach, though i think the "returning a boolean" is cleaner... but then again what if someone doesn't want to printf but wants to do something else with the extra args?

chuck it in. we'll fix it in post.

src/benchmark.cc
```diff
@@ -664,4 +664,10 @@ void Initialize(int* argc, char** argv) {
   internal::LogLevel() = FLAGS_v;
 }
 
+void ReportUnrecognizedArguments(int argc, char** argv) {
+  for (int i = 1; i < argc; ++i) {
+    printf("%s: error: unrecognized command-line flag: %s\n", argv[0], argv[i]);
```
pleroy: This should go to `stderr` IMO because, guess what, it's an error.

include/benchmark/benchmark_api.h
NiklasRosenstein: > I don't really understand why this was made into a separate function in the first place as i think it > was much clearer inline (albeit in a macro).

I agree on that.

> but then again what if someone doesn't want to printf but wants to do something else with the > extra args?

That's when they don't use `BENCHMARK_MAIN()`.

src/benchmark.cc
NiklasRosenstein: ✔️

include/benchmark/benchmark_api.h
NiklasRosenstein: > You could have ReportUnrecognizedArguments(...) return true if there are unrecognized arguments and false otherwise. Then change this to if (ReportUnrecognizedArguments(...)) exit(1);

✔️

dominichamon:
LGTM

dominichamon:
(travis error looks like a travis failure... cmake started failing to build a test program).

PR #335, opened 2017-01-24
Disable iOS CPU Frequency readout

sieren:
`hw.cpufrequency` is not available on iOS devices. As there is no way to reliably
retrieve the CPU frequency on iOS we are printing out a warning
that we were unable to detect the CPU frequency and set it to 0.

This only disables cpu frequency readouts on actual physical iOS devices.
Running this code on the simulator still gives the cpu architecture of
the host computer as the simulator passes down the `sysctl` calls to OSX.

### DIFF
--- a/src/internal_macros.h
+++ b/src/internal_macros.h
@@ -30,9 +30,13 @@
 #elif defined(_WIN32)
 #define BENCHMARK_OS_WINDOWS 1
 #elif defined(__APPLE__)
-// TODO(ericwf) This doesn't actually check that it is a Mac OSX system. Just
-// that it is an apple system.
-#define BENCHMARK_OS_MACOSX 1
+#include "TargetConditionals.h"
+  #if defined(TARGET_OS_MAC)
+    #define BENCHMARK_OS_MACOSX 1
+    #if defined(TARGET_OS_IPHONE)
+      #define BENCHMARK_OS_IOS 1
+    #endif
+  #endif
 #elif defined(__FreeBSD__)
 #define BENCHMARK_OS_FREEBSD 1
 #elif defined(__linux__)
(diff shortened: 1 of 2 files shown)


src/internal_macros.h
```diff
@@ -30,9 +30,13 @@
 #elif defined(_WIN32)
 #define BENCHMARK_OS_WINDOWS 1
 #elif defined(__APPLE__)
-// TODO(ericwf) This doesn't actually check that it is a Mac OSX system. Just
-// that it is an apple system.
-#define BENCHMARK_OS_MACOSX 1
+#include "TargetConditionals.h"
+  #if defined(TARGET_OS_MAC)
+    #define BENCHMARK_OS_MACOSX 1
```
dominichamon: is it correct that an iphone should have both MACOSX and IOS defined?

sieren: Yeah, I think so. Apple see's `TARGET_OS_MAX` as a superset of all variants (as documented in `targetconditionals.h`:

```
        TARGET_OS_MAC             - Generated code will run under Mac OS X variant
           TARGET_OS_OSX          - Generated code will run under OS X devices
           TARGET_OS_IPHONE          - Generated code for firmware, devices, or simulator
              TARGET_OS_IOS             - Generated code will run under iOS 
              TARGET_OS_TV              - Generated code will run under Apple TV OS
              TARGET_OS_WATCH           - Generated code will run under Apple Watch OS
                 TARGET_OS_BRIDGE          - Generated code will run under Bridge devices
           TARGET_OS_SIMULATOR      - Generated code will run under a simulator
           TARGET_OS_EMBEDDED       - Generated code for firmware
```

sieren: Unless you think it makes more sense to check for OSX and iOS separately, but I suppose the benchmark codebase will have a bunch of more complex/longer `ifdef`s then.

dominichamon: i think this is fine for now, but if there are any other ios oddities we should consider using this hierarchy as a reference.

thanks for the link!

PR #337, opened 2017-01-30
Fix typename for powerpc

grypp:
fixed typo of typename. The suite could be compiled successfully on powerpc now.


grypp:
I signed it!

dominichamon:
thanks!

PR #340, opened 2017-02-11
Support for Web platforms

Maratyszcza:
Included are three commits to make Google Benchmark usable in Web environments:
- Implementation of `cycleclock::Now()` for Portable Native Client.
- Fix of `cycleclock::Now()` for Native Client on ARM architecture.
- Implementation of `cycleclock::Now()`, `ProcessCPUUsage`, and `ThreadCPUUsage` for Emscripten (Asm.js and WebAssembly) platforms.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -230,7 +230,7 @@ BENCHMARK_UNUSED static int stream_init_anchor = InitializeStreams();
 // expression from being optimized away by the compiler. This function is
 // intended to add little to no overhead.
 // See: https://youtu.be/nXaxk27zwlk?t=2441
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__pnacl__) && !defined(EMSCRIPTEN)
 template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
   asm volatile("" : : "g"(value) : "memory");
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -45,6 +45,10 @@ extern "C" uint64_t __rdtsc();
 #include <sys/time.h>
 #endif
 
+#if defined(__pnacl__)
+#include <time.h>
+#endif
+
 namespace benchmark {
 // NOTE: only i386 and x86_64 have been well tested.
 // PPC, sparc, alpha, and ia64 are based on
@@ -132,6 +136,14 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   struct timeval tv;
   gettimeofday(&tv, nullptr);
   return static_cast<int64_t>(tv.tv_sec) * 1000000 + tv.tv_usec;
+#elif defined(__pnacl__)
+  // Portable Native Client compiles to architecture-agnostic bytecode,
+  // and does not provide any API to access cycle counter.
+
+  // Initialize to always return if clock_gettime fails
+  struct timespec ts = { 0, 0 };
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  return static_cast<int64_t>(ts.tv_sec) * 1000000000 + ts.tv_nsec;
 #else
 // The soft failover to a generic implementation is automatic only for ARM.
 // For other platforms the developer is expected to make an attempt to create
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -45,6 +45,10 @@ extern "C" uint64_t __rdtsc();
 #include <sys/time.h>
 #endif
 
+#ifdef BENCHMARK_OS_EMSCRIPTEN
+#include <emscripten.h>
+#endif
+
 #if defined(__pnacl__)
 #include <time.h>
 #endif
@@ -69,6 +73,10 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   // counter pauses; it does not continue counting, nor does it
   // reset to zero.
   return mach_absolute_time();
+#elif defined(BENCHMARK_OS_EMSCRIPTEN)
+  // this goes above x86-specific code because old versions of Emscripten
+  // define __x86_64__, although they have nothing to do with it.
+  return static_cast<int64_t>(emscripten_get_now() * 1e+6);
 #elif defined(__i386__)
   int64_t ret;
   __asm__ volatile("rdtsc" : "=A"(ret));
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -112,7 +112,9 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   asm volatile("mrs %0, cntvct_el0" : "=r"(virtual_timer_value));
   return virtual_timer_value;
 #elif defined(__ARM_ARCH)
-#if (__ARM_ARCH >= 6)  // V6 is the earliest arch that has a standard cyclecount
+  // V6 is the earliest arch that has a standard cyclecount
+  // Native Client validator doesn't allow MRC instructions.
+#if (__ARM_ARCH >= 6) && !defined(BENCHMARK_OS_NACL)
   uint32_t pmccntr;
   uint32_t pmuseren;
   uint32_t pmcntenset;
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -43,14 +43,11 @@ extern "C" uint64_t __rdtsc();
 
 #ifndef BENCHMARK_OS_WINDOWS
 #include <sys/time.h>
+#include <time.h>
 #endif
 
 #ifdef BENCHMARK_OS_EMSCRIPTEN
 #include <emscripten.h>
-#endif
-
-#if defined(__pnacl__)
-#include <time.h>
 #endif
 
 namespace benchmark {
@@ -111,6 +108,22 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   _asm rdtsc
 #elif defined(COMPILER_MSVC)
   return __rdtsc();
+#elif defined(BENCHMARK_OS_NACL)
+  // Native Client validator on x86/x86-64 allows RDTSC instructions,
+  // and this case is handled above. Native Client validator on ARM
+  // rejects MRC instructions (used in the ARM-specific sequence below),
+  // so we handle it here. Portable Native Client compiles to
+  // architecture-agnostic bytecode, which doesn't provide any
+  // cycle counter access mnemonics.
+
+  // Native Client does not provide any API to access cycle counter.
+  // Use clock_gettime(CLOCK_MONOTONIC, ...) instead of gettimeofday
+  // because is provides nanosecond resolution (which is noticable at
+  // least for PNaCl modules running on x86 Mac & Linux).
+  // Initialize to always return 0 if clock_gettime fails.
+  struct timespec ts = { 0, 0 };
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  return static_cast<int64_t>(ts.tv_sec) * 1000000000 + ts.tv_nsec;
 #elif defined(__aarch64__)
   // System timer of ARMv8 runs at a different frequency than the CPU's.
   // The frequency is fixed, typically in the range 1-50MHz.  It can be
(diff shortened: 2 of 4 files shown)


src/timers.cc
```diff
@@ -127,7 +138,8 @@ double ProcessCPUUsage() {
 double ThreadCPUUsage() {
 // FIXME We want to use clock_gettime, but its not available in MacOS 10.11. See
 // https://github.com/google/benchmark/pull/292
-#if defined(CLOCK_THREAD_CPUTIME_ID) && !defined(BENCHMARK_OS_MACOSX)
+#if defined(CLOCK_THREAD_CPUTIME_ID) && !defined(BENCHMARK_OS_MACOSX) && \
```
EricWF: EMSCRIPTEN defines `CLOCK_THREAD_CPUTIME_ID` but doesn't provide `clock_gettime`? That seems weird... Apple only does that is to support targeting older platforms with `-mmacosx-version-min=<ver>` .

EricWF: This case feels like the `#else` case., but it definitely shouldn't be the primary `#if`. Please restructure these ifdefs so that the most specialized implementations are at the top, and the generic one is the `#else`.

src/cycleclock.h
```diff
@@ -45,6 +45,14 @@ extern "C" uint64_t __rdtsc();
 #include <sys/time.h>
 #endif
 
+#ifdef BENCHMARK_OS_EMSCRIPTEN
+#include <emscripten.h>
+#endif
+
+#if defined(__pnacl__)
```
EricWF: You don't need to guard includes when they're standard C++ headers.

src/timers.cc
```diff
@@ -102,7 +106,8 @@ BENCHMARK_NORETURN static void DiagnoseAndExit(const char* msg) {
 double ProcessCPUUsage() {
 // FIXME We want to use clock_gettime, but its not available in MacOS 10.11. See
 // https://github.com/google/benchmark/pull/292
-#if defined(CLOCK_PROCESS_CPUTIME_ID) && !defined(BENCHMARK_OS_MACOSX)
+#if defined(CLOCK_PROCESS_CPUTIME_ID) && !defined(BENCHMARK_OS_MACOSX) && \
```
EricWF: Please restructure these ifdefs so that the most specialized implementations are at the top, and the generic one is the #else.

src/cycleclock.h
```diff
@@ -108,7 +120,9 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   asm volatile("mrs %0, cntvct_el0" : "=r"(virtual_timer_value));
   return virtual_timer_value;
 #elif defined(__ARM_ARCH)
-#if (__ARM_ARCH >= 6)  // V6 is the earliest arch that has a standard cyclecount
+  // V6 is the earliest arch that has a standard cyclecount
+  // Native Client validator doesn't allow MRC instructions.
+#if (__ARM_ARCH >= 6) && !defined(BENCHMARK_OS_NACL)
```
EricWF: Please restructure these ifdefs so that the most specialized implementations are at the top, and the generic one is the #else. If you have to add restrictions to the primary `#if` then it's probably better suited as an `#elif` or `#else`.

Maratyszcza:
Fixed #ifdef structure. Also made a minor change to ensure `benchmark::DoNotOptimize` does not use inline assembly on PNaCl and Emscripten platforms (these platforms do not support inline assembly, and using it results in either link-time errors, or runtime errors).

include/benchmark/benchmark_api.h
```diff
@@ -230,7 +230,7 @@ BENCHMARK_UNUSED static int stream_init_anchor = InitializeStreams();
 // expression from being optimized away by the compiler. This function is
 // intended to add little to no overhead.
 // See: https://youtu.be/nXaxk27zwlk?t=2441
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__pnacl__) && !defined(EMSCRIPTEN)
```
EricWF: This change seems wrong. `#if defined(__GNUC__)` guards against *compilers* that don't support inline assembly. `NaCl` and `EMSCRIPTEN` are both platforms, not compilers.

What's the rational for this change?

EricWF:
LGTM except for the question left in the inline comment.

include/benchmark/benchmark_api.h
Maratyszcza: Both PNaCl and Emscripten use Clang, which defines `__GNUC__`. However, as they compile to virtual ISA (PNaCl bytecode, Asm.js, or WebAssembly bytecode), they don't support inline assembly.

EricWF: Does this actually fail to compile? If so that's a Clang bug. The inline assembly doesn't actually generate instructions, so I would hope it would still work regardless of Clang's target.

If not can you provide a full error log so I can used it to file a Clang bug?

EricWF: OK I probably overstated "not supporting inline assembly" as a bug. But if Clang parses "asm volatile(...)" it should probably also support generating code, at least for empty instructions.

I just tried compiling an file which used inline assembly with `clang++ -target x86_64-unknown-nacl` and it accepted it. I can't test linking and running it for obvious reasons, but I'm still skeptical of this change.

Maratyszcza: In case of PNaCl it fails to link, with an explicit message blaming inline assembly:
```
[21/29] LINK build/init-bench.bc
FAILED: /Users/marat/Projects/fxdiv-confu/build/init-bench.bc
/Users/marat/Tools/nacl_sdk/pepper_49/toolchain/mac_pnacl/bin/pnacl-clang++   -o /Users/marat/Projects/fxdiv-confu/build/init-bench.bc /Users/marat/Projects/fxdiv-confu/build/bench/init.cc.bc /Users/marat/Projects/fxdiv-confu/lib/libgooglebenchmark.a
Function _ZL13init_uint32_tRN9benchmark5StateE disallowed: bad result type: %struct.fxdiv_divisor_uint32_t*   %divisor.bc2 = bitcast i8* %divisor to %struct.fxdiv_divisor_uint32_t*
Function _ZL13init_uint32_tRN9benchmark5StateE disallowed: inline assembly:    call void asm sideeffect "", "*imr,~{memory}"(%struct.fxdiv_divisor_uint32_t* %divisor.bc2), !dbg !100433
```

In case of Emscripten/Asm.js, it links, but fails with a strange error at runtime:
```
Run on (1 X 1000.15 MHz CPU )
2017-02-10 21:33:11
Benchmark                           Time           CPU Iterations
-----------------------------------------------------------------
exception thrown: ReferenceError: $5 is not defined,ReferenceError: $5 is not defined
    at Array.Rb (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:434)
    at Array.Pc (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:36040)
    at mc (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:17877)
    at ic (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:10462)
    at hc (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:5620)
    at Object.Wb [as _main] (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:1543)
    at Object.callMain (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:5036)
    at doRun (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:5742)
    at run (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:5910)
    at Object.<anonymous> (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:7050)
/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:1
(function (exports, require, module, __filename, __dirname) { var Module;if(!Module)Module=(typeof Module!=="undefined"?Module:null)||{};var moduleOverrides={};for(var key in Module){if(Module.hasOwnProperty(key)){moduleOverrides[key]=Module[key]}}var ENVIRONMENT_IS_WEB=false;var ENVIRONMENT_IS_WORKER=false;var ENVIRONMENT_IS_NODE=false;var ENVIRONMENT_IS_SHELL=false;if(Module["ENVIRONMENT"]){if(Module["ENVIRONMENT"]==="WEB"){ENVIRONMENT_IS_WEB=true}else if(Module["ENVIRONMENT"]==="WORKER"){ENVIRONMENT_IS_WORKER=true}else if(Module["ENVIRONMENT"]==="NODE"){ENVIRONMENT_IS_NODE=true}else if(Module["ENVIRONMENT"]==="SHELL"){ENVIRONMENT_IS_SHELL=true}else{throw new Error("The provided Module['ENVIRONMENT'] value is not valid. It must be one of: WEB|WORKER|NODE|SHELL.")}}else{ENVIRONMENT_IS_WEB=typeof window==="object";ENVIRONMENT_IS_WORKER=typeof importScripts==="function";ENVIRONMENT_IS_NODE=typeof process==="object"&&typeof require==="function"&&!ENVIR

ReferenceError: $5 is not defined
    at Array.Rb (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:434)
    at Array.Pc (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:36040)
    at mc (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:17877)
    at ic (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:10462)
    at hc (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:5620)
    at Object.Wb [as _main] (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:8:1543)
    at Object.callMain (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:5036)
    at doRun (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:5742)
    at run (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:5910)
    at Object.<anonymous> (/Users/marat/Projects/fxdiv-confu/bin/multiply-bench.js:18:7050)
```

Both Native Client SDK and Emscripten are based on forks of Clang, not upstream Clang. The lack of inline assembly support is documented ([for PNaCl](https://developer.chrome.com/native-client/reference/pnacl-c-cpp-language-support#inline-assembly), [for Emscripten](https://kripken.github.io/emscripten-site/docs/getting_started/FAQ.html#why-does-my-code-fail-to-compile-with-an-error-message-about-inline-assembly-or-text-asm)) so I don't think they will fix it.

Maratyszcza: `x86_64-unknown-nacl` is (non-Portable) Native Client target, it compiles to regular x86-64 instructions (with some limitations due to sandboxing), and supports inline assembly. You need PNaCl toolchain from Native Client SDK or `emcc` toolchain from Emscripten SDK to try it.

EricWF: The `PNaCl` docs call out `asm("" ::: "memory")` as being explicitly supported, which is what `ClobberMemory` uses, so I guess their's hope that these will be supported eventually.

However for now it seems OK to disable them. However in the long run it's important that we find a way to provide a low/no overhead way of preventing optimizations; Calling `internal::UseChar` isn't sufficient. I'll reach out to friends working on NaCL to request a fix.

Maratyszcza: I checked that the current implementation of `benchmark::DoNotOptimize` results in *seemingly* reasonable numbers on PNaCl and Asm.js, but it would certainly help to have documented built-in support for this idiom in compilers.

EricWF:
> x86_64-unknown-nacl is (non-Portable) Native Client target, it compiles to regular x86-64 instructions (with some limitations due to sandboxing), and supports inline assembly. You need PNaCl toolchain from Native Client SDK or emcc toolchain from Emscripten SDK to try it.

I can successfully compile code using `clang -cc1 -triple le32-unknown-nacl`, but as you mentioned the error happens at link time. Honestly I'm really surprised that this doesn't result in an error when compiling or optimizing object files, because it must emit the inline assembly at that point.

Anyway... Thanks for the patch. And thank you for working on WebAssembly!

include/benchmark/benchmark_api.h
EricWF: I don't think Compiler vendors will even give us builtins with better behavior than the inline assembly.

The results of testing the fallback `DoNotOptimize` aren't going to be outrageous, but they're going to be orders of magnitude more expensive than the nop assembly instructions.

PR #358, opened 2017-03-27
Fix ICC compiler warnings

### LINKED ISSUE #354: Build fails with ICC17u2
The build fails with ICC17 because of warnings and Werror. What is the correct solution to fix it?
Should a patch
1) disable Werror for ICC (or maybe all non known compilers)
2) disable the false postive warnings for all files. This could be done using:
  add_cxx_compiler_flag(-wd2102) #ICC17u2: Many false positives for Wstrict-aliasing
  add_cxx_compiler_flag(-wd2259) #ICC17u2: non-pointer conversion from "long" to "int" may lose significant bits (even for explicit static cast, sleep.cc(44))
  add_cxx_compiler_flag(-wd654)  #ICC17u2: overloaded virtual function "benchmark::Fixture::SetUp" is only partially overridden (because of deprecated overload)
3) disable warnings at file level or some other granularity

Another warning which isn't clearly a false positive is 
../test/output_test_helper.cc(34): error #3280: declaration hides variable "<unnamed>::dec_re" (declared at line 67 of "../test/output_test.h")
Should the local dec_re be renamed or should that warning be suppressed too for 2/3? 

If I know which solution is preferred I'm happy to create a pull request.

rolandschulz:
fixes #354


src/sleep.cc
```diff
@@ -41,7 +41,7 @@ void SleepForMicroseconds(int microseconds) {
 }
 
 void SleepForMilliseconds(int milliseconds) {
-  SleepForMicroseconds(static_cast<int>(milliseconds) * kNumMicrosPerMilli);
+  SleepForMicroseconds(static_cast<int>(milliseconds * kNumMicrosPerMilli));
```
EricWF: What's the rational for this change, and the others like it?

test/output_test_helper.cc
```diff
@@ -31,21 +31,21 @@ TestCaseList& GetTestCaseList(TestCaseID ID) {
 
 SubMap& GetSubstitutions() {
   // Don't use 'dec_re' from header because it may not yet be initialized.
-  static std::string dec_re = "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?";
+  static std::string dec_re2 = "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?";
```
EricWF: I don't see a point in this change either, can you further explain?

CMakeLists.txt
```diff
@@ -92,8 +92,13 @@ else()
...
   endif()
   if (HAVE_CXX_FLAG_FSTRICT_ALIASING)
-    add_cxx_compiler_flag(-Wstrict-aliasing)
+    if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "Intel") #ICC17u2: Many false positives for Wstrict-aliasing
+      add_cxx_compiler_flag(-Wstrict-aliasing)
+    endif()
   endif()
+  # ICC17u2: overloaded virtual function "benchmark::Fixture::SetUp" is only partially overridden
+  # (because of deprecated overload)
+  add_cxx_compiler_flag(-wd654)
```
EricWF: Is there a better, more GCC-like, name for this warning?

src/sleep.cc
rolandschulz: Otherwise ICC produces: non-pointer conversion from "long" to "int" may lose significant bits. I could suppress the warning. But I assumed that a cast would be better. Also it seemed that the static_cast as it was made no sense. Given that milliseconds is already int.

test/output_test_helper.cc
rolandschulz: Otherwise ICC produces: declaration hides variable "<unnamed>::dec_re" (declared at line 67 of "../test/output_test.h"). I understood Dominic on #354 that it is better to fix the warning then suppress it.

CMakeLists.txt
rolandschulz: No.

src/sleep.cc
EricWF: Ah sorry I should have figured that out from #354. Go ahead and remove this change. I'm going to commit a different fix upstream. 

I think the actual bug here is that `kNumMicrosPerMilli` is specified as `int64_t`

test/benchmark_test.cc
```diff
@@ -150,7 +150,7 @@ static void BM_LongTest(benchmark::State& state) {
 BENCHMARK(BM_LongTest)->Range(1 << 16, 1 << 28);
 
 static void BM_ParallelMemset(benchmark::State& state) {
-  int size = state.range(0) / sizeof(int);
+  int size = static_cast<int>(state.range(0) / sizeof(int));
```
EricWF: I think it would be more correct to do `st.range(0) / static_cast<int>(sizeof(int))`

test/output_test_helper.cc
EricWF: @dominichamon is probably right. It's easier to fix than it is to suppress. However maybe a name like `safe_dec_re` is more appropriate given the above comment?

However I must add that I think ICC is *INSANE* emitting a -Wshadow diagnostic between a global symbol and a function-local variable. God forbid you ever write `#include <math.h> void foo() { int abs = bar(); }`.

src/sleep.cc
EricWF: See PR #359 for the aforementioned fix.

EricWF:
Thanks for the report and pull request!

PR #359, opened 2017-03-28
Replace int64_t usages with 'int' instead.

EricWF:
Previously the constants used for converting between
different units of time were declared using int64_t. However
we should only use explicitly sized integer types when they
are required, and should use 'int' everwhere else, and there is
no good reason to use int64_t here.

For that reason this patch changes the type of the constants.
This should help address issue #354 as well.


EricWF:
Merging since all of the bots have since passed.

PR #360, opened 2017-03-28
Add BENCHMARK_BUILD_32_BITS option and add builders to test it

EricWF:


EricWF:
The appveyor failure seems unrelated, and instead caused by flaxey tests.

PR #363, opened 2017-04-03
Add CMake Package Config files during install

piribes:
- Remove target_include_directories of ${PROJECT_SOURCE_DIR}/include to
  fix error: Target "benchmark" INTERFACE_INCLUDE_DIRECTORIES property
  contains path which is prefixed in the source directory.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,18 +27,45 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   target_link_libraries(benchmark Shlwapi)
 endif()
 
-# Expose public API
-target_include_directories(benchmark PUBLIC ${PROJECT_SOURCE_DIR}/include)
+set(include_install_dir "include/doctest/")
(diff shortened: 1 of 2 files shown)


piribes:
I signed it!

src/CMakeLists.txt
```diff
@@ -27,18 +27,45 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   target_link_libraries(benchmark Shlwapi)
 endif()
 
-# Expose public API
-target_include_directories(benchmark PUBLIC ${PROJECT_SOURCE_DIR}/include)
+set(include_install_dir "include/doctest/")
```
dominichamon: include/benchmark ?

piribes: You're right.

dominichamon:
This installs the headers into /usr/local/include/benchmark/benchmark so i removed the one extra benchmark subfolder before merging.

thank you so much!

PR #364, opened 2017-04-07
Fix path problem with latest changes

kunitoki:
Config.cmake.in was searched in the main source dir instead of the project dir (the actual benchmark project).


kunitoki:
I signed it!

PR #365, opened 2017-04-07
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

piribes:
As pointed out [here](https://github.com/google/benchmark/commit/824bbb818e2d34e546c77d719687ab732264603c#commitcomment-21670288) we should use PROJCT_SOURCE_DIR instead of CMAKE_SOURCE_DIR.


kunitoki:
already fixed here https://github.com/google/benchmark/pull/364

PR #366, opened 2017-04-07
Changed configure_file path

horttanainen:
When adding benchmark as a subproject CMAKE_SOURCE_DIR causes cmake to look for /cmake/Config.cmake.in parent project's root instead of the intended benchmark project's root.


horttanainen:
I signed it!

horttanainen:
Closing because there is 2 similar pull requests.

PR #367, opened 2017-04-08
Fix cmake file not to use CMAKE_SOURCE_DIR

vargad:
Using ```CMAKE_SOURCE_DIR``` benchmark cannot be built as part of a project.
This change allows to add benchmark using add_subdirectory to a bigger
project.


kunitoki:
already reported in #365 and #366

PR #372, opened 2017-04-17
fix android compilation

Slonegg:
I'm working on GRPC for Android. Here are changes to build benchmark with CMake for Android. Instructions (MSVS + Ninja), related to [this](https://github.com/google/protobuf/pull/2980):

1. Install Android SDK
2. Install Android NDK. Generally only NDK is necessary, SDK just may be useful for other purposes.
3. Set environment variable ANDROID_NDK
4. Install Ninja. Just drop ninja.exe somewhere where it can be accessed from command line, for example, alongside cmake binaries.
5. Build benchmark:
```
    mkdir _android_build
    cd _android_build
    cmake .. -G Ninja -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=21 -DCMAKE_ANDROID_ARCH_ABI=arm64-v8a -DCMAKE_ANDROID_STL_TYPE=gnustl_static -DRUN_HAVE_STD_REGEX=0 -DRUN_HAVE_POSIX_REGEX=0 -DRUN_HAVE_STEADY_CLOCK=0
    cmake --build .
```

### DIFF
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -164,7 +164,7 @@ bool BenchmarkFamilies::FindBenchmarks(
             }
           }
           
-#if defined(__ANDROID__) && (defined(__GLIBCXX__) || defined(__GLIBCPP__))
+#if defined(__ANDROID__) && defined(__GLIBCXX__)
           // workaround for Android, http://stackoverflow.com/questions/22774009/android-ndk-stdto-string-support
           std::ostringstream ss;
           ss << arg;
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -31,6 +31,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <sstream>
 #include <thread>
 
 #include "check.h"
@@ -162,8 +163,15 @@ bool BenchmarkFamilies::FindBenchmarks(
                   StringPrintF("%s:", family->arg_names_[arg_i].c_str());
             }
           }
-
+          
+#ifdef __ANDROID__
+          // workaround for Android, http://stackoverflow.com/questions/22774009/android-ndk-stdto-string-support
+          std::ostringstream ss;
+          ss << arg;
+          instance.name += ss.str();
+#else
           instance.name += std::to_string(arg);
+#endif
           ++arg_i;
         }
 
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -164,7 +164,7 @@ bool BenchmarkFamilies::FindBenchmarks(
             }
           }
           
-#ifdef __ANDROID__
+#if defined(__ANDROID__) && (defined(__GLIBCXX__) || defined(__GLIBCPP__))
           // workaround for Android, http://stackoverflow.com/questions/22774009/android-ndk-stdto-string-support
           std::ostringstream ss;
           ss << arg;
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -164,14 +164,7 @@ bool BenchmarkFamilies::FindBenchmarks(
             }
           }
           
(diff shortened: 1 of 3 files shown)


src/colorprint.cc
```diff
@@ -89,7 +89,11 @@ std::string FormatString(const char* msg, va_list args) {
 
   std::size_t size = 256;
   char local_buff[256];
+#ifdef __ANDROID__
+  auto ret = ::vsnprintf(local_buff, size, msg, args_cp);
```
dominichamon: is this part of a broader set of checks that should be guarded by  #if __cplusplus >= 201103L or is it truly Android specific?

Slonegg: Tested this approach, doesn't work. This is quirk of Android GNU STL library. This problem doesn't appear if compiling with  `-DCMAKE_ANDROID_STL_TYPE=c++_static` which uses clang stl library. But gnustl is default for Android now. I can replace it with `#if defined(__ANDROID__) && defined(__GLIBCXX__)` - it works, if you think it is better.

EricWF: Yeah, I would rather see the `__GLIBCXX__` check in there as well. As you mentioned this works fine with libc++ :-D

src/colorprint.cc
```diff
@@ -89,7 +89,11 @@ std::string FormatString(const char* msg, va_list args) {
 
   std::size_t size = 256;
   char local_buff[256];
+#if defined(__ANDROID__) && (defined(__GLIBCXX__) || defined(__GLIBCPP__))
```
EricWF: Let's just use `vsnprintf` unconditionally. I see no reason to prefer the version in `std::` .

src/colorprint.cc
```diff
@@ -104,7 +108,11 @@ std::string FormatString(const char* msg, va_list args) {
     // we did not provide a long enough buffer on our first attempt.
     size = (size_t)ret + 1;  // + 1 for the null byte
     std::unique_ptr<char[]> buff(new char[size]);
+#if defined(__ANDROID__) && (defined(__GLIBCXX__) || defined(__GLIBCPP__))
```
EricWF: Same comment as above.

src/benchmark_register.cc
```diff
@@ -162,8 +163,15 @@ bool BenchmarkFamilies::FindBenchmarks(
                   StringPrintF("%s:", family->arg_names_[arg_i].c_str());
             }
           }
-
+          
+#if defined(__ANDROID__) && (defined(__GLIBCXX__) || defined(__GLIBCPP__))
```
EricWF: `__GLIBCPP__` was removed in favor of `__GLIBCXX__` in GCC 3.4. I can't imagine anybody is still using that old of a GCC version. Are you sure it's needed?

dominichamon:
LGTM. @EricWF i'll let you have the final say/merge.

src/benchmark_register.cc
```diff
@@ -162,8 +163,15 @@ bool BenchmarkFamilies::FindBenchmarks(
...
           }
-
+          
+#if defined(__ANDROID__) && defined(__GLIBCXX__)
+          // workaround for Android, http://stackoverflow.com/questions/22774009/android-ndk-stdto-string-support
+          std::ostringstream ss;
+          ss << arg;
+          instance.name += ss.str();
+#else
           instance.name += std::to_string(arg);
```
EricWF: How about `instance.name += FormatString("%d", arg)` instead?

I just want to avoid as much conditional compilation as possible. The more of it the library accumulates the harder it gets to maintain.

EricWF:
This LGTM minus my last inline comment. Thanks for the quick turnaround time.

src/benchmark_register.cc
Slonegg: Fair, didn't find FormatString, but there is StringPrintF.

Issue #379, opened 2017-04-28
compare_bench fails if Complexity was computed


felixduvallet:
When the benchmarks are are run with `Complexity()` enabled, the output JSON file contains additional members that cause the `compare_bench.py` script to fail:

> Invalid input file: 'out.json' does not name a valid benchmark executable or JSON file

The offending lines in the JSON look like:

    {
      "name": "ComputationBenchmarks_BigO",
      "cpu_coefficient": 672,
      "real_coefficient": 672,
      "big_o": "N",
      "time_unit": "ns"
    },
    {
      "name": "ComputationBenchmarks_RMS",
      "rms": 2%
    }

Removing them manually works fine, so presumably the script should just ignore complexity-related entries in the JSON.

PR #380, opened 2017-04-28
Ensure all the necessary keys are present before parsing JSON data

felixduvallet:
This prevents errors when additional non-timing data are present in
the JSON that is loaded, for example when complexity data has been
computed (see #379).


dominichamon:
Please sign the CLA so we can take a look at this change. Thanks!

felixduvallet:
Sorry about the delay @dominichamon and @googlebot , I filled out the CLA.

PR #362, opened 2017-04-01
Fixes #357: broken RMS values when time unit is set.

biojppm:

Issue #357, opened 2017-03-27
RMS calculation seems broken, takes time Unit into account?


LebedevRI:
Hi.
Consider the [`BM_Complexity_O_N`](https://github.com/google/benchmark/blob/36a251ab3ee4e6dc668129e0ef5fa62b1b8d5d49/test/complexity_test.cc#L80-L113)
Vanilla version outputs:
```
Benchmark                       Time           CPU Iterations
-------------------------------------------------------------
BM_Complexity_O_N/1024        441 ns        441 ns    1583329
BM_Complexity_O_N/2k          881 ns        881 ns     797104
BM_Complexity_O_N/4k         1767 ns       1767 ns     395988
BM_Complexity_O_N/8k         3691 ns       3690 ns     189567
BM_Complexity_O_N/16k        7377 ns       7376 ns      95031
BM_Complexity_O_N/32k       14813 ns      14801 ns      47507
BM_Complexity_O_N/64k       29605 ns      29602 ns      23608
BM_Complexity_O_N_BigO       0.45 N       0.45 N
BM_Complexity_O_N_RMS           0 %          0 %
BM_Complexity_O_N/1024        446 ns        446 ns    1587672
BM_Complexity_O_N/2k          884 ns        884 ns     792614
BM_Complexity_O_N/4k         1787 ns       1786 ns     393225
BM_Complexity_O_N/8k         3684 ns       3683 ns     189791
BM_Complexity_O_N/16k        7373 ns       7369 ns      95181
BM_Complexity_O_N/32k       14716 ns      14711 ns      47732
BM_Complexity_O_N/64k       29714 ns      29707 ns      23656
BM_Complexity_O_N_BigO       0.45 f(N)       0.45 f(N)
BM_Complexity_O_N_RMS           1 %          1 %
BM_Complexity_O_N/1024        442 ns        442 ns    1589204
BM_Complexity_O_N/2k          885 ns        884 ns     788984
BM_Complexity_O_N/4k         1781 ns       1781 ns     390535
BM_Complexity_O_N/8k         3691 ns       3688 ns     189901
BM_Complexity_O_N/16k        7360 ns       7360 ns      93646
BM_Complexity_O_N/32k       14770 ns      14769 ns      47687
BM_Complexity_O_N/64k       29631 ns      29629 ns      23604
BM_Complexity_O_N_BigO       0.45 N       0.45 N
BM_Complexity_O_N_RMS           0 %          0 %
```

[...]

PR #350, opened 2017-03-02
Add command line option to print user counters in tabular format

biojppm:
This PR adds the option `--benchmark_counters_tabular={true|false}`, which makes the ConsoleReporter print the user counters in tabular format. The option defaults to false. This follows discussions in PR #262 (notably [this](https://github.com/google/benchmark/pull/262#issuecomment-244257349) and [this](https://github.com/google/benchmark/pull/262#issuecomment-248969221)).

To be clear:

- by default, counters are printed at the end, the same way as bytes_processed and items_processed, as agreed in #262 .  This is best for cases in which there are few counters, or where there are only a couple of lines per benchmark.
- when `--benchmark_counters_tabular=true` is given, then the counters are printed as a column of the table. This is best for cases in which there are a lot of counters, or a lot of lines per benchmark.

Here's an example comparison between the two versions (Note that I removed some whitespace from both examples).

First the default `--benchmark_counters_tabular=false`:

```
------------------------------------------------------------------------------
Benchmark                        Time           CPU Iterations UserCounters...
------------------------------------------------------------------------------
BM_UserCounter/threads:8      2248 ns      10277 ns      68808 Bar=16 Bat=40 Baz=24 Foo=8
BM_UserCounter/threads:1      9797 ns       9788 ns      71523 Bar=2 Bat=5 Baz=3 Foo=1024m
BM_UserCounter/threads:2      4924 ns       9842 ns      71036 Bar=4 Bat=10 Baz=6 Foo=2
BM_UserCounter/threads:4      2589 ns      10284 ns      68012 Bar=8 Bat=20 Baz=12 Foo=4
BM_UserCounter/threads:8      2212 ns      10287 ns      68040 Bar=16 Bat=40 Baz=24 Foo=8
BM_UserCounter/threads:16     1782 ns      10278 ns      68144 Bar=32 Bat=80 Baz=48 Foo=16
BM_UserCounter/threads:32     1291 ns      10296 ns      68256 Bar=64 Bat=160 Baz=96 Foo=32
BM_UserCounter/threads:4      2615 ns      10307 ns      68040 Bar=8 Bat=20 Baz=12 Foo=4
BM_Factorial                    26 ns         26 ns   26608979 40320
BM_Factorial/real_time          26 ns         26 ns   26587936 40320
BM_CalculatePiRange/1           16 ns         16 ns   45704255 0
BM_CalculatePiRange/8           73 ns         73 ns    9520927 3.28374
[...]

PR #381, opened 2017-04-29
Add unit tests for user counters.

### LINKED ISSUE #348: Add tests to user counters
Now that user counters were integrated, tests should be added. During review of #262, #276 was suggested which added some tests. @EricWF Could you add your tests so that I pick up from there?

### LINKED ISSUE #378: User Counters ignore Flag parameter
When creating a user counter it display the count as if `kDefaults` flag was used even if explictly setting a different flag
`st.counters["Foo"] = benchmark::Counter(FooCount, benchmark::Counter::kIsRate);`

biojppm:
Finally added unit tests for user counters. The tests cover both expected output formatting and checking of numerical values of benchmark result fields. This required adding facilities for getting at the benchmark results; this is done by parsing the CSV output. It also adds fuzzy floating point comparison CHECK macros, required for checking the counter values when rates are used because of lossy formatting of counters.

Also:
-Fixes #348
-Fixes #378. This PR contains a commit by @vladoovtcharov which completes the fixing of  #378. (The tests successfully detect this bug).
-The result checking is also needed to test RMS values #357, as discussed in #362.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -252,6 +252,7 @@ BenchmarkReporter::Run CreateRunReport(
     report.complexity = b.complexity;
     report.complexity_lambda = b.complexity_lambda;
     report.counters = results.counters;
+    internal::Finish(&report.counters, seconds, b.threads);
   }
   return report;
 }
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -285,7 +285,8 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
     std::vector<BenchmarkReporter::Run>* complexity_reports) {
   std::vector<BenchmarkReporter::Run> reports;  // return value
 
-  size_t iters = 1;
+  const bool has_explicit_iteration_count = b.iterations != 0;
+  size_t iters = has_explicit_iteration_count ? b.iterations : 1;
   std::unique_ptr<internal::ThreadManager> manager;
   std::vector<std::thread> pool(b.threads - 1);
   const int repeats =
@@ -295,7 +296,7 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
       (b.report_mode == internal::RM_Unspecified
            ? FLAGS_benchmark_report_aggregates_only
            : b.report_mode == internal::RM_ReportAggregatesOnly);
-  for (int i = 0; i < repeats; i++) {
+  for (int repetition_num = 0; repetition_num < repeats; repetition_num++) {
     for (;;) {
       // Try benchmark
       VLOG(2) << "Running " << b.name << " for " << iters << "\n";
@@ -331,10 +332,20 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
 
       const double min_time =
           !IsZero(b.min_time) ? b.min_time : FLAGS_benchmark_min_time;
-      // If this was the first run, was elapsed time or cpu time large enough?
-      // If this is not the first run, go with the current value of iter.
-      if ((i > 0) || results.has_error_ || (iters >= kMaxIterations) ||
-          (seconds >= min_time) || (results.real_time_used >= 5 * min_time)) {
+
+      // Determine if this run should be reported; Either it has
+      // run for a sufficient amount of time or because an error was reported.
+      const bool should_report =  repetition_num > 0
+        || has_explicit_iteration_count // An exact iteration count was requested
+        || results.has_error_
+        || iters >= kMaxIterations
+        || seconds >= min_time // the elapsed time is large enough
+        // CPU time is specified but the elapsed real time greatly exceeds the
+        // minimum time. Note that user provided timers are except from this
+        // sanity check.
+        || ((results.real_time_used >= 5 * min_time) && !b.use_manual_time);
+
+      if (should_report) {
         BenchmarkReporter::Run report =
             CreateRunReport(b, results, iters, seconds);
         if (!report.error_occurred && b.complexity != oNone)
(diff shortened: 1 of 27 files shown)


biojppm:
@vladoovtcharov I think you need to add a CLA for your commit to be accepted: https://cla.developers.google.com/

vladoovtcharov:
ok, I added the CLA, hopefully it will pick it up...

dominichamon:
@biojppm @vladoovtcharov I think you'll need to post something like "I signed!" for the cla bot to recheck.

src/benchmark.cc
```diff
@@ -252,6 +252,7 @@ BenchmarkReporter::Run CreateRunReport(
     report.complexity = b.complexity;
     report.complexity_lambda = b.complexity_lambda;
     report.counters = results.counters;
+    internal::Finish(&report.counters, seconds, b.threads);
```
dominichamon: for future TODO perhaps:

report.counters = results.counters.Finish(seconds, b.threads);

src/check.h
```diff
@@ -68,4 +69,11 @@ class CheckHandler {
 #define CHECK_GT(a, b) CHECK((a) > (b))
 #define CHECK_LT(a, b) CHECK((a) < (b))
 
+#define CHECK_EQ_EPS(a, b, eps) CHECK(std::fabs((a) - (b)) <  (eps))
```
dominichamon: CHECK_FLOAT_EQ or CHECK_DOUBLE_EQ are more aligned with googletest.

alternatively, we might want to start using googletest ;) another TODO!

src/console_reporter.cc
```diff
@@ -53,11 +53,10 @@ bool ConsoleReporter::ReportContext(const Context& context) {
 
 void ConsoleReporter::PrintHeader(const Run& run) {
   std::string str =
-      FormatString("%-*s %13s %13s %10s\n", static_cast<int>(name_field_width_),
-                   "Benchmark", "Time", "CPU", "Iterations");
-  if(!run.counters.empty()) {
-    str += " UserCounters...";
-  }
+      FormatString("%-*s %13s %13s %10s%s\n", static_cast<int>(name_field_width_),
```
dominichamon: for consistency, why not put the " " before the last %s and not have the space in the UserCounters string below?

test/output_test.h
```diff
@@ -59,6 +61,125 @@ int SetSubstitutions(
...
+  // as a double, and only then converted to the asked type.
+  template <class T>
+  T GetCounterAs(const char* entry_name) const {
+    double dval = GetAs< double >(entry_name);
+    T tval = static_cast< T >(dval);
+    return tval;
+  }
+
+  // get cpu_time or real_time in seconds
+  double GetTime(const char* which) const;
```
dominichamon: an enum might be a better fit for the `which` argument type.

test/output_test.h
```diff
@@ -59,6 +61,125 @@ int SetSubstitutions(
...
+    << "expected (" << #var_type << ")" << (var_name)                   \
+    << "=" << (entry).getfn< var_type >(var_name)                       \
+    << " to be " #relationship " to " << (value) << "\n"                \
+    << __FILE__ << ":" << __LINE__ << ": "                              \
+    << "with tolerance of " << (eps_factor) * (value)                   \
+    << " (" << (eps_factor)*100. << "%), "                              \
+    << "but delta was " << ((entry).getfn< var_type >(var_name) - (value)) \
+    << " (" << (((entry).getfn< var_type >(var_name) - (value))         \
+               /                                                        \
+               ((value) > 1.e-5 || value < -1.e-5 ? value : 1.e-5)*100.) \
```
dominichamon: should the `1.e-5` here be `eps_factor`?

test/output_test_helper.cc
```diff
@@ -140,9 +149,180 @@ class TestReporter : public benchmark::BenchmarkReporter {
...
+}
+
+// add a results checker for a benchmark
+void ResultsChecker::Add(const std::string& entry_pattern, ResultsCheckFn fn) {
+  check_patterns.emplace_back(entry_pattern, fn);
+}
+
+// check the results of all subscribed benchmarks
+void ResultsChecker::CheckResults(std::stringstream& output)
+{
```
dominichamon: nit: bring this brace up a line please.

(use clang-format with google style to catch these for you)

vladoovtcharov:
I signed the CLA. [clabot:check]

src/check.h
biojppm: Yes, I too found it strange that you were not using googletest!

test/output_test.h
biojppm: Good catch! The magic number here is mainly a hack to prevent overflow when printing the percentage when the check fails. In general, I expect eps_factor to be generally two orders of magnitude larger than this, so that's why I just fixed it here. OTOH, the multiplication by 100 does bring this into the same region as eps_factor. But I still see them as unrelated.

biojppm:
I signed!

src/console_reporter.cc
biojppm: The reporter output tests are expecting the regex `Iterations$`, so adding a space in the format string would make those tests fail, or the tests would have to be adjusted. For that reason, I chose using the space in " UserCounters" as a good balance,

biojppm:
@dominichamon The Appveyor fail is again with MinGW in the reporter output test. This fail is intermittent, and is the same as [the one discussed in #350](https://github.com/google/benchmark/pull/350#issuecomment-290948588). Apparently, BM_Repeat in MinGW sometimes returns a negative time. I've submitted the issue #382 for this.

src/console_reporter.cc
dominichamon: SGTM

dominichamon:
Thank you so much!

PR #383, opened 2017-05-03
Use the sample version of standard deviation

### LINKED ISSUE #371: Standard deviation calculation
Hi all,
The current calculation for standard deviation uses the **population** version, i.e., `sqrt(sum((x_i - x_mean)^2)/n)`. But by using repetitions, it's more sensible to use the **sample** version `sqrt(sum((x_i - x_mean)^2)/(n - 1))`. This is especially important when computing the confidence interval for the mean values.

yixuan:
Based on the discussions in #371 , this PR fixes the `Stat1` class to compute the sample version of standard deviation.
Closes #371 .

### DIFF
--- a/src/stat.h
+++ b/src/stat.h
@@ -119,18 +119,23 @@ class Stat1 {
     if (numsamples_ == 0) return VType();
     VType mean = sum_ * (1.0 / numsamples_);
     if (stddev) {
-      VType avg_squares = sum_squares_ * (1.0 / numsamples_);
-      *stddev = Sqrt(avg_squares - Sqr(mean));
+      // Sample standard deviation is undefined for n = 1
+      if (numsamples_ == 1) {
+        *stddev = VType();
+      } else {
+        VType avg_squares = sum_squares_ * (1.0 / numsamples_);
+        *stddev = Sqrt(numsamples_ / (numsamples_ - 1.0) * (avg_squares - Sqr(mean)));
+      }
     }
     return mean;
   }
 
   // Return the standard deviation of the sample set
   VType StdDev() const {
-    if (numsamples_ == 0) return VType();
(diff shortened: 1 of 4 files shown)


src/stat.h
```diff
@@ -119,18 +119,23 @@ class Stat1 {
...
   }
 
   // Return the standard deviation of the sample set
   VType StdDev() const {
-    if (numsamples_ == 0) return VType();
+    if (numsamples_ <= 1) return VType();
     VType mean = Mean();
     VType avg_squares = sum_squares_ * (1.0 / numsamples_);
-    return Sqrt(avg_squares - Sqr(mean));
+    return Sqrt(numsamples_ / (numsamples_ - 1.0) * (avg_squares - Sqr(mean)));
```
dominichamon: can you replace this method with:

VType stddev;
VType mean = Mean(&stddev);
return stddev;

and remove the extra checks and calculations?

yixuan: Sounds good.

yixuan: And probably we need to use `Mean(&stddev);` for the second line, since `mean` is an unused variable and the Makefile converts warnings into errors.

EricWF:
@dominichamon Can you make the final call on this?

dominichamon:
it wasn't clear to me if the appveyor failures are real.. i believe they are though so i'm waiting for @yixuan to address them.

yixuan:
Those failures seem to be due to abnormal running time (e.g. 195313 ns) that goes beyond the regular expression limit (five digits), usually occurring in multi-threaded programs. Is it possible to re-launch those tests and then see the results?

dominichamon:
nice. thanks!

yixuan:
Looks great. Thank you @dominichamon !

PR #385, opened 2017-05-09
Change `config_install_dir` to share/benchmark

KindDragon:
Hi.

Can we change `config_install_dir` to `share/benchmark`?
CMake works well with both paths, but new path allows me to use google benchmark without patching your CMakeLists.txt on [vcpkg](https://github.com/Microsoft/vcpkg)


LebedevRI:
Better yet, why not use `GNUInstallDirs` cmake module?

PR #393, opened 2017-05-23
Associate the required include directory with the benchmark library

dkruger:
Using target_include_directories CMake will implicitly add the the
necessary include paths to targets which link against the benchmark
library. This is useful when the benchmark repo is included as a
subdirectory in another CMake build.


dkruger:
I signed it!

dominichamon:
thank you!

PR #396, opened 2017-05-30
Adds support for adding name to each benchmark run.

GeorgeARM:
The reason for this pull request is that I couldn't associate a distinct name to each of my benchmark runs. I know that we can distinct each run though the passed arguments but when testing complicated interfaces (template fixtures and datasets) this wasn't enough and sometimes I wanted to add more meaningful names to the runs.
The only way to achieve this was to register different benchmarks over and over again.


GeorgeARM:
I signed it!

include/benchmark/benchmark_api.h
```diff
@@ -512,9 +516,9 @@ class Benchmark {
   // method calls can be chained together in one expression.
 
   // Run this benchmark once with "x" as the extra argument passed
-  // to the function.
+  // to the function. Add a name to the benchmark run if provided.
   // REQUIRES: The function passed to the constructor must accept an arg1.
-  Benchmark* Arg(int x);
+  Benchmark* Arg(int x, const std::string& name = std::string());
```
dominichamon: i would prefer separate methods to make it clear what the string argument is:

```
Benchmark* Arg(int x);
Benchmark* NamedArg(const std::string& name, int x);
```

GeorgeARM: Makes sense @dominichamon, will add the appropriate methods.
Thanks.

EricWF:
@GeorgeARM Is there a reason you can't use `State::Label` to give the benchmark a name when it's run?

GeorgeARM:
@EricWF can I filter tests based on their Label?

EricWF:
> @EricWF can I filter tests based on their Label?

@GeorgeARM No you cannot, because the Label is provided at the end of running the benchmark.

Another way to achieve your goal would be to use `RegisterBenchmark`, which allows you to register benchmarks with arbitrary names.

EricWF:
I believe this PR/Issue would be fixed if we adopted the JSON interface I'm proposing. See [this test case](https://github.com/efcs/benchmark/blob/overhaul-reporting/test/new_api_test.cc#L22) for an example.

PR #398, opened 2017-06-02
Fix #342: DoNotOptimize causes compile errors on older GCC versions.

### LINKED ISSUE #342: benchmark::DoNotOptimize does not compile on gcc4.8 + ppc64
I'm trying to build one of my projects on a ppc64le system, and compilation of benchmarks fails on assembly inside `benchmark::DoNotOptimize`:

```
/home/mdukhan3/NNPACK/deps/fxdiv/deps/googlebenchmark/include/benchmark/benchmark_api.h:236:45: error: cannot reload integer constant operand in ‘asm’
   asm volatile("" : : "g"(value) : "memory");
```

Results for `g++ -v`:
```
mdukhan3@power8:~/NNPACK/deps/fxdiv$ g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/powerpc64le-linux-gnu/4.8/lto-wrapper
Target: powerpc64le-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 4.8.4-2ubuntu1~14.04.3' --with-bugurl=file:///usr/share/doc/gcc-4.8/README.Bugs --enable-languages=c,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-4.8 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --with-gxx-include-dir=/usr/include/c++/4.8 --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --enable-gnu-unique-object --disable-libmudflap --disable-libsanitizer --disable-libsanitizer --disable-libquadmath --enable-plugin --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-4.8-ppc64el/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-4.8-ppc64el --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-4.8-ppc64el --with-arch-directory=ppc64el --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --enable-objc-gc --enable-secureplt --with-cpu=power7 --with-tune=power8 --enable-targets=powerpcle-linux --disable-multilib --enable-multiarch --disable-werror --with-long-double-128 --enable-checking=release --build=powerpc64le-linux-gnu --host=powerpc64le-linux-gnu --target=powerpc64le-linux-gnu
Thread model: posix
gcc version 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.3)
```

EricWF:
DoNotOptimize uses inline assembly contraints to tell
the compiler what the type of the input variable. The 'g'
operand allows the input to be any register, memory, or
immediate integer operand. However this constraint seems
to be too weak on older GCC versions, and certain inputs
will cause compile errors.

This patch changes the constraint to 'X', which is documented
as "any operand whatsoever is allowed". This appears to fix
the issues with older GCC versions.

However Clang doesn't seem to like "X", and will attempt
to put the input into a register even when it can't/shouldn't;
causing a compile error. However using "g" seems to work like
"X" with GCC, so for this reason Clang still uses "g".

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -236,7 +236,7 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 #if defined(__clang__)
   asm volatile("" : : "g"(value) : "memory");
 #else
-  asm volatile("" : : "X"(value) : "memory");
+  asm volatile("" : : "i,r,m"(value) : "memory");
 #endif
 }
 // Force the compiler to flush pending writes to global memory. Acts as an
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -231,7 +231,13 @@ BENCHMARK_UNUSED static int stream_init_anchor = InitializeStreams();
 #ifndef BENCHMARK_HAS_NO_INLINE_ASSEMBLY
 template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
+  // Clang doesn't like the 'X' constraint on `value` and certain GCC versions
+  // don't like the 'g' constraint. Attempt to placate them both.
+#if defined(__clang__)
   asm volatile("" : : "g"(value) : "memory");
+#else
+  asm volatile("" : : "X"(value) : "memory");
+#endif
 }
 // Force the compiler to flush pending writes to global memory. Acts as an
 // effective read/write barrier
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,6 +1,7 @@
 # Enable the tests
 
 find_package(Threads REQUIRED)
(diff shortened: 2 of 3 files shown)


dominichamon:
ice, ice, baby. https://travis-ci.org/google/benchmark/jobs/238921070

EricWF:
Ouch, That's probably my fault. Writing a "portable" version of DoNotOptimize is awfully hard.

dominichamon:
you did pretty well to get all but one... 32-bit, old g++, optimized.

dominichamon:
thaaanks!

PR #404, opened 2017-06-15
Fix #403 HAVE_${VAR} not passed to makefile

### LINKED ISSUE #403: CXXFeatureCheck for regex for cross compiling failing
CXXFeature testing is not possible for cross compiling to another platform. 

Specifying -DRUN_HAVE_POSIX_REGEX=flase -DHAVE_POSIX_REGEX=true does not pass these definitions to the makefiles.

jernkuan:
Add definition of ${VAR} to makefiles if specified as part of cmake
parameter.


EricWF:
@jernkuan Please sign the CLA. We can't move forward until that is done.

EricWF:
@jernkuan I'm not sure what you mean "not passed to Makefiles". Can you clarify? Maybe provide an example?

jernkuan:
I have signed the CLA, somehow it's not picking it up.
I have checked my email address of the signed CLA and with the github.com account. Any other clues?

For example, if you are cross compiling you might do something like this.
cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=../../Toolchain.cmake -DHAVE_STD_REGEX=true

You will get an error, as the 
CMake Error at CMakeLists.txt:183 (message):

As the CXXFeatureCheck.cmake will do an early exit if the variable is define,  it does not add the definition to be used in the generation of the Makefile. The adding of the definition and setting of cache is executed only when RUN_HAVE_STD_REGEX is true.

dominichamon:
the author and commit appear to be jernkuanl@nvidia.com. Is that the email under which you signed the CLA?

jernkuan:
Ah yes, i have a mixed up author email address. I have made the changes, and pushed again to my branch. But seems that this is not picking up the changes to the author email. Should i close this and reissue a new pull request?

EricWF:
@jernkuan  that would likely work to resolve the issue. feel free to open a new PR.

PR #406, opened 2017-06-19
Adding support for batch iteration

astrelni:
This adds two new modes of iteration for benchmark::State.

The first is to allow batch operation benchmarking with State::KeepRunningBatch(size_t). This allows comparing per element performance of an algorithm that takes a varying number of elements as input. It also allows comparing algorithm performance as a function of some integer based parameter.

The second addition is support for range-based for loop iteration over State. The motivation is that running state.KeepRunning() every loop creates a limit on how fast very fast benchmarks can run. Approximately 2 - 3 ns / iteration. The cause is unoptimizable memory load / store operations. The goal with the range based for loop is to take advantage of the new batch mode to cache a number of benchmark iterations locally. This way, very fast benchmarks can amortize the load / store cost to zero.


astrelni:
@dominichamon

dominichamon:
clang release builds aren't happy:

```
google/benchmark/test/batch_test.cc:20:55: error: implicit
      conversion loses integer precision: 'size_t' (aka 'unsigned long') to
      'value_type' (aka 'int') [-Werror,-Wshorten-64-to-32]
  for (size_t i = 0; i < param; ++i) values.push_back(i);
```

README.md
```diff
@@ -190,6 +190,49 @@ Three macros are provided for adding benchmark templates.
...
+BENCHMARK(BM_ParameterizedRoutine)->Range(1<<10, 1<<18);
+```
+
+The advantage of using `KeepRunningBatch` is that each batch is counted as
+`param` number of iterations. This means the results for different batch
+sizes are directly comparable as a function of the parameter value.
+
+### Very fast benchmarks
+
+When benchmarking very fast operations (less than 3 ns per iteration), use the
```
dominichamon: is there any downside to using this method? perhaps we can be more firm on using this as the default?

EricWF:
I'm really skeptical that the claim that the range based for loop is quicker, and about these changes in general

include/benchmark/benchmark_api.h
```diff
@@ -493,8 +525,40 @@ class State {
         internal::ThreadManager* manager);
 
  private:
-  void StartKeepRunning();
-  void FinishKeepRunning();
+  friend class internal::StateIterator;
+
+  size_t GetBatch(size_t batch_size, bool atomic_batch) {
```
dominichamon: boolean parameters can cause callsite readability issues. it's usually better to define an enum instead:

```
enum class AtomicBatch { kNo, kYes };

int GetBatch(int size, AtomicBatch atomic);

...

GetBatch(n, AtomicBatch::kYes);
```

include/benchmark/benchmark_api.h
```diff
@@ -493,8 +525,40 @@ class State {
...
+    // atomic_batch == true means the batch size will always be batch_size, and
+    // iteration will stop at the first multiple of batch_size greater than or
+    // equal to max_iterations.
+    // atomic_batch == false means always iterate to max_iterations. The last
+    // batch will be smaller than batch_size if it does not go evenly into
+    // max_iterations.
+    assert(batch_size != 0);
```
dominichamon: consider CHECK_NE(batch_size, 0) from src/check.h.

note, you should probably move the implementation of this method to a .cc anyway.

test/batch_test.cc
```diff
@@ -0,0 +1,38 @@
...
+#include <algorithm>
+#include <cassert>
+
+static void BM_batch_iteration_count(benchmark::State& state) {
+  const size_t batch_size = state.range(0);
+  size_t actual_iterations = 0;
+  while (state.KeepRunningBatch(batch_size)) {
+    actual_iterations += batch_size;
+  }
+  assert(state.iterations() >= state.max_iterations);
```
dominichamon: you should be able to use CHECK in tests

dominichamon:
@astrelni should be able to provide some benchmark numbers (ha) to support the claims. This is something that has been tested internally.

include/benchmark/benchmark_api.h
```diff
@@ -693,8 +757,59 @@ class Benchmark {
   Benchmark& operator=(Benchmark const&);
 };
 
+// Incomplete iterator-like type with dummy value type so that benchmark::State
+// can support iteration with a range-based for loop. See comments on
+// benchmark::begin() / benchmark::end() concerning usage.
+class StateIterator {
+ public:
+  struct Value {
+    ~Value() {}  // Non-trivial destructor to avoid unused variable warnings
```
EricWF: This seems like a bad way to suppress unused value warnings. Having a non-trivial destructor could, when it's not optimized away, potentially undermine the entire point of this; which is writing the most efficient benchmark possible.

Using `struct BENCHMARK_UNUSED Value` suppresses the warning for GCC, we should use that instead.

EricWF:
> The motivation is that running state.KeepRunning() every loop creates a limit on how fast very fast benchmarks can run. Approximately 2 - 3 ns / iteration. The cause is unoptimizable memory load / store operations

Hmm, it does indeed. I guess that's because the State object is passed in by reference, and compilers might assume that the state is changed elsewhere in the program between iterations. The ranged-for solution works because it creates a temporary that the compiler can now prove isn't accessed elsewhere.

@astrelni Could you please separate the ranged-for addition from the main part of this patch, which is the `getBatch` change? I think they are separate concerns and I would like to review the separately.

include/benchmark/benchmark_api.h
```diff
@@ -693,8 +757,59 @@ class Benchmark {
...
+  };
+
+  StateIterator() : parent_(), cached_(0) {}
+  explicit StateIterator(State* parent) : parent_(parent), cached_(0) {}
+
+  StateIterator& operator++() {
```
EricWF: Should be use `BENCHMARK_ALWAYS_INLINE` here?

include/benchmark/benchmark_api.h
```diff
@@ -693,8 +757,59 @@ class Benchmark {
...
+    assert(cached_ > 0);
+    --cached_;
+    return *this;
+  }
+
+  bool operator!=(const StateIterator& other) {
```
EricWF: This should very likely have `BENCHMARK_ALWAYS_INLINE` too.

include/benchmark/benchmark_api.h
```diff
@@ -693,8 +757,59 @@ class Benchmark {
...
+    assert(parent_);
+    if (BENCHMARK_BUILTIN_EXPECT(cached_ != 0, true)) {
+      return true;
+    }
+    const size_t kRangeBasedBatchSize = 1000;
+    cached_ = parent_->GetBatch(kRangeBasedBatchSize, false);
+    return cached_ != 0;
+  }
+
+  Value operator*() { return Value(); }
```
EricWF: `BENCHMARK_ALWAYS_INLINE`

include/benchmark/benchmark_api.h
```diff
@@ -693,8 +757,59 @@ class Benchmark {
...
+// Functions to enable support for iteration using a range-based for loop.
+//
+// The only supported usage:
+//
+//   static void BM_test(benchmark::State& state) {
+//     for (auto _ : state) {
+//       // perform single iteration
+//     }
+//   }
+inline internal::StateIterator begin(State& state) {
```
EricWF: `BENCHMARK_ALWAYS_INLINE`

include/benchmark/benchmark_api.h
```diff
@@ -693,8 +757,59 @@ class Benchmark {
...
+  return internal::StateIterator(&state);
+}
+inline internal::StateIterator end(State&) { return internal::StateIterator(); }
```
EricWF: `BENCHMARK_ALWAYS_INLINE`

astrelni:
@EricWF Thanks for having a look. I'm not too familiar with the pull request workflow: is there any con to adding new commits to remove the range-based for from this pull request? Or do you prefer to start with a new pull request for both?

include/benchmark/benchmark_api.h
astrelni: Yeah, moving to cc saw a small performance hit (~1ns). Though we could move the slow path of the last ten or so lines into a separate method if that makes things a little cleaner.

test/batch_test.cc
astrelni: Was following along with all the other tests which use asserts. Tried it and we don't have visibility of check.h in test/

README.md
astrelni: My reasoning was that the range based for hurts readability a little bit with the unused opaque variable, and the vast majority of cases are not affected by KeepRunning. If that's a nonissue, I can reword the documentation to put the range-based for front and center.

The only other thing is that this library supports C++03 as far as I can tell. So I'm not sure where that leaves things. Add a separate section in the documentation for "old API + C++03"?

astrelni: One other thought. This may not be the time to resolve it since we're likely separating the range-based for addition from the first batch support, but I'd like to not forget to mention this.

I think there needs to be some clear documentation on State::iterations(). I may be mistaken, but I cannot find a place saying don't bother calling this inside of the benchmarking loop. The current implementation would return the number of actual iterations + 1 which is only slightly wrong. But, range-based for would be way off and for no apparent (to the user) reason. If checking iterations inside the benchmarking loop is something that should be supported, then I don't see how the range-based for can replace KeepRunning(). Please let me know if I'm misunderstanding this.

include/benchmark/benchmark_api.h
astrelni: Fixed

include/benchmark/benchmark_api.h
astrelni: Fixed

include/benchmark/benchmark_api.h
astrelni: Fixed

include/benchmark/benchmark_api.h
astrelni: Fixed

include/benchmark/benchmark_api.h
astrelni: Fixed

include/benchmark/benchmark_api.h
astrelni: Fixed

include/benchmark/benchmark_api.h
astrelni: Fixed. Had to wrap in a struct because of C++03 support.

README.md
dominichamon: The library supports C++03, and those users will have to use the `KeepRunning` api. Other users should strongly prefer the range loop, given the reduced overhead and simplicity of the call (even with the slightly unusual opaque variable).

I don't know why a user might want to call `State::iterations` inside the loop, but we can check that case and log a warning. we do this for other calls.

test/batch_test.cc
dominichamon: you might need `src/check.h` or `../src/check.h`.

See https://github.com/google/benchmark/blob/4cfe790a253d2cb47f33b29676c0a27c462704cb/test/output_test_helper.cc#L7

test/range_test.cc
```diff
@@ -0,0 +1,33 @@
+#include "benchmark/benchmark_api.h"
+
+#include <cassert>
+
+static void BM_range_empty(benchmark::State& state) {
+  for (auto _ : state) {
```
dominichamon: do you need to only run these tests for C++11 runs?

astrelni:
@dominichamon Since @EricWF hasn't had a chance to respond, what is your standing on splitting this up into two pull requests? And also, quoting my previous question to him:

> I'm not too familiar with the pull request workflow: is there any con to adding new commits to remove the range-based for from this pull request? Or do you prefer to start with a new pull request for both?

dominichamon:
there's no cons to commits that remove parts, and then adding a second PR for the other part.

test/batch_test.cc
astrelni: Done.

astrelni:
@dominichamon @EricWF I've updated to remove the range-based for from this pull request, please take a look.

astrelni:
Hi @EricWF, I just realized @dominichamon is likely on vacation right now. I have a sneaking suspicion you may be out as well. But in case otherwise, I'm pinging this thread. Please let me know if I accidentally missed addressing something.

src/benchmark.cc
```diff
@@ -145,6 +145,7 @@ class ThreadManager {
     double real_time_used = 0;
     double cpu_time_used = 0;
     double manual_time_used = 0;
+    int64_t iterations = 0;
```
EricWF: This seems like the wrong type, since we usually use `size_t` when dealing with iterations. It appears that the use of `int64_t` below is probably wrong too, but no need to address that in this patch.

src/benchmark.cc
```diff
@@ -270,14 +269,15 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
...
-      << "Benchmark returned before State::KeepRunning() returned false!";
+  CHECK(st.finished()) << "Benchmark returned before State::KeepRunning() / "
+                          "State::KeepRunningBatch() returned false!";
   {
     MutexLock l(manager->GetBenchmarkMutex());
     internal::ThreadManager::Result& results = manager->results;
     results.cpu_time_used += timer.cpu_time_used();
     results.real_time_used += timer.real_time_used();
     results.manual_time_used += timer.manual_time_used();
+    results.iterations += static_cast<int64_t>(st.iterations());
```
EricWF: The cast should be unnecessary here one `results.iterations` is made to have the type `size_t`.

src/benchmark.cc
```diff
@@ -325,6 +325,10 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
       results.real_time_used /= b.threads;
       results.manual_time_used /= b.threads;
 
+      // Set iters to actual number of iterations run per thread, which may have
+      // been higher if the benchmark used KeepRunningBatch.
+      iters = static_cast<size_t>(results.iterations / b.threads);
```
EricWF: Cast should be unnecessary here as well.

test/batch_test.cc
```diff
@@ -0,0 +1,39 @@
...
+
+#include "../src/check.h"  // NOTE: check.h is for internal use only!
+
+static void BM_batch_iteration_count(benchmark::State& state) {
+  const size_t batch_size = state.range(0);
+  size_t actual_iterations = 0;
+  while (state.KeepRunningBatch(batch_size)) {
+    actual_iterations += batch_size;
+  }
+  CHECK_GE(state.iterations(), state.max_iterations);
```
EricWF: Sorry I know @dominichamon just suggested the opposite, but I think using `<cassert>` is this test is preferable, just so we can avoid any potential ODR issues for non-debug builds of `libbenchmark.a`.

Ideally the other tests shouldn't use `check.h` either, but the added output turns out to be infinitely helpful in those cases. In this case `<cassert>` should do fine. I'll look into fixing the other tests, or fixing `check.h` to avoid ODR violations in the near future.

include/benchmark/benchmark_api.h
```diff
@@ -331,16 +344,19 @@ class State {
...
-    }
-    bool const res = total_iterations_++ < max_iterations;
-    if (BENCHMARK_BUILTIN_EXPECT(!res, false)) {
-      FinishKeepRunning();
-    }
-    return res;
-  }
+  bool KeepRunning();
+
+  // Returns true if the benchmark should process another 'n' iterations. The
```
EricWF: This seems to explicitly allow the batch size to change within the same run of a benchmark, ie the argument passed to the first call differs from the argument passed to the second (within the same run). Is this behavior intentional? Is it needed?

It appears to me that if the batch size is fixed we can choose a `max_iteration` count that is exactly divisible by `kBatchSize`. Not sure how much utility that will have, but I just wanted to inquire.

test/batch_test.cc
EricWF: Actually disregard me. There are no ODR violations when using `check.h` so this is fine. My mistake.

include/benchmark/benchmark_api.h
astrelni: Thanks for bringing this up. It does sound like probably a cleaner solution. There is one situation where I'm seeing it complicate things: 

As much as I'd like it not to be the case, I think the range-based for loop API will depend on this behavior somewhat. Especially if we want to document it as the API of choice. The range-based for uses 1000 as the batch size, but the API could easily be used by a benchmark which is slow enough to only run in the single digits to tens of iterations in a reasonable amount of time. So we want to easily be able to cut the batch size short when that feature comes.

I guess thinking more, we could give a bool flag for State for whether KeepRunningBatch was called or not and use that when calculating the number of iterations. (Also would add the batch size as a State attribute and update the contract that the batch size must remain the same)

Thanks again for bringing this up. I will work on this tomorrow, but I'll leave this comment here already in case you have some thoughts to add in the meantime.

src/benchmark.cc
astrelni: size_t is the natural type to provide in the API, but I think int64_t is more appropriate for the internals because we do arithmetic with these values. In 32-bit applications, size_t could overflow and wrap*, giving wrong but potentially plausible benchmark readouts. I think it's the same reason bytes_processed and items_processed are stored as int64_t's.

* kMaxIterations seems to be 1 billion now, so > 4 threads could do it. Or in case kMaxIterations is increased in the future.

include/benchmark/benchmark_api.h
EricWF: > As much as I'd like it not to be the case, I think the range-based for loop API will depend on this behavior somewhat.

Could you submit a separate PR (likely derived from this branch) with the range-based for loop implementation so it's visible and therefore easier to discuss?

>  The range-based for uses 1000 as the batch size,  but the API could easily be used by a benchmark which is slow enough to only run in the single digits to tens of iterations in a reasonable amount of time.

Shouldn't the range-based for loop simply use the total iteration count as it's batch size? I'm assuming the range-based for loop can only be used to execute all of the iterations. Having the range-based for loop execute larger batch sizes then there are max iterations doesn't make sense and conflicts with the fundamental design of the library. Perhaps I'm not understanding something?

> So we want to easily be able to cut the batch size short when that feature comes.

IDK what "cutting the batch size short" really means. Can you give a motivating example?

> I guess thinking more, we could give a bool flag for State for whether KeepRunningBatch was called or not and use that when calculating the number of iterations. (Also would add the batch size as a State attribute and update the contract that the batch size must remain the same)

Hmm, that sounds reasonable to me but it does complicate things, because before the first attempted run of the benchmark the library doesn't know the batch size that will be used (if any). I guess the solution would be to have a special first run of the benchmark where `KeepRunning()` and `KeepRunningBatch(n)` only return true once, just so we can determine if `KeepRunningBatch` is used and if so then the potential batch size. The library would then calculate the actual number of iterations to use as a multiple of the batch size. In order for this to work it's critical the batch size always equal the batch size specified during the first iteration of the first run.

Using this approach ensures that the library still has full control over the number of iterations run, and that the final batch of iterations doesn't "overflow" the max iteration count selected by the library. Ex. The library selects 15 iterations, but the batch size used by the user is 10, meaning either 10 or 20 iterations will be performed but not the 15 selected by the library.

I'm also concerned that allowing variable batch counts will allow users to write benchmarks that have misleading or inaccurate results, because the size of the batch reported to the library doesn't actually reflect the amount of work done by the actual loop iteration.

EricWF:
@astrelni Do you think you'll have time to revive this soon?

astrelni:
@EricWF Yes, very sorry for the long hiatus. I am going on parental any day now and some priorities had to be juggled around. I am working on updates in response to your comments now and hope to have something for you to look at soon. (I don't plan on dropping off entirely and want to finish this out even if it has to be while on leave)

EricWF:
@astrelni Do you mind if I go ahead with my own version of the ranged-based-for loop patch? Since I think that has the performance benefits you suggested even without batch iteration (which can come later)?

astrelni:
@EricWF thanks a lot for getting the range-for in. Sorry that you had to do it yourself. I will come back later to add the batch iteration. I hope very much not to have you looking at it for as much time as you worked on this PR.

EricWF:
@astrelni No need to apologize. Thanks for letting me hijack your idea. I hope you don't mind :-)

astrelni:
@EricWF Not at all, you're helping me out. No need to credit me, it was actually a fleshed out implementation of something that was originally Dmitry Vyukov's idea. I probably should have credited him in the original description of this PR, but this was my first attempt at an OSS contribution. Is that regular etiquette for PRs? If so, maybe I should go add his name to your PR.

PR #407, opened 2017-06-20
Make Benchmark a single header library (but not header-only)

EricWF:
This patch refactors benchmark into a single header, to allow
for slightly easier usage.

The initial reason for the header split was to keep C++ library
components from being included by benchmark_api.h, making that
part of the library STL agnostic. However this has since changed
and there seems to be little reason to separate the reporters from
the rest of the library.


EricWF:
@dominichamon I'm not sure this is the right thing to do, and I would love your input.

dominichamon:
once this lands, we should do a release. then we can do a later release with the deprecated headers removed.

dominichamon:
@EricWF  i think it's a good thing to do (it used to be one header) at least for the api/benchmark split. reporters are something that only some users would need, i think, so it may be adding more to the compile overhead...

EricWF:
> @EricWF i think it's a good thing to do (it used to be one header) at least for the api/benchmark split. 

Agreed. at minimum that case is better. However we should also consider that changing `benchmark.h` to no longer include `reporter.h` would be a breaking change. 

> reporters are something that only some users would need, i think, so it may be adding more to the compile overhead...

The actual reporter class declarations shouldn't add a significant amount of overhead, they're pretty small. However it does require including `<set>`, but we already include `<map>` in `benchmark_api.h`, so we're already pulling in most of `<set>` anyway (because STL's implement them both using the same internals).

@dominichamon I'll do some tests to see if compile times have slowed noticeably.

dominichamon:
I appreciate the adherence to correctness but it's not going to be something i'm hung up on.

EricWF:
> I appreciate the adherence to correctness but it's not going to be something i'm hung up on.

Normally I can't even consider such a breaking change (in libc++). I'm perfectly happy to keep the reporters header separate. I'll revert that part of the change.

EricWF:
Although it does kind of suck that a change that didn't fold in `reporter.h` would actually cause the library to be, at least technically, less of a single header library than it started. Because previously `benchmark.h` actually acted like a single header, without actually being one.

dominichamon:
true. 

merge 'em.

EricWF:
@dominichamon Thanks! Merging with `reporter.h` being merged into `benchmark.h`.

PR #414, opened 2017-07-06
Fix ThreadCPUUsage when running on RTEMS.

tommadams:
Change ThreadCPUUsage to call ProcessCPUUsage if __rtems__ is defined.

RTEMS real time OS doesn't support CLOCK_THREAD_CPUTIME_ID. See
https://github.com/RTEMS/rtems/blob/master/cpukit/posix/src/clockgettime.c#L58-L59

Prior to this change, ThreadCPUUsage would fail when running on RTEMS with:
  ERROR: clock_gettime(CLOCK_THREAD_CPUTIME_ID, ...) failed


dominichamon:
appveyor looks like a false positive

PR #416, opened 2017-07-13
Google-style cleanups

dominichamon:
Internal tooling suggested these. It'll help us as we keep the internal and external versions in sync.


EricWF:
LGTM. Feel free to merge once the bots sign off.

PR #421, opened 2017-07-20
Add html reports

vgvassilev:
This patch is based on #193. It adds highcharts visualization to the benchmark results by making minimal changes to the code base.

image

It still has a few rough edges which we will iron out soon.


vgvassilev:
I signed it!

dominichamon:
@LepelTsmok Can you confirm you are ok with this PR (as you originally created the first commit).

include/benchmark/benchmark.h
```diff
@@ -1181,6 +1201,39 @@ class CSVReporter : public BenchmarkReporter {
   std::set< std::string > user_counter_names_;
 };
 
+class HTMLReporter : public BenchmarkReporter {
```
dominichamon: i'm still a little uncomfortable with this as a reporter. given we output in JSON (and can add more fields there if necessary) an HTML view could be a server that runs as a standalone binary pointed at a JSON result. That could then be generalised into a service to which people might upload their results and see them visualised over time.

src/html_reporter.cc
```diff
@@ -0,0 +1,487 @@
...
+    "Calculated values: <b>{point.key}</b><br/>Real time: <b>{point.y} "
+    "ns</b><br/>Argument 1: <b>{point.x}</b>",
+    "Calculated values: <b>{point.key}</b><br/>Bytes per second: "
+    "<b>{point.y}</b><br/>Argument 1: <b>{point.x}</b>"};
+
+static const char *const column_tool_tip_bodies[4] = {
+    "CPU time: <b>{point.y} ns</b>", "Items per second: <b>{point.y}</b>",
+    "Real time: <b>{point.y} ns</b>", "Bytes per second: <b>{point.y}</b>"};
+
+static const char *const high_chart_bar_function =
```
dominichamon: examples like this are why i think having this be a separate service is the right approach. any changes to how the data should be rendered now require updates to the benchmark library. that's the wrong abstraction layer.

include/benchmark/benchmark.h
vgvassilev: Thanks for the feedback. Are you saying that such a service should be outside of the project?

src/html_reporter.cc
vgvassilev: I am not sure I can follow, this seems to be true for any other reporter. I'd think of this as adding 'colored' console output. It could be done by a separate service but it is very useful to have this in the library, available to everybody.

OTOH the html reports seem like a popular user request. I'd expect that we will not change much of this code. We want to be able to plot benchmark families and counters.

oshadura:
I signed it!

dominichamon:
still working on this?

dominichamon:
Given how far behind this is from head, i'm going to close it. i'm still open to the idea of having an HTML reporter, but i do think it should be something that is generated from the JSON output instead of being embedded.

Issue #420, opened 2017-07-20
ODR violations on gcc 5.4.0


gnarlie:
Building master with gcc 5.4, I'm getting ODR violations. I've traced the issue to a9a66c85bbfda1d744c267c5e5aa073ef3d1c1d5

gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

```
[51%] Linking CXX executable cxx03_test
/usr/include/c++/5/bits/stl_tree.h:134:12: error: type ‘struct _Rb_tree_node’ violates one definition rule [-Werror=odr]
     struct _Rb_tree_node : public _Rb_tree_node_base
            ^
/usr/include/c++/5/bits/stl_tree.h:134:12: note: a different type is defined in another translation unit
     struct _Rb_tree_node : public _Rb_tree_node_base
            ^
/usr/include/c++/5/bits/stl_tree.h:139:12: note: the first difference of corresponding definitions is field ‘_M_value_field’
       _Val _M_value_field;
            ^
/usr/include/c++/5/bits/stl_tree.h:149:41: note: a field with different name is defined in another translation unit
       __gnu_cxx::__aligned_membuf<_Val> _M_storage;

                                         ^
[... continues ...]
```

dominichamon:
i just ran a check under the AddressSanitizer (which can detect these) and
nothing got flagged. I'll keep looking.

Dominic Hamon | Google
*There are no bad ideas; only good ideas that go horribly wrong.*

gnarlie:
The error appears to be valid ORD violation - the test cxx03_test (built with std=c++03) is linking libbenchmark.a, built with C++11. At least for this version of gcc, the tree nodes in a std::map appear to have a different layout based on the value of `__cplusplus`.

Disabling LTO works around the problem.

```c++
  template<typename _Val>
    struct _Rb_tree_node : public _Rb_tree_node_base
    {
      typedef _Rb_tree_node<_Val>* _Link_type;

#if __cplusplus < 201103L
      _Val _M_value_field;

      _Val*
      _M_valptr()
      { return std::__addressof(_M_value_field); }

      const _Val*
      _M_valptr() const
      { return std::__addressof(_M_value_field); }
#else
      __gnu_cxx::__aligned_membuf<_Val> _M_storage;

      _Val*
      _M_valptr()
      { return _M_storage._M_ptr(); }

      const _Val*
      _M_valptr() const
      { return _M_storage._M_ptr(); }
#endif
    };
```

[...]

PR #422, opened 2017-07-20
Cleaner UserCounter interface (instead of typedef std::map)

dominichamon:
I think this is cleaner, but I'm also open to the notion it might be extra abstraction with no benefit.


gnarlie:
Moving the implementation of UserCounters to the .cc file would likely fix #420 as well.

EricWF:
I'm not sure I think this is an improvement. People know and understand the STL interface and I think it's a benefit to offer it directly.

dominichamon:
I'm wary of users accidentally breaking user counters by (ab)using the map
interface. This is also inspired in part by #420.

Dominic Hamon | Google

EricWF:
> I'm wary of users accidentally breaking user counters by (ab)using the map
interface. 

That feels like a "Doctor, it hurts when I do this" situation, where users are free to make their lives difficult for themselves.

Your suggested interface was a part of the original review, but was removed for a number of reasons. One in particular being when you want to be able to change the value of an existing counter without having to always lookup up the key in the map. Ex:

```
static void BM_Foo(benchmark::State &S) {
  auto &CounterRef = (S.counters["foo"] = 42);
  while (S.KeepRunning()) {
    int Res = Foo();
    CounterRef += Res;
  } 
}
```

> This is also inspired in part by #420.

I don't think this will have an affect on #420, since the `std::map` definitions will still be in `libbenchmark.a` and will be visible during link time.

dominichamon:
fair enough :)

PR #424, opened 2017-07-23
Disable cxx03_test for GCC 5.x and above. (#420)

gnarlie:
GCC 5.x does not keep the same definitions for std::map between
C++11 and C++03, leading to ODR violations for anything using
std::map with link type optimizations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ matrix:
     - compiler: gcc
       env: COMPILER=g++ C_COMPILER=gcc BUILD_TYPE=Release
     - compiler: gcc
-      env: COMPILER=g++-5.8 C_COMPILER=gcc-5.8 BUILD_TYPE=Release
-    - compiler: gcc
       addons:
         apt:
           packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ matrix:
     - compiler: gcc
       env: COMPILER=g++ C_COMPILER=gcc BUILD_TYPE=Release
     - compiler: gcc
+      env: COMPILER=g++-5.8 C_COMPILER=gcc-5.8 BUILD_TYPE=Release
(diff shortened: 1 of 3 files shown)


EricWF:
> GCC 5.x does not keep the same definitions for std::map between
C++11 and C++03, leading to ODR violations for anything using
std::map with link type optimizations.

I'm not sure disabling the test really does anything of value. Either way the definitions that libstdc++ provides, although different, should be compatible if you can somehow suppress the diagnostic.

dominichamon:
@EricWF I agree that disabling is a short term fix. Is it ever correct though to link a binary with the benchmarks compiled in cxx03 and the library in cxx11 mode?

gnarlie:
@EricWF I'm not sure that the diagnostic can be suppressed - its an error, not a warning. Perhaps we can just disallow LTO for that test?

 @dominichamon - to be valid, the C++03 binary would need to be compiled with the same ABI settings introduced in GCC 5 to support C++11. It can be done, but I can't think of a use case.

EricWF:
@gnarlie Disabling LTO for that test sounds reasonable.

> to be valid, the C++03 binary would need to be compiled with the same ABI settings introduced in GCC 5 to support C++11. It can be done, but I can't think of a use case.

I'm almost certain this isn't actually an ABI issue; both versions should compile to ABI compatible code. Typically standard libraries don't break ABI compatibility between dialects.

However this is technically a ODR violation as the diagnostic correctly points out, but the violation should be benign.

EricWF:
@gnarlie I went ahead and fixed this in abafced9909c7e5e8f6b8236eecd953caa4f8e6f, so your patch is no longer needed.

Thanks for the report and PR. My apologies for hijacking the patch.

gnarlie:
@EricWF looks good - thanks for fixing

PR #426, opened 2017-07-24
Json reporter: don't cast floating-point to int; adjust tooling

LebedevRI:
`Json` output format is generally meant for further processing
using some automated tools. Thus, it makes sense not to
intentionally limit the precision of the values contained
in the report.

As it can be seen, `FormatKV()` for doubles, used `%.2f` format,
which was meant to preserve at least some of the precision.
However, before that function is ever called, the doubles
were already cast to the integer via `RoundDouble()`...

This is also the case for `console` reporter, where it makes
sense because the screen space is limited, and this reporter,
however the `CSV` reporter does output some decimal digits.

Thus i can only conclude that the loss of the precision
was not really considered, so i have decided to adjust the
code of the `json` reporter to output the full fp precision.

There can be several reasons why that is the right thing
to do, the bigger the time_unit used, the greater the
precision loss, so i'd say any sort of further processing
(like e.g. `tools/compare_bench.py` does) is best done
on the values with most precision.

Also, that cast skewed the data away from zero, which
i think may or may not result in false- positives/negatives
in the output of `tools/compare_bench.py`

---

Testing: 
```
$ ctest
$ python2 tools/gbench/report.py
```

### DIFF
--- a/src/json_reporter.cc
+++ b/src/json_reporter.cc
@@ -53,9 +53,8 @@ std::string FormatKV(std::string const& key, double value) {
   std::stringstream ss;
   ss << '"' << key << "\": ";
 
-  auto max_digits10 = std::numeric_limits<decltype (value)>::max_digits10;
-  auto one_digit10_in_integer_part = 1;
-  auto max_fractional_digits10 = max_digits10 - one_digit10_in_integer_part;
+  const auto max_digits10 = std::numeric_limits<decltype (value)>::max_digits10;
+  const auto max_fractional_digits10 = max_digits10 - 1;
 
   ss << std::scientific << std::setprecision(max_fractional_digits10) << value;
   return ss.str();
--- a/src/json_reporter.cc
+++ b/src/json_reporter.cc
@@ -21,6 +21,8 @@
 #include <string>
 #include <tuple>
 #include <vector>
+#include <iomanip> // for setprecision
+#include <limits>
 
 #include "string_util.h"
 #include "timers.h"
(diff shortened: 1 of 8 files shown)


src/json_reporter.cc
```diff
@@ -48,7 +50,15 @@ std::string FormatKV(std::string const& key, int64_t value) {
 }
 
 std::string FormatKV(std::string const& key, double value) {
-  return StringPrintF("\"%s\": %.2f", key.c_str(), value);
+  std::stringstream ss;
+  ss << '"' << key << "\": ";
+
+  auto max_digits10 = std::numeric_limits<decltype (value)>::max_digits10;
+  auto one_digit10_in_integer_part = 1;
```
dominichamon: just `- 1` on the next line. this variable doesn't add anything to the code.

dominichamon:
thank you!

LebedevRI:
Thank *you*.

KindDragon:
This PR broke Windows build: https://ci.appveyor.com/project/google/benchmark/build/692

LebedevRI:
Those builds seem to be flaky, and if you look into history, i'd say it was like that even before this PR..

KindDragon:
@LebedevRI I can't find broken builds from master branch before your changes

KindDragon:
Ah, it's because of line: 

    BM_CounterSet2_Tabular/threads:8          231 ns    1953125 ns          8         30         40         10

`1953125 ns` can't match regex `[ ]*%hrfloat` which is `[ ]*[0-9]{1,5} ns`

LebedevRI:
FWIW the build for this PR succeeded, as you can see in https://github.com/google/benchmark/pull/426#issuecomment-317555267

test/reporter_output_test.cc
```diff
@@ -27,9 +27,9 @@ BENCHMARK(BM_basic);
 
 ADD_CASES(TC_ConsoleOut, {{"^BM_basic %console_report$"}});
 ADD_CASES(TC_JSONOut, {{"\"name\": \"BM_basic\",$"},
-                       {"\"iterations\": %int,$", MR_Next},
-                       {"\"real_time\": %int,$", MR_Next},
-                       {"\"cpu_time\": %int,$", MR_Next},
+                       {"\"iterations\": %float,$", MR_Next},
```
KindDragon: Why iterations `%float`?

LebedevRI: Hmm, you're right, this is not right, likely too greedy s///
https://github.com/google/benchmark/pull/431

Issue #324, opened 2016-12-19
Generalize SetItemsProcessed and SetBytesProcessed


baryluk:
Simple feature frequest.

It would be nice to say something like this:

    SetProcessed("requests", 123, "QPS");
    SetProcessed("keys", 12300, "KPS");
    SetProcessed("bytes", 90000, "B/s");
    SetProcessed("errors", 1, "EPS");

etc.

Benchmark framework would sort them by key (or just use in specified order) in each benchmark, compute rates (taking into accounting wall time or cpu time as needed, possibly configurable per SetProcessed call), separate them by semicolons, and if needed format them properly for human consumption.

Proposal:

    SetProcessed(string key, int64 processed, string unit, Options options = Options());

Options could have things like: should average, min, max or sum, be taken from different threads; should we use cpu time or wall time clock, should k, M, G, prefixes be used, or maybe floating point with exponent, or just integer with fixed rounding, should additional statistic be compute (percentiles, median, std dev, iqr, normality test - i.e. from different benchmark repetitions, etc).

Key would be used in other formats like json, csv, to name columns.

We can also extend Benchmark, to allow extending columns explicitly, i.e.

    ->AddProcessedKey("requests", 10)
    ->AddProcessedKey("keys", 10)
    ->AddProcessedKey("bytes")

This could be used by the console to make nice table, including a table header row. 10 in an example above would be optional column width.

Not sure, what would be approach, if different benchmarks declare different key names. (Reemit new header, if it is different than a previous one?)

baryluk:
BTW. SetLabel would remain unchanged, and still be printed after all key rates.

dominichamon:
this is #262 i think?

baryluk:
Yes indeed. I was not aware of that PR, or issue #240 

There are few differences in my proposal, i.e. ability to report a rates (not just counters, which doesn't play well when using with automatic iteration count computation or time based iterations), and ability to set options and suffixes for example

dominichamon:
perhaps it can be extended to cover your use cases (once it lands.. longest PR ever)?

LebedevRI:
Question: is there any status here?
I would like to have a median in addition to the currently-available mean.
If i just add it, would that be upstreamable, or is the full rework like suggested here is the only way to go?

LebedevRI:
Edit: i didn't know `Stat1` was optimized to not keep the values themselves...
I did implement median, but ATM at the cost of **adding** `std::vector<VType> values_;`
[...]

PR #427, opened 2017-07-25
Tooling: generate_difference_report(): show old/new for both values. Fixes #425

### LINKED ISSUE #425: compare_bench.py: percents and values are for different columns
Hi. This is with v1.2.0
the output of the script is like:
```
...
Comparing ./bench-old to ./src/bench
Benchmark                       Time           CPU           Old           New
------------------------------------------------------------------------------
stuff/real_time                -0.01         +0.07      23671928      25389282

```
However if you manually search for the 23671928 in this case, you'll see that the percentages are displayed for the first column, `Time`, while the values are from the second column, `CPU`.
In cases where `b->UseRealTime();` is used, this is at the very least highly confusing.
Roman.

LebedevRI:
While the percentages are displayed for both of the columns,
the old/new values are only displayed for the second column,
for the CPU time. And the column is not even spelled out.

In cases where b->UseRealTime(); is used, this is at the
very least highly confusing. So why don't we just
display both the old/new for both the columns?

The obvious reason *not* to do this would be the
line lenght considerations. But since the bench name
is not limited in length, i'm not sure it matters...

Fixes #425, also, Python3 fix.

### DIFF
--- a/tools/gbench/Inputs/test1_run1.json
+++ b/tools/gbench/Inputs/test1_run1.json
@@ -71,6 +71,13 @@
       "time_unit": "ns"
     },
     {
+      "name": "BM_10PercentCPUToTime",
+      "iterations": 1000,
+      "real_time": 100,
+      "cpu_time": 100,
+      "time_unit": "ns"
+    },
+    {
       "name": "BM_DifferentTimeUnit",
       "iterations": 1,
       "real_time": 1,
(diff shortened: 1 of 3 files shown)


LebedevRI:
Thank you!
@dominichamon  Is there any chance i could get some feedback on https://github.com/google/benchmark/issues/324#issuecomment-317276303
I'd like to also have median output in addition to mean, and that is impossible since `Stat1` does not store the actual values. I suppose if it is extended to store the values, then the interface could be extended to also support any user-provided metric over all the measurements.

PR #428, opened 2017-07-26
Drop Stat1, refactor statistics to be user-providable, add median.

### LINKED ISSUE #255: Calculating Deviation
Hi,
I'm thinking of adding a min/max to the calculations. This is useful to me as we have a project with a maximum latency requirement. Other measurements like standard deviation could also be included.

Would this sort of thing be acceptable?

LebedevRI:
My main goal was to add **median** statistic. Since `Stat1`
calculated the stats incrementally, and did not store
the values themselves, it is was not possible. Thus,
i have replaced `Stat1` with simple `std::vector<double>`,
containing all the values.

As discussed in https://github.com/google/benchmark/issues/324#issuecomment-317811146, this should not be _too_ intrusive...

Then, i have refactored current **mean**/**stdev** to be a
function that is provided with values vector, and
returns the statistic. While there, it seemed to make
sense to deduplicate the code by storing all the
statistics functions in a map, and then simply iterate
over it. And the interface to add new statistics is
intentionally exposed, so they may be added easily,
which is requested in #255.

The notable change is that Iterations are no longer
displayed as 0 for stdev. Is could be changed, but
i'm not sure how to nicely fit that into the API.

Similarly, this dance about sometimes (for some fields,
for some statistics) dividing by run.iterations, and
then multiplying the calculated stastic back is also
dropped, and if you do the math, i fail to see why
it was needed there in the first place.

Since that was the only use of `stat.h`, it is removed.

I suppose this needs more tests?

Maybe Fixes #255.


src/complexity.cc
```diff
@@ -193,63 +201,50 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
...
+      it->second.s.emplace_back(cnt.second);
     }
   }
 
-  // Get the data from the accumulator to BenchmarkReporter::Run's.
-  Run mean_data;
-  mean_data.benchmark_name = reports[0].benchmark_name + "_mean";
-  mean_data.iterations = run_iterations;
-  mean_data.real_accumulated_time =
-      real_accumulated_time_stat.Mean() * run_iterations;
```
LebedevRI: Hmm, if i think really hard about this, i'm not 100% sure this `run_iterations` trickery is never ever needed?
So maybe `run_iterations` should be provided as a parameter into the lambda?

dominichamon: We've had attempts in the past to clean up this `run_iterations` thing as it does seem like it isn't necessary. However, and I wish i could remember or dig up the reasoning, it ends up coming in to play for some combinations of runs/threads/repetitions.

However, if your change shows expected results when we run the tests we have, i'm comfortable with the notion of the change.

LebedevRI:
(Do not merge yet, still want to at least attempt to look into test coverage...)

src/complexity.cc
LebedevRI: > expected results when we run the tests we **have**

I'm not actually seeing any test that fails if i cripple mean/stddev calculation, so i'm not sure it is actually tested.
But yes, *so far* **i** have not seen any wrong output.

LebedevRI:
I did manually check several stdev results, and they appear to match.
I'm not sure what else to do here, any review comments?

README.md
```diff
@@ -384,27 +383,48 @@ the minimum time, or the wallclock time is 5x minimum time. The minimum time is
...
 Calling `ReportAggregatesOnly(bool)` on a registered benchmark object overrides
 the value of the flag for that benchmark.
 
+## User-defined statistics for repeated benchmarks
+While having mean, median and standard deviation is nice, this may not be
+enough for everyone. For example you may want to know what is the largest
+observation, e.g. because you have some real-time constraints. This is easy.
+
+```c++
+const auto StatisticsMax = [](const std::vector<double>& v) {
```
dominichamon: `auto` return type should probably be "double". And a `const` return type doesn't make a whole lot of sense given how it can be copied into a mutable value at the callsite.

README.md
```diff
@@ -384,27 +383,48 @@ the minimum time, or the wallclock time is 5x minimum time. The minimum time is
...
+
+void BM_spin_empty(benchmark::State& state) {
+  while (state.KeepRunning()) {
+    for (int x = 0; x < state.range(0); ++x) {
+      benchmark::DoNotOptimize(x);
+    }
+  }
+}
+
+BENCHMARK(BM_spin_empty)->ComputeStatistics("max", StatisticsMax)->Arg(512);
```
dominichamon: i wonder if we should (and this is total idle thought) make the standard stats methods wrappers public so a user can just use:

```
BENCHMARK(BM_foo)
  ->ReportMean()
  ->ReportMedian()
  ->ReportPercentile(90)
  ->Repetitions(100);
```

include/benchmark/benchmark.h
```diff
@@ -378,6 +378,18 @@ enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
...
 
+// StatisticsFunc is passed to a benchmark in order to compute some descriptive
+// statistics over all the measurements of some type
+typedef double(StatisticsFunc)(const std::vector<double>&);
+
+struct Statistics {
+  std::string name;
+  StatisticsFunc* compute;
+
+  Statistics(std::string name_, StatisticsFunc* compute_)
```
dominichamon: no _ on the args please

include/benchmark/benchmark.h
```diff
@@ -378,6 +378,18 @@ enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
 // computational complexity for the benchmark.
 typedef double(BigOFunc)(int);
 
+// StatisticsFunc is passed to a benchmark in order to compute some descriptive
+// statistics over all the measurements of some type
+typedef double(StatisticsFunc)(const std::vector<double>&);
+
+struct Statistics {
+  std::string name;
```
dominichamon: any reason these are public?

include/benchmark/benchmark.h
```diff
@@ -378,6 +378,18 @@ enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
 // computational complexity for the benchmark.
 typedef double(BigOFunc)(int);
 
+// StatisticsFunc is passed to a benchmark in order to compute some descriptive
+// statistics over all the measurements of some type
+typedef double(StatisticsFunc)(const std::vector<double>&);
+
+struct Statistics {
```
dominichamon: i think this can be a fwd decl here with the implementation in the `benchmark.cc` file. this helps discourage library users from trying to instantiate one of these, which given the API below they shouldn't need to.

src/benchmark_register.cc
```diff
@@ -409,6 +413,12 @@ Benchmark* Benchmark::Complexity(BigOFunc* complexity) {
   return this;
 }
 
+Benchmark* Benchmark::ComputeStatistics(std::string name,
```
dominichamon: i think this belongs in the benchmark.cc file.

src/complexity.cc
```diff
@@ -163,25 +162,34 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
     // We don't report aggregated data if there was a single run.
     return results;
   }
+
```
dominichamon: i'm not sure how this method ended up here, but i'm wondering if it belongs in here, or in a new `statistics.cc` implementation file.

src/complexity.h
```diff
@@ -56,5 +59,55 @@ struct LeastSq {
...
+const auto StatisticsSum = [](const std::vector<double>& v) {
+  return std::accumulate(v.begin(), v.end(), double());
+};
+
+const auto StatisticsMean = [](const std::vector<double>& v) {
+  if (v.size() == 0) return double();
+  return StatisticsSum(v) * (1.0 / v.size());
+};
+
+const auto StatisticsMedian = [](const std::vector<double>& v) {
```
dominichamon: why are these lambdas? it seems they could just as easily be concrete functions declared here (or in a `src/statistics.h`) and defined in an implementation file.

LebedevRI:
Just want to note that i did fix most of the review notes, except a few, and i'm not sure about those...
I'd love to have some feedback, please :)

LebedevRI:
Friendly ping :)

LebedevRI:
Friendly weekly ping :)

dominichamon:
(i was away for a couple of weeks. i'm still catching up but i'll get to this review in the next couple of days)

LebedevRI:
Maybe the last ping, pretty please? ^^
I would gladly change the code, i just need to know what is the vision/what to change...

dominichamon:
I'm so sorry. Totally my fault. I'll see if I can take a look tonight.

src/benchmark_register.cc
```diff
@@ -159,6 +159,7 @@ bool BenchmarkFamilies::FindBenchmarks(
         instance.use_manual_time = family->use_manual_time_;
         instance.complexity = family->complexity_;
         instance.complexity_lambda = family->complexity_lambda_;
+        instance.statistics = &family->statistics_;
```
dominichamon: everything else here takes a copy rather than a reference (pointer). is it possible the lifetime of family is shorter than that of instance? should you take a copy (even though it'll be a memory/cpu hit)?

src/statistics.cc
```diff
@@ -0,0 +1,177 @@
...
+#include "statistics.h"
+
+namespace benchmark {
+
+auto StatisticsSum = [](const std::vector<double>& v) {
+  return std::accumulate(v.begin(), v.end(), double());
+};
+
+double StatisticsMean(const std::vector<double>& v) {
+  if (v.size() == 0) return double();
```
dominichamon: further down you return explicit 0.0 instead. i think i prefer that for readability, but either way be consistent please.

src/statistics.cc
```diff
@@ -0,0 +1,177 @@
...
+  std::vector<double> partial;
+  // we need roundDown(count/2)+1 slots
+  partial.resize(1 + (v.size() / 2));
+  std::partial_sort_copy(v.begin(), v.end(), partial.begin(), partial.end());
+  double median;
+  // did we have odd number of samples?
+  // if yes, then the last element of partially-sorted vector is the median
+  // it no, then the average of the last two elements is the median
+  if(v.size() % 2 == 1)
+    median = partial.back();
```
dominichamon: you could simplify this to

`
if(v.size() % 2 == 1)
  return partial.back();
return partial....;
`

src/statistics.cc
```diff
@@ -0,0 +1,177 @@
...
+  std::vector<double> items_per_second_stat;
+
+  real_accumulated_time_stat.reserve(reports.size());
+  cpu_accumulated_time_stat.reserve(reports.size());
+  bytes_per_second_stat.reserve(reports.size());
+  items_per_second_stat.reserve(reports.size());
+
+  // All repetitions should be run with the same number of iterations so we
+  // can take this information from the first benchmark.
+  int64_t const run_iterations = reports.front().iterations;
```
dominichamon: this should be size_t as that's the type as stored.

dominichamon:
just a few nits. it looks really good, thank you.

src/benchmark_register.cc
LebedevRI: Hmm. I'm using `-fsanitize=address -fsanitize=undefined`, and at least with all the current tests, there are no complaints.

LebedevRI: This one would break too:
https://github.com/google/benchmark/blob/d70417994a3c845c49c4443e92b26a52b320a759/src/benchmark_register.cc#L150

LebedevRI: But then looking at https://github.com/google/benchmark/blob/d70417994a3c845c49c4443e92b26a52b320a759/src/benchmark.cc#L619-L633
(which is seemingly the only consumer of `FindBenchmarksInternal()` -> `BenchmarkFamilies::FindBenchmarks()` that we are talking about here) i think the current answer should be **no**.

Unless of course the consumer calls `ClearRegisteredBenchmarks()` in the middle of benchmarking, but then that probably won't the only thing that will break. I'm not sure there are any current locks that could prevent that.

src/statistics.cc
```diff
@@ -0,0 +1,177 @@
+// Copyright 2016 Ismael Jimenez Martinez. All rights reserved.
```
LebedevRI: I'm not sure about this copyright header. It was copied from `src/complexity.cc`
But i'll just add myself here...

src/benchmark_register.cc
LebedevRI: Hm, correction, https://github.com/google/benchmark/blob/d70417994a3c845c49c4443e92b26a52b320a759/src/benchmark_register.cc#L86
`BenchmarkFamilies::mutex_` does partially help (but no thread safety analysis annotations are used), but it will still not help against the call to `ClearRegisteredBenchmarks()` right after `FindBenchmarks()`. I'm not sure it can happen in reality though.
But i guess `Benchmark::Instance::benchmark` could be a `std::shared_ptr`, and then it would all be fine, with *some* performance penalty.

dominichamon: let's try to avoid shared_ptr.

as you've pointed out, it should be fine and i'd rather avoid the copy if we can as you've done.

src/statistics.cc
dominichamon: https://github.com/google/benchmark/blob/e8fc2a2b8ccd33fefdd7f95e1a2aabee0788c4d6/src/benchmark.cc#L235

i missed this cast.

dominichamon:
thank you!

LebedevRI:
Thank _you_ very much!

PR #430, opened 2017-07-31
added meta data to benchmarks

DerThorsten:
With this little hack one can add meta-data to a benchmark
and store the meta data also to json.

```
static void BM_memcpy(benchmark::State& state) {
  char* src = new char[state.range(0)];
  char* dst = new char[state.range(0)];
  memset(src, 'x', state.range(0));
  while (state.KeepRunning())
    memcpy(dst, src, state.range(0));
  state.SetBytesProcessed(int64_t(state.iterations()) *
                          int64_t(state.range(0)));
  delete[] src;
  delete[] dst;
}
BENCHMARK(BM_memcpy)->Range(8, 8<<10)->AddMetaData("key1","val1")->AddMetaData("foo", "bar")
```

If you consider to merge this PR I can do some cleanup

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -570,6 +570,9 @@ class State {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(State);
 };
 
+
+
+
 namespace internal {
 
 typedef void(Function)(State&);
@@ -583,7 +586,7 @@ typedef void(Function)(State&);
 class Benchmark {
  public:
   virtual ~Benchmark();
-
+  typedef std::map<std::string, std::string> MetaDataType;
   // Note: the following methods all return "this" so that multiple
   // method calls can be chained together in one expression.
 
@@ -705,6 +708,10 @@ class Benchmark {
   // Run one instance of this benchmark concurrently in t threads.
   Benchmark* Threads(int t);
 
+
+  Benchmark* AddMetaData(const std::string & key, const std::string & val);
+  const MetaDataType & MetaData()const;
+
   // Pick a set of values T from [min_threads,max_threads].
   // min_threads and max_threads are always included in T.  Run this
   // benchmark once for each value in T.  The benchmark run for a
@@ -759,7 +766,7 @@ class Benchmark {
   BigO complexity_;
   BigOFunc* complexity_lambda_;
   std::vector<int> thread_counts_;
-
+  MetaDataType meta_data_;
   Benchmark& operator=(Benchmark const&);
 };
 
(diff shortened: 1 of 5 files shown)


DerThorsten:
I signed it!

src/json_reporter.cc
```diff
@@ -163,10 +163,29 @@ void JSONReporter::PrintRunData(Run const& run) {
...
+      ++c;
+    }
+    out<<indent<<"}";
+  }
   for(auto &c : run.counters) {
     out << ",\n"
-        << indent
-        << FormatKV(c.first, c.second);
+        << indent 
+        << FormatKV(c.first, RoundDouble(c.second));
```
LebedevRI: This change seems wrong, probably rebase defect.

include/benchmark/benchmark.h
```diff
@@ -570,6 +570,9 @@ class State {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(State);
 };
 
+
```
LebedevRI: Seems unnessesary

include/benchmark/benchmark.h
```diff
@@ -583,7 +586,7 @@ typedef void(Function)(State&);
 class Benchmark {
  public:
   virtual ~Benchmark();
-
+  typedef std::map<std::string, std::string> MetaDataType;
```
LebedevRI: Not commenting on the typedef itself, but the placement seems not ideal, maybe move down to getters/setters?

EricWF:
I would like to see a lot more rational for this change. Plenty of users have proposed adding various meta-data to the interface, but almost never in the same way, and never in a way that's as helpful to other users as intended.

The first thing this patch needs is documentation; both examples and function specifications. I'm not even really sure what you're trying to do here that `SetLabel` or `RegisterBenchmarkName` couldn't possibly do.

My ideal implementation for storing meta-data would be a `std::any` like type, but since we can't depend on C++17 `std::any` is out of the question. However a lesser implementation in the library might be acceptable.

egpbos:
With a label and/or name you still have only one metadata field. Being able to specify multiple will create far cleaner output in case you want to reuse the metadata, e.g. for aggregating over certain metadata fields. This would significantly ease analysis of large amounts of benchmarks over parameter matrices. I would really like to see this feature included!

DerThorsten:
I think an ``std::any`` is overkill.  Since we store the benchmarks in a string-ish fashion one would need to convert all metadata to a string anyhow.

vincecr0ft:
Whilst I'm appreciative of the time I've just spent brushing up on my data wrangling. Making it easier to plot the outputs of these parameter matrices would be nice...

DerThorsten:
I'll implement the missing CSV reporter these days .

LebedevRI:
Can you please not push every one-line change as a new commit?
Really spams notifications.

DerThorsten:
Sure, sry for that

EricWF:
I believe this PR/Issue would be fixed if we adopted the JSON interface I'm proposing. See [this test case](https://github.com/efcs/benchmark/blob/overhaul-reporting/test/new_api_test.cc#L22) for an example.

DerThorsten:
@EricWF  I agree

PR #431, opened 2017-07-31
reporter_output_test: json: iterations is int, not float

LebedevRI:
May be relevant for flakiness of win builds

Noted by @KindDragon in https://github.com/google/benchmark/pull/426/files#r130466898


EricWF:
LGTM. Thanks for the patch.

PR #435, opened 2017-08-10
compare_bench.py: fixup benchmark_options.

LebedevRI:
@EricWF 

https://github.com/google/benchmark/commit/2373382284918fda13f726aefd6e2f700784797f
reworked parsing, and introduced a regression
in handling of the optional options that
should be passed to both of the benchmarks.

Now, unless the *first* optional argument starts with
'-', it would just complain about that argument:
	Unrecognized positional argument arguments: '['q']'
which is wrong. However if some dummy arg like '-q' was
passed first, it would then happily passthrough them all...

This commit fixes benchmark_options behavior, by
restoring original passthrough behavior for all
the optional positional arguments.

### DIFF
--- a/tools/compare_bench.py
+++ b/tools/compare_bench.py
@@ -39,21 +39,20 @@ def main():
     parser.add_argument(
         'test2', metavar='test2', type=str, nargs=1,
         help='A benchmark executable or JSON output file')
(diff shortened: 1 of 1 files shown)


LebedevRI:
As requested, ping :)

PR #439, opened 2017-08-23
CMake: fix subproject regression

LebedevRI:
Google Benchmark is a library. It has nice CMake build system. CMake provides a way to build projects (ideally, which have cmake build system themselves) as part of some bigger master build. This PR fixes a few bugs related to that possibility.

1. `set(CACHE FORCE)` (and `set(... PARENT_SCOPE)` in top-level src dir) shall be used with **great** care. 
    Google Benchmark could be being built as a subproject. The parent project could have already specified it's variables. By storing our variables into cache, we could be potentially overriding the values which are already set by the master project, and possibly breaking everything.
    Instead, if we really need `CACHE`, we should prefix our variables with some prefix, e.g. `BENCHMARK`.
2. `set(... CACHE INTERNAL ... PARENT_SCOPE)` makes no sense, and as per [documentation](https://cmake.org/cmake/help/v3.0/command/set.html), is invalid: ```PARENT_SCOPE cannot be combined with CACHE.```. And since as the previous point states, `CACHE` shall not be abused, i have fixed it to **only** specify `PARENT_SCOPE`. Since these `set()` are in a function (which results in a new scope), it is the correct fix i believe.

Without this patch, v1.2.0 regressed (broke) coverage build of my project (when the Google Benchmark is built as a subproject). I did test this patch, and it appears to resolve that regression.


dominichamon:
$ cmake ..
-- Found Git: /usr/bin/git (found version "2.14.1.342.g6490525c54-goog") 
-- git Version: v1.2.0-28c0e9f9
CMake Error at CMakeLists.txt:28 (string):
  string sub-command REGEX, mode MATCH needs at least 5 arguments total to
  command.

followed by many more error statements

LebedevRI:
Oh, it must then be 
```
  set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_DEBUG}" CACHE STRING
    "Flags used by the C++ compiler during coverage builds."
    FORCE)
```
That is so bad :/

LebedevRI:
Alright, i think it is better now.
It builds, ci's are happy, and appears to fix the Problem i was hitting.
PR description updated :)

EricWF:
> CMake provides a way to build projects (ideally, which have cmake build system themselves) as part of some bigger master build.

Yes, it does. It's called [`ExternalProject`](https://cmake.org/cmake/help/v3.8/module/ExternalProject.html#externalproject). You should be using that, not configuring it directly as a subdirectory.

LebedevRI:
> Yes, it does. It's called ExternalProject. 

Yes...

> You should be using that, not configuring it directly as a subdirectory.

You are wrong.

If i use `ExternalProject` only, then the entire library is built as part of the `cmake ../` execution. This is rather suboptimal. But if i first use `ExternalProject` to *prepare* cmake, and then do `add_subdirectory()`, then it is built as part of my buildsystem.

It just works for googletest, and is the recommended approach: https://github.com/google/googletest/tree/master/googletest#incorporating-into-an-existing-cmake-project
```
This is done by making the GoogleTest source code available to the main build and adding it using CMake's add_subdirectory() command.
```
It just worked for googlebenchmark until the last release.

**This is a bug.**

dominichamon:
@EricWF this is a workaround to balance the external project and subdirectory approaches, and it does work well.

given this doesn't break anything and improves the experience, and aligns the project with googletest, i'm happy.

LebedevRI:
@dominichamon thank you!

PR #452, opened 2017-10-09
Always use inline asm DoNotOptimize with clang.

mwinterb:
clang-cl masquerades as MSVC but not GCC, so it was using the
MSVC-compatible definitions of DoNotOptimize and ClobberMemory.
Presumably, it's better in general to use the targeted assembly for
this functionality (the codegen is different), but the specific issue
is that clang-cl deprecates the usage of _ReadWriteBarrier, and this
gets rid of that warning.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -295,6 +295,7 @@ BENCHMARK_UNUSED static int stream_init_anchor = InitializeStreams();
 # define BENCHMARK_HAS_NO_INLINE_ASSEMBLY
 #endif
 
+
 // The DoNotOptimize(...) function can be used to prevent a value or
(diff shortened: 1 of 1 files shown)


EricWF:
This LGTM, but could you push a dummy commit to re-trigger the Appveyor bot? I think the failure was spurious.

mwinterb:
Well, this is bothersome.

dominichamon:
It may be unrelated to your change. one second while i check.

dominichamon:
Yes. this is happening already. i'll merge this now (and fix the appveyor test i guess).

dominichamon:
thank you!

mwinterb:
Thank you.

PR #453, opened 2017-10-09
Convert negative 0 to 0 in console reporter.

dominichamon:
Windows builds will occasionally show -0 for CPU times when the time is
very short. This is likely related to the weaknesses in GetThreadTimes.

As a workaround, check for small negative numbers and ensure they're 0
in the console reporter.

### DIFF
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -127,8 +127,19 @@ void ConsoleReporter::PrintRunData(const Run& result) {
         StrCat(" ", HumanReadableNumber(result.items_per_second), " items/s");
   }
 
-  const double real_time = result.GetAdjustedRealTime();
-  const double cpu_time = result.GetAdjustedCPUTime();
+  double real_time = result.GetAdjustedRealTime();
+  if (real_time < 0) {
+    if (std::fabs(real_time) < std::numeric_limits<double>::epsilon()) {
+      real_time = 0.0;
+    }
+  }
+
+  double cpu_time = result.GetAdjustedCPUTime();
+  if (cpu_time < 0) {
+    if (std::fabs(cpu_time) < std::numeric_limits<double>::epsilon()) {
+      cpu_time = 0.0;
+    }
+  }
 
   if (result.report_big_o) {
     std::string big_o = GetBigOString(result.complexity);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,11 @@ if (BENCHMARK_BUILD_32_BITS)
   add_required_cxx_compiler_flag(-m32)
 endif()
 
+
+if (BENCHMARK_MINGW)
+  add_definitions(-D__USE_MINGW_ANSI_STDIO=1)
+endif (BENCHMARK_MINGW)
+
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   # Turn compiler warnings up to 11
   string(REGEX REPLACE "[-/]W[1-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
@@ -51,10 +56,6 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     add_cxx_compiler_flag(-EHs-)
     add_cxx_compiler_flag(-EHa-)
   endif()
-
-  if (BENCHMARK_MINGW)
-    add_definitions(-D__USE_MINGW_ANSI_STDIO=1)
-  endif (BENCHMARK_MINGW)
 
   # Link time optimisation
   if (BENCHMARK_ENABLE_LTO)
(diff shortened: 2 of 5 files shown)


src/console_reporter.cc
```diff
@@ -127,8 +127,19 @@ void ConsoleReporter::PrintRunData(const Run& result) {
         StrCat(" ", HumanReadableNumber(result.items_per_second), " items/s");
   }
 
-  const double real_time = result.GetAdjustedRealTime();
-  const double cpu_time = result.GetAdjustedCPUTime();
+  double real_time = result.GetAdjustedRealTime();
```
EricWF: Something like this seems cleaner:

```c++
auto clamp = [](double V) {
  if (V < 0 && std::fabs(V) < std::numeric_limits<double>::epsilon())
    return 0.0;
  return V;
};
const double real_time = clamp(result.GetAdjustedRealTime());
const double cpu_time = clamp(result.GetAdjustedCPUTime());
```

EricWF: *Actually, since this should probably apply to all reporters, maybe the clamping should be done in `GetAdjustedCPUTime()` et al.

What do you think?

dominichamon: I think so, yes.

Actually this clamping doesn't work because `(-0 == 0) == true`, which suggests i can do something like:

```c++
auto clamp = [](double v) {
  if (std::fabs(v) < std::numeric_limits<double>::epsilon()) return 0.0;
  return v;
};
```

instead.

EricWF: Yeah, that seems to work.

pleroy:
It seems to me that this PR is papering over a problem that we don't really understand.

Negative zeroes don't show up for random reasons.  In particular, they _never_ arise out of subtractions.  The only ways that you can construct a negative zero are by (1) multiplying a positive zero by a negative number or (2) multiplying two numbers of opposite signs in a way that results in an underflow.

Looking at the code it's unclear to me how/where this could happen.  I would be more comfortable if we were to pepper the code with checks or traces to understand exactly where the negative zero comes from (this is easy to do with `std::signbit`).

dominichamon:
That's fair, and it is. I thought at first it was an artifact of the
integer printing rules that were changed recently such that negative zero
is no longer printed as zero[1], but we print these as floats[2] so i'm
extra confused.

I think a PR with signbit checks around and logging that will show up in
appveyor is a great idea. I'll put something together to see if i can
repro. Unfortunately i have no windows machine to test it directly.

1.
https://social.msdn.microsoft.com/Forums/vstudio/en-US/a3b18c01-00d7-44ea-b4ef-61e44234ee94/negative-zero-valid?forum=vcgeneral
2.
https://github.com/google/benchmark/blob/master/src/console_reporter.cc#L142

Dominic Hamon | Google

EricWF:
I did a bunch of investigation, and I found one important bit of information: We're not actually dealing with negative zero. The values that are getting through are negative numbers larger than `epsilon()` but smaller than `round_error()`.

I'm not sure exactly what's causing them yet, but I suspect the measured CPU times on MinGW are close to zero they sometimes result in negative numbers, but honestly I have no clue.

However we should be able to target a workaround to MinGW only, where we check `std::signbit` and return `0.0` whenever we accidentally get a negative time.

EricWF:
@dominichamon To make this easier on you, it might make sense to move the MinGW builders to the top of the `appveyor.yml` file so they run first.

pleroy:
I have the feeling, like @EricWF, that this issue is somehow related to MinGW.  I am using VS2015 and I don't remember ever getting negative numbers, even when benchmarking computations in the 5-10 ns range.

EricWF:
My educated guess is that MinGW's g++ uses a different floating point model than MSVC does normally, 
and because we're using the MSVC runtimes (in particular to provide stdio). Likely related to this bug:

https://sourceforge.net/p/mingw-w64/bugs/520/

It seems that the GNU printf behavior we expect (ie rounding to positive zero) doesn't occur with the MSVC stdio implementation.

pleroy:
@EricWF: You mention "the GNU printf behavior we expect (ie rounding to positive zero)".  Well, your expectations are bad and you should feel bad.  Note 276 in section 7.21.6.1 of the latest C standard, ISO/IEC 9899:2011, says "The results of all floating conversions of a negative zero, and of negative values that round to zero, include a minus sign".

It's also amusing that the MinGW page that you link mentions "The vc6.0 msvcrt.dll that MinGW-w64 targets".  VC++ 6.0 was released in 1989 (not a typo); it's hard to believe that it would be compliant with modern C/C++ standards in all cases.

EricWF:
@pleroy Point granted; and I feel bad. But since when was *GNU* behavior

EricWF:
Woops, didn't mean to close this. Sorry.

dominichamon:
looks like @EricWF has an alternative approach in #475

PR #454, opened 2017-10-09
Add C++11 Ranged For loop alternative to KeepRunning

EricWF:
As pointed out by @astrelni and @dominichamon, the KeepRunning
loop requires a bunch of memory loads and stores every iterations,
which affects the measurements.

The main reason for these additional loads and stores is that the
State object is passed in by reference, making its contents externally
visible memory, and the compiler doesn't know it hasn't been changed
by non-visible code.

It's also possible the large size of the State struct is hindering
optimizations.

This patch allows the `State` object to be iterated over using
a range-based for loop. Example:

```c++
void BM_Foo(benchmark::State& state) {
	for (auto _ : state) {
		[...]
	}
}
```

This formulation is much more efficient, because the variable counting
the loop index is stored in the iterator produced by `State::begin()`,
which itself is stored in function-local memory and therefore not accessible
by code outside of the function. Therefore the compiler knows the iterator
hasn't been changed every iteration.

This initial patch and idea was from Alex Strelnikov.

### DIFF
--- a/test/basic_test.cc
+++ b/test/basic_test.cc
@@ -96,4 +96,23 @@ void BM_empty_stop_start(benchmark::State& state) {
 BENCHMARK(BM_empty_stop_start);
 BENCHMARK(BM_empty_stop_start)->ThreadPerCpu();
 
+
+void BM_KeepRunning(benchmark::State& state) {
+  size_t iter_count = 0;
+  while (state.KeepRunning()) {
+    ++iter_count;
+  }
+  assert(iter_count == state.max_iterations);
+}
+BENCHMARK(BM_KeepRunning);
+
+void BM_RangedFor(benchmark::State& state) {
+  size_t iter_count = 0;
+  for (auto _ : state) {
+    ++iter_count;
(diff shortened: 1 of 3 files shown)


test/basic_test.cc
```diff
@@ -96,4 +96,23 @@ void BM_empty_stop_start(benchmark::State& state) {
 BENCHMARK(BM_empty_stop_start);
 BENCHMARK(BM_empty_stop_start)->ThreadPerCpu();
 
+
```
dominichamon: should all the tests (but this one) switch to the range loop to show it's the default?

EricWF: Could I do that as a separate refactoring, and use it in my assignment? :-D

EricWF:
Failures due to unrelated `-0 ns` output failure on MinGW. This patch should be good to go.

test/basic_test.cc
dominichamon: ha!

yes, that's fine.

PR #457, opened 2017-10-13
Avoid implicit float to double conversion

Algunenano:
Fixing build error triggered by `-Wdouble-promotion`:
```
Scanning dependencies of target benchmark
[  1%] Building CXX object src/CMakeFiles/benchmark.dir/benchmark.cc.o
/home/raul/dev/public/benchmark/src/benchmark.cc: In function 'void benchmark::internal::{anonymous}::RunBenchmarks(const std::vector<benchmark::internal::Benchmark::Instance>&, benchmark::BenchmarkReporter*, benchmark::BenchmarkReporter*)'
/home/raul/dev/public/benchmark/src/benchmark.cc:494:43: error: implicit conversion from 'float' to 'double' to match other operand of binary expression [-Werror=double-promotion]
   context.mhz_per_cpu = CyclesPerSecond() / 1000000.0f;
                         ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,6 +10,7 @@
 
 Albert Pretorius <pretoalb@gmail.com>
 Arne Beer <arne@twobeer.de>
+Carto
 Christopher Seymour <chris.j.seymour@hotmail.com>
 David Coeurjolly <david.coeurjolly@liris.cnrs.fr>
 Dirac Research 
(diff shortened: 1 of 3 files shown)


dominichamon:
thanks!

PR #459, opened 2017-10-17
Refactor most usages of KeepRunning to use the perfered ranged-for.

EricWF:
Recently the library added a new ranged-for variant of the KeepRunning
loop that is much faster. For this reason it should be preferred in all
new code.

Because a library, its documentation, and its tests should all embody
the best practices of using the library, this patch changes all but a
few usages of `KeepRunning()` into `for (auto _ : state)`.

The remaining usages in the tests and documentation persist only
to document and test behavior that is different between the two formulations.

Also note that because the range-for loop requires C++11, the KeepRunning
variant has not been deprecated at this time.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -18,7 +18,7 @@
 // Define a function that executes the code to be measured a
 // specified number of times:
 static void BM_StringCreation(benchmark::State& state) {
-  while (state.KeepRunning())
+  for (auto _ : state)
     std::string empty_string;
 }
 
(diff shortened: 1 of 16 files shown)


include/benchmark/benchmark.h
```diff
@@ -611,7 +616,7 @@ struct State::StateIterator {
 
   BENCHMARK_ALWAYS_INLINE
   explicit StateIterator(State* st)
-      : cached_(st->max_iterations), parent_(st) {}
+      : cached_(st->error_occurred_ ? 0 : st->max_iterations), parent_(st) {}
```
dominichamon: should this be a separate PR to keep the functional changes separate from the refactor?

EricWF: Yep. I'll commit separately.

dominichamon:
when tests pass, go ahead and merge it please :)

EricWF:
Merging pre-test completion. Appveyor only failed on the non-functioning MinGW tests, and Travis is hung waiting on OS X test environments to become available, but I tested this patch on OS X personally.

PR #460, opened 2017-10-17
Improve KeepRunning loop performance to be similar to the range-based…

EricWF:
… for.

This patch improves the performance of the KeepRunning loop in two ways:

(A) it removes the dependency on the max_iterations variable, preventing
it from being loaded every iteration.

(B) it loops to zero, instead of to an upper bound. This allows a single
decrement instruction to be used instead of a arithmetic op followed by a
comparison.


EricWF:
@dominichamon Merging w/o review due to deadline. :-S sorry.

PR #463, opened 2017-10-19
Add option to install benchmark

Yangqing:
This is similar to the setting in gtest - for projects that directly embed it, one would not need to install gbenchmark.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmar
 option(BENCHMARK_ENABLE_LTO "Enable link time optimisation of the benchmark library." OFF)
 option(BENCHMARK_USE_LIBCXX "Build and test using libc++ as the standard library." OFF)
 option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library." OFF)
-option(INSTALL_BENCHMARK "Enable installation of benchmark. (Projects embedding benchmark may want to turn this OFF.)" ON)
+option(BENCHMARK_ENABLE_INSTALL "Enable installation of benchmark. (Projects embedding benchmark may want to turn this OFF.)" ON)
 
 # Make sure we can import out CMake functions
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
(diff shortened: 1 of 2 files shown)


dominichamon:
you can already install it.

https://github.com/google/benchmark/blob/master/src/CMakeLists.txt#L58

dominichamon:
Gah, sorry. moving too fast.

Can you explain a bit more why this would be necessary? Does this cause an outer project's install step to also install benchmark?

Yangqing:
Ah, so basically, it is similar to the use case of googletest - for some projects (github.com/caffe2/caffe2 to be specific), we want to use test and benchmark in our build when we do USE_TEST, and

(1) we do not want to burden the user to install googletest and benchmark.
(2) we don't really want to install googletest and benchmark, as we will just link them statically.

In these cases having an option INSTALL_{GTEST,BENCHMARK} comes really handy.

By the way, I am a bit confused by the Appveyor failure - this should have introduced any behavior change with default settings, should I ignore it?

Thanks - YQ

dominichamon:
yes.. our mingw failure is flaky. we're working on it.

CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmark library." ON)
 option(BENCHMARK_ENABLE_LTO "Enable link time optimisation of the benchmark library." OFF)
 option(BENCHMARK_USE_LIBCXX "Build and test using libc++ as the standard library." OFF)
-option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library" OFF)
+option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library." OFF)
+option(INSTALL_BENCHMARK "Enable installation of benchmark. (Projects embedding benchmark may want to turn this OFF.)" ON)
```
dominichamon: `s/INSTALL_BENCHMARK/BENCHMARK_ENABLE_INSTALL/`

for consistency and so it's clear that it's not doing the actual installation.

Yangqing: Thanks - will do.

Yangqing:
Done - PTAL.

dominichamon:
thanks!

leokoppel:
Is there some way this is better than using `EXCLUDE_FROM_ALL` in the enclosing project, as suggested here? https://github.com/google/googletest/issues/868

PR #465, opened 2017-10-30
 Improve BM_SetInsert example

### LINKED ISSUE #464: BM_SetInsert example is incorrect
I got unexpected results while writing a benchmark based on the `BM_SetInsert` example, "for measuring the speed of set insertion". Even when the `insert` operation is removed from the loop, time grows with set size:

```cpp

std::set<int> ConstructRandomSet(int size) {
    std::set<int> s;
    for (int i = 0; i < size; ++i) s.insert(i);
    return s;
}

static void BM_SetNoop(benchmark::State &state) {
    for (auto _ : state) {
        state.PauseTiming();
        std::set<int> data = ConstructRandomSet(state.range(0));
        state.ResumeTiming();

        // Nothing here
    }
}

BENCHMARK(BM_SetNoop)->Range(1 << 14, 1 << 22)->Complexity();
BENCHMARK_MAIN()
```
```
-------------------------------------------------------------------------------------
Benchmark                                              Time           CPU Iterations
-------------------------------------------------------------------------------------
BM_SetNoop/16384                                  230085 ns     230054 ns       3085
BM_SetNoop/32768                                  486330 ns     486272 ns       1354
BM_SetNoop/262144                                7016412 ns    7016211 ns        100
BM_SetNoop/2097152                              49388757 ns   49388656 ns         14
BM_SetNoop/4194304                              98787818 ns   98784961 ns          7
BM_SetNoop_BigO                                    23.56 N      23.56 N 
BM_SetNoop_RMS                                         1 %          1 % 
```

I think the destructor `~set()` is being timed. I will submit a PR.

leokoppel:
Fix #464 

Move declaration of the container outside the timing loop, so the destructor is not timed.

The catch is the test now takes an unreasonably long time. `ConstructRandomSet` takes much longer than `insert`, and only `insert` is used to determine the number of iterations.

There are a few ways to speed it up:
  - Increase the `Ranges()` parameters
  - Cache `ConstructRandomSet()` result (it's not random anyway), and just copy it every iteration

Any feedback is appreciated.

### DIFF
--- a/test/benchmark_test.cc
+++ b/test/benchmark_test.cc
@@ -83,16 +83,17 @@ BENCHMARK(BM_CalculatePi)->ThreadPerCpu();
 
 static void BM_SetInsert(benchmark::State& state) {
   std::set<int> data;
+  std::set<int> orig = ConstructRandomSet(state.range(0));
   for (auto _ : state) {
     state.PauseTiming();
-    data = ConstructRandomSet(state.range(0));
+    data = orig;
     state.ResumeTiming();
     for (int j = 0; j < state.range(1); ++j) data.insert(rand());
   }
   state.SetItemsProcessed(state.iterations() * state.range(1));
   state.SetBytesProcessed(state.iterations() * state.range(1) * sizeof(int));
 }
-BENCHMARK(BM_SetInsert)->Ranges({{1 << 10, 8 << 10}, {1, 10}});
+BENCHMARK(BM_SetInsert)->Ranges({{1 << 12, 1 << 15}, {100, 500}});
 
 template <typename Container,
           typename ValueType = typename Container::value_type>
--- a/test/benchmark_test.cc
+++ b/test/benchmark_test.cc
@@ -42,7 +42,7 @@ double CalculatePi(int depth) {
 
 std::set<int> ConstructRandomSet(int size) {
   std::set<int> s;
-  for (int i = 0; i < size; ++i) s.insert(i);
+  for (int i = 0; i < size; ++i) s.insert(s.end(), i);
   return s;
 }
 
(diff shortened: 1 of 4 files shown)


test/benchmark_test.cc
```diff
@@ -82,16 +82,18 @@ BENCHMARK(BM_CalculatePi)->ThreadRange(1, 32);
 BENCHMARK(BM_CalculatePi)->ThreadPerCpu();
 
 static void BM_SetInsert(benchmark::State& state) {
+  std::set<int> data;
+  std::set<int> orig = ConstructRandomSet(state.range(0));
```
dominichamon: this changes things a bit, as it means the data will be the same for every iteration. this may drop variance within a run but increase it across runs.

dominichamon: oh. `ConstructRandomSet` is a lie. never mind then.

dominichamon:
the runtime of `test/benchmark_test` on my machine:

before: 2m36.588s
after: 2m12.475s

I don't see this slowdown you've suggested so this looks good to merge. any last thoughts?

leokoppel:
Yes, the slowdown occurs without the two "optimizations" I list. Running the BM_SetInsert variant in the README:

before: 40s
after: 10 minutes later it's on the third item, `8192/1`

Should I edit the ranges in the README as well? I don't want to make it over-complicated but it should be possible to run the readme code.

dominichamon:
I'd like to keep everything consistent between the README and the test code. Having the examples at least be 'best practice' is a good idea.

leokoppel:
OK, I made them consistent. Runtime of `benchmark_test --benchmark_filter=BM_SetInsert`:

before: 0m33.991s
after: 0m46.226s

dominichamon:
thanks!

PR #475, opened 2017-11-06
Fix #382 - MinGW often reports negative CPU times.

### LINKED ISSUE #382: reporter output test: MinGW intermittently returns negative times for BM_Repeat, test fails
Error message is (note the "-0ns"):
```
C:\projects\benchmark\test\output_test_helper.cc:93: CheckCase: Check `remaining_output.eof() == false' failed. End of output reached before match for regex "^BM_Repeat/repeats:3 %console_report$" was found
    actual regex string "^BM_Repeat/repeats:3 [ ]*[0-9]{1,5} ns [ ]*[0-9]{1,5} ns [ ]*[0-9]+$"
    started matching near: BM_Repeat/repeats:3                            4 ns         -0 ns    1000000
```
For an intermittency example, see #350 or #381 .

EricWF:
When stopping a timer, the current time is subtracted
from the start time. However, when the times are identical,
or sufficiently close together, the subtraction can result
in a negative number.

For some reason MinGW is the only platform where this problem
manifests. I suspect it's due to MinGW specific behavior in either
the CPU timing code, floating point model, or printf formatting.

Either way, the fix for MinGW should be correct across all platforms.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -175,7 +175,9 @@ class ThreadTimer {
     CHECK(running_);
     running_ = false;
     real_time_used_ += ChronoClockNow() - start_real_time_;
-    cpu_time_used_ += ThreadCPUUsage() - start_cpu_time_;
+    // Floating point error can result in the subtraction producing a negative
+    // time. Guard against that.
+    cpu_time_used_ += std::max<double>(ThreadCPUUsage() - start_cpu_time_, 0);
   }
 
   // Called by each thread


EricWF:
The Appveyor failure is separate from, and unrelated to, this change! Hurray!

It looks like the new failure is caused because the regex used only accepts times of 5 digits or less, but for some reason the last run of `BM_Counter2_Tabular` with 16 threads produces a 6 digit time in nanoseconds.

That should probably be investigated and dealt with in a separate commit.

src/benchmark.cc
```diff
@@ -175,7 +175,9 @@ class ThreadTimer {
     CHECK(running_);
     running_ = false;
     real_time_used_ += ChronoClockNow() - start_real_time_;
-    cpu_time_used_ += ThreadCPUUsage() - start_cpu_time_;
+    // Floating point error can result in the subtraction producing a negative
+    // time. Guard against that.
```
pleroy: This comment makes me [sad](https://twitter.com/stephentyrone/status/849398027089580032).  I think it should say something like 

> Monotonic clocks are hard, MinGW doesn't manage to get them right, so sometimes the subtraction will yield a negative value.  Guard against that.

EricWF: @pleroy I don't think that's whats happening here.

First, the `ThreadCPUUsage` is using `GetProcessTimes` from the Windows Kernel library; which is not MinGW rolling it's own clock AFAIK.

Second, the negative numbers don't come from the clock itself. When the problem occurs when two consecutive calls to `GetProcessTimes` return the *exact same value*. (See [here](https://ci.appveyor.com/project/EricWF/appveyor-test/build/job/px9hj28swawmd8ng))

Therefore the problem is caused by the FP operations we perform. Whether GCC's FP model in MinGW is partially to blame is up for debate. But it's certainly not the fault of the clock.

eggrobin: @EricWF, you're right, something is afoot with [the floating point arithmetic](https://twitter.com/whitequark/status/670022688526589952). Unfortunately, while it explains a nonzero difference, it does not explain its negative sign.

The issue is not that `d1` and `d2` (from [your traces](https://ci.appveyor.com/project/EricWF/appveyor-test/build/job/xdu18ob9tx44rhj9)) are actually equal, even though, when you print their bit representation as IEEE 754 binary64 (`double`), you get the same thing (yes, this is terrible). If they were equal, any remotely sane floating-point arithmetic would give you a difference of 0.

What is instead happening is that you are seeing x87's 80-bit extended precision format hiding beneath those `double`s.

The following analysis was performed using [my soft-float *Mathematica* package](https://github.com/mockingbirdnest/Principia/blob/273266344021f8d0360c5376a64d75adc0a0d018/mathematica/floating_point.wl), for easy inspection of representations and selection of floating point formats.
Representations below are shown in the form
`sign bit|exponent bits|significand bits;`,
non-representable numbers are shown extending the significand,
`sign bit|exponent bits|significand bits;more bits`.
A caveat on the x87 representations, I do not print [the infamous redundant bit 63](https://en.wikipedia.org/wiki/Extended_precision#x86_extended_precision_format); those will thus be 79 bits long.

The number printed for `d1` and `d2` (which I shall call `d`, to avoid confusion with the actual underlying values of `d1` and `d2`) is 1/64, binary64 representation
`0|01111111001|0000000000000000000000000000000000000000000000000000;`
The number printed for d3 is 835/2048, binary64 representation
`1|01111000010|1010000110000000000000000000000000000000000000000000;`
The ULP distance between `d` and `d+d3` is 835/4096, which less 1, i.e., it was not the result of a subtraction of two nearby binary64 floating-point numbers. Since this distance is not an integer, `d+d3` is not representable in binary64:
`0|01111111000|1111111111111111111111111111111111111111111111111111;10010111101`

Note that we expect the difference to be exact by Sterbenz's lemma, as @pleroy said; indeed, `d3` uses very few of its significand bits. It is likely that its value is actually a double. 

Let us look at `d+d3` in x87 extended precision: it is 
`0|011111111111000|111111111111111111111111111111111111111111111111111110010111101;`
exactly.
This sum takes the entire significand to represent, which is an unmistakable clue that we are indeed dealing with x87 extended format.

That is all very well and good, and explains why the difference of numbers that print equal is nonzero: printing goes through memory rather than staying on the x87 stack, so extended format gets rounded to binary64.
The question remains of why it is negative: even in x87 arithmetic, if `d2-d1 < 0`, `d2 < d1`.
This means that the output of [`ProcessCPUUsage`](https://github.com/efcs/appveyor_test/blob/810cb0f98eeb5d87f50152b0633121ff68fbf8d1/test.cpp#L30) is not monotonic. Further, since we are getting extended format woes, it means that the output of `ProcessCPUUsage` is not binary64.

The latter is easily explained by the presence of arithmetic in `MakeTime`:
```cpp
  return (static_cast<double>(kernel.QuadPart) +
          static_cast<double>(user.QuadPart)) *
         1e-7;
```
while the casts will probably really return `double`, the rest of the arithmetic gets done in x87.

What about the nonmonotonicity?
Composition of monotonic functions is monotonic;
in any rounding mode, rounding is monotonic;
a sum is correctly rounded;
therefore if the exact sum is monotonic, the rounded sum will be monotonic;
multiplication by a constant is monotonic;
a product is correctly rounded;
therefore if the exact sum of `static_cast<double>(kernel.QuadPart)` and `static_cast<double>(user.QuadPart)` is monotonic, the result of `MakeTime` will be monotonic.

Since both are positive, if both `kernel.QuadPart` and  `user.QuadPart` are monotonic, their casts to double and their sum will be too.

It follows that the root issue is nonmonotonicity of either kernel time or user time returned by `GetProcessTimes`, in sufficiently insignificant bits that it only makes it out of the significance-killing `MakeTime` when using x87 arithmetic.

It should be possible to verify that (isolating the issue from the hard-to-trace x87 floating-point) by looking at the actual integers in `MakeTime`.

EricWF: Thank you for the in-depth and excellent analysis.

> What is instead happening is that you are seeing x87's 80-bit extended precision format hiding beneath those doubles.

Agreed.

> It should be possible to verify that (isolating the issue from the hard-to-trace x87 floating-point) by looking at the actual integers in MakeTime.

[The trace](https://ci.appveyor.com/project/EricWF/appveyor-test/build/job/xdu18ob9tx44rhj9) previously mentioned *does* print out the integer values as suggested. It verifies the opposite. 

```
Kernel: 0
User: 156250
Kernel: 0
User: 156250
d1: 00111111 10010000 00000000 00000000 00000000 00000000 00000000 00000000 
d2: 00111111 10010000 00000000 00000000 00000000 00000000 00000000 00000000 
d3: 10111100 00101010 00011000 00000000 00000000 00000000 00000000 00000000 
```

My less technical guess as to what's happening is this: One of the two operands to the subtraction has been written to memory truncating it, while the other has not, and therefore still has extended precision. 

Within the library, and not my example, this makes a lot more sense. Since [`start_cpu_time_`](https://github.com/google/benchmark/blob/master/src/benchmark.cc#L180) is an lvalue that would have to be loaded from memory. Whereas the return value of `ThreadCPUUsage()`, calling conventions permitting, could still have extended precision. 

I'm nowhere close to as knowledgeable about FP conventions. Does this sound reasonable, or am I out to lunch?

EricWF: Also, I should mention that I couldn't reproduce the negative result when I wrote both `d1` and `d2` through volatile, forcing truncation.

pleroy: @EricWF: This makes total sense.  It's a constant pain with compilers using x87 arithmetic to implement `double`, as they spill to memory in random places.

I believe that the fix is to force flushing to `double` at the end of `MakeTime` using a `volatile` variable.  (Oh, and remove the max with 0 ;-)

EricWF: I think the real fix here is to store the integer representation instead of converting to double, at least until as late as possible.

pleroy: Fine, but that's going to be a bigger change, with all the platform-specific code and such.

EricWF: Yes. And for now this change is correct.

eggrobin: > One of the two operands to the subtraction has been written to memory truncating it, while the other has not

Good point, I had overlooked the possibility of additional roundings to binary64 in my analysis. You found the bug.
>> I believe that the fix is to force flushing to double at the end of MakeTime using a volatile variable.

> I think the real fix here is to store the integer representation instead of converting to double, at least until as late as possible.

Switching to integers makes things clumsier, since it prevents using the SI base unit everywhere, which means that identifiers need to be peppered with SI prefixes to avoid mistaking nanoseconds for seconds, microseconds or Windows ticks.
A more proper alternative would be to switch to a wrapper class, such as `absl::Duration`.

As discussed, this would be a much larger scale change (especially if it involves introducing a dependency on `absl`).

In the meantime I would join @pleroy in favouring forcing a rounding (which solves the root issue of inconsistent rounding introducing nonmonotonicity) rather than maxing this difference with 0, since this would allow `ThreadCPUUsage` to be usable safely anywhere (as well as `ProcessCPUUsage`, which would suffer from the same issue). In their current state, these functions seem to be waiting to cause more hard-to-diagnose bugs.

Adding a `volatile` has, unsurprisingly, the right effect both on [gcc 5.3](https://godbolt.org/g/SxXShT) (which is what you seem to be using) and on [a more modern version](https://godbolt.org/g/qHeGax).

PR #482, opened 2017-11-17
Add NetBSD support

krytarowski:
Define BENCHMARK_OS_NETBSD for NetBSD.

Add detection of cpuinfo_cycles_per_second and cpuinfo_num_cpus.
This code shared detection of these properties with FreeBSD.


dominichamon:
thanks!

PR #483, opened 2017-11-21
Refactor System information collection -- Add CPU Cache Info

EricWF:
This patch refactors the system information collection,
and in particular information about the target CPU. The
motivation is to make it easier to access CPU information,
and easier to add new information as need be.

This patch additionally adds information about the cache
sizes of the CPU.


src/sysinfo.cc
```diff
@@ -49,225 +53,349 @@
...
+BENCHMARK_NORETURN void PrintErrorAndDie(Args&&... args) {
+  PrintImp(std::cerr, std::forward<Args>(args)...);
+  std::exit(EXIT_FAILURE);
 }
+
+#ifdef BENCHMARK_HAS_SYSCTL
+
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
```
dominichamon: i assume this is the aliasing issue below in the union? but it's worth a comment for disabling this warning.

src/sysinfo.cc
```diff
@@ -49,225 +53,349 @@
...
+
+  ValueUnion(ValueUnion&& other) = default;
+
+  explicit operator bool() const { return bool(Buff); }
+
+  char* data() const { return Buff->bytes; }
+
+  std::string getAsString() const { return std::string(data()); }
+
+  long long getAsInteger() const {
```
dominichamon: are you assuming something about bit widths of `long long` here? should it be a ssize_t instead?

src/sysinfo.cc
```diff
@@ -49,225 +53,349 @@
...
+    if (Size == sizeof(Buff->uint32_value))
+      return static_cast<int32_t>(Buff->uint32_value);
+    else if (Size == sizeof(Buff->uint64_value))
+      return static_cast<int64_t>(Buff->uint64_value);
+    CHECK(false) << "invalid size";
+    return 0;
+  }
+
+  unsigned long long getAsUnsigned() const {
```
dominichamon: size_t?

src/sysinfo.cc
```diff
@@ -49,225 +53,349 @@
...
+        StrCat("/sys/devices/system/cpu/cpu", cpu, "/cpufreq/scaling_governor");
+    if (ReadFromFile(governor_file, &res) && res != "performance") return true;
   }
-  return Value.compare(0, Key.size(), Key) == 0;
+#endif
+  return false;
 }
+
+BENCHMARK_MAYBE_UNUSED
+std::vector<CPUInfo::CacheInfo> GetCacheSizesFromKVFS() {
```
dominichamon: does this need to be defined for OSX? Maybe it should be

```c++
#if defined BENCHMARK_OS_MACOSX
... GetCacheSizesMacOSX() {
...
}
#else
... GetCacheSizesFromKVFS() {
...
}
#endif
```

src/sysinfo.cc
```diff
@@ -277,89 +405,34 @@ void InitializeSystemInfo() {
...
+}  // end namespace
 
-// Must use a string literal for prefix.
-#define memprefix(str, len, prefix)                       \
-  ((((len) >= strliterallen(prefix)) &&                   \
-    std::memcmp(str, prefix, strliterallen(prefix)) == 0) \
-       ? str + strliterallen(prefix)                      \
-       : nullptr)
+const CPUInfo& CPUInfo::Get() {
+  static const CPUInfo& inited_info = InitializeCPUInfo(GetUninitialized());
```
dominichamon:
```c++
static const CPUInfo* info = new CPUInfo();
return *info;
```

this is a more google-ish pattern for static initialization of non-POD things, and is simpler to reason about with a well-defined `CPUInfo` constructor.

src/sysinfo.cc
EricWF: It doesn't need to be defined, but I find code that's always exposed to the compiler is easier to maintain. Especially when you're not actually developing on that platform.

src/sysinfo.cc
EricWF: Ack.

src/sysinfo.cc
EricWF: Ack.

src/sysinfo.cc
EricWF: Roughly I'm assuming that `long long` can't get any longer, but apart from that the specific bit width isn't too important.

src/sysinfo.cc
EricWF: See comment above.

src/sysinfo.cc
dominichamon: https://google.github.io/styleguide/cppguide.html#Integer_Types

src/sysinfo.cc
```diff
@@ -49,225 +53,349 @@
...
+      PrintErrorAndDie("Failed to read from file ", FPath, "type");
+    if (!ReadFromFile(StrCat(FPath, "level"), &info.level))
+      PrintErrorAndDie("Failed to read from file ", FPath, "level");
+    res.push_back(info);
+  }
+
+  return res;
+}
+
+#ifdef BENCHMARK_OS_MACOSX
```
dominichamon: remove this ifdef so we compile the function even if it isn't called.

dominichamon:
you know what would be great? tests would be great. maybe it's time to integrate with gtest and write some actual tests. maybe not for this PR i guess.

EricWF:
> you know what would be great? tests would be great. maybe it's time to integrate with gtest and write some actual tests. maybe not for this PR i guess.

Agreed. Adding GTest so it's there to use would be great.

I was looking for opportunities to better test this patch. But it seems unfortunately hard to actually write portable unit tests where the expected output is system specific.

src/sysinfo.cc
```diff
@@ -49,225 +53,348 @@
...
+      : Size(sizeof(DataT) + BuffSize),
+        Buff(::new (std::malloc(Size)) DataT(), &std::free) {}
+
+  ValueUnion(ValueUnion&& other) = default;
+
+  explicit operator bool() const { return bool(Buff); }
+
+  char* data() const { return Buff->bytes; }
+
+  std::string getAsString() const { return std::string(data()); }
```
dominichamon: i didn't spot these naming issues before:

`as_string` or `AsString` (or `GetAsString`).

same below with integer accessors.

src/sysinfo.cc
```diff
@@ -49,225 +53,348 @@
...
+
+  int64_t getAsInteger() const {
+    if (Size == sizeof(Buff->uint32_value))
+      return static_cast<int32_t>(Buff->uint32_value);
+    else if (Size == sizeof(Buff->uint64_value))
+      return static_cast<int64_t>(Buff->uint64_value);
+    BENCHMARK_UNREACHABLE();
+  }
+
+  uint64_t getAsUnsigned() const {
```
dominichamon: why unsigned? this shouldn't be representing a bit pattern, and i don't think you need defined overflow modulo 2^64.

EricWF: Some sysctl fields are explicitly specified as being unsigned integers, and this attempts to allow the user to read them as such.

It also seems entirely reasonable that a field could represent a bit pattern as opposed to an integer value.

src/sysinfo.cc
dominichamon: we do. that PR might be wrong. or this might be a special case.

PR #485, opened 2017-11-25
Add support for GTest based unit tests.

EricWF:
As Dominic and I have previously discussed, there is some
need/desire to improve the testing situation in Google Benchmark.

One step to fixing this problem is to make it easier to write
unit tests by adding support for GTest, which is what this patch does.

By default it looks for an installed version of GTest. However the
user can specify -DBENCHMARK_BUILD_EXTERNAL_GTEST=ON to instead
download, build, and use copy of gtest from source. This is
quite useful when Benchmark is being built in non-standard configurations,
such as against libc++ or in 32 bit mode.


cmake/HandleGTest.cmake
```diff
@@ -0,0 +1,47 @@
...
+      CMAKE_CACHE_ARGS
+        -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+        -DCMAKE_CXX_FLAGS:STRING=${GTEST_FLAGS}
+      )
+
+  ExternalProject_Get_Property(googletest install_dir)
+  add_library(gtest UNKNOWN IMPORTED)
+  add_library(gtest_main UNKNOWN IMPORTED)
+  set_target_properties(gtest PROPERTIES
+    IMPORTED_LOCATION ${install_dir}/lib/libgtest.a
```
LebedevRI:
```
INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/googletest/include
```

cmake/HandleGTest.cmake
```diff
@@ -0,0 +1,47 @@
...
+  )
+  set_target_properties(gtest_main PROPERTIES
+    IMPORTED_LOCATION ${install_dir}/lib/libgtest_main.a
```
LebedevRI:
```
INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/googletest/include
```

cmake/HandleGTest.cmake
```diff
@@ -0,0 +1,47 @@
...
+  )
+  add_dependencies(gtest googletest)
+  add_dependencies(gtest_main googletest)
+  set(GTEST_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/googletest/include)
```
LebedevRI:
```diff
- set(GTEST_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/googletest/include)
```

test/CMakeLists.txt
```diff
@@ -135,6 +134,32 @@ endif()
 compile_output_test(complexity_test)
 add_test(complexity_benchmark complexity_test --benchmark_min_time=${COMPLEXITY_MIN_TIME})
 
+###############################################################################
+# GoogleTest Unit Tests
+###############################################################################
+
+if (BENCHMARK_ENABLE_GTEST_TESTS)
+  include(HandleGTest)
+  include_directories(SYSTEM ${GTEST_INCLUDE_DIRS})
```
LebedevRI:
```diff
- include_directories(SYSTEM ${GTEST_INCLUDE_DIRS})
```

test/CMakeLists.txt
```diff
@@ -135,6 +134,32 @@ endif()
...
+
+  macro(compile_gtest name)
+    add_executable(${name} "${name}.cc")
+    if (TARGET googletest)
+      add_dependencies(${name} googletest)
+    endif()
+    target_link_libraries(${name} benchmark
+        ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
```
LebedevRI: so now `${GTEST_BOTH_LIBRARIES}` will properly add gtest include dirs.

cmake/HandleGTest.cmake
EricWF: Ack.

cmake/HandleGTest.cmake
EricWF: Ack.

test/CMakeLists.txt
EricWF: This line still needs to stay for when out-of-tree GTest installations are used. However the SYSTEM can be removed, since it seems to be implicitly specified after adding the `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` line.

EricWF: Nevermind... You're right.

CMakeLists.txt
```diff
@@ -18,6 +19,14 @@ option(BENCHMARK_USE_LIBCXX "Build and test using libc++ as the standard library
 option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library." OFF)
 option(BENCHMARK_ENABLE_INSTALL "Enable installation of benchmark. (Projects embedding benchmark may want to turn this OFF.)" ON)
 
+# This option is useful when building Benchmark in odd configurations, such as 32bit mode
+# or against libc++.
+option(BENCHMARK_BUILD_GTEST_INTREE "Download and build a custom copy of GTest to use" ON)
```
LebedevRI: This should be two options:
1. build gtest
2. if build gtest, is downloading allowed? this really really should default to false. build systems downloading stuff from internet is such a disgrace :/

README.md
```diff
@@ -13,6 +13,28 @@ IRC channel: https://freenode.net #googlebenchmark
...
 [Additional Tooling Documentation](docs/tools.md)
 
+## Building
+
+The basic steps for configuring and building the library look like this:
+
+```bash
+$ git clone https://github.com/google/benchmark.git
+# Benchmark requires GTest as a dependency. Add the source tree as a subdirectory.
+$ git clone https://github.com/google/googletest.git benchmark/googletest
```
LebedevRI: I'd **highly** recommend
1. providing an option specifying the path to the gtest sources
2. defaulting to `/usr/src/googletest/`, since that is the natural location, and is where debian puts it.

cmake/HandleGTest.cmake
```diff
@@ -0,0 +1,77 @@
...
+  endif()
+  split_list(GTEST_FLAGS)
+  ExternalProject_Add(googletest
+      EXCLUDE_FROM_ALL ON
+      GIT_REPOSITORY https://github.com/google/googletest.git
+      GIT_TAG master
+      PREFIX "${CMAKE_BINARY_DIR}/googletest"
+      INSTALL_DIR "${CMAKE_BINARY_DIR}/googletest"
+      CMAKE_CACHE_ARGS
+        -DCMAKE_BUILD_TYPE:STRING=${GTEST_BUILD_TYPE}
```
LebedevRI: Since all this is just to propagate the current params, maybe it would be simpler to just use [2-step approach](https://github.com/google/googletest/tree/master/googletest#incorporating-into-an-existing-cmake-project)?
Here is how it looks in practice: [[0]](https://github.com/darktable-org/rawspeed/blob/develop/cmake/Modules/GoogleTest.cmake) [[1]](https://github.com/darktable-org/rawspeed/blob/develop/cmake/Modules/GoogleTest.cmake.in)

CMakeLists.txt
EricWF: SGTM. Thanks for the feedback.

README.md
EricWF: First, I don't really care to allow users to specify random/arbirtrary out-of-tree GTest source directories. We already have a mechanism to allow users to specify arbitrary GTest installation using `CMAKE_MODULE_PATH` to influence `find_package(GTest)`.

I've changed the build, as you've suggested, to only download dependencies if `-DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON` is specified. So GTest installations among `/usr/` and `/usr/local/` should work by default.

cmake/HandleGTest.cmake
EricWF: I've looked into 2 step approaches, but I think that they're less idiomatic CMake.
If CMake meant for you to download and load subdirectories during configuration time, they would have provided an actually supported way to do so.

README.md
dominichamon: why not git submodules?

README.md
```diff
@@ -13,6 +13,30 @@ IRC channel: https://freenode.net #googlebenchmark
...
+```
+
+Note that Google Benchmark requires GTest to build and run the tests. This
+dependency can be provided three ways:
+
+* Checkout the GTest sources into `benchmark/googletest`.
+* Otherwise, if `-DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON` is specified during
+  configuration, the library will automatically download and build any required
+  dependencies.
+* Otherwise, if nothing is done, CMake will use `find_package(GTest REQUIRED)`
```
dominichamon: this should be the default, no? ie, why do we have it required to have it in-tree instead of just letting cmake do the right thing?

LebedevRI: `/usr/share/doc/googletest/README.Debian`:
```
	Use of precompiled libgtest Not Recommended
	-------------------------------------------

The Google C++ Testing Framework uses conditional compilation for some
things.  Because of the C++ "One Definition Rule", gtest and gmock
must be compiled with exactly the same flags as your C++ code under
test.  Because this is hard to manage, upstream no longer recommends
using precompiled libraries [1].


	Using GTest with your project
	-----------------------------

See the upstream README for instructions on using gtest with your
project.  The sources for libgtest are installed into
/usr/src/googletest/googletest along with CMakeLists.txt for use with
cmake.

If your build system uses CMake, the ExternalProject command can be
used to build gtest, then FindGTest can be used to find the built
library.


	Using gmock with your project
	-----------------------------

See the upstream README for instructions on using gmock with your
project.  The sources for libgmock are installed into
/usr/src/googletest/googlemock along with CMakeLists.txt for use with
cmake.

With this Debian package something like the following should be enough to build 
a static library (which also includes gtest):

g++ -I/usr/src/googletest/googlemock -c /usr/src/googletest/googlemock/src/gmock-all.cc
g++ -I/usr/src/googletest/googletest -c /usr/src/googletest/googletest/src/gtest-all.cc
ar -rv libgmock.a gmock-all.o gtest-all.o



[1] http://groups.google.com/group/googletestframework/browse_thread/thread/668eff1cebf5309d

 -- Steve M. Robbins <smr@debian.org>, Sat, 19 Nov 2016 21:58:04 -0600
```

PR #486, opened 2017-11-26
Improve CPU Cache info reporting -- Add Windows support.

EricWF:
This patch does a couple of thing regarding CPU Cache reporting.

First, it adds an implementation on Windows. Second it fixes
the JSONReporter to correctly (and actually) output the CPU
configuration information.

And finally, third, it detects and reports the number of
physical CPU's that share the same cache.


EricWF:
Bots are passing. This change LGTM. Committing.

PR #488, opened 2017-11-26
Deprecate CSVReporter - A first step to overhauling reporting.

EricWF:
As Dominic and I have discussed, the current reporter interface
is poor at best. And something should be done to fix it.

I strongly suspect such a fix will require an entire reimagining
of the API, and therefore breaking backwards compatibility fully.

For that reason we should start deprecating and removing parts
that we don't intend to replace. One of these parts, I argue,
is the CSVReporter. I propose that the new reporter interface
should choose a single output format (JSON) and traffic entirely
in that. If somebody really wanted to replace the functionality
of the CSVReporter they would do so as an external tool which
transforms the JSON.

For these reasons I propose deprecating the CSVReporter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,6 @@ else()
 
   # Turn compiler warnings up to 11
   add_cxx_compiler_flag(-Wall)
-
   add_cxx_compiler_flag(-Wextra)
   add_cxx_compiler_flag(-Wshadow)
   add_cxx_compiler_flag(-Werror RELEASE)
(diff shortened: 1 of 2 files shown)


pleroy:
This is moving in the wrong direction.  The API may be broken and require rework, and the reporters may need to be rewritten from scratch, but this deprecation a feature that people use is a serious usability regression.

See also the discussion on #443: it's hard to believe that there is consensus for this change.

EricWF:
@pleroy Mind if I proceed with this now that some misunderstandings have been cleared up?

I plan to remove CSVReporter entirely from the `v2` branch today.

pleroy:
@EricWF: Sure, go ahead.  I need more time to digest your plans regarding JSON (will look at that carefully over the week-end) but in the meantime CSV can die.

EricWF:
Thanks. No rush on the JSON stuff. we've created a `v2` branch to stage the changes on. Honestly it might be better to look later rather than sooner, once there is a more clear vision of the full feature.

PR #489, opened 2017-11-28
Added the installation guide for Ubuntu

Kumar-Kishan:
Added installation guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ IRC channel: https://freenode.net #googlebenchmark
 
 
 ## Installation Guide
+
+For Ubuntu and Debian Based System
+
 First make sure you have git and cmake installed (If not please install it)
 
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,6 +13,78 @@ IRC channel: https://freenode.net #googlebenchmark
 
 [Additional Tooling Documentation](docs/tools.md)
 
+
+## Installation Guide
+First make sure you have git and cmake installed (If not please install it)
+
+```
```
dominichamon: make sure it's clear this bit is for ubuntu/debian.

README.md
```diff
@@ -13,6 +13,78 @@ IRC channel: https://freenode.net #googlebenchmark
...
+Now, let's clone the repository and build it
+
+```
+git clone https://github.com/google/benchmark.git
+cd benchmark
+mkdir build
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=RELEASE
+make
+cmake ..
```
dominichamon: why do you need the second cmake?

README.md
```diff
@@ -13,6 +13,78 @@ IRC channel: https://freenode.net #googlebenchmark
...
+```
+
+We need to now place the built library to the location where our compiler can find it
+
+```
+cd ..
+sudo mkdir /usr/include/benchmark
+sudo cp include/benchmark/benchmark.h /usr/include/benchmark/
+sudo cp build/src/libbenchmark.a /usr/local/lib/
+```
```
dominichamon: replace all this with

`make install`

README.md
```diff
@@ -13,6 +13,78 @@ IRC channel: https://freenode.net #googlebenchmark
...
+
+Let's test it now. We create a directory anywhere in our system
```
dominichamon: i think you can remove the test instructions as that's covered by the rest of the docs already.

Kumar-Kishan: It's 3 in the morning here. Will recommit the fix tomorrow

Kumar-Kishan: Added the changes, as per told

PR #491, opened 2017-11-30
Fix compilation error with GCC on OSX (issue #490).

### LINKED ISSUE #490: Compilation errors on GCC with OSX
27e0b439cf0f9be329f617056795b4e5c3612f34 appears to have introduced a compilation error when building with GCC on OSX.

Reproduction steps:

```bash
brew install gcc@7
git clone https://github.com/google/benchmark.git
cd benchmark
mkdir build && cd build
CC=gcc-7 CXX=g++-7 cmake .. && cmake --build .
```

Build failure:

```
[ 26%] Building CXX object src/CMakeFiles/benchmark.dir/sysinfo.cc.o
/Users/pwnall/workspace/benchmark/src/sysinfo.cc:91:16: error: flexible array member in union
     char bytes[];
                ^
/Users/pwnall/workspace/benchmark/src/sysinfo.cc:180:2: error: extra ';' [-Wpedantic]
 };
  ^
make[2]: *** [src/CMakeFiles/benchmark.dir/sysinfo.cc.o] Error 1
```

It appears that `#pragma GCC diagnostic ignore '-Wpedantic'` in `sysinfo.cc` doesn't work. While I'm pretty sure this is a GCC bug, I'll try to put together a workaround that avoids the non-standard feature.

pwnall:
Fixes #490 

Tested by building with OSX's default clang and with GCC 6 and 7, and running tests.

```bash
xcode-select --install
brew install gcc@6 gcc@7
cd build
cd .. && rm -rf build && mkdir build && cd build
cmake .. && cmake --build . && ctest
cd .. && rm -rf build && mkdir build && cd build
CC=gcc-6 CXX=g++-6 cmake .. && cmake --build . && ctest
cd .. && rm -rf build && mkdir build && cd build
CC=gcc-7 CXX=g++-7 cmake .. && cmake --build . && ctest
```


pwnall:
The Travis build here doesn't appear to run GCC on OSX, so I tested this PR by using it in a project where Travis is configured to include GCC on OSX.

With PR: https://travis-ci.org/google/crc32c/builds/309377910
Without PR: https://travis-ci.org/google/crc32c/builds/309350965

pwnall:
My changes are contained in a block that is guarded by an `#ifdef BENCHMARK_HAS_SYSCTL`, which shouldn't be defined on Windows... so I think the AppVeyor failure is unrelated.

dominichamon:
if it's simple to, can you add a travis build with gcc on osx?

dominichamon:
(actually, maybe that can wait for another PR, but it would be great to have)

pwnall:
@dominichamon Thank you very much for the quick turnaround time! Will submit another PR for GCC on Travis.

PR #493, opened 2017-12-02
Fix warnings on MSVC about double `inline`

eliaskosunen:
When `_MSC_VER` is defined, `BENCHMARK_ALWAYS_INLINE` is defined as `__forceinline`. Some functions were defined as `inline BENCHMARK_ALWAYS_INLINE`, which after preprocessing became `inline __forceinline`. This PR fixes the warnings caused by that by removing the extra `inline`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ Christopher Seymour <chris.j.seymour@hotmail.com>
 David Coeurjolly <david.coeurjolly@liris.cnrs.fr>
 Dirac Research 
 Dominik Czarnota <dominik.b.czarnota@gmail.com>
+Elias Kosunen <elias.kosunen@gmail.com>
 Eric Fiselier <eric@efcs.ca>
 Eugene Zhuk <eugene.zhuk@gmail.com>
 Evgeny Safronov <division494@gmail.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -30,6 +30,7 @@ Christopher Seymour <chris.j.seymour@hotmail.com>
 David Coeurjolly <david.coeurjolly@liris.cnrs.fr>
 Dominic Hamon <dma@stripysock.com> <dominic@google.com>
 Dominik Czarnota <dominik.b.czarnota@gmail.com>
+Elias Kosunen <elias.kosunen@gmail.com>
 Eric Fiselier <eric@efcs.ca>
 Eugene Zhuk <eugene.zhuk@gmail.com>
 Evgeny Safronov <division494@gmail.com>
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -302,7 +302,7 @@ BENCHMARK_UNUSED static int stream_init_anchor = InitializeStreams();
 // See: https://youtu.be/nXaxk27zwlk?t=2441
 #ifndef BENCHMARK_HAS_NO_INLINE_ASSEMBLY
 template <class Tp>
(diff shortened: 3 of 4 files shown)


EricWF:
I'm not sure this is correct. inline is a linkage specifier, __force_inline
is not

eliaskosunen:
@EricWF I'm fairly certain that I'm not mistaken here, but I may as well be. [Reading Microsoft documentation](https://msdn.microsoft.com/en-us/library/bw1hbe6y.aspx), it seems to me that `__forceinline`, in some cases, simply overrides compiler analysis regarding the effectiveness of function inlining and subtituting one for the other should not have any behavioral effect or influence on linkage.

Nonetheless, without this PR, `inline __forceinline` gives the warning C141 (Inline used more than once), while `__forceinline` doesn't.

Appveyor build [with patch](https://ci.appveyor.com/project/varuna-lang/spio/build/0.1.50/job/ldsf7yrnydqrm9yn/messages), [without patch](https://ci.appveyor.com/project/varuna-lang/spio/build/0.1.49/job/qo56817dv3orf90x/messages)

eliaskosunen:
Appveyor bot seems to agree

dominichamon:
Given all checks are passing, I don't see any concerns with merging. @EricWF do you still have concerns?

eliaskosunen:
@EricWF @dominichamon Ping?

eliaskosunen:
Is there any sort of plan to resolve this issue?

dominichamon:
waiting for @EricWF because he's much smarter than i am.

EricWF:
@eliaskosunen I can't duplicate the warning on Godbolt. Could you clarify what version of MSVC you're using?

EricWF:
@eliaskosunen Nevermind. I found your Appveyor links. You're using a newer compiler that Godbolt. So this hasn't been fixed in newer versions, like I initially though; instead it's only been added in newer versions.

I still don't like this PR, since I think MSVC is being dumb here. I'm going to open a dialog with the MSVC devs to see what they think.

pleroy:
This PR looks odd.

I am using VS2015 and I don't see any warning about double inline.  Also, as @EricWF indicated, `inline` and `__forceinline` are completely different things.  Removing `inline`, a linkage directive, looks like it could cause ODR nightmares.

The line numbers in the Appveyor logs point at lines 644 and 647, but the declarations have not been at these lines since commit fa341e51cb7f6bce69a7577f4000381a03f61c70.  Are you sure that the problems still exist from the latest master commit?  (There _was_ a problem with these declaration that got fixed in #469.)

pleroy:
(Unrelated comment: my preference would be to remove `inline` from all the `template`s because it has really no effect whatsoever there.)

EricWF:
@pleroy I also considered removing inline on templates. But (A) that doesn't really fix this issue for non-templates, and (B) what about compilers that actually take `inline` as an optimizer hint and don't support either `__forceinline` or `__attribute__((always_inline))`?

pleroy:
(A) I was not trying to fix this issue (which I think is a red herring addressed by #469), just stating a preference (B) do we know that there are really compilers like that around? in this day and age, I think it's best to assume that `inline` doesn't have anything to do with performance.

EricWF:
Just spoke to an MSVC compiler dev. They couldn't reproduce with the latest version of MSVC either (19.13.xxx).

I think we should go ahead and close this PR. The warning appears to be a briefly shipped bug in MSVC.

eliaskosunen:
Looking more into this issue, I agree.

PR #495, opened 2017-12-04
Enforce using a semicolon after BENCHMARK_MAIN to remove compiler warnings

### LINKED ISSUE #494: Extra semicolon warning emitted with GCC when using BENCHMARK_MAIN
The following code triggers a warning with GCC when `-Wpedantic` is enabled:

```c++
BENCHMARK_MAIN();
```

The warning is about the semicolon being unnecessary, because `BENCHMARK_MAIN()` expands to

```c++
int main(int, char**) {
  ...
}
```

[(reproduction)](https://wandbox.org/permlink/C80p8zmjacQQQqaS)

This is unfortunate because the macro is used everywhere in the documentation with a semicolon. I propose changing the macro to the following, such that the semicolon becomes mandatory and compilers stop warning:

```c++
#define BENCHMARK_MAIN()              \
  int main(int , char** ) {           \
    ...                               \
  }                                   \
  int main(int, char**)
```

[(example)](https://wandbox.org/permlink/MiUL7hlJfef20yMn)

This should not be a breaking change since AFAICT the macro is always documented with a semicolon after it. For those that will be broken by it, the fix is trivial (just add the semicolon).

ldionne:
This fixes #494.

In #494, I claim that `BENCHMARK_MAIN()` without semicolon is probably never used. Funnily enough, it seems like the tests of the library itself were using it without a semicolon, as can be seen from the fixes I had to make.

Since the documentation used a semicolon, it may be safe to assume that people have not been relying on it, but it's not 100% sure either. Another possibility is to change the documentation to _not_ use a semicolon (that may be unfriendly to syntax highlighters & al, but that's a tradeoff).

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1146,7 +1146,8 @@ class Fixture : public internal::Benchmark {
     ::benchmark::Initialize(&argc, argv);  \
     if (::benchmark::ReportUnrecognizedArguments(argc, argv)) return 1; \
     ::benchmark::RunSpecifiedBenchmarks(); \
-  }
+  }                                        \
+  int main(int, char**)
 
 
 // ------------------------------------------------------
--- a/test/basic_test.cc
+++ b/test/basic_test.cc
@@ -115,4 +115,4 @@ void BM_RangedFor(benchmark::State& state) {
 }
 BENCHMARK(BM_RangedFor);
 
-BENCHMARK_MAIN()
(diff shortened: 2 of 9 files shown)


ldionne:
I signed the CLA.

ldionne:
Looking at the appveyor log, it's not clear the failure is related to this PR.

EricWF:
It's not related. Some tests are still flaky.

PR #496, opened 2017-12-04
bazel build support

monkeynova:
basic bazel workspace and buildfile -- to accomplish this tests now add an include path of "../src"

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,393 @@
+cc_library(
+    name = "benchmark",
+    srcs = ["src/benchmark.cc"],
+    hdrs = ["include/benchmark/benchmark.h"],
+    strip_include_prefix = "include",
+    deps = [
+        ":benchmark_api_internal",
+        ":benchmark_header",
+        ":benchmark_register",
+        ":check",
+        ":colorprint",
+        ":commandlineflags",
+        ":complexity",
+        ":console_reporter",
+        ":counter",
+        ":csv_reporter",
+        ":internal_macros",
+        ":json_reporter",
+        ":log",
+        ":mutex",
+        ":re",
+        ":reporter",
+        ":statistics",
+        ":string_util",
+        ":sysinfo",
+        ":timers",
+    ],
+    visibility = [
+        "//visibility:public",
+    ],
+)
+
+cc_library(
+    name = "arraysize",
+    hdrs = ["src/arraysize.h"],
+    strip_include_prefix = "src",
+ )
+
+cc_library(
+    name = "benchmark_api_internal",
+    hdrs = ["src/benchmark_api_internal.h"],
+    strip_include_prefix = "src",
+)
+
+cc_library(
+    name = "benchmark_header",
+    hdrs = ["include/benchmark/benchmark.h"],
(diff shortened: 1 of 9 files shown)


BUILD
```diff
@@ -0,0 +1,393 @@
...
+    ]
+)
+
+cc_library(
+    name = "re",
+    hdrs = ["src/re.h"],
+    strip_include_prefix = "src",
+    deps = [
+    ],
+    defines = ["HAVE_POSIX_REGEX"],
```
dominichamon: note, this won't work for everyone. this is one of the reasons i held off; no way to do this kind of dynamic configuration.

test/CMakeLists.txt
```diff
@@ -3,6 +3,8 @@
 find_package(Threads REQUIRED)
 include(CheckCXXCompilerFlag)
 
+include_directories("../src")
```
dominichamon: this should not be necessary. We don't want the tests to have blanket access to the private parts of src.

for the same reason, you should be using visibility = ["//visibility:private"] on most of the cc_library targets above.

BUILD
monkeynova: It's not perfect agreed, but it's also not unprecedented. gflags has a similar hardcoding (https://github.com/gflags/gflags/blob/master/bazel/gflags.bzl). I started a patch for moving some of the defines there to run off of a config but haven't gotten to a PR for that yet (googletest does something similar for linkopts: https://github.com/google/googletest/blob/master/BUILD.bazel).

But I know I don't have complete access to all the build modes I might need. Is there an automated build tool I can use to ensure coverage?

test/CMakeLists.txt
monkeynova: A fair number of the tests #include "../src/<whatever>" so the encapsulation is actually pretty bad. Since this was tests as part of the package I wasn't particularly concerned with the model here, but it's a reasonable concern.

As far as the visibility:private, isn't that the default in the build file? When I was trying to make use of this from an external dependency, I had to add the explicit visibility:public to the one rule I wanted to make visible.

BUILD
dominichamon: the travis and appveyor builds are probably the closest we have for covering the various platforms/compilers. there are other platforms that people run on, but we don't have a way to cover every combination.

test/CMakeLists.txt
dominichamon: ah, you're right. i wasn't sure how they'd set the default.

they include _some things_ from ../src, but don't have full access. i don't know enough about bazel, but would it be better to add a BUILD in src, a BUILD in test, and manage the visibility of certain src/ libraries to test/ that way?

dominichamon:
(i just pushed https://github.com/google/benchmark/commit/7f2d2cd5b9d8f892862e9c7974ede413c76f9bff which should help the travis xcode builds)

dominichamon:
Can you also add a travis entry to test the config? Example here: https://github.com/korfuri/bazel-travis/blob/master/.travis.yml

dominichamon:
Made obsolete by #533

PR #499, opened 2017-12-14
Begin JSON reporter change: Allow users to specify arbitrary JSON input/output.

EricWF:
This patch begins a set of changes to convert the library to use JSON as
the primary data type used to communicate with users; but for custom user
input and output.

The patch introduces the json.h header taken, and modified, from
https://github.com/nlohmann/json . The header was modified to no
longer use the namespace nlohmann, nor any macros containing NLOHMANN
so as to not interfere with other users who already incidentally
use this header.

Additionally, this patch adds `Benchmark::WithInput` to allow
the passing of arbitrary JSON as an input argument to a benchmark.
The results of which can be accessed using `State::GetInput()`.

This patch also adds `State::operator[](std::string)`, which allows
users to report arbitrary JSON as output from a benchmark. See
the documentation for examples.


EricWF:
@pleroy Here is the first set of changes I'm proposing in order to move toward  using JSON as the primary format.

EricWF:
The coverage decrease is very likely caused by the introduction of the new header, which we aren't testing.

dominichamon:
the json header you're using is licensed under MIT. This means we'd have to update our license to indicate that the software is now multiple-licensed.

Can we instead depend on this JSON library in some other way (submodules, findpackage (it uses cmake)) which would also give us the opportunity to get updates?

biojppm:
Hey @EricWF,

The unit tests I wrote for the counters are currently reading their data from the CSV output. This will need to be changed as well. FWIW I can move that logic to JSON if it helps.

EricWF:
@biojppm I was planning on converting the tests myself in a later commit, since it's a good opportunity for me to demonstrate how I believe the API should be used and tested.

EricWF:
Urg. This Google bot thing is going to be a problem using the v2 branch.

PR #500, opened 2017-12-14
Remove CSV Reporter on V2 Branch

EricWF:
This patch removes the CSV Reporter entirely from the V2 branch.

This change is needed to support the work of converting to JSON.


dominichamon:
LGTM

KindDragon:
Why you remove CSV support? CSV support really easy to open in LibreOffice Calc and build some charts

dominichamon:
@KindDragon We're removing support from the core library, but retaining JSON output. It is relatively to write an external tool to read the JSON and generate the CSV should it be necessary, but the library is getting very complicated with the various non-JSON reporters.

KindDragon:
> It is relatively to write an external tool to read the JSON and generate the CSV should it be necessary,

Yes, but not as easy as it is now 😞

Issue #497, opened 2017-12-11
static member variable initialized in constructor of benchmark fixture


spinosae:
When using a custom class in constructor of a benchmark fixture, I noticed its static member variable was not properly defined.

Upon examination of benchmark.h, I realized the constructor was called in macro REGISTER_BENCHMARK_F. It is called before main function and creates static benchmark objects. According to answers on stack-overflow, initialization of static member variable before first line of main function is implementation dependent. So problem I had can be explained.

Meanwhile the static member variable is correctly defined within main function. Is it possible to register benchmarks in main function? I think it applies to google test too.

Here is a minimal example. Basically I am using a static member variable to store hash code of the class. 
```
#include <iostream>
#include <benchmark/benchmark.h>
#include <typeinfo>
using namespace std;

// helper hashing class
template <class T>
class Hashed {
 public:
  static const size_t Hash;
};

template <class T>
const size_t Hashed<T>::Hash = typeid(T).hash_code();

class Tmp1 : public Hashed<Tmp1> {};

// access static member value
template <class T>
void PrintT() {
  cout << "class: " << typeid(T).name() << " hash: " << T::Hash << endl;
}

class A : public benchmark::Fixture {
 public:
  static const int var1;
  A() { PrintT<Tmp1>(); } // prints 0
};

BENCHMARK_F(A, JustToCallConstructor)
(benchmark::State& st) {}

int main() {
  PrintT<Tmp1>(); // prints correctly
  return 0;
}
```
output of above program
```
class: 4Tmp1 hash: 0
class: 4Tmp1 hash: 776362582
```

EricWF:
I'm not really sure what to do with this. 

Obviously, you've correctly identified this as an classic example of the static initialization order fiasco.

>  Is it possible to register benchmarks in main function?

Yes. https://github.com/google/benchmark#using-registerbenchmarkname-fn-args

spinosae:
Hi @EricWF, thanks for your prompt reply. I am a new user of Google Benchmark. Therefore it took me a while to notice the problem and figure out where the trap was. It is not so obvious that initialization of fixtures happens before main. Wouldn't it be better if benchmarks are registered in main function by default (i.e. in BENCHMARK_MAIN)? 

`RegisterBenchmark` doesn't seem to support fixture benchmark, unless I missed something.

Anyway, this is probably not a common problem. Feel free to close it if you feel it's not relevant.

EricWF:
[...]

PR #504, opened 2017-12-23
fix static initialization order problem

lijinpei:
This commit can fix the static initalization problem (see https://github.com/google/benchmark/issues/497 ), the commit pass all the tests and won't break sane codes.
Two problem the commit tries to solve is (1) Benchmark.Apply() calls user defined function during static initialization. (2) Fixture's construtor is user defined, and can do anything during static initialization.
    To solve problem (1), Benchmark.Apply() doesn't run the the user defined function, but push it to a vector to run later(specifically, in BenchmarkFamilies.FindBenchmarks() just before benchmark's argument is checked, see https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L130 and https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L130 )
    To solve problem (2), the class hierarchy is changed(but the change won't break existing code, after recompiling). The old classs hierarchy is Benchmark->Fixture->UserDefinedFixture->FixtureDefinedThroughMacro, now the hierarchy is Benchmark->FixtureBenchmark(the other two classed derived from Benchmark are FunctionBenchmark and LambdaBenchmark), Fixture->UserdefinedFixture->FixtureDefinedThroughMacro, and now FixtureDefinedThroughMacro has a static method which takes no argument and return a Fixture*. FixtureBenchmark stores a Fixture* and a function pointer(the static member of FixtureDefinedThroughMacro). During static initialization, FixtureBenchmark is created and registered, no user defined class involved, just a function to generate a user defined class is recorded. When the benchmark is run, the function pointer is used to generate the fixture if necessary. See https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L489 ,  https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L525 , and https://github.com/lijinpei/benchmark/blob/v2/include/benchmark/benchmark.h#L1043 .

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -274,6 +274,8 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter,
 
 namespace internal {
 class Benchmark;
+class FunctionBenchmark;
+class FixtureBenchmark;
 class BenchmarkImp;
 class BenchmarkFamilies;
 
@@ -662,6 +664,7 @@ typedef void(Function)(State&);
 // chained into one expression.
 class Benchmark {
  public:
+  typedef Benchmark* (*Creator_t)();
   virtual ~Benchmark();
 
   // Note: the following methods all return "this" so that multiple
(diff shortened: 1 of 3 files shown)


lijinpei:
The current commit has race condition when FixtureBenchmark.Run() is called from multiple threads, please don't merge it.

Issue #498, opened 2017-12-13
Benchmark Registration API, and in particular Fixtures, beg users to write static initialization bugs.


EricWF:
The `BENCHMARK` registration macros execute during program startup, ie static initialization. Since the order of static initialization is unspecified, users who execute arbitrary code while registering benchmarks may encounter the static initialization order fiasco. Two common way users execute arbitrary code when registering benchmarks are (A) using the `Benchmark::Apply` method, and (B) in the constructor of user-created Fixtures. The latter case causes problems a lot more frequently than the former.

Although users are partly to blame, especially in the fixture case, for even attempting to execute code during static initialization; it is still the libraries fault for providing API's which encourage users to do so.

We should attempt to come up with a fully functional API to register benchmarks at runtime, *after* static initialization is complete. And then we should re-write the documentation to encourage users to
prefer the new interface over the old.

However... I still think the old interface is very useful, especially to those who are aware of it's dark corners. So I'm not suggesting removing it; but it would be nice if we were no longer responsible
for causing initialization order bugs in users code.

Any and all ideas are welcome on how to improve the status quo. @dominichamon @pleroy Any ideas?

dominichamon:
What is missing from our current runtime API?

PR #505, opened 2017-12-24
fix static initialization order problem

lijinpei:
This commit can fix the static initalization problem (see [issue #498 ](https://github.com/google/benchmark/issues/498) ), the commit pass all the tests and won't break existing sane codes.

The commit tries to solve two problem

1. Benchmark.Apply() calls user defined function during static initialization. 
2. Fixture's construtor is user defined, and can do anything during static initialization. 

To solve problem 1, Benchmark.Apply() doesn't run the the user defined function, but push it to a vector to run later(specifically, in BenchmarkFamilies.FindBenchmarks() just before benchmark's argument is checked, see [Benchmark::Apply()](https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L382) and [BenchmarkFamilies::FindBenchmarks](https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L138).

To solve problem 2, the class hierarchy is changed(but the change won't break existing code, after recompiling). The old classs hierarchy is Benchmark->Fixture->UserDefinedFixture->FixtureDefinedThroughMacro, now the hierarchy is Benchmark->FixtureBenchmark(the other two classed derived from Benchmark are FunctionBenchmark and LambdaBenchmark), Fixture->UserdefinedFixture->FixtureDefinedThroughMacro. FixtureDefinedThroughMacro has a static method which takes no argument and return a Fixture*. FixtureBenchmark stores a unique_ptr<Fixture> and a creator pointer(FixtureDefinedThroughMacro's static method). During static initialization, FixtureBenchmark is created and registered, no user defined class constructed, just a function to generate a user defined class is recorded. When the benchmark is find, the function pointer is used to generate the fixture if necessary. See [creator](https://github.com/lijinpei/benchmark/blob/v2/include/benchmark/benchmark.h#L1031), [FixtureBenchmark](https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L75), [BenchmarkFamilies::FindBenchmarks](https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L159)

Also I change FixtureBenchmark and FunctionBenchmark's definition to benchmark_register.cc(https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L62), because user need not see them, and I add two function overloads(https://github.com/lijinpei/benchmark/blob/v2/src/benchmark_register.cc#L525) for dynamically register benchmark.


include/benchmark/benchmark.h
```diff
@@ -662,6 +664,7 @@ typedef void(Function)(State&);
 // chained into one expression.
 class Benchmark {
  public:
+  typedef Benchmark* (*Creator_t)();
```
dominichamon: we generally don't do _ in type names: https://google.github.io/styleguide/cppguide.html#Type_Names

include/benchmark/benchmark.h
```diff
@@ -813,6 +817,8 @@ class Benchmark {
 
   virtual void Run(State& state) = 0;
 
+  void ApplyFunctions();
```
dominichamon: I'm a little surprised this is public. does it need to be?

src/benchmark_register.cc
```diff
@@ -124,6 +156,10 @@ bool BenchmarkFamilies::FindBenchmarks(
     // Family was deleted or benchmark doesn't match
     if (!family) continue;
 
+    if (auto p_fixture_benchmark = dynamic_cast<FixtureBenchmark*>(family.get())) {
```
dominichamon: i think this is the first time we've had an RTTI requirement in the library.
https://google.github.io/styleguide/cppguide.html#Run-Time_Type_Information__RTTI_

ideally we'd resolve this through virtuals (though i understand this is hard given you don't want to define `TryCreateFixture` on the base class.

is it completely onerous to have a separate list of fixture benchmarks on which this can be called while also having them added to the full set of benchmarks?

dominichamon: if you need a p_ to indicate it's a pointer, maybe use `auto*` instead. ie, prefer types to naming to indicate, well, type.

src/benchmark_register.cc
```diff
@@ -461,13 +490,50 @@ int Benchmark::ArgsCnt() const {
   return static_cast<int>(args_.front().size());
 }
 
-//=============================================================================//
-//                            FunctionBenchmark
-//=============================================================================//
+void Benchmark::ApplyFunctions() {
+  if (functions_to_apply_.size()) {
```
dominichamon:
```
if (functions_to_apply_.empty()) {
  return;
}
```
?

src/benchmark_register.cc
dominichamon: alternatively, perhaps `ApplyFunctions` can also create the fixture if necessary. ie, use a virtual `ApplyFunctions` method (maybe renamed to `SetUp` ;)) that can do more in the case of a fixture benchmark.

src/benchmark_register.cc
```diff
@@ -461,13 +490,50 @@ int Benchmark::ArgsCnt() const {
...
+Benchmark* RegisterBenchmarkInternal(Benchmark* p_benchmark) {
+  BenchmarkFamilies::GetInstance()->AddBenchmark(std::unique_ptr<Benchmark>(p_benchmark));
+  return p_benchmark;
+}
+
+} // end namespace internal
+
+// register a function benchmark
+internal::Benchmark* RegisterBenchmark(const char* name, internal::Function* func) {
+  auto p_benchmark = new internal::FunctionBenchmark(name, func);
```
dominichamon: `auto* benchmark = ...`

src/benchmark_register.cc
```diff
@@ -461,13 +490,50 @@ int Benchmark::ArgsCnt() const {
...
+}
+
+Benchmark* RegisterBenchmarkInternal(Benchmark* p_benchmark) {
+  BenchmarkFamilies::GetInstance()->AddBenchmark(std::unique_ptr<Benchmark>(p_benchmark));
+  return p_benchmark;
+}
+
+} // end namespace internal
+
+// register a function benchmark
```
dominichamon: this documentation doesn't add anything.

src/benchmark_register.cc
```diff
@@ -461,13 +490,50 @@ int Benchmark::ArgsCnt() const {
...
+internal::Benchmark* RegisterBenchmark(const char* name, internal::Function* func) {
+  auto p_benchmark = new internal::FunctionBenchmark(name, func);
+  internal::BenchmarkFamilies::GetInstance()->AddBenchmark(std::unique_ptr<internal::Benchmark>(p_benchmark));
+  return p_benchmark;
+}
+
+// register a fixture benchmark
```
dominichamon: also a pointless comment.

EricWF:
I'm not really concerned with "fixing" the `Apply`, since it should be immediately apparent to callers that it is executed during static initialization. This is unlike Fixtures.

I would prefer to see a minimal version of this patch that just handles fixtures and leaves `Apply` untouched.

include/benchmark/benchmark.h
lijinpei: I will change the code.

include/benchmark/benchmark.h
lijinpei: This method should be private.

src/benchmark_register.cc
lijinpei: I think it is good practice to add a  private virtual member function named setUp/prepare or something like this.

lijinpei: I will change the code.

src/benchmark_register.cc
lijinpei: I will change the code.

src/benchmark_register.cc
lijinpei: I will change the code.

src/benchmark_register.cc
lijinpei: I will change the code.

src/benchmark_register.cc
lijinpei: I will change the code.

lijinpei:
I will rewrite a commit to just handle the fixtures, the class hierarchy will be like this commit, and the Benchmark base class will have a virtual member named like "private: virtual void prepare();" to give sub-class a chance to construct fixtures before benchmark is run.
Apply() won't be changed.
Plus multiple coding style changes.

PR #507, opened 2018-01-03
Use uintXX_t instead of size_t for items/bytes/iterations

Maratyszcza:
7a767012f1c423b37069f6d315b97164b5850271 seemingly unintentionally changed the data types for bytes_processed, items_processed, and iterations to size_t. On 32-bit systems, the bytes_processed and items_processed variables easily overflow, and often report fewer items/second for faster code. The fact that overflow chances depends on architecture and even ABI (e.g. x64 vs x32 ABI on x86-64) makes it particularly annoying. iterations counter is made to use uint32_t because it is incremented in the hot loop, and incrementing uint64_t is slow on some 32-bit platforms (e.g. Asm.js).


include/benchmark/benchmark.h
```diff
@@ -565,17 +565,17 @@ class State {
   int range_y() const { return range(1); }
 
   BENCHMARK_ALWAYS_INLINE
-  size_t iterations() const { return (max_iterations - total_iterations_) + 1; }
+  uint32_t iterations() const { return (max_iterations - total_iterations_) + 1; }
```
pleroy: This change seems unnecessary.  On a 32-bit system it has no effect and on a 64-bit system there is no harm in using `size_t`.  Same comment for all the places that touch iterations counters.

Maratyszcza: With this change, if a benchmark has an overflow bug on 32-bit systems, it will also manifest on a 64-bit system. E.g. `state.SetItemsProcessed(1000000 * state.iterations())` would have identical overflow behavior on both 32-bit and 64-bit systems.

pleroy: And I claim that it's a bad property.

I happen to only ever run benchmarks on 64-bit systems.  Why would you force me to have to deal with 32-bit overflows just for a misconceived notion of uniformity?

Maratyszcza:
I changed interface for `State.iterations()` to return `uint64_t`, so code that relied on `iterations` being 64-bit on x86-64 will keep working, and will also work on 32-bit ABIs. In the private classes, I left `iterations` as `uint32_t`. It is upper bounded by 1e+9 iterations (`kMaxIterations`), so this wouldn't break anything on either 32-bit or 64-bit systems.

dominichamon:
for one thing, these should not be explicitly unsigned unless we're representing bitfields or closed-group arithmetic. We're not, so we should be using signed integers if we're going to move away from the standard size_t type (which is unsigned by the standard, but that was an unfortunate decision).

while signed integer overflow is undefined behaviour, this is detectable with ubsan and is safer (in a way) than the wrapping that unsigned integers would give us.

ie, i'm comfortable with the bytes/items processed being intxx_t instead of size_t, but not uintxx_t.

EricWF:
@Maratyszcza Thanks for the patch. Don't worry about ABI breaks. The project is free to break the ABI.

First, I agree with @dominichamon about using signed integer types. And I agree about making this change for things we explicitly want to be 64 bits.

However, I don't know if using explicit 32 bit types makes sense. In these cases, it seems to make the most sense to simply use `int`, which I think is what we should try to do universally except where 64 bits are needed.  What do you think?

Maratyszcza:
@dominichamon I changed types in the interface to `int64_t`, can we merge now? Internally, I left `iterations` as `uint32_t`.

@EricWF Using `int` would hurt portability. We need a type that can hold values up to `1e+9`. `uint32_t`/`int32_t` guarantees that, `int` does not (in fact, `INT_MAX` is usually just 32767 on 16-bit systems).

include/benchmark/benchmark.h
```diff
@@ -565,17 +565,21 @@ class State {
...
+    return static_cast<int64_t>(
+      static_cast<uint64_t>(
+        (max_iterations - total_iterations_) + 1));
+  }
 
  private:
   bool started_;
   bool finished_;
-  size_t total_iterations_;
+  uint32_t total_iterations_;
```
dominichamon: i buy your 'int' argument and if we need a guarantee, which we do, then this should be ?int32_t. The Google C++ style guide, interestingly, states that we can assume an int is [at least 32 bits](https://google.github.io/styleguide/cppguide.html#Integer_Types), which is a bit surprising.

However, in the same section it also specifically recommends avoiding unsigned unless we need overflow modulo 2^N or representing bit patterns. Given we're doing neither, this should be signed.

So: `int32_t` please. And then promoting once to int64_t in 'iterations' seems reasonable, though i'd promote the individual parts of the calculation. ie, `max_iterations - total_iterations_` could be INT_MAX already if total is 0. the +1 will cause an overflow before you get the chance to promote to int64.

pleroy: I'm sorry, I don't buy the `int` vs. `int32_t` argument.

According to [cppreference](http://en.cppreference.com/w/cpp/language/types), `int` can only be a 16-bit type on LP32 systems, e.g., Win16 or, maybe, some microcontrollers.  We have no evidence that the library was ever ported to such a system.  If software hasn't been ported, it's not portable, period.  The code is choke-full of usages of `int` so there is no point in an illusory attempt at improving it by using `int32_t` in a handful of places.

If someone wanted to port the library to Win16 that might be an interesting exercise, but that's not what this PR does.

EricWF: ++pleroy. @Maratyszcza I maintain an STL, and I honestly I don't think we even worry about 16 bit integers. Let's go with the Google style guide and assume that int is at least 32 bits.

PR #508, opened 2018-01-05
Merge recent changes into v2 branch.

EricWF:


dominichamon:
I approve the changes that I made to this PR in the branch.

EricWF:
I don't think we actually need to go through this process, because every contributor to the master branch already signed the CLA. This seems like a bug in the bot. Maybe we could report it to somebody (since we're going to be doing these kind of merges a lot).

PR #511, opened 2018-01-06
handle fixture construction after static initialization

lijinpei:
This pr is a revision of [pr #505 ](https://github.com/google/benchmark/pull/505). It just tries to do fixture 's construction before benchmark runs. And hopefully, I have fixed the coding style problem.
See pr #505  for discussion history and implementation details.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -279,10 +279,6 @@ class BenchmarkFamilies;
 
 void UseCharPointer(char const volatile*);
 
-// Take ownership of the pointer and register the benchmark. Return the
-// registered benchmark.
-Benchmark* RegisterBenchmarkInternal(Benchmark*);
-
 // Ensure that the standard streams are properly initialized in every TU.
 int InitializeStreams();
 BENCHMARK_UNUSED static int stream_init_anchor = InitializeStreams();
@@ -826,6 +822,9 @@ class Benchmark {
   static void AddRange(std::vector<int>* dst, int lo, int hi, int mult);
 
  private:
+  // Init() is called before this benchmark is run.
+  virtual void Init();
+
   friend class BenchmarkFamilies;
 
   std::string name_;
@@ -849,7 +848,7 @@ class Benchmark {
 
 }  // namespace internal
 
-// Create and register a benchmark with the specified 'name' that invokes
+// Create and register a function benchmark with the specified 'name' that invokes
 // the specified functor 'fn'.
 //
 // RETURNS: A pointer to the registered benchmark.
@@ -866,18 +865,6 @@ internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn);
 void ClearRegisteredBenchmarks();
 
 namespace internal {
-// The class used to hold all Benchmarks created from static function.
-// (ie those created using the BENCHMARK(...) macros.
-class FunctionBenchmark : public Benchmark {
- public:
-  FunctionBenchmark(const char* name, Function* func)
-      : Benchmark(name), func_(func) {}
-
-  virtual void Run(State& st);
-
- private:
-  Function* func_;
-};
 
 #ifdef BENCHMARK_HAS_CXX11
 template <class Lambda>
@@ -902,13 +889,14 @@ class LambdaBenchmark : public Benchmark {
 
 }  // namespace internal
 
-inline internal::Benchmark* RegisterBenchmark(const char* name,
-                                              internal::Function* fn) {
-  return internal::RegisterBenchmarkInternal(
-      ::new internal::FunctionBenchmark(name, fn));
+#ifdef BENCHMARK_HAS_CXX11
+
+namespace internal {
+// Take ownership of the pointer and register the benchmark. Return the
+// registered benchmark.
+Benchmark* RegisterBenchmarkInternal(Benchmark*);
 }
 
-#ifdef BENCHMARK_HAS_CXX11
 template <class Lambda>
 internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn) {
   using BenchType =
@@ -931,16 +919,9 @@ internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn,
 #endif
 
 // The base class for all fixture tests.
-class Fixture : public internal::Benchmark {
+class Fixture {
  public:
-  Fixture() : internal::Benchmark("") {}
-
-  virtual void Run(State& st) {
-    this->SetUp(st);
-    this->BenchmarkCase(st);
-    this->TearDown(st);
-  }
-
+  virtual ~Fixture();
   // These will be deprecated ...
   virtual void SetUp(const State&) {}
   virtual void TearDown(const State&) {}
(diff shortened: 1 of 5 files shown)


lijinpei:
Also, I think Benchmark::SetName() can be removed, because it's a protected method, and currently no sub-class uses it. But I didn't remove it in this pr.

lijinpei:
Travis CI's failure is caused by apt https://travis-ci.org/google/benchmark/jobs/325788181 , it succeeded in my repo https://travis-ci.org/lijinpei/benchmark/builds/325782844 .

include/benchmark/benchmark.h
```diff
@@ -826,6 +822,9 @@ class Benchmark {
   static void AddRange(std::vector<int>* dst, int lo, int hi, int mult);
 
  private:
+  // prepare() is called before this benchmark is run.
+  virtual void prepare();
```
EricWF: This should be upper case as with the rest of the methods. I would vote for the name `Init()`

include/benchmark/benchmark.h
```diff
@@ -931,27 +919,30 @@ internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn,
...
 
+namespace internal {
+typedef Fixture* (FixtureCreator)();
+}
+
+// Create and register a fixture benchmark with the specified 'name' whose fixture
+// is created by creator
+//
+// RETURNS: A pointer to the registered benchmark.
+internal::Benchmark* RegisterBenchmark(const char* name,
```
EricWF: I'm not sure I agree in making this function part of the public API. Nobody should be calling this directly, right?

include/benchmark/benchmark.h
```diff
@@ -1019,32 +1006,29 @@ class Fixture : public internal::Benchmark {
...
 #else
 #define BENCHMARK_TEMPLATE(n, a) BENCHMARK_TEMPLATE1(n, a)
 #endif
 
 #define BENCHMARK_PRIVATE_DECLARE_F(BaseClass, Method)        \
   class BaseClass##_##Method##_Benchmark : public BaseClass { \
    public:                                                    \
-    BaseClass##_##Method##_Benchmark() : BaseClass() {        \
-      this->SetName(#BaseClass "/" #Method);                  \
+    BaseClass##_##Method##_Benchmark() : BaseClass() {}       \
```
EricWF: These `Create` methods can be replaced with a single template.

```c++
template <class FixtureT>
Fixture *CreateFixture() { return new FixtureT(); }
```

src/benchmark_register.cc
```diff
@@ -461,16 +463,77 @@ int Benchmark::ArgsCnt() const {
...
+//=============================================================================//
+//
+// The class used to hold all Benchmarks created from fixture.
+// (ie those created using the BENCHMARK_F(...) macros.
+class FixtureBenchmark : public Benchmark {
+ public:
+  FixtureBenchmark(const char* name, FixtureCreator* creator)
+      : Benchmark(name), creator_(creator) {}
+
+  void Run(State& st) override;
```
EricWF: You might as well define the functions inline.

src/benchmark_register.cc
```diff
@@ -461,16 +463,77 @@ int Benchmark::ArgsCnt() const {
   return static_cast<int>(args_.front().size());
 }
 
+void Benchmark::prepare() {}
+
 //=============================================================================//
 //                            FunctionBenchmark
 //=============================================================================//
 
+// The class used to hold all Benchmarks created from static function.
```
EricWF: Nice job moving these classes out of the header.

They can probably put in an unnamed namespace now, since we don't really want to leak their symbols.

src/benchmark_register.cc
```diff
@@ -461,16 +463,77 @@ int Benchmark::ArgsCnt() const {
...
+// (ie those created using the BENCHMARK(...) macros.
+class FunctionBenchmark : public Benchmark {
+ public:
+  FunctionBenchmark(const char* name, Function* func)
+      : Benchmark(name), func_(func) {}
+
+  virtual void Run(State& st);
```
EricWF: Lose virtual add `override` to be consistent with below?

EricWF:
Also could you add this test to the bottom of `fixture_test.cpp`:

https://gist.github.com/EricWF/e033d66c024fe65f9e15705b653bd3f7

EricWF:
Also this change should be against the main branch, not v2.

include/benchmark/benchmark.h
lijinpei: This function is in the public api because there is this function https://github.com/google/benchmark/blob/master/include/benchmark/benchmark.h#L856 in the public api, I just keep them similar.  If users decide to write their own main function, and register benchmarks themselves, they may need this function. But class Benchmark itself is in the internal namespace, and I think it's wired for a function in the public api to return or accept parameters of types defined in the internal namespace.

src/benchmark_register.cc
lijinpei: Inline virtual function?  The compiler has to be able to determine the pointee's type at compile time to truly inline the function, as the benchmarks are run()ed through a std::vector<std::unique_ptr<Benchmark>>, I don't think the compiler much of the function call?

src/benchmark_register.cc
lijinpei: OK, I will do this.

include/benchmark/benchmark.h
lijinpei: I will change it to upper case, and use "Init()" if no others object. @dominichamon

lijinpei:
I will fix those problems, add the test and re-pr against the main branch.

include/benchmark/benchmark.h
lijinpei: I will move the three RegisterBenchmark() functions(one for function, one for fixture, and if CXX11, one for lambda) to internal namespace.

lijinpei:
I will add the new test https://gist.github.com/EricWF/e033d66c024fe65f9e15705b653bd3f7 as fixture_init_test.cc, is this OK?

src/benchmark_register.cc
EricWF: What I meant by "define inline" here is that you could put the function body *inside* the class. I don't see much value in re-declaring it only a few lines later.

EricWF:
> I will add the new test https://gist.github.com/EricWF/e033d66c024fe65f9e15705b653bd3f7 as fixture_init_test.cc, is this OK?

Yes.

include/benchmark/benchmark.h
EricWF: Please keep the lambda and function one as part of the public API.

The fixture overload works entirely differently that the existing ones, since you have to pass it a special function pointer *which isn't the benchmark, but a function which creates it*. The fact that this is how we handle registering fixtures feels like an internal implementation detail; not a part of the public API.

EricWF: > But class Benchmark itself is in the internal namespace,

Yeah, that's kind of dumb. I think the initial intent was that users were never supposed to name or create the type themselves; so the name was effectively "internal".

PR #512, opened 2018-01-07
handle fixture construction after static initialization: rev1

lijinpei:
a revision of #511 and #505


include/benchmark/benchmark.h
```diff
@@ -902,13 +889,14 @@ class LambdaBenchmark : public Benchmark {
...
-inline internal::Benchmark* RegisterBenchmark(const char* name,
-                                              internal::Function* fn) {
-  return internal::RegisterBenchmarkInternal(
-      ::new internal::FunctionBenchmark(name, fn));
+#ifdef BENCHMARK_HAS_CXX11
+
+namespace internal {
+// Take ownership of the pointer and register the benchmark. Return the
+// registered benchmark.
+Benchmark* RegisterBenchmarkInternal(Benchmark*);
```
dominichamon: why is this now under HAS_CXX11? I don't think it was before.

include/benchmark/benchmark.h
```diff
@@ -931,27 +919,38 @@ internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn,
...
-  Fixture() : internal::Benchmark("") {}
-
-  virtual void Run(State& st) {
-    this->SetUp(st);
-    this->BenchmarkCase(st);
-    this->TearDown(st);
-  }
-
+  virtual ~Fixture();
   // These will be deprecated ...
```
dominichamon: can you link to an issue here that describes why they'll be deprecated?

src/benchmark_register.cc
```diff
@@ -461,16 +463,78 @@ int Benchmark::ArgsCnt() const {
   return static_cast<int>(args_.front().size());
 }
 
+void Benchmark::Init() {}
```
dominichamon: this should be in benchmark.cc, right?

src/benchmark_register.cc
```diff
@@ -461,16 +463,78 @@ int Benchmark::ArgsCnt() const {
...
+namespace {
+
 //=============================================================================//
 //                            FunctionBenchmark
 //=============================================================================//
 
-void FunctionBenchmark::Run(State& st) { func_(st); }
+// The class used to hold all Benchmarks created from static function.
+// (ie those created using the BENCHMARK(...) macros.
+class FunctionBenchmark : public Benchmark {
```
dominichamon: i'd be quite happy if these could be move out to their own header/source files (in src/).

src/benchmark_register.cc
```diff
@@ -461,16 +463,78 @@ int Benchmark::ArgsCnt() const {
...
+  FixtureCreator* creator_;
+  friend BenchmarkFamilies;
+};
 
+}  // namespace
 }  // end namespace internal
 
+internal::Benchmark* RegisterBenchmark(const char* name, internal::Function* func) {
+  auto* p_benchmark = new internal::FunctionBenchmark(name, func);
+  internal::BenchmarkFamilies::GetInstance()->AddBenchmark(std::unique_ptr<internal::Benchmark>(p_benchmark));
```
dominichamon: line length should be <= 80

src/benchmark_register.cc
```diff
@@ -461,16 +463,78 @@ int Benchmark::ArgsCnt() const {
...
+  return p_benchmark;
+}
+
+namespace internal {
+
+Benchmark* RegisterBenchmark(const char *name, FixtureCreator* creator) {
+  auto* p_benchmark = new FixtureBenchmark(name, creator);
+  internal::BenchmarkFamilies::GetInstance()->AddBenchmark(std::unique_ptr<internal::Benchmark>(p_benchmark));
```
dominichamon: line length

test/fixture_init_test.cc
```diff
@@ -0,0 +1,35 @@
...
+BENCHMARK_REGISTER_F(StaticInitFixtureTest, ConstructionHappensAtRuntime)
+    ->Arg(42);
+
+BENCHMARK_F(StaticInitFixtureTest, ConstructionHappensAtRuntime2)
+(benchmark::State &st) {
+  for (auto _ : st) {
+  }
+}
+
+int main(int argc, char **argv) {
```
dominichamon: maybe you can use googletest for this instead of asserts.

include/benchmark/benchmark.h
lijinpei: Because this is only needed if you use lambda benchmarks, function and fixture benchmarks are registered using the non-internal RegisterBenchmark().

src/benchmark_register.cc
lijinpei: All Benchmakr:: member functions are already in benchmark_register.cc, maybe we shall use a seperate commit to move them to benchmark.cc

src/benchmark_register.cc
lijinpei: OK, maybe I shall also move benchmark's member function definitions to benchmark.cc?

test/fixture_init_test.cc
lijinpei: OK, I will do it.

src/benchmark_register.cc
lijinpei: OK

src/benchmark_register.cc
lijinpei: OK

src/benchmark_register.cc
```diff
@@ -461,16 +463,78 @@ int Benchmark::ArgsCnt() const {
...
+  FunctionBenchmark(const char* name, Function* func)
+      : Benchmark(name), func_(func) {}
+
+  void Run(State& st) override { func_(st); }
+
+ private:
+  Function* func_;
+};
+
+//=============================================================================//
```
lijinpei: just to mention it, all those comments are 81 characters width(not including the newline)

test/fixture_init_test.cc
lijinpei: Of the existing 35 testcases, only one(statistics_test) uses gtest. I add two testcases which does essentially the same thing, fixture_init_test.cpp doesn't use gtest while fixture_init_gtest.cpp does.

src/benchmark_register.cc
dominichamon: huh. yeah, probably move benchmark member functions to benchmark.cc and the stuff there to benchmark_main.cc or something.

include/benchmark/benchmark.h
```diff
@@ -931,27 +919,38 @@ internal::Benchmark* RegisterBenchmark(const char* name, Lambda&& fn,
...
   virtual void TearDown(const State&) {}
   // ... In favor of these.
   virtual void SetUp(State& st) { SetUp(const_cast<const State&>(st)); }
   virtual void TearDown(State& st) { TearDown(const_cast<const State&>(st)); }
 
- protected:
   virtual void BenchmarkCase(State&) = 0;
 };
 
+namespace internal {
```
dominichamon: can this be put inside one of the existing internal namespaces instead of adding another block? at some point i tried to clean these blocks up.. it'd be nice to not make it messier.

test/fixture_init_test.cc
dominichamon: is this file necessary any more?

gtest is a new addition so we're just starting to move testing over to it. you're ahead of the curve :)

EricWF:
@lijinpei When you want to update a PR, please use the same branch and PR on github -- don't create new ones every time.

You can either squash the old commits and force-push yourself, or you can simply push the new commits. When we merge the PR we'll squash the commits into a single one anyway.

src/benchmarks.cc
```diff
@@ -0,0 +1,87 @@
...
+// limitations under the License.
+
+/* This file implements FunctionBenchmark and FixtureBenchmark */
+
+#include <memory>
+#include "benchmark/benchmark.h"
+#include "benchmark_api_internal.h"
+
+namespace {
+using namespace benchmark;
```
EricWF: I would just put the unnamed namespace inside `benchmark::internal` instead of adding `using` statements.

test/fixture_init_gtest.cc
```diff
@@ -0,0 +1,45 @@
...
+  ASSERT_EQ(num_constructed, 0);
+  have_entered_main = true;
+  ::benchmark::RunSpecifiedBenchmarks();
+  ASSERT_EQ(num_constructed, 2);
+}
+
+int main(int argc, char **argv) {
+  ::testing::InitGoogleTest(&argc, argv);
+  benchmark::Initialize(&argc, argv);
+  return RUN_ALL_TESTS();
```
EricWF: The test LGTM. Thanks.

lijinpei:
I can revert to the version which doesn't change the header file very much if you don't want to merge the newest change.

dominichamon:
it would be nice to keep the PR focused. so one PR to handle the issue here, and another to clean up the header would be most appreciated.

dominichamon:
Are you still looking in to this, @lijinpei ?

lijinpei:
I can re-start to work on this.
BTW, on my machine(gcc 8.1.1), state_assembly_test.cc compiles to the following [state_assembly_test.s](https://gist.github.com/lijinpei/58c9113aff3951d2d926350a20a0e678) output, I think it is just a reorder of the loop header/body of what FileCheck means to check, but FileCheck won't accept it.

dominichamon:
I'm going to close this as it appears abandoned. Please feel free to try again.

PR #514, opened 2018-01-09
Wrap COMPILER macros.

### LINKED ISSUE #513: COMPILER_MSVC macro redefinition warning in Windows
When building google benchmark in Windows with Bazel and VS2017 I get the following warning:

src\internal_macros.h(13): warning C4005: 'COMPILER_MSVC': macro redefinition
src\internal_macros.h(13): note: command-line arguments:  see previous definition of 'COMPILER_MSVC'

Just in case, this is my bazel BUILD file for google benchmark:

```
cc_library(
    name = "benchmark",

    srcs = glob(["src/*.cc"]),
    hdrs = glob(["include/**/*.h", "src/*.h"]),

    includes = ["include", "."],

    copts = ["-DHAVE_STD_REGEX"],
    linkstatic = 1,

    visibility = ["//visibility:public"],
)
```

dominichamon:
Some command line or build systems may already set these (eg, bazel) so
make sure that takes priority.

Fixes #513

### DIFF
--- a/src/internal_macros.h
+++ b/src/internal_macros.h
@@ -11,29 +11,35 @@
 #endif
 
 #if defined(__clang__)
-#define COMPILER_CLANG
+  #if !defined(COMPILER_CLANG)
+    #define COMPILER_CLANG
+  #endif
 #elif defined(_MSC_VER)
-#define COMPILER_MSVC
+  #if !defined(COMPILER_MSVC)
+    #define COMPILER_MSVC
+  #endif
 #elif defined(__GNUC__)
-#define COMPILER_GCC
+  #if !defined(COMPILER_GCC)
+    #define COMPILER_GCC
+  #endif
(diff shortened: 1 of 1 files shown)


EricWF:
Should we change the compiler macros to use the `BENCHMARK_` prefix to avoid conflicts?

dominichamon:
I thought about that, but i didn't want to risk confusion in the case that it's set differently.

maybe we could set the BENCHMARK_ one to whatever is set on commandline, and if nothing is set then detect?

EricWF:
I should have clarified that I think that the preventing re-definition part of this patch is correct. Feel free to land this as-is and we can discuss adding a prefix later.

I'm just a bit uncomfortable that certain build systems are defining these macros magically while others are not. So we have slightly different build behavior for each.

PR #517, opened 2018-01-16
Use int64_t throughout to facilitate probing larger inputs.

NAThompson:


NAThompson:
I signed it!

include/benchmark/benchmark.h
```diff
@@ -572,27 +572,27 @@ class State {
...
   // Number of threads concurrently executing the benchmark.
-  const int threads;
+  const int64_t threads;
   const size_t max_iterations;
 
   // TODO(EricWF) make me private
-  State(size_t max_iters, const std::vector<int>& ranges, int thread_i,
-        int n_threads, internal::ThreadTimer* timer,
+  State(size_t max_iters, const std::vector<int64_t>& ranges, int64_t thread_i,
+        int64_t n_threads, internal::ThreadTimer* timer,
```
dominichamon: i'm not sure we need more than 2^32 threads, do we?

src/benchmark.cc
```diff
@@ -293,14 +293,14 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
   size_t iters = has_explicit_iteration_count ? b.iterations : 1;
   std::unique_ptr<internal::ThreadManager> manager;
   std::vector<std::thread> pool(b.threads - 1);
-  const int repeats =
+  const int64_t repeats =
```
dominichamon: repetitions certainly shouldn't head north of 2^32.

src/benchmark_register.cc
```diff
@@ -453,12 +453,12 @@ Benchmark* Benchmark::ThreadPerCpu() {
 
 void Benchmark::SetName(const char* name) { name_ = name; }
 
-int Benchmark::ArgsCnt() const {
+int64_t Benchmark::ArgsCnt() const {
```
dominichamon: also here: if someone is passing this many args around we're in trouble.

include/benchmark/benchmark.h
NAThompson: I got a bunch of compiler warnings while only changing the `Range` arguments. But yes it is ridiculous to have 64-bit threadcounts.

NAThompson:
Looks like the threads are passed to the `AddRange` function:

```
benchmark/src/benchmark_register.cc:432:12: error: cannot initialize a parameter of type 'std::vector<int64_t> *'
      (aka 'vector<long long> *') with an rvalue of type 'std::vector<int> *'
  AddRange(&thread_counts_, min_threads, max_threads, 2);
           ^~~~~~~~~~~~~~~
benchmark/src/benchmark_register.cc:247:48: note: passing argument to parameter 'dst' here
void Benchmark::AddRange(std::vector<int64_t>* dst, int64_t lo, int64_t hi, int64_t mult)
```

So the type of the threads looks like it must be the same as the type as the argument to the `AddRange`, or the `AddRange` function must be templated on an integer. Do you have a preference?

(The other alternative-that my mental model of this code is unsophisticated and I've totally missed the point-is of course the most likely possibility.)

dominichamon:
Oh right: Enthusiastic reuse of code.

Templating AddRange on integer sgtm, given it's a generic utility function.

include/benchmark/benchmark.h
```diff
@@ -1212,7 +1213,7 @@ class BenchmarkReporter {
     bool error_occurred;
     std::string error_message;
 
-    int64_t iterations;
+    int iterations;
```
dominichamon: this was int64_t before and probably still should be.

src/benchmark.cc
```diff
@@ -229,7 +229,7 @@ BenchmarkReporter::Run CreateRunReport(
   report.error_message = results.error_message_;
   report.report_label = results.report_label_;
   // Report the total iterations across all threads.
-  report.iterations = static_cast<int64_t>(iters) * b.threads;
+  report.iterations = static_cast<int>(iters) * b.threads;
```
dominichamon: same here.. keep as int64_t

include/benchmark/benchmark.h
EricWF: What's this change?

src/mutex.h
```diff
@@ -98,7 +98,7 @@ class SCOPED_CAPABILITY MutexLock {
 
 class Barrier {
  public:
-  Barrier(int num_threads) : running_threads_(num_threads) {}
+  Barrier(int64_t num_threads) : running_threads_(num_threads) {}
```
EricWF: We're never going to have more than 4 billion threads.

src/mutex.h
```diff
@@ -120,11 +120,11 @@ class Barrier {
  private:
   Mutex lock_;
   Condition phase_condition_;
-  int running_threads_;
+  int64_t running_threads_;
```
EricWF: All these changes for thread counts seem unneeded.

src/statistics.cc
```diff
@@ -99,7 +99,7 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
 
   // All repetitions should be run with the same number of iterations so we
   // can take this information from the first benchmark.
-  int64_t const run_iterations = reports.front().iterations;
+  int const run_iterations = reports.front().iterations;
```
EricWF: Why are you doing the reverse here, and changing an `int64_t` into an int?

src/json_reporter.cc
```diff
@@ -49,6 +49,12 @@ std::string FormatKV(std::string const& key, int64_t value) {
   return ss.str();
 }
 
+std::string FormatKV(std::string const& key, int value) {
```
dominichamon: if this is for outputting the args (which i don't think we do other than as part of the name) then it should be int64_t value, right?

src/statistics.cc
dominichamon: int64_t

test/benchmark_test.cc
```diff
@@ -72,7 +72,7 @@ static void BM_CalculatePiRange(benchmark::State& state) {
 BENCHMARK_RANGE(BM_CalculatePiRange, 1, 1024 * 1024);
 
 static void BM_CalculatePi(benchmark::State& state) {
-  static const int depth = 1024;
+  static const int64_t depth = 1024;
```
dominichamon: definitely doesn't need to be updated :)

include/benchmark/benchmark.h
NAThompson: I figured the iteration count wouldn't go past 4 billion. It reverts the previous change, which is admittedly annoying.

src/json_reporter.cc
NAThompson: This fixed a build error so that either type would work.

NAThompson: J/K, that error went away once I made your other suggested changes.

dominichamon:
works for me once tests pass and @pleroy and @EricWF have weighed in.

i think there's a broader approach here to templatize range args but i may try that after this lands to see if it works out.

dominichamon:
I think this PR has been obsoleted by other changes.

PR #519, opened 2018-01-25
resolves #516: update Range argument size from int to int64_t

### LINKED ISSUE #516: `Range(int start, int limit)`-maximum int is too small for number theory.
Many number theoretic algorithms are O(log(log(N)). When benchmarking these using (say)

    BENCHMARK(BM_FloorSqrt)->Range(2, 9223372036854775808LL)->RangeMultiplier(2)->Complexity();

the upper limit overflows. Using only 32 bits for the upper range does not give very much test data.

Changing the arguments to `Range` to `long long`s or `size_t`s would dramatically expand the range or data values that can be probed.

manparvesh:
This PR updates Range argument size from int to int64_t


manparvesh:
Sorry, it seems that there are some things that I missed. I'll close this and create another PR when I'm done.

PR #521, opened 2018-01-31
Support State::KeepRunningBatch().

sam-panzer:
State::KeepRunning() can take large amounts of time relative to quick
operations (on the order of 1ns, depending on hardware). For such
sensitive operations, it is recommended to run batches of repeated
operations.

This commit simplifies handling of total_iterations_. Rather than
predecrementing such that total_iterations_ == 1 signals that
KeepRunning() should exit, total_iterations_ == 0 now signals the
intention for the benchmark to exit.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -623,15 +623,16 @@ bool State::KeepRunning() {
     --total_iterations_;
     return true;
   }
-  if (!started_ && !error_occurred_) {
+  if (!started_) {
     StartKeepRunning();
-    // max_iterations > 0. The first iteration is always valid.
-    --total_iterations_;
-    return true;
-  } else {
-    FinishKeepRunning();
-    return false;
+    if (!error_occurred_) {
+      // max_iterations > 0. The first iteration is always valid.
+      --total_iterations_;
+      return true;
+    }
   }
+  FinishKeepRunning();
+  return false;
 }
 
 inline BENCHMARK_ALWAYS_INLINE
@@ -642,24 +643,25 @@ bool State::KeepRunningBatch(size_t n) {
     total_iterations_ -= n;
     return true;
   }
-  if (!started_ && !error_occurred_) {
+  if (!started_) {
     StartKeepRunning();
-    if (total_iterations_ >= n) {
-      total_iterations_-= n;
-    } else {
-      batch_leftover_  = n - total_iterations_;
-      total_iterations_ = 0;
+    if (!error_occurred_) {
+      if (total_iterations_ >= n) {
+        total_iterations_-= n;
+      } else {
+        batch_leftover_  = n - total_iterations_;
+        total_iterations_ = 0;
+      }
+      return true;
     }
-    return true;
   }
   if (total_iterations_ != 0) {
     batch_leftover_  = n - total_iterations_;
     total_iterations_ = 0;
     return true;
-  } else {
-    FinishKeepRunning();
-    return false;
   }
+  FinishKeepRunning();
+  return false;
 }
 
 struct State::StateIterator {
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -172,6 +172,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 
 #include <stdint.h>
 
+#include <algorithm>
 #include <cassert>
 #include <cstddef>
 #include <iosfwd>
@@ -429,16 +430,18 @@ class State {
   // Returns true if the benchmark should continue through another iteration.
   // NOTE: A benchmark may not return from the test until KeepRunning() has
   // returned false.
-  bool KeepRunning() {
-    if (BENCHMARK_BUILTIN_EXPECT(!started_, false)) {
-      StartKeepRunning();
-    }
-    bool const res = (--total_iterations_ != 0);
-    if (BENCHMARK_BUILTIN_EXPECT(!res, false)) {
-      FinishKeepRunning();
-    }
-    return res;
-  }
+  bool KeepRunning();
+
+  // Returns true iff the benchmark should run n more iterations.
+  // NOTE: A benchmark must not return from the test until KeepRunningBatch()
+  // has returned false.
+  // NOTE: KeepRunningBatch() may overshoot by up to 'n' iterations.
+  //
+  // Intended usage:
+  //   while (state.KeepRunningBatch(1000)) {
+  //     // process 1000 elements
+  //   }
+  bool KeepRunningBatch(int n);
 
   // REQUIRES: timer is running and 'SkipWithError(...)' has not been called
   //           by the current thread.
@@ -565,12 +568,15 @@ class State {
   int range_y() const { return range(1); }
 
   BENCHMARK_ALWAYS_INLINE
-  size_t iterations() const { return (max_iterations - total_iterations_) + 1; }
+  size_t iterations() const { return (completed_iterations_ - total_iterations_); }
 
  private:
   bool started_;
   bool finished_;
+  // When total_iterations_ is 0, KeepRunning() and friends will return false.
   size_t total_iterations_;
+  // May be larger than max_iterations.
+  size_t completed_iterations_;
 
   std::vector<int> range_;
 
@@ -580,6 +586,11 @@ class State {
   int complexity_n_;
 
   bool error_occurred_;
+
+  // When using KeepRunningBatch(), batch_leftover_ holds the number of
+  // iterations beyond max_iters that were run. Used to track
+  // completed_iterations_ accurately.
+  size_t batch_leftover_;
 
  public:
   // Container for user-defined counters.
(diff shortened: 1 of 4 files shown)


include/benchmark/benchmark.h
```diff
@@ -429,16 +429,18 @@ class State {
...
+  // Returns true iff the benchmark should run n more iterations.
+  // NOTE: A benchmark must not return from the test until KeepRunningBatch()
+  // has returned false.
+  // NOTE: KeepRunningBatch() may overshoot by up to 'n' iterations.
+  //
+  // Intended usage:
+  //   while (state.KeepRunningBatch(1000)) {
+  //     // process 1000 elements
+  //   }
+  bool KeepRunningBatch(size_t n);
```
dominichamon: this should likely be an `int`. `int64_t` if we expect very large batches.

include/benchmark/benchmark.h
```diff
@@ -597,12 +607,41 @@ class State {
...
+  if (BENCHMARK_BUILTIN_EXPECT(!res, false)) {
+    FinishKeepRunning(batch_leftover_);
+  }
+  // If this batch has 0 < n < total_iterations iterations, (i.e. it's the last
+  // batch), track how many extra iterations past max_iters we ran.
+  batch_leftover_ = leftover;
+  return res;
+}
+
+inline bool State::KeepRunning() {
```
dominichamon: BENCHMARK_ALWAYS_INLINE ?

include/benchmark/benchmark.h
sam-panzer: Done.

include/benchmark/benchmark.h
sam-panzer: Done.

sam-panzer:
I appreciate the quick review! This is my first time using Github in years, so I apologize in advance for any procedural mistakes.

EricWF:
Ide like to take a look at this, but I'm busy until Friday.

EricWF:
So, this patch actually results in worse codegen for the `KeepRunning` loop as a result of adding the extra `completed_iterations_` variable. Every loop now includes an extra load from memory.

Sample Code:

```c++
using namespace benchmark;
extern "C" void foo();
void BM_TestKeepRunning(State &S) {
  while (S.KeepRunning())
    foo();
}
```

ASM Before Change:
```asm
BM_TestKeepRunning(benchmark::State&): # @BM_TestKeepRunning(benchmark::State&)
  pushq %rbx
  movq %rdi, %rbx
  cmpb $1, (%rbx)
  je .LBB0_3
  jmp .LBB0_2
.LBB0_4: # %while.body
  callq foo
  cmpb $1, (%rbx)
  jne .LBB0_2
.LBB0_3: # %if.end.i
  addq $-1, 8(%rbx)
  jne .LBB0_4
  jmp .LBB0_5
.LBB0_2: # %if.then.i
  movq %rbx, %rdi
  callq benchmark::State::StartKeepRunning()
  addq $-1, 8(%rbx)
  jne .LBB0_4
.LBB0_5: # %while.end
  movq %rbx, %rdi
  popq %rbx
  jmp benchmark::State::FinishKeepRunning() # TAILCALL
```

ASM After Change:
```asm
BM_TestKeepRunning(benchmark::State&): # @BM_TestKeepRunning(benchmark::State&)
  pushq %rbx
  movq %rdi, %rbx
  cmpb $1, (%rbx)
  je .LBB0_3
  jmp .LBB0_2
.LBB0_4: # %while.body
  callq foo
  cmpb $1, (%rbx)
  jne .LBB0_2
.LBB0_3: # %if.end.i
  movq 8(%rbx), %rax # Worse codegen starts here.
  leaq -1(%rax), %rcx
  movq %rcx, 8(%rbx)
  testq %rax, %rax
  jne .LBB0_4
  jmp .LBB0_5
.LBB0_2: # %if.then.i
  movq %rbx, %rdi
  callq benchmark::State::StartKeepRunning()
  jmp .LBB0_3
.LBB0_5: # %while.end
  xorl %esi, %esi
  movq %rbx, %rdi
  popq %rbx
  jmp benchmark::State::FinishKeepRunning(unsigned long) # TAILCALL
```

The  simple benchmark doesn't exactly show a concerning performance difference in terms of "reported iteration time", but I would still like to keep this loop as tight as possible (and an empty benchmark runs fewer iterations than it previously did).

More investigation/suggestions coming later.

EricWF:
I'm wondering if it might be better to formulate this as a `for` loop with special iterators. By using iterators to count we can avoid the problem of non optimizable memory accesses in `State`.

 For example:

```c++
for (auto It=State.begin_batch(1000); It != State.end_batch(); ++It) {
  /* Do Stuff */
}
```

sam-panzer:
My goal with this PR is to unify the open-source Google Benchmark API with the internal version. KeepRunningBatch() is missing from the open-source version.

I've taken a cue from the internal version to create a fast path in the common case that the benchmark has already started and will continue running. What do you think?

include/benchmark/benchmark.h
dominichamon: still should be an int or int64_t. So should the iterations tracking really. I'd say leave this, but i'd like the public API not to change very often, even in the signedness of a parameter.

include/benchmark/benchmark.h
```diff
@@ -642,24 +643,25 @@ bool State::KeepRunningBatch(size_t n) {
...
-  if (!started_ && !error_occurred_) {
+  if (!started_) {
     StartKeepRunning();
-    if (total_iterations_ >= n) {
-      total_iterations_-= n;
-    } else {
-      batch_leftover_  = n - total_iterations_;
-      total_iterations_ = 0;
+    if (!error_occurred_) {
+      if (total_iterations_ >= n) {
```
dominichamon: how do we get here? on line 642 if total_iterations_ >= n, we return. if we're here then total_iterations_ must be < n.

i think you can avoid this check and improve the code generation.

sam-panzer: Good question. total_iterations_ is now set to 0 when State is initialized. StartKeepRunning() can set total_iterations_ to any positive number of iterations. For example, if max_iterations is 10 and KeepRunningBatch() is given a batch size of 100, we should hit this path. I see that I can fold it into the  if block just above the call to FinishKeepRunning(), though I wouldn't be surprised if it doesn't change generated code with optimizations enabled. Does this address your comment, or were you looking for something else?

On another front, how much does code generation matter this far down? The fast path should be hit every time within a benchmark except for up to two calls:
 1. The first call, which starts the benchmark
 2. The last call, either due to error or because it's the last iteration/batch.

include/benchmark/benchmark.h
```diff
@@ -603,6 +615,50 @@ class State {
...
+    --total_iterations_;
+    return true;
+  }
+  if (!started_) {
+    StartKeepRunning();
+    if (!error_occurred_) {
+      // max_iterations > 0. The first iteration is always valid.
+      --total_iterations_;
+      return true;
+    }
```
dominichamon: I think you can replace this with something like:

```c++
bool State::KeepRunningInternal(int batch) {
  if (BENCHMARK_BUILTIN_EXPECT(total_iterations_ >= n, true)) {
    total_iterations -= n;
    return true;
  }
  if (!started_) {
    StartKeepRunning();
    if (!error_occurred_ && total_iterations_ >= n) {
      total_iterations_ -= n;
      return true;
    }
  }
  if (total_iterations_ > 0) {
    batch_leftover_ = n - total_iterations_;
    total_iterations_ = 0;
    return true;
  }
  FinishKeepRunning();
  return false;
}

bool State::KeepRunning() { return KeepRunningInternal(1); }
bool State::KeepRunningBatch(int n) { return KeepRunningInternal(n); }
```

EricWF:
This LGTM. The codegen looks amazing! Thanks.

EricWF:
@dominichamon Woops, sorry I missed your last comment. Could we address it in a follow up commit since I just merged this change :-S

@sam-panzer ^

include/benchmark/benchmark.h
```diff
@@ -429,16 +430,18 @@ class State {
...
-    bool const res = (--total_iterations_ != 0);
-    if (BENCHMARK_BUILTIN_EXPECT(!res, false)) {
-      FinishKeepRunning();
-    }
-    return res;
-  }
+  bool KeepRunning();
+
+  // Returns true iff the benchmark should run n more iterations.
+  // NOTE: A benchmark must not return from the test until KeepRunningBatch()
```
EricWF: We should note the requirement that `n != 0` since we'll enter an infinite loop if a user passes `0`. And users are silly, they might attempt to do that.

include/benchmark/benchmark.h
sam-panzer: I'll send a PR to implement that.

include/benchmark/benchmark.h
sam-panzer: I'll send a PR to assert and note it.

sam-panzer: Fixes are in https://github.com/google/benchmark/pull/526.

PR #523, opened 2018-02-04
Ignore 32bit build option when using msvc

### LINKED ISSUE #447: CMakeLists.txt BENCHMARK_BUILD_32_BITS fails with MSVC
CMakeLists.txt has an option BENCHMARK_BUILD_32_BITS. This works when gcc is used, but it fails with MSVC. In fact this option tries to add -m32 to the compiler options, and this is gcc-specific and doesn't work with MSVC.

With MSVC this option doesn't really make sense, because the CMake MSVC generator determines whether the build is 32-bit or 64-bit. An MSVC user probably doesn't want or expect that setting BENCHMARK_BUILD_32_BITS would make the build fail. Therefore, I propose that BENCHMARK_BUILD_32_BITS be ignored when using MSVC. Doing this is a trivial change to CMakeLists.txt, moving the -m32 stuff so that it is not used with MSVC.

Here's a diff:

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 4265a6c..74fe2ea 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,10 +37,6 @@ include(CheckCXXCompilerFlag)
 include(AddCXXCompilerFlag)
 include(CXXFeatureCheck)

-if (BENCHMARK_BUILD_32_BITS)
-  add_required_cxx_compiler_flag(-m32)
-endif()
-
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   # Turn compiler warnings up to 11
   string(REGEX REPLACE "[-/]W[1-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
@@ -72,6 +68,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     set(CMAKE_EXE_LINKER_FLAGS_MINSIZEREL "${CMAKE_EXE_LINKER_FLAGS_MINSIZEREL} /LTCG")
   endif()
 else()
+  if (BENCHMARK_BUILD_32_BITS)
+    add_required_cxx_compiler_flag(-m32)
+  endif()
+
   # Try and enable C++11. Don't use C++14 because it doesn't work in some
   # configurations.
   add_cxx_compiler_flag(-std=c++11)

aJetHorn:
Resolves #447

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmark library." ON)
 option(BENCHMARK_ENABLE_LTO "Enable link time optimisation of the benchmark library." OFF)
 option(BENCHMARK_USE_LIBCXX "Build and test using libc++ as the standard library." OFF)
-option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library." OFF)
+option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library." OFF) # Ignored if using MSVC
 option(BENCHMARK_ENABLE_INSTALL "Enable installation of benchmark. (Projects embedding benchmark may want to turn this OFF.)" ON)
 
 # Allow unmet dependencies to be met using CMake's ExternalProject mechanics, which
@@ -48,7 +48,8 @@ include(CheckCXXCompilerFlag)
 include(AddCXXCompilerFlag)
 include(CXXFeatureCheck)
 
-if (BENCHMARK_BUILD_32_BITS)
+# Exclude -m32 when using MSVC as it is incompatible
+if (BENCHMARK_BUILD_32_BITS AND NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   add_required_cxx_compiler_flag(-m32)
 endif()
 


EricWF:
This seems like an odd approach to me. Why are people specifying this option for compilers that don't support it?

Looking at how LLVM handles this, I think their approach is more correct. For example, IDK if your patch correctly handles clang-cl on Windows. Perhaps the check should be:
```
 if( CMAKE_SIZEOF_VOID_P EQUAL 8 AND NOT WIN32 )
```

CMakeLists.txt
```diff
@@ -48,7 +48,8 @@ include(CheckCXXCompilerFlag)
 include(AddCXXCompilerFlag)
 include(CXXFeatureCheck)
 
-if (BENCHMARK_BUILD_32_BITS)
+# Exclude -m32 when using MSVC as it is incompatible
+if (BENCHMARK_BUILD_32_BITS AND NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
```
chfast: You can simplify it to `NOT MSVC`.

dominichamon:
I don't mind the belt-and-braces approach here. Can you simplify the check as per the comment from @chfast?

EricBackus:
I'd really like to see this merged into the official release, as it fixes a build issue that we've run into several times within Keysight. Any chance this can be merged in?

Croydon:
@dominichamon @chfast Is simplifying this condition the only reason why this didn't get merged yet?

If so I would be happy to open a new pull request.

dominichamon:
@Croydon yes.

PR #526, opened 2018-02-12
Fixups following 296ec5693e7dae47c50244a70297172fab69ccae

sam-panzer:


dominichamon:
LGTM. @EricWF ?

EricWF:
@sam-panzer This again [generates worse assembly](https://gist.github.com/EricWF/afd29e7620d71d7b159f37d8038dc193) for the `KeepRunning` case, although the assembly is never taken (so naively it should have no impact on performance). 

As @dominichamon discussed, plenty of people read the assembly the library generates, so it would be nice to keep it as small and simple as possible.

However, there is a way to fix this. Having both `KeepRunning` and `KeepRunningBatch` call an internal function `KeepRunningImpl(size_t n, bool is_batch)` seems to allow the compiler to eliminate the bits only needed for batch the batch function. I tested this and it does solve the problem.

Could we go ahead and implement that?

sam-panzer:
Indeed. I tried to mention the worse codegen above, but it looks like the comment didn't go through. As you suggested, a separate is_batch parameter restored nicer codegen.

EricWF:
@sam-panzer Can you add a dummy commit to force the bots to re-build? I should have fixed the problems making them spuriously fail.

sam-panzer:
@EricWF Done. The build bots seem happy now. Thanks for fixing them!

PR #527, opened 2018-02-12
Improve State packing: put important members on first cache line.

EricWF:
This patch does a few different things to ensure commonly accessed
data is on the first cache line of the `State` object.

First, it moves the `error_occurred_` member to reside after
the `started_` and `finished_` bools, since there was internal
padding there that was unused.

Second, it moves `batch_leftover_` and `max_iterations` further up
in the struct declaration. These variables are used in the calculation
of `iterations()` which users might call within the loop. Therefore
it's more important they exist on the first cache line.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -573,9 +573,9 @@ class State {
   }
 
 private: // items we expect on the first cache line (ie 64 bytes of the struct)
-  bool started_ : 1;
-  bool finished_ : 1;
-  bool error_occurred_ : 1;
+  bool started_;
+  bool finished_;
+  bool error_occurred_;
   // When total_iterations_ is 0, KeepRunning() and friends will return false.
   // May be larger than max_iterations.
   size_t total_iterations_;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -573,9 +573,7 @@ class State {
   }
 
 private: // items we expect on the first cache line (ie 64 bytes of the struct)
-  bool started_;
-  bool finished_;
-  bool error_occurred_;
+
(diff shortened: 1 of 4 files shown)


include/benchmark/benchmark.h
```diff
@@ -572,35 +572,38 @@ class State {
     return (max_iterations - total_iterations_ + batch_leftover_);
   }
 
- private:
-  bool started_;
-  bool finished_;
+private: // items we expect on the first cache line (ie 64 bytes of the struct)
+  bool started_ : 1;
```
dominichamon: how about not doing this yet, but adding a "TODO: make these bitfields if required".

reasoning: on some platforms (at least this used to be true), bitfield accesses required quite a bit more ASM than bool accesses. Ie, the bitfield work might negate the work to get these in the cache line in the first place.

EricWF: Sounds reasonable to me.

Perhaps we could still attack the low hanging fruit and enable it on x86?

dominichamon:
do you have any benchmark results for this change in benchmarks? :)

EricWF:
> do you have any benchmark results for this change in benchmarks? :)

Ha. Ironically I don't -- at least not a portable one. But when I get around to adding `FlushCache(ptr)` it should be easy enough to write. Until then, this ia32 specific benchmark shows a difference of about 2x. From `140 ns` to `70 ns`.

```c++
static void BM_Flush(State &S) {
  DoNotOptimize(&S);
  while (S.KeepRunning()) {
    __builtin_ia32_clflush(&S.threads); // the object previously before `max_iterations`
    DoNotOptimize(S.iterations());
    ClobberMemory();
  }
}
BENCHMARK(BM_Flush)->Repetitions(3);

```

dominichamon:
LGTM. Merge when ready.

PR #528, opened 2018-02-13
Attempt to fix travis timeouts during apt-get.

EricWF:
During some builds, travis fails to update the apt-get indexes.
This causes the build to fail in different ways.

This patch attempts to avoid this issue by manually calling
apt-get update. I'm not sure if it'll work, but it's worth a try.


EricWF:
The travis bot seems to be good. Re-running to see if I can trigger a failure again.

If not I'll merge this tonight or tomorrow.

PR #530, opened 2018-02-14
Add tests to verify assembler output -- Fix DoNotOptimize.

EricWF:
For things like `DoNotOptimize`, `ClobberMemory`, and even `KeepRunning()`,
it is important exactly what assembly they generate. However, we currently
have no way to test this. Instead it must be manually validated every
time a change occurs -- including a change in compiler version.

This patch attempts to introduce a way to test the assembled output automatically.
It's mirrors how LLVM verifies compiler output, and it uses LLVM FileCheck to run
the tests in a similar way.

The tests function by generating the assembly for a test in CMake, and then
using FileCheck to verify the // CHECK lines in the source file are found
in the generated assembly.

Currently, the tests only run on 64-bit x86 systems under GCC and Clang,
and when FileCheck is found on the system.

Additionally, this patch tries to improve the code gen from DoNotOptimize. This should probably be a separate change, but I needed something to test.

This PR is largely just to run the new patch against the bots to ensure nothing has broken.


dominichamon:
Can you add a markdown documentation file about AssemblyTests or whatever name you choose, essentially with the contents of your commit message. Maybe more detail.

And then cross-link to it from the README?

pleroy:
TIL about LLVM FileCheck.  Pretty cool!

EricWF:
@dominichamon Yep. Can do.

@pleroy:
> TIL about LLVM FileCheck. Pretty cool!

Indeed. It's an amazing tool to have when you need to test string output.

EricWF:
@dominichamon I added documentation as requested. Please take a look at your leisure.

docs/AssemblyTests.md
```diff
@@ -0,0 +1,147 @@
+# Assembly Tests
+
+The Benchmark library provides a number of functions who's primary
```
dominichamon: 'whose'

docs/AssemblyTests.md
```diff
@@ -0,0 +1,147 @@
...
+    return ExternInt + 1;
+
+    // CHECK: movl ExternInt(%rip), %eax
+    // CHECK: addl %eax
+    // CHECK: ret
+}
+
+```
+
+#### LLVM Filecheck
```
dominichamon: this should be ### i think, as you're only at ## above.

EricWF: If it's no matter, I actually prefer having the smaller section headers for these bits. IMHO it's looks nicer and flows better.

EricWF:
@dominichamon Any final words?

dominichamon:
I'd prefer if the bazel TODO linked to a github issue so we can track it, but otherwise, Ship It.

PR #531, opened 2018-02-14
Don't include <sys/resource.h> on Fuchsia.

ianloic:
It doesn't support POSIX resource measurement and timing APIs.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -17,7 +17,9 @@
 #include "internal_macros.h"
 
 #ifndef BENCHMARK_OS_WINDOWS
+#ifndef __Fuchsia__
 #include <sys/resource.h>
+#endif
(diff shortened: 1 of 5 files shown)


src/benchmark.cc
```diff
@@ -17,7 +17,9 @@
 #include "internal_macros.h"
 
 #ifndef BENCHMARK_OS_WINDOWS
+#ifndef __Fuchsia__
```
dominichamon: can you define a BENCHMARK_OS_FUSCHIA alongside the BENCHMARK_OS_XXX macros elsewhere? that'll help be consistent and track any other per-OS changes.

ianloic: Ok, done.

EricWF:
This LGTM. Just waiting on the bots.

PR #501, opened 2017-12-14
Build benchmark and all tests with Bazel without modifying C++ source code

qzmfranklin:
This PR attempts to build all libraries and tests of benchmark with Bazel without modifying the source code.

Tested on a Ubuntu 16.04 x86_64 machine with GCC 4.7.0 and clang 5.0.0:
```
bazel test :all
```

Also, this BUILD file is written to allow this repository to be included in other Bazel based projects in source format.

Of the 17 tests, 16 are passing as of this writing.

The failed one, `user_counters_tabular_test` has the following error log:
```
Testing ConsoleReporter Output
-------------------------------
2017-12-14 21:49:08
Run on (8 X 3900 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** Library was built as DEBUG. Timings may be affected.
test/output_test_helper.cc:97: CheckCase: Check `TC.match_rule != MR_Next' failed. Expected line "Benchmark                                   Time           CPU Iterations UserCounters..." to match regex "^Benchmark %s Time %s CPU %s Iterations %s Bar %s Bat %s Baz %s Foo %s Frob %s Lob$"
    actual regex string "^Benchmark[ ]+Time[ ]+CPU[ ]+Iterations[ ]+Bar[ ]+Bat[ ]+Baz[ ]+Foo[ ]+Frob[ ]+Lob$"
    started matching near: Benchmark                                   Time           CPU Iterations UserCounters...
------------------------------------------------------------------------------------------
Benchmark                                   Time           CPU Iterations UserCounters...
------------------------------------------------------------------------------------------
BM_Counters_Tabular/threads:1               2 ns          2 ns  287283167 Bar=2 Bat=8 Baz=4 Foo=1 Frob=16 Lob=32
BM_Counters_Tabular/threads:2               1 ns          3 ns  296119936 Bar=2 Bat=8 Baz=4 Foo=1 Frob=16 Lob=32
BM_Counters_Tabular/threads:4               1 ns          2 ns  273998224 Bar=2 Bat=8 Baz=4 Foo=1 Frob=16 Lob=32
BM_Counters_Tabular/threads:8               0 ns          3 ns  285606144 Bar=2 Bat=8 Baz=4 Foo=1 Frob=16 Lob=32
[...]

PR #533, opened 2018-02-21
Add support for building with Bazel.

jmillikin-stripe:
Limitations compared to existing CMake rules:
* Defaults to using C++11 `<regex>`, with an override via Bazel flag `--define` of `google_benchmark.have_regex`. The TravisCI config sets the regex implementation to `posix` because it uses ancient compilers.
* Debug vs Opt mode can't be set per test. TravisCI runs all the tests in debug mode to satisfy `diagnostics_test`, which depends on `CHECK` being live.

If accepted, this would obsolete PRs https://github.com/google/benchmark/pull/496 and https://github.com/google/benchmark/pull/501.

I plan to use this for [Envoy](https://github.com/envoyproxy/envoy) to replace [build_recipes/benchmark.sh](https://github.com/envoyproxy/envoy/blob/22e0d08309bdb904287d0cb0f9c8550f35ecbb2b/ci/build_container/build_recipes/benchmark.sh).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,3 +1,5 @@
+workspace(name = "com_github_google_benchmark")
+
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 git_repository(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,7 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "com_google_googletest",
+    commit = "3f0cf6b62ad1eb50d8736538363d3580dd640c3e",  # HEAD
+    remote = "https://github.com/google/googletest",
+)
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,9 @@ build.ninja
 install_manifest.txt
 rules.ninja
 
+# bazel output symlinks.
+bazel-*
+
 # out-of-source build top-level folders.
 build/
 _build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -159,11 +159,22 @@ install:
       brew update;
       brew install gcc@7;
     fi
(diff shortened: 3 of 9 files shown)


jmillikin-stripe:
cc @dominichamon

Dominic, will you have time to review this PR, or should I find someone else?

WORKSPACE
```diff
@@ -0,0 +1,7 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
```
dominichamon: as per bazel, it may be worth changing this to http_archive: https://docs.bazel.build/versions/master/be/workspace.html

dominichamon:
Other than the minor thing above (which could be a followup) I'm comfortable with this PR. And thank you for including the tests and travis changes.

It seems that maintenance is relatively sane too given the use of globs.

Let me know if you want to replace the git_repository or not and then I'll merge.

WORKSPACE
jmillikin-stripe: I'd like to leave this as `git_repository`, because the auto-generated tarballs created by GitHub don't have stable checksums. See https://github.com/tensorflow/tensorflow/issues/12979 and https://github.com/bazelbuild/bazel/issues/3737 for some examples of GitHub breaking builds by tweaking their tarball builder.

`http_repository` does work well for GitHub "release" tarballs uploaded by maintainers, such as the Protobuf releases. Googletest doesn't have any of those, though.

jmillikin-stripe:
That AppVeyor failure looks unrelated to this PR, since it occurred in the cmake build.

dominichamon:
I agree, it looks like the appveyor failure is due to flakiness in the regex where the tests run slower than expected.

dominichamon:
Leaving this a day for more comments... @pleroy @EricWF any thoughts?

pleroy:
Looks good.  Thanks for doing this.

dominichamon:
thank you!

PR #535, opened 2018-02-21
Fix typo in README.md

jwakely:


jwakely:
I don't understand why a Travis check failed, but it doesn't seem to be caused by my change (which only touches the README).

dominichamon:
thanks!

Issue #67, opened 2014-11-12
Example fails to run (std::system_error)


DaimonPl:
Tried to run first example (master @ d750144542de0f35321f066f03afc8d873122c6d) and got following result:

```
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
```

As I checked any of following lines (even when run alone) cause that problem:

```
benchmark::Initialize(&argc, (const char **) argv);
benchmark::RunSpecifiedBenchmarks();
```

Full file for reference:

```
#include <string>
#include "benchmark/benchmark.h"

static void BM_StringCreation(benchmark::State& state) {
  while (state.KeepRunning())
    std::string empty_string;
}
// Register the function as a benchmark
BENCHMARK(BM_StringCreation);

// Define another benchmark
static void BM_StringCopy(benchmark::State& state) {
  std::string x = "hello";
  while (state.KeepRunning())
    std::string copy(x);
}
BENCHMARK(BM_StringCopy);

// Augment the main() program to invoke benchmarks if specified
// via the --benchmarks command line flag.  E.g.,
//       my_unittest --benchmark_filter=all
//       my_unittest --benchmark_filter=BM_StringCreation
//       my_unittest --benchmark_filter=String
//       my_unittest --benchmark_filter='Copy|Creation'
int main(int argc, char** argv) {
  benchmark::Initialize(&argc, (const char **) argv);
  benchmark::RunSpecifiedBenchmarks();
  return 0;
}
```

facontidavide:
my fault... linking pthread fixed the problem.

odeits:
`g++ -o google_benchmark google_benchmark.cpp -Wall -Wextra --std=c++11 -lbenchmark -lpthread` worked but
`g++ -o google_benchmark google_benchmark.cpp -Wall -Wextra --std=c++11  -lpthread -lbenchmark` did not... apparently if you link pthreads BEFORE benchmark bad things happen.

jwakely:
@odeits you should use GCC's `-pthread` option instead of linking with `-lpthread` (GCC will ensure it's at the right place in the link command).

PR #536, opened 2018-02-21
Update note about linking with pthreads

jwakely:
Since GCC 5.1 using `std::thread` without linking to libpthread causes a linker error, not runtime exceptions. This was changed by https://gcc.gnu.org/PR61841 in 2014, so update the README info.


jwakely:
IMHO referring to issue #67 here isn't actually very helpful, since that issue contains lots of back and forth trying to identify the problem, when the solution is simply "you need to use `-pthread`".

dominichamon:
thank you!

PR #539, opened 2018-02-26
Add Solaris support

alekseyshl:
Define BENCHMARK_OS_SOLARIS for Solaris.

Platform specific implementations added:
* number of CPUs detection
* CPU cycles per second detection
* Thread CPU usage

### DIFF
--- a/src/timers.cc
+++ b/src/timers.cc
@@ -122,10 +122,6 @@ double ProcessCPUUsage() {
   // same as total time, but this is ok because there aren't long-latency
   // syncronous system calls in Emscripten.
   return emscripten_get_now() * 1e-3;
-#elif defined(BENCHMARK_OS_SOLARIS)
-  struct rusage ru;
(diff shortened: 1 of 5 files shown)


src/timers.cc
```diff
@@ -122,6 +122,10 @@ double ProcessCPUUsage() {
   // same as total time, but this is ok because there aren't long-latency
   // syncronous system calls in Emscripten.
   return emscripten_get_now() * 1e-3;
+#elif defined(BENCHMARK_OS_SOLARIS)
+  struct rusage ru;
```
EricWF: Isn't this identical to the `#else` case? Is that intended?

alekseyshl: Yes, it is, thank you for catching it. Removed.

EricWF:
LGTM. I'll leave this for 24 hours to see if anybody else has any comments, but after that I think it's good to merge.

PR #544, opened 2018-03-06
Rename StrCat to StringCat

winksaville:
StrCat is a macro in Shlwapi.h that conflicts with StrCat in string_util.h:

   #defines StrCat lstrcatA

When Shlwapi.h is included in sysinfo.cc and all references to StrCat are
renamed to lstrcatA in string_util.h and sysinfo.cc, therefore this
renaming is innocuous, although unintended.

But if string_util.h is moved above Shlwapi.h then errors occur because
StrCat is not renamed in string_util.h but are renamed in sysinfo.cc:

  ClCompile:
    C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.12.25827\bin\HostX86\x64\CL.exe /c /I"C:\Users\winks\prgs\google-benchmark\include" /I"C:\Users\winks\prgs\google-benchmark\src" /I"C:\Users\winks\prgs\google-benchmark\src\..\include" /Zi /nologo /W4 /WX- /diagnostics:classic /Od /Ob0 /D WIN32 /D _WINDOWS /D _CRT_SECURE_NO_WARNINGS /D HAVE_STD_REGEX /D HAVE_STEADY_CLOCK /D "CMAKE_INTDIR=\"Debug\"" /D _MBCS /Gm- /EHsc /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR /Fo"benchmark.dir\Debug\\" /Fd"benchmark.dir\Debug\benchmark.pdb" /Gd /TP /errorReport:queue "C:\Users\winks\prgs\google-benchmark\src\sysinfo.cc"
    sysinfo.cc
  C:\Users\winks\prgs\google-benchmark\src\sysinfo.cc(228): error C2660: 'lstrcatA': function does not take 4 arguments [C:\Users\winks\prgs\google-benchmark\build\src\benchmark.vcxproj]
  C:\Users\winks\prgs\google-benchmark\src\sysinfo.cc(229): error C2664: 'LPSTR lstrcatA(LPSTR,LPCSTR)': cannot convert argument 1 from 'std::string' to 'LPSTR' [C:\Users\winks\prgs\google-benchmark\build\src\benchmark.vcxproj]
  ...
  Done Building Project "C:\Users\winks\prgs\google-benchmark\build\test\basic_test.vcxproj" (default targets) -- FAILED.

This PR renames StrCat to StringCat in string_util.h, sysinfo.cc and
console_reporter.cc to fix this issue.

### DIFF
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -117,14 +117,14 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   // Format bytes per second
   std::string rate;
   if (result.bytes_per_second > 0) {
-    rate = StrCat(" ", HumanReadableNumber(result.bytes_per_second), "B/s");
+    rate = StringCat(" ", HumanReadableNumber(result.bytes_per_second), "B/s");
   }
 
   // Format items per second
   std::string items;
   if (result.items_per_second > 0) {
-    items =
-        StrCat(" ", HumanReadableNumber(result.items_per_second), " items/s");
+    items = StringCat(" ", HumanReadableNumber(result.items_per_second),
+        " items/s");
   }
 
   const double real_time = result.GetAdjustedRealTime();
--- a/src/string_util.h
+++ b/src/string_util.h
@@ -26,7 +26,7 @@ inline std::ostream& StringCatImp(std::ostream& out, First&& f,
 }
 
 template <class... Args>
-inline std::string StrCat(Args&&... args) {
+inline std::string StringCat(Args&&... args) {
   std::ostringstream ss;
   StringCatImp(ss, std::forward<Args>(args)...);
   return ss.str();
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -191,8 +191,8 @@ bool CpuScalingEnabled(int num_cpus) {
   // running on Linux, so we silently ignore all the read errors.
   std::string res;
   for (int cpu = 0; cpu < num_cpus; ++cpu) {
-    std::string governor_file =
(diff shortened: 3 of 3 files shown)


pleroy:
(FWIW I am using this library on Windows.)

This PR is moving in the wrong direction in my opinion as it would complicate the transition to Abseil, which I believe should be our (distant) goal as it would make it much easier to use Google libraries together.

The Windows headers are well known for polluting the macro space (it's the 90s, go for it!).  If that gets in the way, just `#undef` the conflicting symbols (I'd vote for doing that just after the Windows includes for clarity).  After all, this is `sysinfo.cc`, it already contains a lot of platform-specific crud.

winksaville:
@pleroy, if it’s decided that to use the undef solution, should we change other StringXxx names to StrXxx? With this change there is consistency, which seems a good thing.

dominichamon:
I agree that moving to absl is the right approach, which is dependent on the (multiple?) outstanding PRs to use bazel (or absl moving to support non-bazel build systems). However, until then, we shouldn't be blocking PRs that fix breakages.

Either the rename or the #undef are fine with me, though I do like the consistency argument of the rename (as a strictly short-term solution). @pleroy Is this ok with you too?

pleroy:
I'd rather undef, and rename String to Str if we care about consistency, as that seems to be what absl has chosen.

dominichamon:
SGTM, and moves us closer to absl.

winksaville:
Created PR #546 and #547, closing this PR.

PR #547, opened 2018-03-06
Make string_util naming more consistent

winksaville:
As discussed in PR #544 making the names consistent is deseriable.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -543,7 +543,7 @@ class State {
   //  static void BM_Compress(benchmark::State& state) {
   //    ...
   //    double compress = input_size / output_size;
-  //    state.SetLabel(StringPrintf("compress:%.1f%%", 100.0*compression));
+  //    state.SetLabel(StringPrintF("compress:%.1f%%", 100.0*compression));
   //  }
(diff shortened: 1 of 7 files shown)


include/benchmark/benchmark.h
```diff
@@ -543,7 +543,7 @@ class State {
   //  static void BM_Compress(benchmark::State& state) {
   //    ...
   //    double compress = input_size / output_size;
-  //    state.SetLabel(StringPrintf("compress:%.1f%%", 100.0*compression));
+  //    state.SetLabel(StringPrintF("compress:%.1f%%", 100.0*compression));
```
dominichamon: this should be StrPrintF, right?

pleroy: This will be called StrFormat in absl, apparently.

src/sysinfo.cc
```diff
@@ -16,6 +16,7 @@
 
 #ifdef BENCHMARK_OS_WINDOWS
 #include <Shlwapi.h>
+#undef StrCat // Don't let StrCat in string_util.h be renamed to lstrcatA
```
pleroy: Two spaces before //

PR #548, opened 2018-03-08
Allow AddRange to work with int64_t.

### LINKED ISSUE #516: `Range(int start, int limit)`-maximum int is too small for number theory.

dominichamon:
Fixes #516

Also, tweak how we manage per-test build needs, and create a standard
_gtest suffix for googletest to differentiate from non-googletest tests.

I also ran clang-format on the files that I changed (but not the
benchmark include or main src as they have too many clang-format
issues).


dominichamon:
Looks like travis and appveyor are clean. @pleroy @EricWF any comments or thoughts?

EricWF:
LGTM.

PR #555, opened 2018-03-22
Rely on compiler intrinsics to identify regex engine.

dominichamon:
Having the copts set on a per-target level can lead to ODR violations
in some cases. Avoid this by ensuring the regex engine is picked
through compiler intrinsics in the header directly.


EricWF:
IDK how any of this works, but I don't see any glaring problems.

Issue #341, opened 2017-02-11
benchmark::MakeUnpredictable


Maratyszcza:
`folly::benchmark` includes [makeUnpredictable](https://github.com/facebook/folly/blob/e0b75e77cc224d3caffbb717342873c858e893fe/folly/Benchmark.h#L203) function to hide a compile-time constant value of a variable from optimizer. This is a very useful feature, and `google::benchmark` should provide something similar.

EricWF:
`DoNotOptimize()` should do that already.

Maratyszcza:
`DoNotOptimize()` is different. Here is my use-case:

```
static void quotient_uint32_t(benchmark::State& state) {
	const uint32_t divisor = benchmark::MakeUnpredictable(3);
	uint32_t x = 0;
	while (state.KeepRunning()) {
		const uint32_t quotient = x++ / divisor;
		benchmark::DoNotOptimize(quotient);
	}
}
BENCHMARK(quotient_uint32_t);
```
Normally compilers would replace division by an integer known at compile-time with a multiplication. I want to hide the value of the variable `divisor` from the optimizing compiler, so that it generates a generic hardware division instruction.

EricWF:
`DoNotOptimize()` may do other things, but it will also do this.

```
const uint32_t divisor x = 3;
DoNotOptimize(&x);
```

EricWF:
*Actually don't declare the variable as const. Woops.

If you want to provide `MakeUnpredictable` as a wrapper for `DoNotOptimize` with a different interface that  would be OK too.

Maratyszcza:
I see in the assembly code that I gets compiled into division-via-multiplication sequence, with both Apple Clang 8.0.0 and gcc 6.3.0. Removing const doesn't make a difference.

EricWF:
No, Clang and GCC both generate `div` instructions.
https://godbolt.org/g/76Rz4K

Maratyszcza:
The difference is I pass `divisor`, not `&divisor` no `DoNotOptimize`. However, the README suggests syntax without ampersand, is it wrong?

EricWF:
No, because you don't always want to escape the stack address of a variable. Sometimes you just want to escape/consume its value. 

`DoNotOptimize` is tricky, but so is the optimizer.

Maratyszcza:
Ok, thanks for the solution! Would be good to document it in the README.

EricWF:
[...]

PR #557, opened 2018-03-23
Add MakeUnpredictable to hide an objects value from the optimizer.

EricWF:
This patch addresses issue #341 by adding the function MakeUnpredictable.
The MakeUnpredictable(...) functions can be used to prevent the optimizer
from knowing the value of the specified 'object'. The function returns
the "laundered" input, either by reference if the input was a non-const
lvalue reference, or by value (when the input was a const lvalue or rvalue).

When MakeUnpredictable is supplied a non-const lvalue, the object
referenced by the input is made unpredictable and the return value
can be ignored. Otherwise, only the return value is considered
"unpredictable". In the latter the MakeUnpredictable function
is marked [[nodiscard]] and a warning is emitted if the return
value is ignored. For example:

```c++
int divide_by_two(int Value) {
  const int Divisor = 2;
  DoNotOptimize(Divisor); // INCORRECT! Has no effect on Divisor
  MakeUnpredictable(Divisor); // INCORRECT! should warn that return is ignored.
  // Correct Usage #1
  return Value / MakeUnpredictable(Divisor);
  // Correct Usage #2
  const int Divisor = MakeUnpredictable(2);
  return Value / Divisor;
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,6 @@ IRC channel: https://freenode.net #googlebenchmark
 [Known issues and common problems](#known-issues)
 
 [Additional Tooling Documentation](docs/tools.md)
-
 [Assembly Testing Documentation](docs/AssemblyTests.md)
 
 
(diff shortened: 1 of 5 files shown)


EricWF:
Urg. This is harder than I thought. I'll fix this up in the next couple days.

dominichamon:
Is this still something you're looking into, @EricWF ?

EricWF:
Yeah, I plan on working on this when I find time.

The problem seems to be coming up with the right set of inline assembly to use as to not produce too much assembly but which also ensures the desired effects.

PR #558, opened 2018-03-24
Cleanup test/CMakeLists.txt

EricWF:
This patch attempts to cleanup and reduce duplication in the CMake
which builds and registers the tests.

The patch adds a `add_benchmark_test` CMake function with a number
of options; allowing almost all tests to use it directly.


dominichamon:
Is this still something you want to pursue @EricWF ?

dominichamon:
ping :)

dominichamon:
treating this as abandoned.

PR #563, opened 2018-04-04
Do not combine inline with BENCHMARK_ALWAYS_INLINE

chfast:
MSVC does not like functions with both inline and __forceinline keywords. This change makes BENCHMARK_ALWAYS_INLINE to decay to inline if "always inline"-like attribute is not supported. So you should not use both inline and BENCHMARK_ALWAYS_INLINE together.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -203,7 +203,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 
 #if defined(__GNUC__)
 #define BENCHMARK_UNUSED __attribute__((unused))
-#define BENCHMARK_ALWAYS_INLINE __attribute__((always_inline))
+#define BENCHMARK_ALWAYS_INLINE inline __attribute__((always_inline))
 #define BENCHMARK_NOEXCEPT noexcept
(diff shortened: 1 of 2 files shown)


pleroy:
This PR looks similar to #493.  The bottom line is that you cannot just remove `inline`, it's going to cause ODR violations.  I can't seem to reproduce a problem on godbolt (see https://godbolt.org/g/WS4We6), can you detail what issue you are running into, with which version of MSVC, etc.?

chfast:
Thanks, let me find the warnings.

chfast:
In https://github.com/google/benchmark/blob/v1.3.0/include/benchmark/benchmark.h#L645

> c:\.hunter\_base\5e7b471\cbd9067\2860324\install\include\benchmark\benchmark.h(645): warning C4141: 'inline': used more than once [c:\projects\ethash\build\test\benchmarks\bench.vcxproj]

It looks it's caused by `inline` being after `__forceinline`.
This was swapped in v1.4.0, so I will check this version. Sorry for not checking it earlier.

chfast:
No warning in v1.4.0, and already in Hunter: https://github.com/ruslo/hunter/pull/1395.

PR #567, opened 2018-04-12
Benchmark ID for Better Tooling

BaaMeow:
This Pull Request implements a Benchmark ID that is output for the various reporting statures.

## What it does

This pull request outputs an additional column (console and csv) and an additional field (json) called "ID" ("id" for csv and json). It looks like this:

Multi-Pretty ID Printing
Pretty ID Printing

## Rationale

Suppose one registers 2 benchmarks functions, as below (implementation purely for exposition):

````

static void BM_StringCreation(benchmark::State& state) {
  for (auto _ : state)
    std::string empty_string;
}
// Register the function as a benchmark
BENCHMARK(BM_StringCreation);

// Define another benchmark
static void BM_StringCreation_Append(benchmark::State& state) {
  for (auto _ : state) {
    std::string empty_string;
    empty_string += "hello";
  }
}
BENCHMARK(BM_StringCreation_Append);

BENCHMARK_MAIN();
````

Now consider that someone runs this code with `--benchmark_repetitions=2`. We get the report for `BM_StringCreation` and `BM_StringCreation_Append`, giving us `stddev`, `mean`, and `median` for the benchmarks. The problem then becomes...

How do you associate a the name of the benchmark statistic with the benchmark it belongs to?

Normally, this wouldn't be a problem... but in the presence of custom statistics allowed by `ComputeStatistics`, we can run into problems. For example, if a statistic called `my_stat` is added, we run into issues where a tool cannot reliably determine which stats belong to which benchmark:

````
--------------------------------------------...
Name
--------------------------------------------...
BM_StringCreation
BM_StringCreation
BM_StringCreation_stddev
BM_StringCreation_median
BM_StringCreation_mean
BM_StringCreation_Append
BM_StringCreation_Append
BM_StringCreation_Append_stddev
BM_StringCreation_Append_median
BM_StringCreation_Append_mean
BM_StringCreation_Append_my_stat
````

Is the statistic `BM_StringCreation_Append_my_stat` called `Append_my_stat` and belong to `BM_StringCreation`? Or is the statistic's name `my_stat` and belongs to `BM_StringCreation_Append`?

A unique identifier solves this ambiguity, paving the way for tools to appropriately group statistic measurements with the right benchmark with a very small, simple change. It also allows tools to use order-independent parsers for json and csv, which is important in tooling languages where the default dictionary types and parsers do not guarantee read-order to be the order of submission.

## Code Submitted

The code submitted here is adds a few extra utility functions and a few new members. Convention was followed as much as possible, and `int` was preferred except where convention (such as inside of `ConsoleReporter`, where the saved width types are of `size_t`) demonstrated a use for otherwise.

### CODE REVIEW

**Comment by dominichamon** [comment]
I like the problem statement, but why not just change the separator in the output name? We already use '/' for threads, maybe we should do '.' for stats suffices.

File: src/benchmark_register.cc [hunk 1]
```diff
@@ -224,12 +225,22 @@ bool FindBenchmarksInternal(const std::string& re,
   return BenchmarkFamilies::GetInstance()->FindBenchmarks(re, benchmarks, Err);
 }
 
+// This function is exclusively for generating a unique identifier for a benchmark
+// This is to uniquely identify both statistics and benchmarks runs
+// in the case where names of benchmarks have whole names and prefixes which overlap
+// when multiple benchmarks are run
+int GetBenchmarkUniqueId() {
+  static std::atomic<int> benchid(0);
+  return benchid.fetch_add(1);
```
chfast: You can use `std::memory_order_relaxed` here.

BaaMeow: Thanks! I specified that memory order.

BaaMeow:
@dominichamon Thank you! Albeit, I had this discussion with someone earlier: because names can conceptually be anything (because of the more generic `RegisterBenchmark` function), I believe it would only be a temporary fix and tooling would run into issues later if they accidentally chose what we use as the "reserved" delimiter (not that it's actually reserved).

I chose this solution because it doesn't rely on the semantics of the name. Not relying on the name means it can be chosen for purely aesthetic and descriptive reasons, without affecting classification and grouping.

dominichamon:
That's fair. I'm a little concerned that it might break existing tooling, though i _hope_ tooling is using the JSON output rather than console.

Maybe it's worth doing that as a first step, actually: Having a unique ID output for the JSON (why isn't it just the "base name" of the benchmark, by the way?) and leave the console output alone?

BaaMeow:
As a side note, if we're concerned about breaking things, should I move the CSV field "id" to the last row, rather than the first? People may expect a certain ordering from the CSV (or is it not not really that much of a big deal, because CSV reporting is frowned upon and soon to be deprecated / removed like other issues indicate?).

BaaMeow:
Oh, and I forgot screenshots of what things look like.

Console output:
Console output right now

JSON Output:
Json output right now

CSV Output:
CSV output right now

dominichamon:
let me give an example:

c_code, c_code_mean, c_code_min would all have a field `base_name` that would be c_code.

i think this is a more obvious ID (and more likely to be unique) than an integer, and it would be deterministic between runs no matter what benchmark_filter you use, which isn't the case (i think) for your current solution.

pleroy:
@BaaMeow said: "because names can conceptually be anything (because of the more generic RegisterBenchmark function), I believe it would only be a temporary fix and tooling would run into issues later if they accidentally chose what we use as the "reserved" delimiter (not that it's actually reserved)".

I don't get this.  The benchmark name better be a C++ identifier, so a separator that cannot show up in an identifier is fine (Dominic proposed '.' which would work).

I for one have a fair amount of tooling that parses the console output (was written long before JSON/CSV output existed).  I do not relish the perspective of updating it for what is in fact a very minor improvement.

BaaMeow:
@pleroy You don't have to! @dominichamon suggested I do not touch the console output: I already put in the code to keep console output displaying as it always has. My concern is with my tooling and working with the JSON, so I only updated the CSV and JSON output as per @dominichamon's earlier comment! I really don't want to break existing tools that do things like parse the console output, so his suggestion made sense to me and I rolled back changing console output.

With respect to Names. The function `RegisterBenchmark` allows you to pick an arbitrary name to register a benchmark with that does not have to be the name of a C++ identifier, and there's nothing to enforce that it is. This is what I was referring to before. It looks like this:

```

auto BM_test = [](benchmark::State& state) {
  std::string x = "hello";
  for (auto _ : state)
    std::string copy(x);
};

::benchmark::RegisterBenchmark("A. Quirky?! Benchmark$$ name!", BM_test);
::benchmark::Initialize(&argc, argv);
  if (::benchmark::ReportUnrecognizedArguments(argc, argv)) {
	  return 1;
  }
  ::benchmark::RunSpecifiedBenchmarks();
}
```

And shows up like this:

````
2018-04-13 07:40:48
Running ./x64/bin/ptrptr_benchmarks
Run on (2 X 2592 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 256K (x2)
  L3 Unified 6144K (x2)
---------------------------------------------------------------------
Benchmark                              Time           CPU Iterations
---------------------------------------------------------------------
A. Quirky?! Benchmark$$ name!          4 ns          4 ns  167304859
```

To confirm, here's the actual visual output from the code:

![Interesting Names.](https://i.imgur.com/uMt9zYt.png)

BaaMeow:
@dominichamon That's a great idea, using `base_name`. You are right that I don't do anything in particular to stabilize the integer ID, just use it as a way of making each unique. However, since this is for the purposes of just making sure I can properly group Benchmarks with their corresponding statistics, `base_name` sounds like a much better idea. I will implement that right away.

BaaMeow:
@dominichamon I was able to implement the `base_name` attribute and it works very well as an identifier, for both `csv` and `json` for my use cases! I also checked over the tests, and made sure the console ones were left alone but the JSON and CSV ones were updated to reflect the addition of 'base_name' in the fields. This should keep @pleroy happy, and anyone else who had to depend on console output in a time before csv/json were supported.

If there's anything I did wrong in the code, please do let me know!

dominichamon:
@BaaMeow yes, the appveyor regex needs to be updated to handle the longer times. it hasn't become an issue important enough to fix yet, but that's in direct proportion to how straightforward the fix is. :face_with_head_bandage:

include/benchmark/benchmark.h
```diff
@@ -578,8 +567,8 @@ class State {
     return (max_iterations - total_iterations_ + batch_leftover_);
   }
 
-private: // items we expect on the first cache line (ie 64 bytes of the struct)
-
+ private
```
dominichamon: this change looks odd. clang-format misbehaving?

include/benchmark/benchmark.h
```diff
@@ -892,9 +878,10 @@ class Benchmark {
  private:
   friend class BenchmarkFamilies;
 
+  int id_;
```
dominichamon: do you still need this?

include/benchmark/benchmark.h
```diff
@@ -1271,6 +1262,7 @@ class BenchmarkReporter {
           report_rms(false),
           counters() {}
 
+    std::string benchmark_base_name;
```
dominichamon: just `base_name` is fine.

src/benchmark.cc
```diff
@@ -111,6 +111,17 @@ namespace internal {
 
 void UseCharPointer(char const volatile*) {}
 
+size_t Base10Digits(size_t num) {
```
dominichamon: still necessary?

src/benchmark.cc
```diff
@@ -174,7 +186,8 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
     results.manual_time_used += timer.manual_time_used();
     results.bytes_processed += st.bytes_processed();
     results.items_processed += st.items_processed();
-    results.complexity_n += st.complexity_length_n();
+    // explicitly avoid warning on Visual C++ about truncating int64_t
+    results.complexity_n += static_cast<int>(st.complexity_length_n());
```
dominichamon: this was fixed in another commit so you shouldn't need it any more (`complexity_n` is 64-bit).

src/benchmark.cc
```diff
@@ -235,15 +248,16 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
...
+      const bool should_report =
+          repetition_num > 0 ||
+          has_explicit_iteration_count  // An exact iteration count was
+                                        // requested
+          || results.has_error_ || iters >= kMaxIterations ||
+          seconds >= min_time  // the elapsed time is large enough
+          // CPU time is specified but the elapsed real time greatly exceeds the
+          // minimum time. Note that user provided timers are except from this
+          // sanity check.
+          || ((results.real_time_used >= 5 * min_time) && !b.use_manual_time);
```
dominichamon: can you put this back how it was? the || being lined up helps readability a lot.

src/console_reporter.cc
```diff
@@ -45,19 +46,26 @@ bool ConsoleReporter::ReportContext(const Context& context) {
...
 }
 
 void ConsoleReporter::PrintHeader(const Run& run) {
-  std::string str = FormatString("%-*s %13s %13s %10s", static_cast<int>(name_field_width_),
-                                 "Benchmark", "Time", "CPU", "Iterations");
-  if(!run.counters.empty()) {
-    if(output_options_ & OO_Tabular) {
-      for(auto const& c : run.counters) {
+  std::string str =
+      (output_options_ & OO_Base_Name)
```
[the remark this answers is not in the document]
BaaMeow: It's not strictly necessary, but I wanted to leave this as a way for someone to (easily) enable printing the base name for the console reporter, maybe gated by an option or something. I didn't add any such option here, it's just in case somebody wants it.

src/benchmark.cc
BaaMeow: Sorry! Clang format kicked in when I was doing the latest commits and, uh. Changed things...

src/benchmark.cc
BaaMeow: Gotcha: will remove!

include/benchmark/benchmark.h
BaaMeow: Yeah.... I had to move the comment to the next line, otherwise clang-format just kind of mangles the `:` on the `private`.

src/benchmark.cc
BaaMeow: Nope, removed!

src/console_reporter.cc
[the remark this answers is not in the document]
dominichamon: We should either make it fully functional with an option, or remove it. We can always create an issue as a "good first issue" to add this support and option as a followup. 

that'll make it easier to understand what was added in this PR.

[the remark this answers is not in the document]
BaaMeow: I removed the code for it: I suppose somebody else can implement the option for console reporting and the like!

dominichamon:
LGTM.. I'll wait for @EricWF or @pleroy to comment as they have Opinions, usually.

BaaMeow:
@LebedevRI The formatting changes are not really something I can control: I make an edit and save, and according to the clang-format rules the entire thing gets changed. I would have hoped that somebody before me made the fixes and formatted the files before they committed: they did not, it happened here. I'm sorry! But, well. The formatting fixes would have to happen at some point, right?

The screenshots above are perhaps misleading because they were done before the final changes. Here it what it looks like in JSON for the final changes:

JSON

The `base_name` is the name of the benchmark exactly as you specified it.

All of the tests were updated to properly test for `base_name`, so you can run those if you'd like to see it in the JSON and CSV reporters. Note that we don't include the printout in the console reporter because this breaks people who parse the console report, as @pleroy stated earlier.

Does that make sense?

LebedevRI:
@BaaMeow 
Can you attach the json produced by the following code please?
```c++
static void NOP(benchmark::State& state) {
  for (auto _ : state) {}
}
BENCHMARK(NOP);
BENCHMARK(NOP)->Repetitions(3);
BENCHMARK(NOP)->Range(8, 8<<10);
BENCHMARK(NOP)->Range(8, 8<<10)->Repetitions(3);
```

BaaMeow:
Console:
Console

JSON:
````json
{
  "context": {
    "date": "05/06/18 00:57:53",
    "executable": "build\\x64-Debug\\x64\\bin\\nop_benchmarks.exe",
    "num_cpus": 8,
    "mhz_per_cpu": 2592,
    "cpu_scaling_enabled": false,
    "caches": [
      {
        "type": "Data",
        "level": 1,
        "size": 32768000,
        "num_sharing": 2
      },
      {
        "type": "Instruction",
        "level": 1,
        "size": 32768000,
        "num_sharing": 2
      },
      {
        "type": "Unified",
        "level": 2,
        "size": 262144000,
        "num_sharing": 2
      },
      {
        "type": "Unified",
        "level": 3,
        "size": 6291456000,
        "num_sharing": 8
      }
    ],
    "library_build_type": "debug"
  },
  "benchmarks": [
    {
      "base_name": "NOP",
      "name": "NOP",
      "iterations": 22400000,
      "real_time": 2.9738841874080077e+01,
      "cpu_time": 2.9994419642857142e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/repeats:3",
      "iterations": 22400000,
      "real_time": 2.7985229106499382e+01,
      "cpu_time": 2.7901785714285715e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/repeats:3",
      "iterations": 22400000,
      "real_time": 2.6264596921724401e+01,
      "cpu_time": 2.6506696428571427e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/repeats:3",
      "iterations": 22400000,
      "real_time": 2.7667434242175659e+01,
      "cpu_time": 2.7204241071428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/repeats:3_mean",
      "iterations": 22400000,
      "real_time": 2.7305753423466481e+01,
      "cpu_time": 2.7204241071428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/repeats:3_median",
      "iterations": 22400000,
      "real_time": 2.7667434242175659e+01,
      "cpu_time": 2.7204241071428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/repeats:3_stddev",
      "iterations": 22400000,
      "real_time": 9.1556187108647824e-01,
      "cpu_time": 6.9754464285706352e-01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8",
      "iterations": 26352941,
      "real_time": 2.7724797660724654e+01,
      "cpu_time": 2.7866908668751620e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64",
      "iterations": 24888889,
      "real_time": 2.6776640090242402e+01,
      "cpu_time": 2.6367187382289341e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512",
      "iterations": 26352941,
      "real_time": 2.5701620931957933e+01,
      "cpu_time": 2.5495256867155739e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096",
      "iterations": 26352941,
      "real_time": 2.7658431784598143e+01,
      "cpu_time": 2.7273995718352651e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192",
      "iterations": 26352941,
      "real_time": 2.5998235983796267e+01,
      "cpu_time": 2.6088169817554707e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8/repeats:3",
      "iterations": 28000000,
      "real_time": 2.5851613463601094e+01,
      "cpu_time": 2.6227678571428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8/repeats:3",
      "iterations": 28000000,
      "real_time": 2.5843542927759700e+01,
      "cpu_time": 2.5669642857142858e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8/repeats:3",
      "iterations": 28000000,
      "real_time": 2.5681342143798251e+01,
      "cpu_time": 2.5669642857142858e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8/repeats:3_mean",
      "iterations": 28000000,
      "real_time": 2.5792166178386349e+01,
      "cpu_time": 2.5855654761904759e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8/repeats:3_median",
      "iterations": 28000000,
      "real_time": 2.5843542927759700e+01,
      "cpu_time": 2.5669642857142858e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8/repeats:3_stddev",
      "iterations": 28000000,
      "real_time": 9.6061221984072728e-02,
      "cpu_time": 3.2218206986057524e-01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64/repeats:3",
      "iterations": 28000000,
      "real_time": 2.6447238641724521e+01,
      "cpu_time": 2.6785714285714285e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64/repeats:3",
      "iterations": 28000000,
      "real_time": 2.6957008571896171e+01,
      "cpu_time": 2.6227678571428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64/repeats:3",
      "iterations": 28000000,
      "real_time": 2.6219090821021901e+01,
      "cpu_time": 2.5669642857142858e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64/repeats:3_mean",
      "iterations": 28000000,
      "real_time": 2.6541112678214198e+01,
      "cpu_time": 2.6227678571428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64/repeats:3_median",
      "iterations": 28000000,
      "real_time": 2.6447238641724521e+01,
      "cpu_time": 2.6227678571428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/64/repeats:3_stddev",
      "iterations": 28000000,
      "real_time": 3.7780934716984538e-01,
      "cpu_time": 5.5803571428565080e-01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512/repeats:3",
      "iterations": 28000000,
      "real_time": 2.7084712072142533e+01,
      "cpu_time": 2.7343750000000000e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512/repeats:3",
      "iterations": 28000000,
      "real_time": 2.5702844750152767e+01,
      "cpu_time": 2.5111607142857142e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512/repeats:3",
      "iterations": 28000000,
      "real_time": 2.6121764714064607e+01,
      "cpu_time": 2.5669642857142858e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512/repeats:3_mean",
      "iterations": 28000000,
      "real_time": 2.6303107178786636e+01,
      "cpu_time": 2.6041666666666664e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512/repeats:3_median",
      "iterations": 28000000,
      "real_time": 2.6121764714064607e+01,
      "cpu_time": 2.5669642857142858e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/512/repeats:3_stddev",
      "iterations": 28000000,
      "real_time": 7.0855708381853810e-01,
      "cpu_time": 1.1616439729164194e+00,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096/repeats:3",
      "iterations": 28000000,
      "real_time": 2.6202540605611699e+01,
      "cpu_time": 2.6227678571428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096/repeats:3",
      "iterations": 28000000,
      "real_time": 2.5716149857284368e+01,
      "cpu_time": 2.5111607142857142e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096/repeats:3",
      "iterations": 28000000,
      "real_time": 2.6736733751022257e+01,
      "cpu_time": 2.6785714285714285e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096/repeats:3_mean",
      "iterations": 28000000,
      "real_time": 2.6218474737972773e+01,
      "cpu_time": 2.6041666666666664e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096/repeats:3_median",
      "iterations": 28000000,
      "real_time": 2.6202540605611699e+01,
      "cpu_time": 2.6227678571428573e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/4096/repeats:3_stddev",
      "iterations": 28000000,
      "real_time": 5.1047849462049211e-01,
      "cpu_time": 8.5241363373437296e-01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192/repeats:3",
      "iterations": 26352941,
      "real_time": 2.6258227648907159e+01,
      "cpu_time": 2.6088169817554707e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192/repeats:3",
      "iterations": 26352941,
      "real_time": 2.7611554437120592e+01,
      "cpu_time": 2.7273995718352651e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192/repeats:3",
      "iterations": 26352941,
      "real_time": 2.5462437000549215e+01,
      "cpu_time": 2.5495256867155739e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192/repeats:3_mean",
      "iterations": 26352941,
      "real_time": 2.6444073028858988e+01,
      "cpu_time": 2.6285807467687697e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192/repeats:3_median",
      "iterations": 26352941,
      "real_time": 2.6258227648907159e+01,
      "cpu_time": 2.6088169817554707e+01,
      "time_unit": "ns"
    },
    {
      "base_name": "NOP",
      "name": "NOP/8192/repeats:3_stddev",
      "iterations": 26352941,
      "real_time": 1.0865451292885608e+00,
      "cpu_time": 9.0568949190772241e-01,
      "time_unit": "ns"
    }
  ]
}
````
CSV:
````
name,iterations,real_time,cpu_time,time_unit,bytes_per_second,items_per_second,label,error_occurred,error_message,base_name
"NOP",20363636,32.0716,32.2266,ns,,,,,,"NOP"
"NOP/repeats:3",26352941,27.8607,27.8669,ns,,,,,,"NOP"
"NOP/repeats:3",26352941,26.3785,25.4953,ns,,,,,,"NOP"
"NOP/repeats:3",26352941,27.601,27.8669,ns,,,,,,"NOP"
"NOP/repeats:3_mean",26352941,27.28,27.0764,ns,,,,,,"NOP"
"NOP/repeats:3_median",26352941,27.601,27.8669,ns,,,,,,"NOP"
"NOP/repeats:3_stddev",26352941,0.791506,1.36927,ns,,,,,,"NOP"
"NOP/8",26352941,26.7635,26.6811,ns,,,,,,"NOP"
"NOP/64",26352941,28.6893,28.4598,ns,,,,,,"NOP"
"NOP/512",26352941,26.8772,27.274,ns,,,,,,"NOP"
"NOP/4096",26352941,25.9611,26.0882,ns,,,,,,"NOP"
"NOP/8192",26352941,26.6772,26.6811,ns,,,,,,"NOP"
"NOP/8/repeats:3",28000000,26.5917,26.7857,ns,,,,,,"NOP"
"NOP/8/repeats:3",28000000,25.7512,25.6696,ns,,,,,,"NOP"
"NOP/8/repeats:3",28000000,25.507,25.6696,ns,,,,,,"NOP"
"NOP/8/repeats:3_mean",28000000,25.95,26.0417,ns,,,,,,"NOP"
"NOP/8/repeats:3_median",28000000,25.7512,25.6696,ns,,,,,,"NOP"
"NOP/8/repeats:3_stddev",28000000,0.569019,0.644364,ns,,,,,,"NOP"
"NOP/64/repeats:3",28000000,27.2579,26.7857,ns,,,,,,"NOP"
"NOP/64/repeats:3",28000000,28.7635,29.0179,ns,,,,,,"NOP"
"NOP/64/repeats:3",28000000,27.7135,27.9018,ns,,,,,,"NOP"
"NOP/64/repeats:3_mean",28000000,27.9116,27.9018,ns,,,,,,"NOP"
"NOP/64/repeats:3_median",28000000,27.7135,27.9018,ns,,,,,,"NOP"
"NOP/64/repeats:3_stddev",28000000,0.772122,1.11607,ns,,,,,,"NOP"
"NOP/512/repeats:3",26352941,26.2543,26.0882,ns,,,,,,"NOP"
"NOP/512/repeats:3",26352941,29.0715,29.0527,ns,,,,,,"NOP"
"NOP/512/repeats:3",26352941,25.8172,26.0882,ns,,,,,,"NOP"
"NOP/512/repeats:3_mean",26352941,27.0477,27.0764,ns,,,,,,"NOP"
"NOP/512/repeats:3_median",26352941,26.2543,26.0882,ns,,,,,,"NOP"
"NOP/512/repeats:3_stddev",26352941,1.76627,1.71159,ns,,,,,,"NOP"
"NOP/4096/repeats:3",26352941,26.1249,26.0882,ns,,,,,,"NOP"
"NOP/4096/repeats:3",26352941,26.5308,26.6811,ns,,,,,,"NOP"
"NOP/4096/repeats:3",26352941,26.593,26.6811,ns,,,,,,"NOP"
"NOP/4096/repeats:3_mean",26352941,26.4162,26.4834,ns,,,,,,"NOP"
"NOP/4096/repeats:3_median",26352941,26.5308,26.6811,ns,,,,,,"NOP"
"NOP/4096/repeats:3_stddev",26352941,0.254198,0.342318,ns,,,,,,"NOP"
"NOP/8192/repeats:3",26352941,26.2092,26.0882,ns,,,,,,"NOP"
"NOP/8192/repeats:3",26352941,26.4208,26.0882,ns,,,,,,"NOP"
"NOP/8192/repeats:3",26352941,25.7008,26.0882,ns,,,,,,"NOP"
"NOP/8192/repeats:3_mean",26352941,26.1103,26.0882,ns,,,,,,"NOP"
"NOP/8192/repeats:3_median",26352941,26.2092,26.0882,ns,,,,,,"NOP"
"NOP/8192/repeats:3_stddev",26352941,0.370045,0,ns,,,,,,"NOP"
````

Note that, as talked about in earlier bikeshedding, the CSV output has the base_name as a parameter at the end of the formal list, primarily to avoid breaking people who use CSV and hard-code their use of the CSV indices.

Does this answer your questions?

**Comment by LebedevRI** [comment]
> Does this answer your questions?

Yes, it does. You confirmed my concerns.
There are **4** different benchmarks there, but they all have **one**, **exactly** same `basename``.
Unless i misunderstand the basic idea of this diff, this is **bad**.

I'd expect for basename to be equal for name with removed suffix of `_mean`, `_median`, `_stdev` so to speak.

Do you disagree?

**Comment by LebedevRI** [comment]
Thank you for replying!

> With the base name you can still accurately determine things like extraneous modifiers (everything after NOP, now), which means you can -- with some effort -- still isolate which benchmark group each of those belongs to and their associated statistics (modifiers like range, threads, etc. all use / and : -- these characters don't show up in regular C function names).

>  **with some effort**

Let me rephrase. Why can't we have the following:
```
name,iterations,real_time,cpu_time,time_unit,bytes_per_second,items_per_second,label,error_occurred,error_message,base_name
"NOP",20363636,32.0716,32.2266,ns,,,,,,"NOP"
"NOP/repeats:3",26352941,27.8607,27.8669,ns,,,,,,"NOP/repeats:3"
"NOP/repeats:3",26352941,26.3785,25.4953,ns,,,,,,"NOP/repeats:3"
"NOP/repeats:3",26352941,27.601,27.8669,ns,,,,,,"NOP/repeats:3"
"NOP/repeats:3_mean",26352941,27.28,27.0764,ns,,,,,,"NOP/repeats:3"
"NOP/repeats:3_median",26352941,27.601,27.8669,ns,,,,,,"NOP/repeats:3"
"NOP/repeats:3_stddev",26352941,0.791506,1.36927,ns,,,,,,"NOP/repeats:3"
"NOP/8",26352941,26.7635,26.6811,ns,,,,,,"NOP/8"
"NOP/64",26352941,28.6893,28.4598,ns,,,,,,"NOP/64"
"NOP/512",26352941,26.8772,27.274,ns,,,,,,"NOP/512"
"NOP/4096",26352941,25.9611,26.0882,ns,,,,,,"NOP/4096"
"NOP/8192",26352941,26.6772,26.6811,ns,,,,,,"NOP/8192"
"NOP/8/repeats:3",28000000,26.5917,26.7857,ns,,,,,,"NOP/8/repeats:3"
"NOP/8/repeats:3",28000000,25.7512,25.6696,ns,,,,,,"NOP/8/repeats:3"
"NOP/8/repeats:3",28000000,25.507,25.6696,ns,,,,,,"NOP/8/repeats:3"
"NOP/8/repeats:3_mean",28000000,25.95,26.0417,ns,,,,,,"NOP/8/repeats:3"
"NOP/8/repeats:3_median",28000000,25.7512,25.6696,ns,,,,,,"NOP/8/repeats:3"
"NOP/8/repeats:3_stddev",28000000,0.569019,0.644364,ns,,,,,,"NOP/8/repeats:3"
"NOP/64/repeats:3",28000000,27.2579,26.7857,ns,,,,,,"NOP/64/repeats:"
"NOP/64/repeats:3",28000000,28.7635,29.0179,ns,,,,,,"NOP/64/repeats:"
"NOP/64/repeats:3",28000000,27.7135,27.9018,ns,,,,,,"NOP/64/repeats:"
"NOP/64/repeats:3_mean",28000000,27.9116,27.9018,ns,,,,,,"NOP/64/repeats:"
"NOP/64/repeats:3_median",28000000,27.7135,27.9018,ns,,,,,,"NOP/64/repeats:"
"NOP/64/repeats:3_stddev",28000000,0.772122,1.11607,ns,,,,,,"NOP/64/repeats:"
"NOP/512/repeats:3",26352941,26.2543,26.0882,ns,,,,,,"NOP/512/repeats:3"
"NOP/512/repeats:3",26352941,29.0715,29.0527,ns,,,,,,"NOP/512/repeats:3"
"NOP/512/repeats:3",26352941,25.8172,26.0882,ns,,,,,,"NOP/512/repeats:3"
"NOP/512/repeats:3_mean",26352941,27.0477,27.0764,ns,,,,,,"NOP/512/repeats:3"
"NOP/512/repeats:3_median",26352941,26.2543,26.0882,ns,,,,,,"NOP/512/repeats:3"
"NOP/512/repeats:3_stddev",26352941,1.76627,1.71159,ns,,,,,,"NOP/512/repeats:3"
"NOP/4096/repeats:3",26352941,26.1249,26.0882,ns,,,,,,"NOP/4096/repeats:3"
"NOP/4096/repeats:3",26352941,26.5308,26.6811,ns,,,,,,"NOP/4096/repeats:3"
"NOP/4096/repeats:3",26352941,26.593,26.6811,ns,,,,,,"NOP/4096/repeats:3"
"NOP/4096/repeats:3_mean",26352941,26.4162,26.4834,ns,,,,,,"NOP/4096/repeats:3"
"NOP/4096/repeats:3_median",26352941,26.5308,26.6811,ns,,,,,,"NOP/4096/repeats:3"
"NOP/4096/repeats:3_stddev",26352941,0.254198,0.342318,ns,,,,,,"NOP/4096/repeats:3"
"NOP/8192/repeats:3",26352941,26.2092,26.0882,ns,,,,,,"NOP/8192/repeats:3"
"NOP/8192/repeats:3",26352941,26.4208,26.0882,ns,,,,,,"NOP/8192/repeats:3"
"NOP/8192/repeats:3",26352941,25.7008,26.0882,ns,,,,,,"NOP/8192/repeats:3"
"NOP/8192/repeats:3_mean",26352941,26.1103,26.0882,ns,,,,,,"NOP/8192/repeats:3"
"NOP/8192/repeats:3_median",26352941,26.2092,26.0882,ns,,,,,,"NOP/8192/repeats:3"
"NOP/8192/repeats:3_stddev",26352941,0.370045,0,ns,,,,,,"NOP/8192/repeats:3"
```
This way, there are **four** distinct benchmark groups, and no effort required to distinguish between them.

**Comment by BaaMeow** [comment]
Because that would require every relevant separating counter, option, and repetition added VIA the `BENCHMARK(...)` API to be accumulated in the base name. And, someone would have to remember to make sure to tag each one appropriately. From my understanding of the code base, that's exceptionally more complicated and might break in small ways when someone forgets to accumulate the "right" options (which ones?) into the `base_name`.

Continually extending `base_name` to encompass all of these is likely more engineering work and less tenable. The succinct solution to your problem is an `id` marker, because each benchmark registered with `BENCHMARK(...)` or `RegisterBenchmark( Name, ... )` will get tagged. It requires no fiddling with the reporters beyond printing the ID, and no identification of which options deserved to be accumulated in the base name or not.

**Comment by LebedevRI** [comment]
> From my understanding of the code base, that's exceptionally more complicated and might break in small ways when someone forgets to accumulate the "right" options (which ones?) into the base_name.

I'm pretty sure that's **not** how things work.

Here, that full name is stored:
https://github.com/google/benchmark/blob/c4858d8012acd54d8ef89053c74e9cdcf0fa6649/src/benchmark.cc#L123

Here, the statistic's suffix is appended:
https://github.com/google/benchmark/blob/50ffc781b1df6d582d6a453b9942cc8cb512db69/src/statistics.cc#L154

So i'm 99.9% sure you just need to change from
https://github.com/BaaMeow/benchmark/blob/bf2f5a5f8618220f4a5e64583bb5564ddadb3931/src/benchmark.cc#L124
```diff
  report.benchmark_name = b.name;
- report.base_name = b.base_name;
+ report.base_name = b.name;
```
(and drop `b.base_name` field, NOT `report.base_name` field)

Try it, _please_ :)

BaaMeow:
That would not bring the desired results. From what you presented and stated before: you have 4 benchmarks, and you want 4 base names to identify each of the 4.

Look at the CSV you posted where you just cut off the statistics name: there's _18_ unique strings for the base name. You've only moved the problem around, and it _still_ requires post-processing to handle correctly. This is what I mean by "accumulating certain fields". Do we drop the range specifier from the base name, but keep the threads? The repetitions? How do we reduce the `name` field to a suitable `base_name` that gets you to the 4 distinct identifiers you want?

This is the cognitive overhead of how the reporters are defined I am talking about.

The answer is not in this PR. I would be happy to make a new PR for an ID number that matches your use case and would give you 4 distinct identifiers for your 4 benchmark registrations, but I still maintain that mangling the base name is not the answer.

LebedevRI:
> From what you presented and stated before: you have 4 benchmarks, and you want 4 base names to identify each of the 4.

Eeeh, i fooled myself there :)
I was talking about the each group for which we print statistics.
So 18 unique strings is indeed the expected result.

> Look at the CSV you posted where you just cut off the statistics name: there's 18 unique strings for the base name.

Yep!

dominichamon:
Let me see if i can cut through the issues here. It really comes down to what bit of the benchmark name you think identifies a unique "benchmark". I can confirm that the benchmark name is the registered name + thread count + number of repetitions. It is not just the registered name.

For tooling reasons, you really want to include these things in the "base name" as you want to compare like-for-like. If you drop these, you will no longer be able to correctly understand what's happening with the benchmark.

Now, the benchmark names were originally written pre json, With json, i imagine we could have:

```json
{
  "name": "NOP",
  "threads": 8,
  "repetitions": 3,
  "iterations": ...
  ...
},
{
  "name": "NOP",
  "threads": 16,
  "iterations": ...
  ...
},
...
```

ie, parse out the various bits that are interesting about the benchmark for ease of parsing and tooling. The console output tries to do this by having a full name that includes the basename and parseable separators, but you've identified it's not ideal.

We can migrate to this by adding these extra fields, leaving the legacy name, and adding a base name that is just the name as @BaaMeow has defined it.

@LebedevRI does that get us to where you can use the json for tooling?

LebedevRI:
> @LebedevRI does that get us to where you can use the json for tooling?

In the case of the bundled tools, the current (git master) status is already sufficient.
So i don't *need* this.

Having more structured json would be nice, but it needs to be either a non-breaking change,
or it needs to be really thought-through first, with json scheme written beforehand.

LebedevRI:
> To be clear, I'm not against what @LebedevRI wants or suggests

I'm just not sure what problem exactly this pr is trying to solve.
If it is as written in https://github.com/google/benchmark/pull/567#issue-181073435:
> Now consider that someone runs this code with `--benchmark_repetitions=2`. We get the report for `BM_StringCreation` and `BM_StringCreation_Append`, giving us `stddev`, `mean`, and `median` for the benchmarks. The problem then becomes...
>
>How do you associate a the name of the benchmark statistic with the benchmark it belongs to?
>
>Normally, this wouldn't be a problem... but in the presence of custom statistics allowed by `ComputeStatistics`, we can run into problems. For example, if a statistic called `my_stat` is added, we run into issues where a tool cannot reliably determine which stats belong to which benchmark:

then i really don't see why the current solution is the appropriate one.
I'm biased, but i'd argue that what i'm suggesting matches the stated problem much more closely.

BaaMeow:
So to solve the problems I brought up and the ones @LebedevRI has, I'm going to implement `base_name`, `id`, and also include all of the modifiers @dominichamon has mentioned (adding "Threads", "repetitions", "iterations" to the JSON and CSV output).

Would it be preferred I close this and open a new PR so this is less messy?

LebedevRI:
That sounds like a major json scheme break.

I'm of no power here, but first i would really want to hear *concrete*
description of the problem you are trying to solve.
If it is what you have described in the https://github.com/google/benchmark/pull/567#issue-181073435
But since i have asked that before, it seems we are going in circles here.

dominichamon:
Adding fields to JSON shouldn't break anything, and it should make tooling easier to have the name broken apart.

I'd do it for JSON but not bother for CSV. You might also want to do it in the v2 branch with @EricWF's blessing, as they're working on JSON changes.

LebedevRI:
> Adding fields to JSON shouldn't break anything, and it should make tooling easier to have the name broken apart.

Yep, **adding** - absolutely.
It's just that in your comment https://github.com/google/benchmark/pull/567#issuecomment-387363969 the full re-design of json scheme is suggested,
and while that will/should happen eventually, it is probably best to refrain to
just addition of new fields, not removal of old fields in this PR.
Just thought i'd mention it

BaaMeow:
@LebedevRI There already is a discrete description of the problems we're trying to solve here. 1 is that associating a statistic with the benchmark it came from is a bit harder than it should be (the original problem), and the other is that there is a desire for a conceptual `base_name` grouping that comes from base_name + thread + repetition count.

Problem 1 is solved by a numeric ID on each benchmark. Repeated runs and statistics with the same ID can be successfully mapped to a single registered benchmark, regardless of if the user has the same name applied 2 or 3 times with separate repetitions / threading numbers. This is the finest granularity I can think of offering.

Problem 2 is solved by producing more fields on the JSON that accurately describe a single benchmark _group_ (or, as the code calls it, a single `BenchmarkFamily` in the `BenchmarkFamilies`). This means _adding_ (not subtracting) JSON fields, and leaving the current ones alone. To do this, you need a combination of `base_name` (the name the benchmark was registered with), `repetitions`, and `threads`.

So, there will be 4 total new fields:
`base_name`
`id`
`threads`
`repetitions`

All of this information will allow as fine or as coarse comparisons as one would like in their tooling, plus arbitrary grouping. I think this satisfies all of the necessary use cases present.

BaaMeow:
This pull request is being closed in favor of #599.

PR #573, opened 2018-04-21
Add caching for cxx_feature_check

Yangqing:
In repeated cmake runs, the result of cxx_feature_check can be cached so we do not do try_run multiple times - this makes cmake reconfiguration much faster. We encountered this issue when using benchmark under pytorch: for repeated cmake runs, this PR reduces the cmake time on benchmark from 2 seconds to 0.2 seconds on my macbook.

### DIFF
--- a/cmake/CXXFeatureCheck.cmake
+++ b/cmake/CXXFeatureCheck.cmake
@@ -27,25 +27,27 @@ function(cxx_feature_check FILE)
     return()
   endif()
 
-  message("-- Performing Test ${FEATURE}")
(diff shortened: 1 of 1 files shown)


dominichamon:
The upshot of this is that someone changing their configuration will need to clean their cmake? works for me but @EricWF fiddles with configs more often so they should weigh in.

Yangqing:
Thanks! Basically, the cxx_feature_check should largely stay the same unless one changes the compiler stack on the machine - in which case, it would probably mandate a cmake clean anyway.

dominichamon:
SGTM

PR #574, opened 2018-04-21
Fix precision loss warning in MSVC

pwnall:
Warning text:

c:\projects\berrydb\third_party\benchmark\src\complexity.cc(31): warning C4244: 'return': conversion from 'int64_t' to 'double', possible loss of data

Thank you very much for maintaining google benchmark!


dominichamon:
nice catch, thanks!

PR #575, opened 2018-04-22
Add statistics to the reporter

FredTingaud:
This PR is a work in progress, but I would need some guidance, to know whether I should go on with it at all and if yes how.

The goal is to keep the minimum iteration time, maximum iteration time and standard deviation. All three of those can be kept with minimal memory footprint, but they require to interogate the timer on each iteration of the benchmark, which from what I understand from the code and the comments, will reduce the number of possible iterations of small tests.
Does it make the development a total no-go? Or could we perhaps hide it behind an option flag? Or do you see a third option?

On top of the information for the end-user, perhaps this standard deviation could be used as a condition to continue or end the benchmark.


dominichamon:
If you run with --benchmark_repetitions set to >0 then you'll get the std deviation reported already. Similarly, there are minimum, maximum, mean, median stats already calculated.

I understand this is the std dev for a given run rather than across runs, but the fundamental notion is that we need some number of iterations in a run to get a good confidence in the reported timing. As such, i'm not sure that this is really going to be useful information for a user, as any particular iteration can be affected by a number of external variables, throwing off the min and max, and possibly the stddev.

I do like the idea of using the stddev to guide when a particular run is "good", though if we're going to do that we really should use standard error to derive confidence intervals for the timing data (assuming the data is normal, etc, etc).

dominichamon:
I'm going to close this because I think my last comment explains where we're coming from. Feel free to open an issue or another PR if you want to discuss it further.

Issue #571, opened 2018-04-19
Improved Benchmark Filter


Tim020:
Currently, the benchmark_filter command option uses POSIX regexes, however these do not allow for negative matching. Unlike the gtest_filter option, you cannot add a '-' in-front of the regex to negate it. This would be a useful addition to the benchmark_filter to easily be able to negate the search and run all benchmarks which do not match the filter.

PR #576, opened 2018-04-25
Issue 571: Allow support for negative regex filtering

Tim020:
This patch allows one to apply a negation to the entire regex filter by appending it with a '-' character, much in the same style as GoogleTest uses.


Tim020:
I signed it!

Tim020:
Implements #571

src/benchmark_register.cc
```diff
@@ -114,7 +114,14 @@ bool BenchmarkFamilies::FindBenchmarks(
   // Make regular expression out of command-line flag
   std::string error_msg;
   Regex re;
-  if (!re.Init(spec, &error_msg)) {
+  std::string prefix("-");
```
dominichamon: is this used?

src/benchmark_register.cc
```diff
@@ -114,7 +114,14 @@ bool BenchmarkFamilies::FindBenchmarks(
...
+  std::string temp_spec = std::string(spec);
```
dominichamon: given you have to take a copy, it might be better to just pass `spec` in by value instead of const reference.

src/benchmark_register.cc
Tim020: Ah not anymore, this was left over from some test code, will remove :)

src/benchmark_register.cc
Tim020: Done :)

dominichamon:
You can say no, but is there any way we can add a test? Now that we integrate with googletest (which was done much later than the original implementation) it should be possible to write a unit test.

Tim020:
@dominichamon done :) Added negative variants for all the existing unit tests

dominichamon:
great, thank you!

chfast:
Was there a way to achieve similar results before this PR?

PR #580, opened 2018-05-01
Add pragma to fix msvc link issue, modify bazel WORKSPACE to work with msvc

astrelni:
Pragma for shlwapi.lib is taken from
https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/sysinfo.cc#L81

Tested this on local Windows 10 machine with Visual Studio 2017. Bazel did not work correctly with git_repository(), so switched it out with http_archive() in the workspace file as found in
https://github.com/abseil/abseil-cpp/blob/master/WORKSPACE#L14

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,9 +1,7 @@
 workspace(name = "com_github_google_benchmark")
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
-
-git_repository(
-    name = "com_google_googletest",
-    commit = "3f0cf6b62ad1eb50d8736538363d3580dd640c3e",  # HEAD
-    remote = "https://github.com/google/googletest",
+http_archive(
+     name = "com_google_googletest",
+     urls = ["https://github.com/google/googletest/archive/master.zip"],
+     strip_prefix = "googletest-master",
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,6 +2,6 @@ workspace(name = "com_github_google_benchmark")
 
 http_archive(
      name = "com_google_googletest",
-     urls = ["https://github.com/google/googletest/archive/master.zip"],
-     strip_prefix = "googletest-master",
+     urls = ["https://github.com/google/googletest/archive/3f0cf6b62ad1eb50d8736538363d3580dd640c3e.zip"],
+     strip_prefix = "googletest-3f0cf6b62ad1eb50d8736538363d3580dd640c3e",
 )
--- a/README.md
+++ b/README.md
@@ -933,11 +933,6 @@ sudo cpupower frequency-set --governor powersave
 
 # Known Issues
 
-### Windows
-
-* Users must manually link `shlwapi.lib`. Failure to do so may result
-in unresolved symbols.
-
 ### Solaris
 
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -1,9 +1,7 @@
...
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
-
-git_repository(
-    name = "com_google_googletest",
-    commit = "3f0cf6b62ad1eb50d8736538363d3580dd640c3e",  # HEAD
-    remote = "https://github.com/google/googletest",
+http_archive(
+     name = "com_google_googletest",
+     urls = ["https://github.com/google/googletest/archive/master.zip"],
```
dominichamon: should this be a specific commit so we don't end up with version mismatch issues as googletest releases?

dominichamon:
perhaps we should accelerate a dependency on absl and pull sysinfo from there instead? and merge any changes we've made independently.... and make absl work for cmake :P

WORKSPACE
astrelni: I figured that diamond dependency isn't an issue here because the benchmark library doesn't actually depend on gtest, only the benchmark tests do. So whatever version of gtest is brought in here does not get forwarded to a project that uses benchmark. Let me know if I'm confused on that and I'll look into adding a specific commit back.

astrelni:
I'm not sure that Abseil would want to take sysinfo.h/.cc out of being an internal implementation detail, and we should probably set a good example and not pull the source out here. Good cmake support is coming really soon<sup>tm</sup> :)

WORKSPACE
dominichamon: I'm more worried about tests starting to fail because of some change in the head version of googletest that is later reverted. It's unlikely, but i'd rather depend on a specific version.

something like https://github.com/google/googletest/archive/3f0cf6b62ad1eb50d8736538363d3580dd640c3e.zip in the URL should work.

astrelni:
//absl/base:base is tricky in that it has some internal circular dependencies. There are headers in absl/base/internal that are needed in other parts of abseil, but we would like to keep them as internal implementation details at least for now. However, we could not split them into their own target with non-public visibility. 

The result right now is that they are in hdrs, but we explicitly document that people should not be using anything internal to abseil. It's not going to stop them all, but it's there to give us flexibility in refactoring that internal code.

Now, all that said. We started out with a lot of this code as internal, because it is much easier to take something from internal to public than vice versa. There would need to be a conversation, but we are open to discussing making certain internal details public. Whether or not that applies to sysinfo, I cannot say.

For the time being, I would prefer to work this out in parallel.

On the other hand, I see that this library is tested with MinGW, so I may need to find another way to link against shlwapi.lib. I saw in the comments that added the issue to the README discussing that the pragma would not work on MinGW, but I did not see MinGW as supported on the README either.

WORKSPACE
astrelni: Done

astrelni:
Closing this one in favor of https://github.com/google/benchmark/pull/581 which only attempts to solve the issue for bazel. There we only need to worry about msvc on windows and not mingw.

PR #581, opened 2018-05-01
Update bazel WORKSPACE and BUILD files to work better on Windows.

astrelni:
Note, bazel only supports MSVC on Windows, and not MinGW, so linking against shlwapi.lib only needs to follow MSVC conventions.

Tested this on local Windows 10 machine with Visual Studio 2017. Bazel did not work correctly with git_repository(), so switched it out with http_archive() in the workspace file as found in
https://github.com/abseil/abseil-cpp/blob/master/WORKSPACE#L14
except it is modified to fetch a particular commit of gtest.

This PR is an alternative to https://github.com/google/benchmark/pull/580 which tried to solve this issue for CMake as well and ran into MinGW issues

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -16,7 +16,7 @@ cc_library(
     ]),
     hdrs = ["include/benchmark/benchmark.h"],
     linkopts = select({
-        ":windows": [],
+        ":windows": ["-DEFAULTLIB:shlwapi.lib"],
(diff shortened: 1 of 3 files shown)


astrelni:
Recommended linking against libraries with msvc and bazel taken from
https://github.com/bazelbuild/bazel/issues/3425#issuecomment-316998146

PR #583, opened 2018-05-02
Use __EMSCRIPTEN__ (rather then EMSCRIPTEN) to check for emscripten

sbc100:
The old EMSCRIPTEN macro is deprecated and not enabled when
EMCC_STRICT is set.

Also fix a typo in EMSCRIPTN (not sure how this ever worked).


dominichamon:
thank you. do you know if there's any way to set up a continuous test for enscripten builds?

sbc100:
Its possible, but you would need to pull down the emscripten SDK as part of your build.
That might well be do-able, but even then it likely would not have found this error unless you were testing with strict option and/or with the upstream LLVM backend.  I guess its up to you to decide how much you care about the emscripten target.

PR #584, opened 2018-05-04
There is no "FATAL" in message(), only "FATAL_ERROR"

LebedevRI:
This turns compile-time build failure into a cmake-time build failure, like it's supposed to be.


dominichamon:
whoops, thanks!

PR #587, opened 2018-05-05
Return a reasonable value from State::iterations() even before starting a benchmark

sam-panzer:
This is the smallest fix I could think of. I'm happy to pursue another approach if this is a dead end.


LebedevRI:
(wow, a lot of nop commits, maybe a rebase is needed?)
Is there a bugreport or some textual description of the problem this is solving?

sam-panzer:
Looks like this PR isn't useful after updating to upstream. Closing.

Issue #589, opened 2018-05-06
GetGitVersion fails when cmake is executed


php1ic:
When you run cmake to configure a build and are in a separate build directory, the default version of the project is shown, rather than the current one.

```
[~/git_repos/google/build]$ cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${HOME}/git_repos/google/install/ ../benchmark/
.
.
.
-- Failed to find LLVM FileCheck
-- git Version: v0.0.0
-- Version: 0.0.0
.
.
```

The problem looks to be in cmake/GetGitVersion.cmake file as it expects to be run from the source directory.

I can see how to fix it, so will work on it and issue a pull-request.

php1ic:
Fixed and merged with PR #590

PR #590, opened 2018-05-06
Run git from the source directory (#589)

php1ic:
Git was being executed in the current directory, so could not get the
latest tag if cmake was run from a build directory. Force git to be
run from with the source directory.


dominichamon:
that's odd... i've been running cmake from a build directory for ever:

```
$ cd _build
$ cmake ../
```

and it was working fine.

I don't think this breaks anything, but i'm curious what other differences you might have on your platform.

dominichamon:
oh i see.. your build directory is parallel to the source directory :)

PR #591, opened 2018-05-07
Remove unnecessary memset functions.

NanXiao:


dominichamon:
why do you think this is unnecessary? isn't the instance of `std::tm` being default-initialized here? i don't believe it has a user-defined constructor that would trigger value-initialization (or zero-initialization).

if you do want to avoid the memset, i think you still need an initializer (`{}`) after the variable name to trigger safe initialization.

NanXiao:
@dominichamon IMHO, no matter set `std::tm` to `0` or not, the following [statement](https://github.com/google/benchmark/blob/718cc91d004e15ce483cb1ba9bcd5e3a1f025184/src/timers.cc#L206) will fill `std::tm` correct value:  

    ::gmtime_r(&now, &timeinfo);

Whether `std::tm` is zero-initialized or not, its value is meaningless unless filled by functions like `localtime_r` or `gmtime_r`. So it is no need to call `memset()` before `gmtime_r`.  

Just my personal opinion, don't mind, thanks!

PR #593, opened 2018-05-12
[Tools] Two-Sided Mann–Whitney U test. P-Value

LebedevRI:
Benchmarking is hard. Making sense of the benchmarking results is even harder.

The first problem you have to solve yourself. The second one can be aided.
The benchmark library can compute some statistics over the repetitions,
which helps with grasping the results somewhat.

But that is only for the one set of results. It does not really help to compare
the two benchmark results, which is the interesting bit. Thankfully, there are
these bundled `tools/compare.py` and `tools/compare_bench.py` scripts.

They can provide a diff between two benchmarking results. Yay!
Except not really, it's just a diff, while it is very informative and better than
nothing, it does not really help answer The Question - am i just looking at the noise?
It's like not having these per-benchmark statistics...

Roughly, we can formulate the question as:
> Are these two benchmarks the same?
> Did my change actually change anything, or is the difference below the noise level?

Well, this really sounds like a [null hypothesis](https://en.wikipedia.org/wiki/Null_hypothesis), does it not?
So maybe we can use statistics here, and solve all our problems?
lol, no, it won't solve all the problems. But maybe it will act as a tool,
to better understand the output, just like the usual statistics on the repetitions...

I'm making an assumption here that most of the people care about the change
of average value, not the standard deviation. Thus i believe we can use T-Test,
be it either [Student's t-test](https://en.wikipedia.org/wiki/Student%27s_t-test), or [Welch's t-test](https://en.wikipedia.org/wiki/Welch%27s_t-test). 
**EDIT**: however, after @dominichamon review, it was decided that it is better
to use more robust [Mann–Whitney U test](https://en.wikipedia.org/wiki/Mann–Whitney_U_test)
I'm using [scipy.stats.mannwhitneyu](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mannwhitneyu.html#scipy.stats.mannwhitneyu).

There are two new user-facing knobs:
```
$ ./compare.py --help
usage: compare.py [-h] [-u] [--alpha UTEST_ALPHA]
                  {benchmarks,filters,benchmarksfiltered} ...

versatile benchmark output compare tool
<...>
optional arguments:
  -h, --help            show this help message and exit

  -u, --utest           Do a two-tailed Mann-Whitney U test with the null
                        hypothesis that it is equally likely that a randomly
                        selected value from one sample will be less than or
                        greater than a randomly selected value from a second
                        sample. WARNING: requires **LARGE** (9 or more)
                        number of repetitions to be meaningful!
  --alpha UTEST_ALPHA   significance level alpha. if the calculated p-value is
                        below this value, then the result is said to be
                        statistically significant and the null hypothesis is
                        rejected. (default: 0.0500)
```

Example output:
screenshot_20180512_175517
As you can guess, the alpha does affect anything but the coloring of the computed p-values.
If it is green, then the change in the average values is statistically-significant.

I'm detecting the repetitions by matching name. This way, no changes to the json are _needed_.
Caveats:
* This won't work if the json is not in the same order as outputted by the benchmark,
   or if the parsing does not retain the ordering.
* This won't work if after the grouped repetitions there isn't at least one row with
  different name (e.g. statistic). Since there isn't a knob to disable printing of statistics
  (only the other way around), i'm not too worried about this.
* **The results will be wrong if the repetition count is different between the two benchmarks being compared.**
* Even though i have added (hopefully full) test coverage, the code of these python tools is staring
  to look a bit jumbled.
* So far i have added this only to the `tools/compare.py`.
  Should i add it to `tools/compare_bench.py` too?
  Or should we deduplicate them (by removing the latter one)?

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -4,6 +4,9 @@ import os
 import re
 import copy
 
+from scipy.stats import mannwhitneyu
+
+
 class BenchmarkColor(object):
     def __init__(self, name, code):
         self.name = name
@@ -16,11 +19,13 @@ class BenchmarkColor(object):
     def __format__(self, format):
         return self.code
 
+
 # Benchmark Colors Enumeration
 BC_NONE = BenchmarkColor('NONE', '')
 BC_MAGENTA = BenchmarkColor('MAGENTA', '\033[95m')
 BC_CYAN = BenchmarkColor('CYAN', '\033[96m')
 BC_OKBLUE = BenchmarkColor('OKBLUE', '\033[94m')
+BC_OKGREEN = BenchmarkColor('OKGREEN', '\033[32m')
 BC_HEADER = BenchmarkColor('HEADER', '\033[92m')
 BC_WARNING = BenchmarkColor('WARNING', '\033[93m')
 BC_WHITE = BenchmarkColor('WHITE', '\033[97m')
@@ -28,6 +33,7 @@ BC_FAIL = BenchmarkColor('FAIL', '\033[91m')
 BC_ENDC = BenchmarkColor('ENDC', '\033[0m')
 BC_BOLD = BenchmarkColor('BOLD', '\033[1m')
 BC_UNDERLINE = BenchmarkColor('UNDERLINE', '\033[4m')
+
 
 def color_format(use_color, fmt_str, *args, **kwargs):
     """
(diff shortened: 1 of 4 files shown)


LebedevRI:
Thank you for taking a look!

> So this means you're comparing the repetitions, which are already averages of iteration times. 

Yep, obviously.

> Idle thought: should we have a mode that outputs every iteration into a JSON file? :D

This comes up every so often..
There are obvious downsides:
1. the timer will need to be stopped each time. i don't think we currently stop it unless the code tells to.
2. this will increase memory (and on-disk) footprint of the measurements

The last one is expected.
But i do expect that it may be useful sometimes.
So if we can somehow not penalize the default case (the current way), **i would be interested in this**.

> but I wonder if you might want to require more than two repetitions to define a distribution.

Yes, clearly, but i'm also not sure *what* would be the sane choice.
Maybe state it in help, and print a warning if repetition count is too small? (less than 5?)

> Do you have any reason to assume a priori that the distributions of benchmark repetition results are normally distributed?

No. Only factual (which means - probably invalid) evidence of t-test being used for this purpose.

> A more general approach is the Mann-Whitney U test. 

Ok, great, will take a look!

LebedevRI:
Ping.

tools/gbench/report.py
```diff
@@ -16,11 +19,13 @@ def __repr__(self):
     def __format__(self, format):
         return self.code
 
+
 # Benchmark Colors Enumeration
 BC_NONE = BenchmarkColor('NONE', '')
 BC_MAGENTA = BenchmarkColor('MAGENTA', '\033[95m')
 BC_CYAN = BenchmarkColor('CYAN', '\033[96m')
 BC_OKBLUE = BenchmarkColor('OKBLUE', '\033[94m')
+BC_OKGREEN = BenchmarkColor('OKBLUE', '\033[32m')
```
dominichamon: s/OKBLUE/OKGREEN/

tools/gbench/report.py
```diff
@@ -78,30 +84,82 @@ def filter_benchmark(json_orig, family, replacement=""):
...
 
 
-def generate_difference_report(json1, json2, use_color=True):
+def generate_difference_report(
+        json1,
+        json2,
+        utest=False,
+        utest_alpha=0.05,
+        use_color=True):
+    assert utest is True or utest is False
```
dominichamon: this should be under the docstring. also, if you care about the type being boolean specifically, maybe add type annotations :)

otherwise, it's up to the caller to be truthy or not. this is Python, after all.

tools/gbench/report.py
```diff
@@ -200,6 +281,41 @@ def test_basic(self):
             self.assertEqual(parts, expect_lines[i])
 
 
+class TestReportDifferenceWithTTest(unittest.TestCase):
```
dominichamon: nit: UTest

tools/gbench/report.py
LebedevRI: True.
I'm just following pre-existing practice e.g. with `color_format()` doing this.

tools/gbench/report.py
LebedevRI: Right.

LebedevRI:
Is there any preferences re enabling this by default?
I have finally got around to actually using all these my changes, and it took me a moment to realize why u test results aren't printed :)

tools/gbench/report.py
```diff
@@ -78,30 +84,82 @@ def filter_benchmark(json_orig, family, replacement=""):
...
         filtered['benchmarks'].append(filteredbench)
     return filtered
 
 
-def generate_difference_report(json1, json2, use_color=True):
+def generate_difference_report(
+        json1,
+        json2,
+        utest=False,
+        utest_alpha=0.05,
```
dominichamon: why are these defaulted? they're always passed in, and the default is defined in the calling binary.

LebedevRI: The unittests later in this file not always pass them

dominichamon:
ok. i'll merge this and then do a followup to default to true? (so it's easier to roll back if we need to. which we likely won't).

LebedevRI:
SGTM!

LebedevRI:
@dominichamon thank you!

PR #596, opened 2018-05-15
Fix compilation on Android with GNU STL

Maratyszcza:
GNU STL in Android NDK lacks string conversion functions from C++11, including `std::stoul`, `std::stoi`, and `std::stod`. This patch replaces calls to these functions with C-style equivalents from C++03.

### DIFF
--- a/src/complexity.cc
+++ b/src/complexity.cc
@@ -26,6 +26,7 @@ namespace benchmark {
 
 // Internal function to calculate the different scalability forms
 BigOFunc* FittingCurve(BigO complexity) {
+  static const double kLog2E = 1.44269504088896340736;
   switch (complexity) {
     case oN:
       return [](int64_t n) -> double { return static_cast<double>(n); };
@@ -34,9 +35,11 @@ BigOFunc* FittingCurve(BigO complexity) {
     case oNCubed:
       return [](int64_t n) -> double { return std::pow(n, 3); };
     case oLogN:
-      return [](int64_t n) { return log2(n); };
+      /* Note: can't use log2 because Android's GNU STL lacks it */
+      return [](int64_t n) { return kLog2E * log(n); };
     case oNLogN:
-      return [](int64_t n) { return n * log2(n); };
+      /* Note: can't use log2 because Android's GNU STL lacks it */
+      return [](int64_t n) { return kLog2E * n * log(n); };
     case o1:
     default:
       return [](int64_t) { return 1.0; };
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -169,4 +169,93 @@ void ReplaceAll(std::string* str, const std::string& from,
   }
 }
 
+#ifdef BENCHMARK_STL_ANDROID_GNUSTL
+/*
+ * GNU STL in Android NDK lacks support for some C++11 functions, including
+ * stoul, stoi, stod. We reimplement them here using C functions strtoul,
+ * strtol, strtod. Note that reimplemented functions are in benchmark::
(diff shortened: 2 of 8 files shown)


dominichamon:
we use c++11 throughout benchmark. perhaps it's worth adding a platform check for this, if it's unique to Android?

Maratyszcza:
Android STL lives in Android NDK, apps include pre-compiled STL along with their own code. The problem fixed by this patch is present since ancient times (since C++ got support in Android NDK) and up to and including Android NDK r17.

dominichamon:
there are some semantic differences between the old and new methods. The newer std:: versions will throw exceptions on error, while the old versions will return 0 (or 0.0) and set `errno`. Strictly speaking, if we're going to use those we should check errno just in case something goes wrong reading the system info.

I think it may be worth wrapping them in a method that uses `#ifdef BENCHMARK_OS_ANDROID` to call the new or old versions, and check the `errno` correctly.

Maratyszcza:
@dominichamon Please take another look

src/string_util.h
```diff
@@ -32,6 +32,100 @@ inline std::string StrCat(Args&&... args) {
   return ss.str();
 }
 
+#if defined(__ANDROID__) && defined(__GLIBCXX__)
```
dominichamon: can you define a BENCHMARK_OS_ANDROID where we define the other OS constants (src/internal_macros.h)?

i'm not sure we currently differentiate on libcxx but maybe we should add something there too.

src/string_util.h
```diff
@@ -32,6 +32,100 @@ inline std::string StrCat(Args&&... args) {
...
+/*
+ * GNU STL in Android NDK lacks support for some C++11 functions, including
+ * stoul, stoi, stod. We reimplement them here using C functions strtoul,
+ * strtol, strtod. Note that reimplemented functions are in benchmark::
+ * namespace, not std:: namespace.
+ */
+inline unsigned long stoul(const std::string& str, size_t* pos = nullptr,
```
dominichamon: why inline and why in the header?

Maratyszcza:
@dominichamon Added `BENCHMARK_STL_ANDROID_GNUSTL` macro, and moved implementation into `.cc` file

src/string_util.cc
```diff
@@ -169,4 +169,93 @@ void ReplaceAll(std::string* str, const std::string& from,
   }
 }
 
+#ifdef BENCHMARK_STL_ANDROID_GNUSTL
```
dominichamon: can you add tests for these please? 

you'll have to create a test/string_util_gtest.cc file, and add it to test/CMakeLists.txt. bazel will pick it up automatically.

Maratyszcza: Added tests and replaces `log2(n)` with `1.44269504088896340736 * log(n)` as Android GNUSTL lacks that too.

src/complexity.cc
```diff
@@ -34,9 +34,11 @@ BigOFunc* FittingCurve(BigO complexity) {
     case oNCubed:
       return [](int64_t n) -> double { return std::pow(n, 3); };
     case oLogN:
-      return [](int64_t n) { return log2(n); };
+      /* Note: can't use log2 because Android's GNU STL lacks it */
+      return [](int64_t n) { return 1.44269504088896340736 * log(n); };
```
dominichamon: can you at least add a kLog2E constant?

Maratyszcza: @dominichamon done

enh:
(note that gnustl is being removed from NDK r18.)

PR #598, opened 2018-05-23
Return 0 from State::iterations() when not yet started.

sam-panzer:
Previously, State::iterations() returned some unexpected value (typically max_iterations). State::iterations() will now return 0 until KeepRunning() or KeepRunningBatch() is called.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -575,7 +575,7 @@ class State {
 
   BENCHMARK_ALWAYS_INLINE
   size_t iterations() const {
-    if (!started_) {
+    if (BENCHMARK_BUILTIN_EXPECT(!started_, false)) {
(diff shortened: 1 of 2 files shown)


include/benchmark/benchmark.h
```diff
@@ -575,7 +575,10 @@ class State {
 
   BENCHMARK_ALWAYS_INLINE
   size_t iterations() const {
-    return (max_iterations - total_iterations_ + batch_leftover_);
+    if (!started_) {
```
dominichamon: maybe the option to slightly improve codegen by assuming branch prediction?

sam-panzer: I'm not sure what you mean. Were you looking for BENCHMARK_BUILTIN_EXPECT or something else?

PR #599, opened 2018-05-23
Benchmark IDs, Names, Grouping Support: Better Tooling

BaaMeow:
This pull request is a follow-up from #567 and the concerns raised therein. There were a number of demonstrations of various needs, and this Pull Request fills all of those needs, and more, by introducing 2 new identifiers and 3 new fields to output (**except the console, which is left the same to not affect individuals parsing Console Output and who need that stability**).

- 5 fields are now in place on JSON/CSV data
    1. "base_name" (the name one registers with the benchmark)
    2. "threads" (the number of threads for that instance of the benchmark)
    3. "repetitions" (the number of repetitions for the benchmark)
    4. "id" (the instance id of the benchmark)
    5. "family" (the id of the family of benchmarks spawned by a single call + parameters passed to a RegisterBenchmark/BENCHMARK()/etc. call).
- All tests were updated to look for, inspect, and demand the new fields be present.
- Console output is **not changed**, but 3 new options were added to have it print the ID, Family, and Base Name attributes if (and ONLY if) the new `benchmark_console_id`, `benchmark_console_base_name`, and `benchmark_console_family` parameters are explicitly set to a 'truthy' value

I believe these changes adequately address all concerns and pave the way for future tooling that may need to add more "canonical identifiers" to the name (rather than append to the name, instead add a field that describes the new axis of information being provided).

This PR also adds a minor fix that addresses #523, since it was blocking some of our work and that PR has been untouched for a few months by the original author...


LebedevRI:
Why so many formatting changes.. Are you using msvs, which always formats the entire files?

.gitignore
```diff
@@ -1,57 +1,58 @@
...
-build/
-_build/
-
-# in-source dependencies
-/googletest/
-
-# Visual Studio 2015/2017 cache/options directory
-.vs/
-CMakeSettings.json
+*.a
```
dominichamon: i'm not sure why this entire file shows as modified, but you might want to look into that.

include/benchmark/benchmark.h
```diff
@@ -164,7 +164,6 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #ifndef BENCHMARK_BENCHMARK_H_
```
dominichamon: could you send a separate PR with the formatting changes? it'll help keep this focused on the actual functional change (and i'm happy to land pure format changes quickly).

src/benchmark.cc
```diff
@@ -386,29 +407,56 @@ void State::FinishKeepRunning() {
 namespace internal {
 namespace {
 
+int base10_digit_count(int64_t n) {
```
dominichamon: watch the naming (`Base10DigitCount`). Also, given this isn't on the critical path, just using `std::log10(n) + 1` should be good for all cases instead of the unrolled version.

src/benchmark.cc
```diff
@@ -562,6 +625,9 @@ void PrintUsageAndExit() {
           "          [--benchmark_out_format=<json|console|csv>]\n"
           "          [--benchmark_color={auto|true|false}]\n"
           "          [--benchmark_counters_tabular={true|false}]\n"
+          "          [--benchmark_console_id={true|false}]\n"
```
dominichamon: please split this out to a separate PR. I'm not sure we will need it, and i'd like to keep this PR focused on the important bits.

src/benchmark.cc
```diff
@@ -571,25 +637,31 @@ void ParseCommandLineFlags(int* argc, char** argv) {
...
+                      &FLAG(benchmark_console_id)) ||
+        ParseBoolFlag(argv[i], "benchmark_console_family",
+                      &FLAG(benchmark_console_family)) ||
+        ParseBoolFlag(argv[i], "benchmark_console_base_name",
+                      &FLAG(benchmark_console_base_name)) ||
+        ParseStringFlag(argv[i], "benchmark_color", &FLAG(benchmark_color)) ||
         // "color_print" is the deprecated name for "benchmark_color".
         // TODO: Remove this.
-        ParseStringFlag(argv[i], "color_print", &FLAGS_benchmark_color) ||
+        ParseStringFlag(argv[i], "color_print", &FLAG(benchmark_color)) ||
```
dominichamon: there's no need to introduce this `FLAG(..)` macro. It doesn't add anything useful.

src/benchmark_register.cc
```diff
@@ -62,6 +62,23 @@ namespace internal {
 //=============================================================================//
 //                         BenchmarkFamilies
 //=============================================================================//
+int64_t GetBenchmarkUniqueId() {
```
dominichamon: Why is the base name and overall benchmark name not enough for this?

src/csv_reporter.cc
```diff
@@ -22,27 +22,31 @@
 #include <tuple>
```
dominichamon: split the CSV changes out to another PR. I don't know that we need it, and it's helpful to keep this PR more focused.

src/internal_macros.h
```diff
@@ -11,79 +11,79 @@
 #endif
```
dominichamon: we should clang-format off this. when you get around to the format PR please make sure this isn't changed.

src/benchmark.cc
BaaMeow: I didn't introduce the FLAG macro; this was present in the source already. It was defined as "the way you are supposed to use it to refer to flags". I simply noticed that it was defined as the way to do things and we didn't use it anywhere. I will revert the changes, but does that mean the `FLAG` macro present [here](https://github.com/google/benchmark/blob/master/src/commandlineflags.h#L8) should not be around...?

BaaMeow:
@dominichamon I went through and made the format PR (#610). I'll remake / rebase this once that PR goes through, with the splits you asked for (CSV in another PR, command line parameters in another PR, etc...)

src/benchmark_register.cc
BaaMeow: The base name can be similar between several benchmarks. See the example posted on the last PR, where a NOP function with only slight variations in repetitions, threads, etc. were applied: https://github.com/google/benchmark/pull/567#issuecomment-386820365

PR #600, opened 2018-05-24
Corrections, additions to initial doc

mattreecebentley:
See individual commits for details, but 3rd option provided appears not to work.
Overall I think the documentation is poor for first-timers, these are merely some minor corrections to reduce some of the confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,15 +31,13 @@ $ cmake -G <generator> [options] ../benchmark
 $ make
 ```
 
-Note that Google Benchmark requires GTest to build and run the tests. This
-dependency can be provided three ways:
+Note that Google Benchmark requires Google Test to build and run the tests. This
+dependency can be provided two ways:
 
-* Checkout the GTest sources into `benchmark/googletest`.
+* Checkout the GTest sources into `benchmark/googletest` as above.
 * Otherwise, if `-DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON` is specified during
   configuration, the library will automatically download and build any required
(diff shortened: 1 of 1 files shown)


mattreecebentley:
I signed

dominichamon:
Thanks so much for helping clean up the documentation for first time users. If you have any other specific issues, please feel free to raise an issue, create a PR, or contact us on freenode IRC at #googlebenchmark.

mattreecebentley:
Thanks - I think it needs a lot more work, it's just that I don't actually understand what's going on enough at this point to re-write it. There's no details about what methodology the benchmarks is using, and I can only pick up anecdotal stuff elsewhere.

dominichamon:
If you have specific questions then I think we'll be able to tackle them,
and then we can use that to add more to the documentation. Feel free to
start a thread on the benchmark-discuss google group so we can capture it.

Dominic Hamon | Google

mattreecebentley:
It's more that there's just no detail there. No information about what the individual macro/functions are doing, nothing. I don't feel like this is something that even requires discussion, it's that bizarre, to me.

README.md
```diff
@@ -88,6 +87,11 @@ to use, test, and provide feedback on the new features are encouraged to try
 this branch. However, this branch provides no stability guarantees and reserves
 the right to change and break the API at any time.
 
+##Prerequisite knowledge
```
LebedevRI: While this may be a valid markdown, it isn't displayed correctly on github,
A space is missing.

PR #601, opened 2018-05-24
Add benchmark_main target.

astrelni:
This provides an alternative to using `BENCHMARK_MAIN();` at the end of benchmark source files. Linking against a default "_main" target makes the workflow more like that of googletest. 

It is also easier for working with benchmark targets that may have benchmark functions split out over several files since `BENCHMARK_MAIN();` must only appear once.

After these changes, users can either continue to use `BENCHMARK_MAIN();` and link to `benchmark`, or they can link to only `benchmark_main`. Existing uses should be unaffected.


pleroy:
Could we use this (and remove `BENCHMARK_MAIN`) in the tests of this project?  The current situation of having one `main` per test makes it hard to link all the tests in a single binary.  (Why you'd want to do that: to have a single test project in Visual Studio.)

dominichamon:
@pleroy This would be a followup, but I'm also not sure we'd want that anyway. I like to run my tests in parallel when sanity checking library changes and a single test binary would take quite a while to run.

Unless we can do both: define a library for each test with a binary target that links in benchmark_main, and then an uber_test binary on top that links all the things?

astrelni:
@dominichamon My understanding of @pleroy's comment (please correct me) is that this project would continue to have the same test targets, but each target would use `benchmark_main` instead of ` BENCHMARK_MAIN();`. That would allow users who want to keep a local copy of the repository to use a custom provided build configuration not generated by CMake. That could allow them to combine multiple benchmark source files into one target without first having to strip out the `BENCHMARK_MAIN();`s every time they sync their copy.

dominichamon:
OK, sure, that works. It's been a while since I used VS but that sounds fine.

@pleroy Are you ok if we do this in a follow up?

pleroy:
Exactly what @astrelni said.  A follow-up is fine by me.

astrelni:
@pleroy @dominichamon I took a quick look this morning and about half of the benchmark tests in the repository use custom a custom `main()`. I don't know if the follow up would be very practical at this point.

pleroy:
Dealing the low-hanging fruit of replacing the macro with the `benchmark_main` target would already be an improvement IMO.

PR #609, opened 2018-05-31
Disable deprecation warnings when -Werror is enabled.

### LINKED ISSUE #608: CSVReporter giving Wdeprecated-declarations warning from inside benchmark.cc
CSVReporter is marked deprecated as of https://github.com/google/benchmark/pull/488, but the type is still used inside benchmark.cc https://github.com/google/benchmark/blob/master/src/benchmark.cc#L451, breaking -Werror builds.

dominichamon:
Fixes #608


dominichamon:
@astrelni I think this should help.

PR #610, opened 2018-05-31
(clang-)format all the things

BaaMeow:
This is a pull request is entirely editorial and simply uses clang-format on everything.

The goal of this pull-request is to make it so people whose clang-format auto-fires when a `.clang-format` is detected for a directory does not absolutely destroy the content of their pull requests due to auto-formatting.

It's also just nice. 😄

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -115,8 +115,7 @@ namespace {
 
 BenchmarkReporter::Run CreateRunReport(
     const benchmark::internal::Benchmark::Instance& b,
-    const internal::ThreadManager::Result& results,
-    double seconds) {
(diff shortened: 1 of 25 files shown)


EricWF:
I'm not thrilled to change the blame on almost every line.

People shouldnt be blindly formatting entire files when they submit pull requests. That's what `git clang-format` is for

dominichamon:
I'm ok with this. git blame can be limited to a range of commits, and git log is a better tool for tracking changes over time anyway.

I agree that git clang-format would be ideal, but given we've had a number of PRs where the entire file ends up being changed, i'd rather get to a 'known good' state.

EricWF:
then I withdraw my objection.

Issue #603, opened 2018-05-25
Use of cmake 3.6 feature with cmake_minimum_required (VERSION 2.8.12)


jszuppe:
https://github.com/google/benchmark/commit/e776aa0275e293707b6a0901e0e8d8a8a3679508 This commit adds `list(FILTER SOURCE_FILES EXCLUDE REGEX "benchmark_main\\.cc")` line to https://github.com/google/benchmark/blob/master/src/CMakeLists.txt. `list(FILTER` was introduced in CMake 3.6, however you script only sets minimum required CMake version to `2.8.12`.

fengggli:
I got the same one!

kfitch:
I can't get to a computer where I can fork/edit/test code at the moment, but I think the proper solution (that should work on cmake 2.8.12)  is something like:

```cmake
file(GLOB BENCHMARK_MAIN "benchmark_main.cc")
foreach(item ${BENCHMARK_MAIN})
    list(REMOVE_ITEM SOURCE_FILES ${item})
endforeach()
```

Maratyszcza:
PR #612 for a fix

EricWF:
I would much rather bump the required CMake version. CMake is trivial to install on all platforms. There is no reason for people to be using CMake versions that old.

Maratyszcza:
@EricWF CMake 3.5.1 is the latest available in Ubuntu 14.04 and 16.04 (LTS releases).

EricWF:
So we should at least bump it to that, but again, it's trivial for users to install newer versions.

LebedevRI:
Documenting (and sticking with a) dependency versioning policy would be nice[r].
Personally i think trying to support 3-years old versions of the deps
(+ debian stable + last two ubuntu lts) is the correct solution.
There can be exceptions of course.

EricWF:
LLVM is currently requiring 3.4.3, with the other information provided above it seems like we can agree to bump it to 3.5.1?

LebedevRI:
[cmake v3.5.1](https://github.com/Kitware/CMake/releases/tag/v3.5.1) was released `Mar 24, 2016`, which is 2 years 2 months. `v3.2.3` is the newest that would fit into that sliding window.
That sliding window seems to be somewhat reasonable.
I don't mean to halt forward progress, i'm just trying to hint that cost-benefit analysis,
and ideally effect on (not just one) users is a good thing.

EricWF:
[...]

PR #612, opened 2018-06-05
Avoid using CMake 3.6 feature list(FILTER ...)

Maratyszcza:
`list(FILTER ...)` is a CMake 3.6 feature, but benchmark targets CMake 2.8.12

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -11,7 +11,10 @@ file(GLOB
     *.cc
     ${PROJECT_SOURCE_DIR}/include/benchmark/*.h
     ${CMAKE_CURRENT_SOURCE_DIR}/*.h)
-list(FILTER SOURCE_FILES EXCLUDE REGEX "benchmark_main\\.cc")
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ file(GLOB
     *.cc
     ${PROJECT_SOURCE_DIR}/include/benchmark/*.h
     ${CMAKE_CURRENT_SOURCE_DIR}/*.h)
-list(FILTER SOURCE_FILES EXCLUDE REGEX "benchmark_main\\.cc")
+list(REMOVE_ITEM SOURCE_FILES "benchmark_main.cc")
```
LebedevRI: Does that actually work though?
`GLOB` should result in absolute filenames..

Maratyszcza: Right, fix coming

Maratyszcza: Switched to solution suggested by @kfitch in #603

dominichamon:
thanks for the quick fix.

PR #616, opened 2018-06-12
[JSON] Benchmark IDs, Names, Grouping Support: Better Tooling

BaaMeow:
This pull request is the same as #599, but is rebased after the formatting changes and **only** includes the JSON changes, leaving out console and CSV for later (or never).

5 fields are now in place on JSON data
- "base_name" (the name one registers with the benchmark)
- "threads" (the number of threads for that instance of the benchmark)
- "repetitions" (the number of repetitions for the benchmark)
- "id" (the instance id of the benchmark)
- "family" (the id of the family of benchmarks spawned by a single call + parameters passed to a RegisterBenchmark/BENCHMARK()/etc. call).

All tests were updated to look for, inspect, and demand the new fields be present.

There are 2 new generated IDs asides from what comes from the composed "name" attribute of each benchmark.

# "id"
this is a number tied to each created instance of a benchmark, and is unique across every single instance of that benchmark. It uniquely ties a single registered benchmark to the statistics generated for that benchmark. It solves the original problem in #567.

# "family"
This is a number tied to a single family of benchmarks: for example, a benchmark registered with `BENCHMARK(my_benchmark)->Range(8, 8 << 10);` will generate many, many different instances for the range, but they all conceptually belong to this same "family". Families are useful for knowing each intentionally registered group. It also gets past potential self-hostile users who register benchmarks under the exact same name: they can be differentiated by their family ID.

The rest of the fields are a destructing of the typical name: `base_name` is the core name that is registered with the benchmark. `threads` is the used thread count for the benchmark. `repetitions` is the number of repetitions performed, set at the command line or directly on the benchmark itself.

Each one of these allows for optimal and easy grouping, and allows individual runs of a repeating benchmark to have their stats group together with others. `name` parsing / shenanigans will continue to
work as before.

Console output changes and CSV output changes are left for another Pull Request, potentially as a follow-up to this one.


BaaMeow:
Some test cases for the edge / adversarial cases:

````

static void adversarial(benchmark::State &state)
{
	for (auto _ : state)
	{
	}
}
BENCHMARK(adversarial);
BENCHMARK(adversarial);

static void many_range_rep(benchmark::State &state)
{
	for (auto _ : state)
	{
	}
}
BENCHMARK(many_range_rep);
BENCHMARK(many_range_rep)->Repetitions(3);
BENCHMARK(many_range_rep)->Range(8, 8 << 10);
BENCHMARK(many_range_rep)->Range(8, 8 << 10)->Repetitions(3);

static void common_prefix(benchmark::State &state)
{
	for (auto _ : state)
	{
	}
}
static void common_prefix_mean(benchmark::State &state)
{
	for (auto _ : state)
	{
	}
}
BENCHMARK(common_prefix);
BENCHMARK(common_prefix_mean);
````

[JSON output from above test-case benchmarks](https://gist.github.com/BaaMeow/f37127fecd5a862f601fcd85b90f680d)

File: src/complexity.cc [hunk 1]
```diff
@@ -182,12 +182,17 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_cpu = MinimalLeastSq(n, cpu_time, reports[0].complexity);
     result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
   }
-  std::string benchmark_name =
-      reports[0].benchmark_name.substr(0, reports[0].benchmark_name.find('/'));
+  std::string benchmark_name = reports[0].base_name;
```
LebedevRI: Hm, this looks like a non-NFC change, but i don't see anything in test changes?

src/json_reporter.cc
```diff
@@ -78,7 +78,12 @@ bool JSONReporter::ReportContext(const Context& context) {
   out << indent << FormatKV("date", walltime_value) << ",\n";
 
   if (Context::executable_name) {
-    out << indent << FormatKV("executable", Context::executable_name) << ",\n";
+    // windows uses backslash for its path separator,
+    // which must be escaped in JSON otherwise it blows up conforming JSON
+    // decoders
+    std::string executable_name = Context::executable_name;
+    ReplaceAll(&executable_name, "\\", "\\\\");
+    out << indent << FormatKV("executable", executable_name) << ",\n";
```
LebedevRI: I'm not seeing this change reflected in tests.

test/user_counters_test.cc
```diff
@@ -142,6 +158,11 @@ BENCHMARK(BM_Counters_Threads)->ThreadRange(1, 8);
 ADD_CASES(TC_ConsoleOut, {{"^BM_Counters_Threads/threads:%int %console_report "
                            "bar=%hrfloat foo=%hrfloat$"}});
 ADD_CASES(TC_JSONOut, {{"\"name\": \"BM_Counters_Threads/threads:%int\",$"},
+                       {"\"base_name\": \"BM_Counters_Threads\",$", MR_Next},
```
LebedevRI: This is still bothering me. It just looks completely wrong.

dominichamon:
In the output example you have:

```
      "repetitions": -3689348814741910324,
      "threads": -3689348814741910324,
```

this doesn't look right.

Also, if the ID only serves to group foo:repeats/3 and foo:repeats/3_mean, but you want to also know when it's a statistical output vs a BM run, you could likely add a field for whether it's statistical. maybe a "stats" field that would be "mean" or "median" that could be queried.

I'm still missing why unique IDs are necessary when we have any number of names that can be generated, that at least will be consistent across runs.

src/complexity.cc
BaaMeow: Base name is equivalent to everything leading up to the first `/`, so there's no reason to run a `substr`.

src/json_reporter.cc
BaaMeow: The tests do not explicitly check the full executable path, just the resulting endname. Properly escaped names show up in the appveyor Windows tests now.

test/user_counters_test.cc
BaaMeow: The full name is ` {base_name}\threads:{thread}\{repetitions}`. This PR provides access to the parts, and lets you re-assemble them and group them arbitrarily to your liking. We provide ID numbers for instances, so you can pick statistics, and provide family ID numbers, so you can differentiate between benchmarks which are -- for whatever reason -- registered with the same name.

BaaMeow:
I'm still working on this, just as an FYI. I've just been terribly busy!

I guess I'm not very good at rebasing / merges (or PGP and commit signing). Sorry about that; it looked okay on my local box but I guess I have some more work to do!

BaaMeow:
@dominichamon I added the name "stats" with the name of the statistic (including for RMS and BigO). That mostly solves the problem, since there's a distinct way to tell runs apart now.

I still think "family" is useful as an ID, since I would want to group things without having to hobble together the name. There are also cases where someone can register things of the exact same name. For example, below is a collection of degenerately named and used benchmarks. [This is what the output looks like](https://gist.github.com/BaaMeow/1cbb468e94c138098bd3cb1cdd7159ca).

With the name "stats" and the family ID, we probably don't need "id" anymore. I'm still a bit hesitant to pull it, because I think that having a unique tag for every single sample is a useful feature when something gets parsed. But, it's no longer completely essential (and I can probably live without having an id -> source location line mapping for the data, and just store the source location next to every sample in my data set for warning reporting of bad samples or bizarre outliers in the tools I'm making).

```c++

static void adversarial(benchmark::State& state) {
	int x = 0;
	for (auto _ : state) {
		x += 2;
		benchmark::DoNotOptimize(x);
	}
}
BENCHMARK(adversarial);
BENCHMARK(adversarial);

static void common_prefix(benchmark::State& state) {
	for (auto _ : state) {
	}
}
static void common_prefix_mean(benchmark::State& state) {
	for (auto _ : state) {
	}
}
static void using_the_underscore_poorly_(benchmark::State& state) {
	for (auto _ : state) {
	}
}
BENCHMARK(common_prefix)->Repetitions(3);
BENCHMARK(common_prefix_mean)->Repetitions(3);
BENCHMARK(using_the_underscore_poorly_)->Repetitions(3);

void complexity(benchmark::State& state) {
	for (auto _ : state) {
		for (int i = 0; i < 1024; ++i) {
			benchmark::DoNotOptimize(&i);
		}
	}
}
BENCHMARK(complexity)->Range(1, 1 << 18)->Complexity(benchmark::o1);
BENCHMARK(complexity)->Range(1, 1 << 18)->Complexity();
BENCHMARK(complexity)->Range(1, 1 << 18)->Complexity([](int64_t) {
	return 1.0;
});

std::vector<int> random_vector(int64_t size) {
	std::vector<int> v;
	v.reserve(static_cast<int>(size));
	for (int i = 0; i < size; ++i) {
		v.push_back(static_cast<int>(std::rand() % size));
	}
	return v;
}

void complexity_rms(benchmark::State& state) {
	auto v = random_vector(state.range(0));
	const int64_t item_not_in_vector = state.range(0) * 2;
	for (auto _ : state) {
		benchmark::DoNotOptimize(std::find(v.begin(), v.end(), item_not_in_vector));
	}
}
BENCHMARK(complexity_rms)
	->RangeMultiplier(2)
	->Range(1 << 10, 1 << 16)
	->Complexity(benchmark::oN);
BENCHMARK(complexity_rms)
	->RangeMultiplier(2)
	->Range(1 << 10, 1 << 16)
	->Complexity([](int64_t n) -> double { return static_cast<double>(n); });
BENCHMARK(complexity_rms)
	->RangeMultiplier(2)
	->Range(1 << 10, 1 << 16)
	->Complexity();

static void many_range_rep(benchmark::State& state) {
	for (auto _ : state) {
	}
}
BENCHMARK(many_range_rep);
BENCHMARK(many_range_rep)->Repetitions(3);
BENCHMARK(many_range_rep)->Range(8, 8 << 10);
BENCHMARK(many_range_rep)->Range(8, 8 << 10)->Repetitions(3);
````

LebedevRI:
PR tips:
* Fetch the remotes (the `google/benchmark`) repo
* Rebase your local `git master` branch on top of `git master` of `google/benchmark` repo
* Push that to your github repo clone
* Rebase your WIP branch ontop of your new `git master`
* [force]Push that to your github repo clone

Else there is a lot of unrelated noise in the PR.

src/json_reporter.cc
BaaMeow: I changed it to check for either `//` or an escaped `\\\\` in the reporter output tests.

BaaMeow:
@LebedevRI Thanks for the tips; I think I managed to clean it up. The googlebot seems to agree; I just need to check over the diffs to make sure they're no longer super messy...

There's a few places where the clang-format er had fun, but it didn't seem to drastically alter the presentation of anything. If something looks wrong do let me know: otherwise, everything looks to be okay (just need to wait for Appveyor and Travis-CI to report that everything is fine).

LebedevRI:
> I think I managed to clean it up.

Thanks, this is indeed better.

It would be best if you could split the `\\\\` change (+ tests!) into a separate pr,
because it **can** be merged right away, separately from the rest of this.

BaaMeow:
@LebedevRI Done! #652

LebedevRI:
Ok, that is great, some progress. Now, how about let's try to split this a bit more.

I think it would be completely uncontroversial to introduce that `stats` field.
@dominichamon, @EricWF ?

Can you split it into yet another PR? :)

BaaMeow:
@LebedevRI Sure, if anyone else chimes in and says just the `stats` part is wanted I will be more than happy to break it off into another PR and leave the ID bits for later.

As a small update, I changed `id` in the JSON to be the much more descriptive `"stats_group"` label. It just makes a bit more sense than `ìd`.

LebedevRI:
I had another look, and i still stand with the same opinion :/
I think this does the right thing, and does move in the right direction.
But. It implements loads of **different** (but related, yes) things at once:
* Store **existing** data in the JSON:
  1. threads
  2. repetitions
  3. aggregate name (statistic, rms, bigo)
These three are **great**, and completely uncontroversial on their own. I'd merge them.
* Add new*ish* data:
  * ID's.
  * Store more `name` variants.
Again, both of these most likely make sense.
But It's really hard to tell (due to the diff's size) if they do the sane thing or not.
Also, they really should have separate tests, that actually check the names/ids.

So i would still suggest to split this up further...

BaaMeow:
@LebedevRI Ping pong: sorry for the quiet! I've been **way** too busy for my own good, but I'm back now and I have quite a bit of free time!

I wanted to get up to speed again and help move forward some of the wanted things. Judging from #675, it looks like a stats_name (aggregate name) was pulled into a separate PR by you and handled nicely, yay!

Should I go ahead and add a new PR for threads and repetition counts, since those are uncontroversial? I can probably get those done by Monday, and then we can figure out if additional identifiers are still worth it!

BaaMeow:
With #748 going live, if its accepted then the only thing that might be needed are some of the family IDs for the case of clashing and degenerate benchmark names. I'll make a new pull request with the last of that after #748 is reviewed.

Happy Holidays to everyone! 🎄

BaaMeow:
I won't be pursuing the IDs. While helpful for bringing me to the exact right place for the data samples, I'm sure I can figure out other things to get around it.

If someone else wants to put IDs in the benchmarks and samples, feel free.

PR #617, opened 2018-06-16
[Tooling] Enable U Test by default, add tooltip about repetition count.

LebedevRI:
[Tooling] Enable U Test by default, add tooltip about repetition count.

As previously discussed, let's flip the switch ^^.

This exposes the problem that it will now be run
for everyone, even if one did not read the help
about the recommended repetition count.

This is not good. So i think we can do the smart thing:
```
$ ./compare.py benchmarks gbench/Inputs/test3_run{0,1}.json
Comparing gbench/Inputs/test3_run0.json to gbench/Inputs/test3_run1.json
Benchmark                   Time             CPU      Time Old      Time New       CPU Old       CPU New
--------------------------------------------------------------------------------------------------------
BM_One                   -0.1000         +0.1000            10             9           100           110
BM_Two                   +0.1111         -0.0111             9            10            90            89
BM_Two                   +0.2500         +0.1125             8            10            80            89
BM_Two_pvalue             0.2207          0.6831      U Test, Repetitions: 2. WARNING: Results unreliable! 9+ repetitions recommended.
BM_Two_stat              +0.0000         +0.0000             8             8            80            80
```
(old screenshot)
image

Or, in the good case (noise omitted):
```
s$ ./compare.py benchmarks /tmp/run{0,1}.json
Comparing /tmp/run0.json to /tmp/run1.json
Benchmark                                            Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------
<99 more rows like this>
./_T012014.RW2/threads:8/real_time                +0.0160         +0.0596            46            47            10            10
./_T012014.RW2/threads:8/real_time_pvalue          0.0000          0.0000      U Test, Repetitions: 100
./_T012014.RW2/threads:8/real_time_mean           +0.0094         +0.0609            46            47            10            10
./_T012014.RW2/threads:8/real_time_median         +0.0104         +0.0613            46            46            10            10
./_T012014.RW2/threads:8/real_time_stddev         -0.1160         -0.1807             1             1             0             0
```
(old screenshot)
image

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -34,6 +34,9 @@ BC_ENDC = BenchmarkColor('ENDC', '\033[0m')
 BC_BOLD = BenchmarkColor('BOLD', '\033[1m')
 BC_UNDERLINE = BenchmarkColor('UNDERLINE', '\033[4m')
 
+UTEST_MIN_REPETITIONS = 2
+UTEST_OPTIMAL_REPETITIONS = 9  # Lowest reasonable number, More is better.
+
 
 def color_format(use_color, fmt_str, *args, **kwargs):
     """
@@ -109,11 +112,11 @@ def generate_difference_report(
                 return b
         return None
 
-    utest_col_name = "U-test (p-value)"
+    utest_col_name = "_pvalue"
     first_col_width = max(
         first_col_width,
-        len('Benchmark'),
-        len(utest_col_name))
+        len('Benchmark'))
+    first_col_width += len(utest_col_name)
     first_line = "{:<{}s}Time             CPU      Time Old      Time New       CPU Old       CPU New".format(
         'Benchmark', 12 + first_col_width)
     output_strs = [first_line, '-' * len(first_line)]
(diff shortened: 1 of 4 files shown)


tools/gbench/report.py
```diff
@@ -126,16 +128,15 @@ def find_test(name):
...
         if last_name != bn['name']:
-            MIN_REPETITIONS = 2
-            if ((len(timings_time[0]) >= MIN_REPETITIONS) and
-                (len(timings_time[1]) >= MIN_REPETITIONS) and
-                (len(timings_cpu[0]) >= MIN_REPETITIONS) and
-                    (len(timings_cpu[1]) >= MIN_REPETITIONS)):
+            if ((len(timings_time[0]) >= UTEST_MIN_REPETITIONS) and
+                (len(timings_time[1]) >= UTEST_MIN_REPETITIONS) and
+                (len(timings_cpu[0]) >= UTEST_MIN_REPETITIONS) and
+                    (len(timings_cpu[1]) >= UTEST_MIN_REPETITIONS)):
```
dominichamon: weird indent here.. is this correct?

tools/gbench/report.py
```diff
@@ -302,7 +310,7 @@ def test_utest(self):
             ['BM_One', '-0.1000', '+0.1000', '10', '9', '100', '110'],
             ['BM_Two', '+0.1111', '-0.0111', '9', '10', '90', '89'],
             ['BM_Two', '+0.2500', '+0.1125', '8', '10', '80', '89'],
-            ['U-test', '(p-value)', '0.2207', '0.6831'],
+            ['U-test', '(p-value)', '0.2207', '0.6831', 'Repetitions:', '2.', 'WARNING:', 'Results', 'unreliable!', '9+', 'repetitions', 'recommended.'],
```
dominichamon: a comment on output: should it have the bm name in the 'u-test' bit so people running the output through grep/ag/whatever will catch the u-test for the bm they want?

tools/gbench/report.py
LebedevRI: This is correct, but the indent is indeed wrong.

tools/gbench/report.py
LebedevRI: Ah, good idea..
How do you want it spelled, `_utest_pvalue` ?

tools/gbench/report.py
LebedevRI: Funny, `autopep8` actually insists on this weird indent.

tools/gbench/report.py
dominichamon: or just `BM_Two_p_value` ?

tools/gbench/report.py
dominichamon: as long as it does the right thing :)

tools/gbench/report.py
LebedevRI: But now that i have actually looked, having such a long suffix as `_utest_pvalue`, causes
the need to realign the columns (since the first column may now be wider).
So `_utest_pvalue` is /better/, but more troublesome.

LebedevRI: So i have thought, and ok, let's just use `_pvalue`.
Avoids the headache with column width readjusting _for now_.

LebedevRI: Pushed the `_pvalue` variant.

LebedevRI:
@dominichamon thank you!

PR #621, opened 2018-06-19
Add Iteration-related Counter::Flags. Fixes #618

### LINKED ISSUE #618: Possibly need more Counter::Flags
I'm thinking of two more flags (well, because i just fixed [two](https://github.com/darktable-org/rawspeed/commit/a1ebe07bea5738f8607b48a7596c172be249590e) [bugs](https://github.com/darktable-org/rawspeed/commit/0891555be56b24f9f4af716604cedfa0da1efc6b) in my code due to the lack of those):
* `* state.iterations()`
  The value is constant for every iteration, and needs to be **multiplied** by the iteration count.
* `/ state.iterations()`
  The is global over all the iterations, and needs to be **divided** by the iteration count.

Additionally, the first one (multiplication) should play nice with `kIsRate`.

LebedevRI:
Inspired by these [two](https://github.com/darktable-org/rawspeed/commit/a1ebe07bea5738f8607b48a7596c172be249590e) [bugs](https://github.com/darktable-org/rawspeed/commit/0891555be56b24f9f4af716604cedfa0da1efc6b) in my code due to the lack of those i have found fixed in my code:
* `kIsIterationInvariant` - `* state.iterations()`
  The value is constant for every iteration, and needs to be **multiplied** by the iteration count.
* `kAvgIterations` - `/ state.iterations()`
  The is global over all the iterations, and needs to be **divided** by the iteration count.

They play nice with `kIsRate`:
* `kIsIterationInvariantRate`
* `kAvgIterationsRate`.

I'm not sure how  meaningful they are when combined with `kAvgThreads`.
I guess the `kIsThreadInvariant` can be added, too, for symmetry with `kAvgThreads`.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -343,12 +343,24 @@ class Counter {
     kDefaults = 0,
     // Mark the counter as a rate. It will be presented divided
     // by the duration of the benchmark.
-    kIsRate = 1,
+    kIsRate = 1U << 0U,
     // Mark the counter as a thread-average quantity. It will be
     // presented divided by the number of threads.
-    kAvgThreads = 2,
+    kAvgThreads = 1U << 1U,
     // Mark the counter as a thread-average rate. See above.
-    kAvgThreadsRate = kIsRate | kAvgThreads
+    kAvgThreadsRate = kIsRate | kAvgThreads,
+    // Mark the counter as a constant value, valid/same for *every* iteration.
+    // When reporting, it will be *multiplied* by the iteration count.
+    kIsIterationInvariant = 1U << 2U,
+    // Mark the counter as a constant rate.
+    // When reporting, it will be *multiplied* by the iteration count
+    // and then divided by the duration of the benchmark.
+    kIsIterationInvariantRate = kIsRate | kIsIterationInvariant,
+    // Mark the counter as a iteration-average quantity.
+    // It will be presented divided by the number of iterations.
+    kAvgIterations = 1U << 3U,
+    // Mark the counter as a iteration-average rate. See above.
+    kAvgIterationsRate = kIsRate | kAvgIterations
   };
 
   double value;
@@ -360,6 +372,14 @@ class Counter {
   BENCHMARK_ALWAYS_INLINE operator double const&() const { return value; }
   BENCHMARK_ALWAYS_INLINE operator double&() { return value; }
 };
+
+// A helper for user code to create unforeseen combinations of Flags, without
+// having to do this cast manually each time, or providing this operator.
+Counter::Flags inline operator|(const Counter::Flags& LHS,
+                                const Counter::Flags& RHS) {
+  return static_cast<Counter::Flags>(static_cast<int>(LHS) |
+                                     static_cast<int>(RHS));
+}
 
 // This is the container for the user-defined counters.
 typedef std::map<std::string, Counter> UserCounters;
--- a/src/counter.cc
+++ b/src/counter.cc
@@ -17,7 +17,8 @@
 namespace benchmark {
 namespace internal {
 
-double Finish(Counter const& c, double cpu_time, double num_threads) {
+double Finish(Counter const& c, int64_t iterations, double cpu_time,
+              double num_threads) {
   double v = c.value;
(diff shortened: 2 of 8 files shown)


include/benchmark/benchmark.h
```diff
@@ -343,12 +343,24 @@ class Counter {
...
     // Mark the counter as a thread-average quantity. It will be
     // presented divided by the number of threads.
-    kAvgThreads = 2,
+    kAvgThreads = 1U << 1U,
     // Mark the counter as a thread-average rate. See above.
-    kAvgThreadsRate = kIsRate | kAvgThreads
+    kAvgThreadsRate = kIsRate | kAvgThreads,
+    // Mark the counter as a constant value, valid/same for *every* iteration.
+    // When reporting, it will be *multiplied* by the iteration count.
+    kIsIterationInvariant = 1U << 2U,
```
dominichamon: can you maybe add some realistic examples of these? i'm having a hard time thinking what an iteration invariant counter would be measuring that's useful.

LebedevRI: https://github.com/darktable-org/rawspeed/commit/a1ebe07bea5738f8607b48a7596c172be249590e - here, `kIsIterationInvariantRate` would be nice
https://github.com/darktable-org/rawspeed/commit/0891555be56b24f9f4af716604cedfa0da1efc6b - `kAvgIterations`

Basically, `kAvgIterations` means that the value is collected over all the iterations, e.g. a timer, thus for it to be meaningful, it needs to be divided by iteration count
And `kIsIterationInvariant` means that every iteration produces this very counter value.
You could use it as is, but if you use anything else other than `kDefaults`, the produced value may be incorrect.

dominichamon:
LGTM but I'll let someone else comment too.

Maybe @EricWF or @pleroy have opinions.

LebedevRI:
To be fair, these aren't _needed_.
As it is seen from the commits i linked, it **can** be solved in user code.
But having these will make users *think*, whether maybe they need them?

include/benchmark/benchmark.h
```diff
@@ -361,6 +373,14 @@ class Counter {
   BENCHMARK_ALWAYS_INLINE operator double&() { return value; }
 };
 
+// A helper for user code to create unforeseen combinations of Flags, without
+// having to do this cast manually each time, or providing this operator.
+Counter::Flags inline operator|(const Counter::Flags& LHS,
```
LebedevRI: @biojppm FTFY

pleroy:
@dominichamon @LebedevRI: `kIsIterationInvariant` seems eminently reasonable to me.  There are many benchmarks where each iteration does a constant number of things.  You can also count them yourself, but having that as a flag seems useful.

src/counter.cc
```diff
@@ -17,20 +17,27 @@
...
 namespace internal {
 
-double Finish(Counter const& c, double cpu_time, double num_threads) {
+double Finish(Counter const& c, int64_t iterations, double cpu_time,
+              double num_threads) {
   double v = c.value;
   if (c.flags & Counter::kIsRate) {
     v /= cpu_time;
   }
   if (c.flags & Counter::kAvgThreads) {
```
LebedevRI: Should i follow up with `kIsThreadInvariant`, for symmetry?

LebedevRI: After thinking about it a bit more, it would make no sense, since that is already what happens.
https://github.com/google/benchmark/blob/master/README.md#user-defined-counters
```
In multithreaded benchmarks, each counter is set on the calling thread only. When the benchmark finishes, the counters from each thread will be summed; the resulting sum is the value which will be shown for the benchmark.
```

BUT i will have to follow-up with `*= cpu_time;`
Now, if the benchmark processes `N` Things (not `items`, more fine-grained),
`kIsRate` would tell you how many Things it processes per second.
But what about how many seconds per `Thing`?

In the same thought, i wonder how `Set{Bytes,Items}Processed()` are useful
given the existance of custom counters. I wonder whether they should be
deprecated v2. One very obvious downside is that they have very generic names,
and it is impossible, unlike custom counters, to customize them.

LebedevRI:

Issue #619, opened 2018-06-19
Documentation: Adding instructions to build the "basic usage" example 


kanak:
Hi benchmark authors,

I saw the library used to great effect in a course on C++ Optimization [1] and was eager to try the examples at home. I'm using Fedora 28 Linux with G++ 8.1.1, cmake 3.11.2 and GNU make 4.2.1 .

The instructions for building and installing the library were clear and easy to follow. I was able to get a build the library itself quite easily. Thank you for this!

I then wanted to build and run the "Basic Usage" example. I immediately ran into a roadblock because the linker kept complaining about things like ``undefined reference to benchmark::State::StartKeepRunning()``. I got some help from the friendly folks at the #googlebenchmark chatroom, but we weren't able to resolve the issue.

I ended up spending a few hours googling 'google benchmark tutorial' to find someone who had successfully compiled a benchmark program. After a few false starts, I found an article from 2014 [2] that helped me resolve the issue. 

The issue was that I was running
```
g++ -std=c++11  -lbenchmark -lpthread  benchmark.cpp   -o benchmark
```
instead of 

```
g++ benchmark.cpp -std=c++11 -lbenchmark -lpthread -O2 -o benchmark
```

(Note the position of `benchmark.cpp` relative to `-lbenchmark`). The reason this happened is that I had set `LDFLAGS=-lbenchmark` instead of using `LDLIBS=-lbenchmark` in my Makefile.

I would like to update the documentation so that it has additional information on how to build the basic usage example. The current instruction to also add `-lpthread` on g++ is helpful, but I think we could flesh it out further.

I was curious if you thought this was a good idea and if so, where I should add this info and what form it should take. I think that spelling out the g++ command or even better, providing an example Makefile would go a long way.

Thanks!

[1] https://www.safaribooksonline.com/library/view/high-performance-computing-and/9781491967560/
[2] http://www.cookandcommit.eu/2014/09/tiny-c-benchmark-framework_29.html

dominichamon:
That would be great. README.md should likely be minimised with links to other docs :)

PR #622, opened 2018-06-19
Add documentation on Running Your First Benchmark

kanak:
In https://github.com/google/benchmark/issues/619 , the need for
documentation aimed at rank beginners was mentioned. Specifically, the
need for exact command line incantations and expected outputs.

This commit adds a new file inside docs that contains this
beginner-friendly documentation. The documentation was verified
against Fedora 28, but can be extended to add ubuntu and other os
information.

### DIFF
--- a/docs/RunningFirstBenchmark.md
+++ b/docs/RunningFirstBenchmark.md
@@ -1,0 +1,114 @@
+# Running Your First Benchmark
+
+This tutorial is aimed at people new to the library and shows you how to run benchmarks with the Google Benchmark library.
+
+# Install the prerequisites
+
+We need the following programs:
+
+* git : to clone google benchmark and gtest libraries
+* cmake : to build google benchmark
+* make : to build google benchmark. Advanced users can use an alternative cmake backend like ninja.
+* g++ : to compile your c++ code. Advanced users can use other compilers like clang instead.
+
+On Fedora 28, the following command installs the requirements:
+
+```
+sudo dnf install --assumeyes wget git-core cmake make gcc-c++
+```
+
+At the time of writing, the versions were as follows:
+* g++ : 8.1.1 20180502
+* cmake : 3.11.2
+* gnu make: 4.2.1
+* git: 2.17.1
+
+# Build and install Google Benchmark
+
+Follow the instructions elsewhere to build and install the google-benchmark library.
+
+On Fedora 28, these commands were sufficient:
+
+```
+git clone https://github.com/google/benchmark.git
+cd benchmark
+git clone https://github.com/google/googletest.git
+mkdir build
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=RELEASE
+make -j
+sudo make install
+```
+
+`make install` should result in lines that look like:
+
+```
+-- Install configuration: "RELEASE"
+-- Installing: /usr/local/lib/libbenchmark.a
+-- Installing: /usr/local/lib/libbenchmark_main.a
+-- Installing: /usr/local/include/benchmark
+-- Installing: /usr/local/include/benchmark/benchmark.h
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkConfig.cmake
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkConfigVersion.cmake
+-- Installing: /usr/local/lib/pkgconfig/benchmark.pc
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkTargets.cmake
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkTargets-release.cmake
+```
+
+Note the locations of the header file (`benchmark.h`) and the static library (`libbenchmark.a`); you might have to tell your compiler to look for libraries in those paths. Things worked out of the box in Fedora 28.
+
+
+# Build your first benchmark
+
+Save this file as `BenchmarkMain.cpp`:
+
+```
+#include <benchmark/benchmark.h>
+#include <string>
+
+static void BM_StringCopy(benchmark::State& state) {
+    std::string x = "hello";
+    for (auto _ : state)
+        std::string copy(x);
+}
+
+BENCHMARK(BM_StringCopy);
+BENCHMARK_MAIN();
+```
+
+Use this command to build the library:
+
+```
+g++ BenchMain.cpp -std=c++11 -lbenchmark -lpthread -O2 -o BenchMain
+```
+
+Notes:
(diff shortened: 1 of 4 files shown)


kanak:
I'm new to the library and also not a native speaker, so I'm open to all feedback.

I'm also happy to add Ubuntu Linux instructions or additional instructions using Clang if necessary. I also did not remove or otherwise refactor any docs in the Readme, I can help with that work as well.

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Use this command to build the library:
+
+```
+g++ BenchMain.cpp -std=c++11 -lbenchmark -lpthread -O2 -o BenchMain
+```
+
+Notes:
+* We use `-std=c++11` because benchmark library requires at least c++11. Replacing it with `c++14` or `c++17` also work.
```
LebedevRI: This isn't true, i think right now, c++11 is still not required for consumers

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=RELEASE
+make -j
+sudo make install
+```
+
+`make install` should result in lines that look like:
+
+```
+-- Install configuration: "RELEASE"
```
LebedevRI: This is quite messy.
I'm thinking that maybe you either want to specify a custom install prefix, or not install at all.
In either case, you'll need to specify the include/link paths.

Or maybe just use the packaged version?

docs/RunningFirstBenchmark.md
kanak: Fixed.

docs/RunningFirstBenchmark.md
kanak: Sadly, google benchmark is not available in the Fedora 28 repository right now. 

I'll prepare instructions for the no install use case. The "install" related bits can then be an "appendix" in this document or just removed.

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
+# Running Your First Benchmark
+
+This tutorial is aimed at people new to the library and shows you how to run benchmarks with the Google Benchmark library.
+
+# Install the prerequisites
```
dominichamon: This should only be necessary if a package is not available. Ie, it might be worth having a section on getting the library first, with installation from source being one of the options.

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+
+We need the following programs:
+
+* git : to clone google benchmark and gtest libraries
+* cmake : to build google benchmark
+* make : to build google benchmark. Advanced users can use an alternative cmake backend like ninja.
+* g++ : to compile your c++ code. Advanced users can use other compilers like clang instead.
+
```
dominichamon: cmake or bazel (bazel is preferred)
make or ninja or visual studio for windows
g++ or clang++ can be replaced by "A C++ compiler with C++11 support"

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+```
+sudo dnf install --assumeyes wget git-core cmake make gcc-c++
+```
+
+At the time of writing, the versions were as follows:
+* g++ : 8.1.1 20180502
+* cmake : 3.11.2
+* gnu make: 4.2.1
+* git: 2.17.1
```
dominichamon: I think it's better to avoid this specificity and just have the users install the packages however they know how. there are too many platforms and variants to go into detail.

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Build and install Google Benchmark
+
+Follow the instructions elsewhere to build and install the google-benchmark library.
```
dominichamon: Add a doc on building and (optionally) installing and link to it, then remove the same instructions from the base README file.

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+-- Installing: /usr/local/include/benchmark
+-- Installing: /usr/local/include/benchmark/benchmark.h
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkConfig.cmake
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkConfigVersion.cmake
+-- Installing: /usr/local/lib/pkgconfig/benchmark.pc
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkTargets.cmake
+-- Installing: /usr/local/lib/cmake/benchmark/benchmarkTargets-release.cmake
+```
+
+Note the locations of the header file (`benchmark.h`) and the static library (`libbenchmark.a`); you might have to tell your compiler to look for libraries in those paths. Things worked out of the box in Fedora 28.
```
dominichamon: this should not be necessary unless the user overrides the install location, in which case they know what they're doing.  :)

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+# Build your first benchmark
+
+Save this file as `BenchmarkMain.cpp`:
```
dominichamon: maybe leave out the specifics of naming if you can. there are many standards around and this is irrelevant for library usage.

docs/RunningFirstBenchmark.md
```diff
@@ -0,0 +1,114 @@
...
+
+BENCHMARK(BM_StringCopy);
+BENCHMARK_MAIN();
+```
+
+Use this command to build the library:
+
+```
+g++ BenchMain.cpp -std=c++11 -lbenchmark -lpthread -O2 -o BenchMain
+```
```
dominichamon: again, we work against many platforms. I think it's better to have more generic instructions about building your executable and linking against the library, then have a section below on requirements like pthread and benchmark. Users on different platforms will have different ways of linking against those libraries.

dominichamon:
is this still something you wanted to pursue?

kanak:
@dominichamon : I would like to abandon this effort. I understand the concerns laid out here, that this documentation would be too specific and overly prescriptive and unsuitable in its form as a piece of official documentation. However, I wrote this when I was running into some really basic problems (in retrospect), such as swapping passing the `-lbenchmark` flag in the wrong place, and want a very rigid set of steps that was guaranteed to work. I think I will probably just publish this as a blog post instead. 

Thank you for your time and for the wonderful library.

dominichamon:
thanks, @kanak

PR #624, opened 2018-06-20
Use EXPECT_DOUBLE_EQ when comparing doubles in tests.

### LINKED ISSUE #623: Build error on string_util_gtest.cc
With the latest commit, 7d03f2d , I'm seeing the following build error:

```
In file included from /home/developer/git/google/benchmark/test/string_util_gtest.cc:6:0:
/home/developer/git/google/benchmark/build/googletest/include/gtest/gtest.h: In instantiation of ‘testing::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = double; T2 = double]’:
/home/developer/git/google/benchmark/build/googletest/include/gtest/gtest.h:1460:64:   required from ‘static testing::AssertionResult testing::internal::EqHelper<lhs_is_null_literal>::Compare(const char*, const char*, const T1&, const T2&) [with T1 = double; T2 = double; bool lhs_is_null_literal = false]’
/home/developer/git/google/benchmark/test/string_util_gtest.cc:117:5:   required from here
/home/developer/git/google/benchmark/build/googletest/include/gtest/gtest.h:1432:11: error: comparing floating point with == or != is unsafe [-Werror=float-equal]
   if (lhs == rhs) {
           ^
```

dominichamon:
Fixes #623

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,7 +138,6 @@ else()
   add_cxx_compiler_flag(-pedantic)
   add_cxx_compiler_flag(-pedantic-errors)
   add_cxx_compiler_flag(-Wshorten-64-to-32)
-  add_cxx_compiler_flag(-Wfloat-equal)
   add_cxx_compiler_flag(-fstrict-aliasing)
   # Disable warnings regarding deprecated parts of the library while building
   # and testing those parts of the library.
--- a/test/string_util_gtest.cc
+++ b/test/string_util_gtest.cc
@@ -114,28 +114,28 @@ TEST(StringUtilTest, stoi) {
 TEST(StringUtilTest, stod) {
   {
     size_t pos = 0;
-    EXPECT_EQ(0.0, benchmark::stod("0", &pos));
+    EXPECT_DOUBLE_EQ(0.0, benchmark::stod("0", &pos));
     EXPECT_EQ(1, pos);
   }
   {
     size_t pos = 0;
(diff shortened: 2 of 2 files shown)


pleroy:
This PR makes me cry tears of blood.

When parsing any of the values in this test, which are all exactly representable in IEEE binary64, I surely hope that I get the *exact* value, not something fudged to within 4 ulps.

Now it appears that some genius at the FSF put in GCC a warning that says "don't compare floats for equality".  Could we please disable this warning?  There are circumstances where comparing for equality is fine, and this is one of them.  And if you don't know what you are doing, compiler warnings are not helping anyway.

For your enjoyment, here are some [words of wisdom](https://twitter.com/stephentyrone/status/856882266408202244) from @stephentyrone, who understands numerics much better than I ever will.

LebedevRI:
Is it intentional that an external project (googletest) is being built with warnings?

dominichamon:
Ahah, I wondered what was going on and why it hasn't failed in the past (or in my local builds). This is something I really hope clang doesn't embrace.

@LebedevRI I think this is not googletest being compiled with warnings, it's our test code being compiled with warnings (which is a good thing) and a macro being expanded from googletest.

Let me see if i can disable the warning (though i can't repro the issue locally to check).

dominichamon:
Oh even better, we're enabling -Wfloat-equal.

pleroy:
I blamed this line and I am looking sternly at @EricWF...

Issue #620, opened 2018-06-19
Timers need some work.


LebedevRI:
I think something much like custom stats, custom counters is needed for timers.
(Personally, i miss the `std::clock`-based timer.)
The `UseManualTime()`/`UseRealTime()` should also be just one `UseTimer(name)`.

Also in v2, once json is here, it would be great to be able to read those values,
and generate custom stats based on them.

I'm also wondering if there are any sane ideas for the interface for these timers.
Currently, there is only a `SetIterationTime()`, which needs to be called every iteration.

BaaMeow:
I'd like to participate in this a little bit, albeit my needs might be a bit different and maybe not exactly aligned.

Many of my company's measurements are tiny enough that I want to count things like cycles (`rtdsc`, for example) rather than nanoseconds, and so I wanted the ability to have a custom TimeUnit alongside the ManualTime. Would the work here allow for something like that, or is this orthogonal?

LebedevRI:
> I'd like to participate in this a little bit, albeit my needs might be a bit different and maybe not exactly aligned.
> 
> Many of my company's measurements are tiny enough that I want to count things like cycles (`rtdsc`, for example) rather than nanoseconds,

That is basically the same thing.
Proper generalized "custom 'times'" should be oblivious to what they count, they are just infrastructure.

> and so I wanted the ability to have a custom TimeUnit alongside the ManualTime. Would the work here allow for something like that, or is this orthogonal?

Hmm yeah. Even the existing user counters lack the "unit" customizability.

The current `ManualTime` does imply that it counts time, so hardcoding of time units makes sense,
but indeed, that will be non-ideal for custom timers. I'm not sure how that would/will look though.

I had some code for custom timers that almost worked,
but v2 branch with all it's json infra changes for some reason derailed that work.
Since v2 is still there where it was back then, i guess maybe i should take another stab at it.

I'm not sure what will it take to extend everything to support custom "time" units,
(what needs changing? just the "s" suffix?), so i think that is orthogonal to this issue..

BaaMeow:
Maybe having ManualTime count (whatever) and then allowing to have a `TimeUnit::Manual`, which will call a user-provided function to serialize the stored value might be a good first start? You could trigger it by using a 2-argument `UseManualTime()` function, like

```
BENCHMARK(BM_ManualTiming)->Range(1, 1<<17)->UseManualTime(time_unit_serialization_function);
```

[...]

PR #625, opened 2018-06-26
Memory management and reporting hooks

dominichamon:
This PR allows one to register a memory management object that can track allocation counts and heap usage, and then inject them into the benchmark report.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -243,6 +243,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 
 namespace benchmark {
 class BenchmarkReporter;
+class MemoryManager;
 
 void Initialize(int* argc, char** argv);
 
@@ -267,12 +268,9 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter);
 size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter,
                               BenchmarkReporter* file_reporter);
 
-// If this routine is called, peak memory allocation past this point in the
-// benchmark is reported at the end of the benchmark report line. (It is
-// computed by running the benchmark once with a single iteration and a memory
-// tracer.)
-// TODO(dominic)
-// void MemoryUsage();
+// Register a MemoryManager instance that will be used to report on allocation
+// data for benchmark runs.
+void RegisterMemoryManager(MemoryManager* memory_manager);
 
 namespace internal {
 class Benchmark;
@@ -1260,7 +1258,9 @@ class BenchmarkReporter {
           complexity_n(0),
           report_big_o(false),
           report_rms(false),
-          counters() {}
+          counters(),
+          allocs_per_iter(0.0),
+          max_bytes_used(0) {}
 
     std::string benchmark_name;
     std::string report_label;  // Empty if not set by benchmark.
@@ -1304,6 +1304,10 @@ class BenchmarkReporter {
     bool report_rms;
 
     UserCounters counters;
+
+    // Memory metrics.
+    double allocs_per_iter;
+    int64_t max_bytes_used;
   };
 
   // Construct a BenchmarkReporter with the output stream set to 'std::cout'
@@ -1416,6 +1420,24 @@ class BENCHMARK_DEPRECATED_MSG("The CSV Reporter will be removed in a future rel
 
   bool printed_header_;
   std::set<std::string> user_counter_names_;
+};
+
+// If a MemoryManager is registered, it can be used to report memory statistics
+// for a run of the benchmark.
+class MemoryManager {
+ public:
+  struct Result {
+    Result() : num_allocs(0), max_bytes_used(0), iterations(0) {}
+    int64_t num_allocs;
+    int64_t max_bytes_used;
+    int64_t iterations;
+  };
+
+  // Implement this to start recording allocation information.
+  virtual void Start() = 0;
+
+  // Implement this to stop recording and fill out the given Result structure.
+  virtual void Stop(Result* result) = 0;
 };
 
 inline const char* GetTimeUnitString(TimeUnit unit) {
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -268,8 +268,8 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter);
 size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter,
                               BenchmarkReporter* file_reporter);
 
-// Register a MemoryManager instance that will be used to report on allocation
-// data for benchmark runs.
+// Register a MemoryManager instance that will be used to collect and report
+// allocation measurements for benchmark runs.
 void RegisterMemoryManager(MemoryManager* memory_manager);
 
 namespace internal {
@@ -1424,23 +1424,21 @@ class BENCHMARK_DEPRECATED_MSG("The CSV Reporter will be removed in a future rel
   std::set<std::string> user_counter_names_;
 };
 
-// If a MemoryManager is registered, it can be used to report memory statistics
-// for a run of the benchmark.
+// If a MemoryManager is registered, it can be used to collect and report
+// allocation metrics for a run of the benchmark.
 class MemoryManager {
  public:
   struct Result {
-    explicit Result(int64_t iters)
-        : num_allocs(0), max_bytes_used(0), iterations(iters) {}
+    Result() : num_allocs(0), max_bytes_used(0) {}
 
     // The number of allocations made in total between Start and Stop.
     int64_t num_allocs;
 
     // The peak memory use between Start and Stop.
     int64_t max_bytes_used;
-
-    // The number of iterations run for the memory test.
-    const int64_t iterations;
   };
+
+  virtual ~MemoryManager() {}
 
   // Implement this to start recording allocation information.
   virtual void Start() = 0;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1259,6 +1259,7 @@ class BenchmarkReporter {
           report_big_o(false),
           report_rms(false),
           counters(),
+          has_memory_result(false),
           allocs_per_iter(0.0),
           max_bytes_used(0) {}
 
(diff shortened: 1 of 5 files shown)


dominichamon:
Thoughts on API welcome. This will be used internally with our closed-source memory management and measurement handlers.

test/memory_manager_test.cc
```diff
@@ -0,0 +1,36 @@
...
+BENCHMARK(BM_empty);
+
+ADD_CASES(TC_ConsoleOut, {{"^BM_empty %console_report$"}});
+ADD_CASES(TC_JSONOut, {{"\"name\": \"BM_empty\",$"},
+                       {"\"iterations\": %int,$", MR_Next},
+                       {"\"real_time\": %float,$", MR_Next},
+                       {"\"cpu_time\": %float,$", MR_Next},
+                       {"\"time_unit\": \"ns\"$", MR_Next},
+                       {"\"allocs_per_iter\": %float,$", MR_Next},
+                       {"\"max_bytes_used\": %float,$", MR_Next},
```
LebedevRI: Why `float`? Should be `int`.

include/benchmark/benchmark.h
```diff
@@ -1418,6 +1424,25 @@ class BENCHMARK_DEPRECATED_MSG("The CSV Reporter will be removed in a future rel
...
 
+// If a MemoryManager is registered, it can be used to report memory statistics
+// for a run of the benchmark.
+class MemoryManager {
+ public:
+  struct Result {
+    explicit Result(int64_t iters)
+        : num_allocs(0), max_bytes_used(0), iterations(iters) {}
+    int64_t num_allocs;
+    int64_t max_bytes_used;
```
LebedevRI: It would be great to have a comment explaining _what_ this is.
Is this the total amount of memory allocated, ever, or the peak memory usage?

For reference, it is possible to get the former natively - e.g. [`llvm::sys::Process::GetMallocUsage()`](https://github.com/llvm-mirror/llvm/blob/09f0f7ccc5eec100c3902c070f674a5c97a17c92/lib/Support/Unix/Process.inc#L87-L114)

test/memory_manager_test.cc
dominichamon: oops, thanks. and json of course lets this past :)

include/benchmark/benchmark.h
dominichamon: I imagine we might have a native implementation of these things, but we might have someone with a custom allocator, or an integration with tcmalloc. I wanted to get the interface in so we can do something, and then worry about a native implementation (and all the cross-platform shenanigans that entails).

test/memory_manager_test.cc
```diff
@@ -0,0 +1,36 @@
...
+BENCHMARK(BM_empty);
+
+ADD_CASES(TC_ConsoleOut, {{"^BM_empty %console_report$"}});
+ADD_CASES(TC_JSONOut, {{"\"name\": \"BM_empty\",$"},
+                       {"\"iterations\": %int,$", MR_Next},
+                       {"\"real_time\": %float,$", MR_Next},
+                       {"\"cpu_time\": %float,$", MR_Next},
+                       {"\"time_unit\": \"ns\"$", MR_Next},
+                       {"\"allocs_per_iter\": %float,$", MR_Next},
+                       {"\"max_bytes_used\": %int,$", MR_Next},
```
LebedevRI: Nit: this does not actually check the values, only their presence

include/benchmark/benchmark.h
```diff
@@ -267,12 +268,9 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter);
 size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter,
                               BenchmarkReporter* file_reporter);
 
-// If this routine is called, peak memory allocation past this point in the
-// benchmark is reported at the end of the benchmark report line. (It is
-// computed by running the benchmark once with a single iteration and a memory
-// tracer.)
-// TODO(dominic)
-// void MemoryUsage();
+// Register a MemoryManager instance that will be used to report on allocation
```
ghemawat: IIUC, the memory manager is also responsible for recording allocation data.  So perhaps reword to:

.. will be used to collect and report allocation measurements for benchmark runs.

include/benchmark/benchmark.h
```diff
@@ -267,12 +268,9 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* console_reporter);
...
+// data for benchmark runs.
+void RegisterMemoryManager(MemoryManager* memory_manager);
```
ghemawat: How about:
   SetMemoryManager
   GetMemoryManager

Getters are often useful to have; e.g., if somebody wants to wrap the existing memory manager.

include/benchmark/benchmark.h
```diff
@@ -1418,6 +1424,31 @@ class BENCHMARK_DEPRECATED_MSG("The CSV Reporter will be removed in a future rel
   std::set<std::string> user_counter_names_;
 };
 
+// If a MemoryManager is registered, it can be used to report memory statistics
```
ghemawat: See note on comment above.

include/benchmark/benchmark.h
```diff
@@ -1418,6 +1424,31 @@ class BENCHMARK_DEPRECATED_MSG("The CSV Reporter will be removed in a future rel
...
+        : num_allocs(0), max_bytes_used(0), iterations(iters) {}
+
+    // The number of allocations made in total between Start and Stop.
+    int64_t num_allocs;
+
+    // The peak memory use between Start and Stop.
+    int64_t max_bytes_used;
+
+    // The number of iterations run for the memory test.
+    const int64_t iterations;
```
ghemawat: From what I see in the .cc file, iterations is only used to convey info from one internal function to another.  So maybe this field should not be in the API (header file) and should be just passed as an extra arg internally.  You will end up with two new args in the internal call:
    int64 memory_iterations, const MemoryManager::Result& result

"memory_iterations > 0" can be used to determine whether or not memory usage was recorded, and therefore as a side-benefit, Result will no longer need to be heap-allocated.

include/benchmark/benchmark.h
```diff
@@ -1418,6 +1424,31 @@ class BENCHMARK_DEPRECATED_MSG("The CSV Reporter will be removed in a future rel
...
+  };
+
+  // Implement this to start recording allocation information.
+  virtual void Start() = 0;
```
LebedevRI: `MemoryManager` has virtual functions but non-virtual destructor :)

include/benchmark/benchmark.h
dominichamon: With the way this works, I'm not sure when they'd do that. I agree it would be useful, but it would have to happen between the registration (which should be before the call to `RunSpecifiedBenchmarks`) and the benchmark being run.

Having a const accessor might be useful if a benchmark instance wants to mutate the memory manager in some way.

If the memory manager object was part of the benchmark instance state, then it could be set on a per benchmark basis, which might be more flexible. WDYT?

Actually, thinking about it, I wonder if users could just use `UserCounters` already for this. If at the start of the instance they start their allocator measuring, and then stop it after the `KeepRunning` loop and add the allocation counters to `UserCounters`. The only issue with that might be the profiling overhead, if there is any.

include/benchmark/benchmark.h
dominichamon: Done

include/benchmark/benchmark.h
dominichamon: Done

include/benchmark/benchmark.h
dominichamon: Done

include/benchmark/benchmark.h
dominichamon: Done

include/benchmark/benchmark.h
ghemawat: 1. I  think it is fine to leave out the accesor.  But maybe rename the setter to SetMemoryManager just in case we add the getter later?

2. The main benefit of registering the memory manager outside the confines of a benchmarks is that then all benchmarks get the benefit of memory measurement automatically.   But your comment does make me think that this API is perhaps too focused one type of measurement: it is only treating one type of measurement somebody might want to add to all benchmarks.  What if somebody wants a different type of measurement; maybe they want to report contention observed in the benchmark.

So an alternative more general API might be something like:

void RegisterMetric(const string& name, Metric* metric);

class Metric {
 public:
   virtual void Start(Counter*) = 0;  // Empties counter
   virtual void Stop(Counter*) = 0;  // Fills counter
};

main (or something called at initialization time) would call RegisterMetric for all relevant metrics that should be measured across all benchmarks.  The benchmark library would call Start/Stop around benchmark execution for every registered metric.

What do you think?

ghemawat: One thing I just realized is that unlike Counters, the memory measurement is collected in a separate run (because it slows down regular benchmark performance too much).  Something to keep in mind if we go with the Metric approach I proposed.  Perhaps Counter flags can be extended to indicate that the particular metric must be collected via a separate run.  Or maybe we could just do separate runs for each metric registered like this?

LebedevRI: Relevant: https://github.com/google/benchmark/issues/620
I wonder if timers should be generalized into metrics,
and you'd just add a second benchmark with `use timer = memory`.
I don't think counters should be taught about that.

src/benchmark.cc
```diff
@@ -249,7 +262,23 @@ std::vector<BenchmarkReporter::Run> RunBenchmark(
       // clang-format on
 
       if (should_report) {
-        BenchmarkReporter::Run report = CreateRunReport(b, results, seconds);
+        MemoryManager::Result memory_result;
```
LebedevRI: @dominichamon why was this added here, and not at the end of the body of per-repetition loop?
```
     }
     // Here
   }
   // Calculate additional statistics	   // Calculate additional statistics
```

test/memory_manager_test.cc
```diff
@@ -0,0 +1,40 @@
+#include <memory>
+
+#include "../src/check.h"
+#include "benchmark/benchmark.h"
+#include "output_test.h"
+
```
LebedevRI: And this is why two build systems is a *horrible* idea.

Issue #568, opened 2018-04-12
Build fails in benchmark.cc


UCLGuichard:
Hello,
I've got to these two errors, while trying to build google benchmarks for the project I'm on:
```
/home/ucapgui/BICO/benchmark/src/benchmark.cc(177): error #2259: non-pointer conversion from "long" to "int" may lose significant bits
      results.complexity_n += st.complexity_length_n();
                           ^

/home/ucapgui/BICO/benchmark/src/benchmark.cc(326): error #1875: offsetof applied to non-POD (Plain Old Data) types is nonstandard
    static_assert(offsetof(State, error_occurred_) <=
                  ^
```
Any idea how to fix it ? Thank you.

dominichamon:
The first is a bug that i'll push a fix for now (#569).

The second is a known issue that was already fixed by disabling the warning in source (#552) .

UCLGuichard:
It works for the first although I still get the second. I'm using the Intel compiler for which this fix is not applicable ?

dominichamon:
Ah maybe... @EricWF who knows more about the compiler side and how we might support that. Are you using CMake or bazel to build? I think we have assumed clang, msvc, or g++ at this point.

UCLGuichard:
Yes, I'm using CMake and Intel compiler (2017).

EricWF:
Urg. I hate this warning.

UB is not allowed in constant expressions, so if the static assert passes, the code must be well formed. It's dumb to emit a warning for it.

I'll look into this more tomorrow.

dominichamon:
How was this resolved?

UCLGuichard:
So far we moved to the GNU compiler for which it works.

Issue #552, opened 2018-03-14
Use of offsetof macro on non standard-layout type is undefined at least prior to C++17


astrelni:
To the best of my understanding [benchmark.cc#L316](https://github.com/google/benchmark/blob/master/src/benchmark.cc#L316), which calls offsetof on benchmark::State, is undefined behavior.

C++11 18.2.4 [support.types]

> The macro offsetof(type, member-designator) accepts a restricted set of type arguments in this International Standard. If type is not a standard-layout class (Clause 9), the results are undefined.

benchmark::State is a non standard-layout type because all of its non-static members do not have the same access control.

I am not sure what would be lost by just removing the static_assert, so I am starting this off as an issue instead of going straight for a PR.

dominichamon:
@EricWF who added the check.

Simply, the static_assert exists to ensure that we keep the performance-necessary struct layout.

astrelni:
For context, I am currently failing to build a benchmark with gcc because it is triggering Werror=invalid-offsetof on that line.

dominichamon:
I noticed that in the bazel builds. We currently disable that warning for our gcc builds: https://github.com/google/benchmark/blob/master/CMakeLists.txt#L123

EricWF:
yeah, my bad. I figured this might come up. However all compilers seem to support the behavior as an extension even though they generate a warning about it.

astrelni:
Would you be in favor of turning it into a runtime CHECK on `error_occured_ - this` to keep standard compliant?

astrelni:
Ping @EricWF @dominichamon

EricWF:
Are we just complaining the the compiler issues a warning? If so we should just disable the warning. As long as the static assert passes there is no undefined behaviour; because there can't be undiagnosed undefined behaviour at compile time.

astrelni:
So if IIUC we just need to add -Winvalid-offsetof to the bazel copts?

EricWF:
Yeah, either that or we could wrap the static assert in a `#pragma diagnostic disable "-Winvalid-offsetof". Either option seems good to me.

dominichamon:
It's disabled in the cmake, so doing that in bazel sgtm. but only if it's gcc (ie, if the warning is supported).

EricWF:
I decided to disable it in the source instead, just to ensure we don't have invalid applications of `offsetof` in other places we don't intend.

astrelni:
Great, thanks a lot.

[...]

PR #629, opened 2018-07-02
Disable Intel warning about use of offsetof on non standard-layout type

nazavode:
This PR disables the *invalid `offsetof`* warning for Intel compiler just like the fix already in place does for GCC and Clang. It should do the job for #568, at least for the second part of the issue (the one related to #552).

Please note that Intel compiler defines `__GNUC__` on Linux, so the check for `__INTEL_COMPILER` must come first.


dominichamon:
Thank you!

PR #631, opened 2018-07-04
Fix build with Intel compiler

### LINKED ISSUE #568: Build fails in benchmark.cc

### LINKED ISSUE #630: Intel compiler support
Since in the benchmark's main page the Intel compiler is listed among the supported toolchains, I'm opening this issue to track references to the PRs I'm about to submit to fix the build for versions I'm currently targeting:

* 2016 (`16.0.3`)
* 2017 (`17.0.4`)
* 2018 (`18.0.3`)
* 2019-beta (`19.0.0.046`, not yet released)

nazavode:
This PR fixes some quirks that prevent the build with the Intel compiler when `-DCMAKE_BUILD_TYPE=Release` is set that in turn adds `-Werror` to build flags. Should fix #630 (and should fix #568 as well) for the following Intel versions:

* 2016 (`16.0.3`)
* 2017 (`17.0.4`)
* 2018 (`18.0.3`)
* 2019-beta (`19.0.0.046`, not yet released)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,7 +141,13 @@ else()
   add_cxx_compiler_flag(-fstrict-aliasing)
   # Disable warnings regarding deprecated parts of the library while building
   # and testing those parts of the library.
-  add_cxx_compiler_flag(-Wno-deprecated-declarations)
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
+    # Intel silently ignores '-Wno-deprecated-declarations',
+    # warning no. 1786 must be explicitly disabled.
+    add_cxx_compiler_flag(-wd1786)
+  else()
+    add_cxx_compiler_flag(-Wno-deprecated-declarations)
+  endif()
   # Disable deprecation warnings for release builds (when -Werror is enabled).
   add_cxx_compiler_flag(-Wno-deprecated RELEASE)
   add_cxx_compiler_flag(-Wno-deprecated RELWITHDEBINFO)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,12 +141,12 @@ else()
   add_cxx_compiler_flag(-fstrict-aliasing)
   # Disable warnings regarding deprecated parts of the library while building
   # and testing those parts of the library.
+  add_cxx_compiler_flag(-Wno-deprecated-declarations)
   if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
     # Intel silently ignores '-Wno-deprecated-declarations',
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -141,7 +141,13 @@ else()
   add_cxx_compiler_flag(-fstrict-aliasing)
   # Disable warnings regarding deprecated parts of the library while building
   # and testing those parts of the library.
-  add_cxx_compiler_flag(-Wno-deprecated-declarations)
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
```
dominichamon: I _think_ you can just try adding the flag. AddCXXCompilerFlag.cmake will check if the flag exists before adding it.

Or, as per the c++11:

```
add_cxx_compiler_flag(-Wno-deprecated-declarations)
if (NOT HAVE_CXX_FLAG_WNO_DEPRECATED_DECLARATIONS)
  add_cxx_compiler_flag(-wd1786)
endif()
```

maybe. it depends if the intel compiler supports the flag but ignores it, or returns an error if you try to set it.

nazavode: @LebedevRI yep, their frontend can get quite weird. A probably safer choice would be to tell the compiler to treat unrecognized flags as errors:

https://godbolt.org/g/SFSuFb

LebedevRI: Point was, it looks like there *is* a way to make it complain.
So sounds like you just need to fix `add_cxx_compiler_flag()`.

nazavode: @LebedevRI I was trying to fix `add_cxx_compiler_flag()` for `icpc` but this looks like an Intel regression to me:

https://godbolt.org/g/tfxWvW

If you're using `icc >= 17.0.4` (the `17.0.0` godbolt is using still works), the `-Wno-deprecated-declarations` option is considered recognized (`-diag-error=10006` is happy) but still ignored.

nazavode: @dominichamon I like obvious and explicit :+1: 

Ok, I will add that change and I'm going to try to open a call to Intel to fix the regression, let's just hope for a better future :)

PR #633, opened 2018-07-09
Include system load average in console and JSON reports

atdt:
High system load can skew benchmark results. By including system load averages in the library's output, we help users identify a potential issue in the quality of their measurements, and thus assist them in producing better (more reproducible) results. I got the idea for this from [Brendan Gregg's checklist for benchmark accuracy](http://www.brendangregg.com/blog/2018-06-30/benchmarking-checklist.html).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,6 +36,7 @@ Maxim Vafin <maxvafin@gmail.com>
 MongoDB Inc.
 Nick Hutchinson <nshutchinson@gmail.com>
 Oleksandr Sochka <sasha.sochka@gmail.com>
+Ori Livneh <ori.livneh@gmail.com>
 Paul Redmond <paul.redmond@gmail.com>
 Radoslav Yovchev <radoslav.tm@gmail.com>
 Roman Lebedev <lebedev.ri@gmail.com>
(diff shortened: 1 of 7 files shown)


LebedevRI:
I'm honestly not sure how useful this is to be automated, some thoughts:

* This has only one measurement, right before the benchmarking begins.
* Nothing is stopping some other cpu-hungry process to start (or awake) during the benchmarking, and cause cpu contention.
* And adding a second load average report afterwards is of not much use, since it is rather expected that benchmarking would cause some load on the system.
* Similarly, if you run two benchmarks back-to-back, the load average would not be very useful..

atdt:
@LebedevRI, valid points. However: while load average won't *always* be pertinent, neither will the count of CPU cores, or the cache hierarchy. You are also right that reporting load averages prior to running the benchmarks does nothing to alert the user to spikes in demand from other processes that occur while the benchmark is in flight. But persistent high demand on the system is important to report for two reasons: first, it can cause a consistent depression of results across multiple benchmark runs, which is a subtler issue to spot than a sporadic spike that manifests in a few outlier results. Second, while a low/zero load average does not guarantee that load won't spike, a *high* load average is almost always a sign that the person performing the benchmark has overlooked something.

dominichamon:
Sorry for the delay on this. I think it's a useful data point to have so I'm happy to merge it.

Issue #635, opened 2018-07-11
[feature] Conan package.


p-groarke:
Any chance we can get a conan package? Thank you.

Croydon:
I'm working on something non-official here, but would be nice if we can get official support https://github.com/Croydon/conan-google_benchmark

Croydon:
This would be also required for https://github.com/grpc/grpc/issues/15363

p-groarke:
Bincrafters may be interested in your package once it's ready. Thx for taking the time.

Croydon:
That was indeed my goal with this recipe :)

Croydon:
Linux, macOS and Windows static builds seem to be fine, Windows shared fails and I don't understand why yet 

Ideas would be appreciated 
https://ci.appveyor.com/project/Croydon/conan-google-benchmark-83vi6/build/job/q9nm3togdwumnct8#L1609

dominichamon:
PRs always welcome.

Dominic Hamon | Google
*There are no bad ideas; only good ideas that go horribly wrong.*

raulbocanegra:
Hi! 
As @Croydon referenced I was asked to include my repo a couple of weeks ago.
I am happy to help and also open to receive any suggestions.
One of the issues I found on Windows is building a shared version of the library. After having a look at the code it seems that the symbols are not being exported, so shared builds are not supported on Windows.

Croydon:
@raulbocanegra This is a general problem, not a Conan recipe specific one. See #640

Croydon:
Benchmark has now both a git Conan recipe version and is also in Conan Center.

Done!

PR #640, opened 2018-07-15
- fix Windows shared build

### LINKED ISSUE #639: Windows shared builds are failing
Tested both with 1.14.1 and master, same result.

cmake .. -A x64 -DBUILD_SHARED_LIBS=On && cmake --build .

> (Link target) ->
 LINK : fatal error LNK1104: cannot open file 'Debug\benchmark.lib'

SSE4:
fixed #639 
related to the #635
/cc @Croydon @p-groarke

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -22,9 +22,16 @@ set_target_properties(benchmark PROPERTIES
   VERSION ${GENERIC_LIB_VERSION}
   SOVERSION ${GENERIC_LIB_SOVERSION}
 )
+
+include(GenerateExportHeader)
+generate_export_header(benchmark)
+
 target_include_directories(benchmark PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
     )
+
+set_target_properties(benchmark PROPERTIES PUBLIC_HEADER ${CMAKE_CURRENT_BINARY_DIR}/benchmark_export.h)
 
 # Link threads.
 target_link_libraries(benchmark  ${BENCHMARK_CXX_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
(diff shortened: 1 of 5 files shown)


SSE4:
I signed it!

src/CMakeLists.txt
```diff
@@ -22,10 +22,16 @@ set_target_properties(benchmark PROPERTIES
   VERSION ${GENERIC_LIB_VERSION}
   SOVERSION ${GENERIC_LIB_SOVERSION}
 )
+
+include(GenerateExportHeader)
+generate_export_header(benchmark EXPORT_FILE_NAME ${CMAKE_CURRENT_SOURCE_DIR}/../include/benchmark_export.h)
```
LebedevRI: It's a very bad idea to write outside of the `CMAKE_CURRENT_BINARY_DIR`

SSE4: thanks, I'll try to change to use CMAKE_CURRENT_BINARY_DIR. initial attempt produced some CMake errors for me. also, build fails hard on tests, I'll take a look.

SSE4:
it seems to be dual build-system project (Bazel + CMake), so CMake's **generate_export_header** doesn't seem to be good fit here (I don't want to re-implement the same in Bazel, especially as I have no experience with mentioned build system). probably, the best thing I can do for now is to just copy generated export definitions into **benchmark/benchmark.h**, does it sound like an acceptable solution?

LebedevRI:
I would suggest looking how googletest handles this issue.
They too have `gtest_main`.

SSE4:
they do not use **generate_export_header** and have their own macro definitions (**GTEST_API_**) :
https://github.com/google/googletest/blob/41f0e24335d8874cc87420a6349bb1844e2a15de/googletest/include/gtest/internal/gtest-port.h#L969

LebedevRI:
I must say that my personal opinion is that this is kinda ugly.
I wonder if it would be better to simply do
```
if(BUILD_SHARED_LIBS AND WIN32)
  message(SEND_ERROR "Building libbenchmark as shared library is not supported")
endif()
```

dominichamon:
How does this work on installation on linux when the public header is copied to /usr and benchmark_export.h doesn't exist? or does it also get created for linux, and get copied to /usr on installation?

This seems awfully intrusive. I think short term not supporting shared libs on windows is reasonable, until we can figure out a better way.

SSE4:
I've asked @Croydon and he is okay with not supporting Windows shared built at all, works for me as well.

dominichamon:
Can this PR be closed then?

SSE4:
yes

PR #644, opened 2018-07-19
(hopefully) helpful changes to readme.md only

mattreecebentley:
Clarifications, reduction of repetitions, and corrections. Movement of some elements to consolidated locations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,9 +561,8 @@ However benchmarks are often noisy and a single result may not be representative
 of the overall behavior. For this reason it's possible to repeatedly rerun the
 benchmark.
 
-The number of runs of each benchmark is specified globally by the
-`--benchmark_repetitions` flag or on a per benchmark basis by calling
-`Repetitions` on the registered benchmark object. When a benchmark is run more
+The number of runs of each benchmark can be specified globally by using the
+`--benchmark_repetitions` command-line option on the executable call (eg. "./benchmark --benchmark_repetitions=20") or on a per benchmark basis by calling `Repetitions` on the registered benchmark object. When a benchmark is run more
 than once the mean, median and standard deviation of the runs will be reported.
 
 Additionally the `--benchmark_report_aggregates_only={true|false}` flag or
--- a/README.md
+++ b/README.md
@@ -31,15 +31,13 @@ $ cmake -G <generator> [options] ../benchmark
 $ make
 ```
 
-Note that Google Benchmark requires GTest to build and run the tests. This
-dependency can be provided three ways:
+Note that Google Benchmark requires Google Test to build and run the tests. This
+dependency can be provided two ways:
 
-* Checkout the GTest sources into `benchmark/googletest`.
+* Checkout the GTest sources into `benchmark/googletest` as above.
 * Otherwise, if `-DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON` is specified during
   configuration, the library will automatically download and build any required
   dependencies.
-* Otherwise, if nothing is done, CMake will use `find_package(GTest REQUIRED)`
-  to resolve the required GTest dependency.
 
 If you do not wish to build and run the tests, add `-DBENCHMARK_ENABLE_GTEST_TESTS=OFF`
 to `CMAKE_ARGS`.
@@ -61,6 +59,7 @@ Now, let's clone the repository and build it
 ```
 git clone https://github.com/google/benchmark.git
 cd benchmark
+git clone https://github.com/google/googletest.git
 mkdir build
 cd build
 cmake .. -DCMAKE_BUILD_TYPE=RELEASE
--- a/README.md
+++ b/README.md
@@ -87,6 +87,11 @@ to use, test, and provide feedback on the new features are encouraged to try
 this branch. However, this branch provides no stability guarantees and reserves
 the right to change and break the API at any time.
 
+##Prerequisite knowledge
+
+Before attempting to understand this framework one should ideally have some familiarity with the structure and format of the Google Test framework, upon which it is based. Documentation for Google Test, including a "Getting Started" (primer) guide, is available here:
+https://github.com/google/googletest/blob/master/googletest/docs/Documentation.md
+
 
 ## Example usage
 ### Basic usage
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The basic steps for configuring and building the library look like this:
 
 ```bash
 $ git clone https://github.com/google/benchmark.git
-# Benchmark requires GTest as a dependency. Add the source tree as a subdirectory.
+# Benchmark requires Google Test as a dependency. Add the source tree as a subdirectory.
 $ git clone https://github.com/google/googletest.git benchmark/googletest
 $ mkdir build && cd build
 $ cmake -G <generator> [options] ../benchmark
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,9 +118,37 @@ BENCHMARK(BM_StringCopy);
 BENCHMARK_MAIN();
 ```
 
-Don't forget to inform your linker to add benchmark library e.g. through `-lbenchmark` compilation flag.
+Don't forget to inform your linker to link to the benchmark library e.g. through `-lbenchmark` compilation flag, and also the pthread library ('-lpthread'). If running google benchmarks on Windows, the shlwapi library must additionally be linked to (`-lshlwapi`) and the library location specified eg. ../../libraries/google_benchmark/lib. If running Solaris, kstat (`-lkstat`) must be linked to.
```
LebedevRI: These a problems, not intended behaviour.
I think this should stay as is in the section at the end.

README.md
```diff
@@ -88,6 +87,11 @@ to use, test, and provide feedback on the new features are encouraged to try
 this branch. However, this branch provides no stability guarantees and reserves
 the right to change and break the API at any time.
 
+##Prerequisite knowledge
+
+Before attempting to understand this framework one should ideally have some familiarity with the structure and format of the Google Test framework, upon which it is based. Documentation for Google Test, including a "Getting Started" (primer) guide, is available here:
```
LebedevRI: But this is not *based* on that. They just happen to have a similar approach. As far as i know.

README.md
```diff
@@ -23,23 +23,21 @@ The basic steps for configuring and building the library look like this:
...
-dependency can be provided three ways:
+Note that Google Benchmark requires Google Test to build and run the tests. This
+dependency can be provided two ways:
 
-* Checkout the GTest sources into `benchmark/googletest`.
+* Checkout the Google Test sources into `benchmark/googletest` as above.
 * Otherwise, if `-DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON` is specified during
   configuration, the library will automatically download and build any required
   dependencies.
-* Otherwise, if nothing is done, CMake will use `find_package(GTest REQUIRED)`
```
LebedevRI: Why are you removing this?

README.md
```diff
@@ -114,9 +118,37 @@ BENCHMARK(BM_StringCopy);
...
+
+### Basic usage code explanation
+
+The following code:
+```BENCHMARK(BM_StringCreation);```
+registers the function as a google benchmark function, to be run when the:
+```BENCHMARK_MAIN();```
+function is called, along with any other registered benchmark functions.
+
+When the benchmark is run, google benchmark first runs it a certain number of times to ascertain how long the function usually takes to run. This serves two purposes: (a) "warming up" the cpu cache/prediction-code by running the code a number of times, (b) finding out how many iterations of the function will be necessary to get a statistically-meaningful result. For example, a function that only takes a couple of nanoseconds to run is far more sensitive during measurement to the effects of other things the OS is doing (networking, background processes) than a function which takes a couple of seconds. So the 2-nanosecond function needs to be run a greater number of times to get accurate timings.
```
LebedevRI: Please linewrap and use proper lists

README.md
mattreecebentley: They may be problems, but the user is not going to see the bottom of the document when they are trying to compile the first time, following the guidelines. When corrected in Gbenchmark, they can be corrected here also.

README.md
mattreecebentley: May have been a mistake. The previous line should've been removed (Otherwise, etc) as it is incorrect in my experience - something is broken there, at least under ubuntu.

README.md
mattreecebentley: When you say 'linewrap', what do you mean exactly.

README.md
mattreecebentley: This is actually part of a previous commit (see current readme.md). Google benchmark is indeed based on the gtest framework - that's why you need it to compile it.

README.md
mattreecebentley: BTW, I spent the better part of an hour googling to find the answer above the first time I tried to compile google benchmark code on windows. So I'm speaking from experience, not theory. You can't expect the user to read the whole document in one go, or assume that parts which relate to linking are somewhere else for some previously unknown reason.

README.md
LebedevRI: Uhm, no?
Can you quote the test you think says that?

mattreecebentley: What?
Have you even compiled google benchmark yourself before?
At any rate, it's part of a previous commit which has Already Been Accepted.

README.md
mattreecebentley: I'm not changing this commit.

README.md
LebedevRI: Uhm, if that is part of the README.md as of current git master, then why/how it is in this diff?
Since that line was added by #600 by You, the question remains unanswered.

LebedevRI:
This diff seems rather messed up.
Half of the changes has already happened before (in #600).
And i have to say, some of those changes are questionable.

README.md
dominichamon: Right now line 134 is a long line. Please wrap at the same width as the rest to make it more readable.

dominichamon:
There are conflicts so syncing to master HEAD and resolving would be useful before we can continue to review the changes.

README.md
mattreecebentley: Why is this important when the HTML will wrap anyway?
Happy to split up the lining to space it out a bit.

README.md
mattreecebentley: It was answered by the commit itself. Look it up kid.

mattreecebentley:
@dominichamon: I'll consider it. I didn't bother because I have low time. 
@LebedevRI: I don't think you're worth listening to, based on your comments.

README.md
dominichamon: Not everyone reads it as HTML.

dominichamon:
@mattreecebentley If you're going to take that attitude when presented with reasonable comments by an established contributor to the project, perhaps it's best you spend your limited time creating PRs for other projects.

Related, perhaps it's time i created code of conduct and respectful communications expectation documentation for the project.

README.md
LebedevRI: Maybe i just can't read. Which one?
* https://github.com/google/benchmark/pull/600/commits/53786ab39e68da9f9c4aeba0f9177fd921b37f6a
  > Third option under this  line: "Note that Google Benchmark requires Google Test to build and run the tests. This dependency can be provided three ways:"
  > Was not true (did not occur). If there is a further option that needs to be specified in order for that functionality to work it needs to be specified.

  That is not correct, it's clearly occurring - https://github.com/google/benchmark/blob/6d74c0625b8e88c1afce72b4f383c91b9a99dbe6/cmake/HandleGTest.cmake#L102
* https://github.com/google/benchmark/pull/600/commits/cf9e96815b298643c99831554e17a1ad6a25bd54 commit msg doesn't actually answer it, it just states it as a fact.

mattreecebentley: You're not being helpful, or posting under the right heading, and I'm going to ignore you.

mattreecebentley:
Well, if that's your attitude, consider your bias first. You've been fine, but he's been obstructive, unhelpful, arrogant, refusing to reply to requests for clarification. I don't have time for this as mentioned and your project can continue on without my unpaid help. Bye. Don't @ me.

EricWF:
This thread disappoints me. It's unfortunate that it seems we need a code of conduct. Personal attacks are never acceptable during code review.

mattreecebentley:
If I can't call a spade a spade - and I believe my previous statement describes his Behaviour accurately - then I quit.

PR #646, opened 2018-07-24
Iteration counts should be int64_t globally.

dominichamon:


dominichamon:
Those assembly tests suggest that this is creating worse assembly. Ie, more instructions being added around the main benchmark loop. going to abandon this PR.

PR #649, opened 2018-08-06
Remove redundant default which causes failures

kbobyrev:
I tried to embed Google Benchmark into the LLVM source tree and have encountered an error. Since the Benchmark CMakeLists contains the following lines:

```CMake
  add_cxx_compiler_flag(-Werror RELEASE)
  add_cxx_compiler_flag(-Werror RELWITHDEBINFO)
  add_cxx_compiler_flag(-Werror MINSIZEREL)
```

The compilation fails when all warnings by latest Clang are enabled because of the warnings "error: default label in switch which covers all enumeration values". This totally makes sense, because the `TimeUnit` only contains the cases which are already handled by the `switch` block:

```C++
enum TimeUnit { kNanosecond, kMicrosecond, kMillisecond };
```
Two functions which currently have this kind of behavior look like this:
```C++
inline const char* GetTimeUnitString(TimeUnit unit) {
  switch (unit) {
    case kMillisecond:
      return "ms";
    case kMicrosecond:
      return "us";
    case kNanosecond:
    default:
      return "ns";
  }
}
```

Unfortunately, the latest release (1.4.1) contains this code, I probably have to manually edit those in the LLVM tree.

Clang version which I used to compile the Benchmark Library: clang r335143.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1476,7 +1476,7 @@ inline const char* GetTimeUnitString(TimeUnit unit) {
     case kNanosecond:
       return "ns";
   }
-  return "ns";
+  __builtin_unreachable();
 }
 
 inline double GetTimeUnitMultiplier(TimeUnit unit) {
@@ -1488,7 +1488,7 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
     case kNanosecond:
       return 1e9;
   }
-  return 1e9;
+  __builtin_unreachable();
 }
 
 }  // namespace benchmark
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -241,6 +241,10 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #define BENCHMARK_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #endif
 
+#ifndef __has_builtin
+#define __has_builtin(x) 0
+#endif
+
 #if defined(COMPILER_GCC) || __has_builtin(__builtin_unreachable)
   #define BENCHMARK_UNREACHABLE() __builtin_unreachable()
 #elif defined(COMPILER_MSVC)
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1474,7 +1474,6 @@ inline const char* GetTimeUnitString(TimeUnit unit) {
     case kMicrosecond:
       return "us";
     case kNanosecond:
-    default:
       return "ns";
   }
 }
@@ -1486,7 +1485,6 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
     case kMicrosecond:
       return 1e6;
     case kNanosecond:
-    default:
       return 1e9;
   }
 }
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1476,6 +1476,7 @@ inline const char* GetTimeUnitString(TimeUnit unit) {
     case kNanosecond:
       return "ns";
   }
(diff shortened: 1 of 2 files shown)


LebedevRI:
Can you append `__builtin_unreachable()` after the switch ?

kbobyrev:
@LebedevRI sure. Are you suggesting to replace the outer `return VALUE;` statement (which was added in the latest commit so that old GCC doesn't complain about control reaching end of non-void function) or just before that, e.g.:

```
inline double GetTimeUnitMultiplier(TimeUnit unit) {
  switch (unit) {
    case kMillisecond:
      return 1e3;
    case kMicrosecond:
      return 1e6;
    case kNanosecond:
      return 1e9;
  }
  __builtin_unreachable();
  return 1e9;
}
```

LebedevRI:
```cpp
enum E {
    X,
    Y
};

int test(E e) {
    switch(e) {
        case X:
            return 0;
        case Y:
            return 1;
    }
    __builtin_unreachable();
}
```
https://godbolt.org/g/GjXwwY

kbobyrev:
@LebedevRI Ah, okay, sounds great! Thank you for the suggestion!

LebedevRI:
Won't work with msvc though IIRC, so you'll need to create
benchmark_unreachable() wrapper function, which is either `__builtin_unreachable()` normally,
or `__assume(0)` if the former is unavaliable.

include/benchmark/benchmark.h
```diff
@@ -241,6 +241,14 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #define BENCHMARK_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #endif
 
+#if defined(COMPILER_GCC) || __has_builtin(__builtin_unreachable)
```
LebedevRI: How will this work?
This has been moved from `src/internal_macros.h`,
which (at the top of the file) has (*)
```
#ifndef __has_feature
#define __has_feature(x) 0
#endif
#ifndef __has_builtin
#define __has_builtin(x) 0
#endif
```
But it includes *this* header before that:
```
```
So i think you also need to get (move?) the (*) block here.

kbobyrev:
Aww, I guess I totally messed this up, should get back to this tomorrow.

I assumed `__has_builtin(x)` is supported by all host compilers and missed this macro in the corresponding files.

Now the problem is also that `#if defined(COMPILER_GCC)` and similar are also defined in the header which I have operated on. This is a mess, I have to refine everything.

kbobyrev:
@LebedevRI these two are the only cases with `default:` placed in the wrong place. I was misguided by the compiler messages in the first place, thinking there were more, but all of the error messages were actually caused by these two pieces of code.

include/benchmark/benchmark.h
```diff
@@ -241,6 +241,32 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
...
+  #if !defined(COMPILER_GCC)
+    #define COMPILER_GCC
+  #endif
+#endif
+
+#ifndef __has_builtin
+#define __has_builtin(x) 0
+#endif
+
+#if defined(COMPILER_GCC) || __has_builtin(__builtin_unreachable)
```
dominichamon: given the rest of this header (above) works on __clang__ and __GNUC__ directly, perhaps you don't need to define the COMPILER_XXX macros?

If you do want to then maybe define them higher up (line 188?) and use them throughout.

But i think it's fine to remove lines 244 to 256 and not define new macros just for this check.

dominichamon:
Thanks!

kbobyrev:
@dominichamon thank you very much for taking a look!
It's way better to use the same macros the top of the file utilizes, I didn't notice them in the first place.

PR #652, opened 2018-08-16
properly escape json names

BaaMeow:
This PR is another piece broken off of #616 -- just a slight fix so that python JSON decoders (and other conforming decoders) parsing windows pathnames generated by Google Benchmark don't suddenly upend and explode with parsing errors because the backslash hasn't been escaped.


src/json_reporter.cc
```diff
@@ -78,7 +78,12 @@ bool JSONReporter::ReportContext(const Context& context) {
   out << indent << FormatKV("date", walltime_value) << ",\n";
 
   if (Context::executable_name) {
-    out << indent << FormatKV("executable", Context::executable_name) << ",\n";
+    // windows uses backslash for its path separator,
+    // which must be escaped in JSON otherwise it blows up conforming JSON
+    // decoders
+    std::string executable_name = Context::executable_name;
+    ReplaceAll(&executable_name, "\\", "\\\\");
```
dominichamon: is this only ever going to be a problem for executable name or should we do this for all FormatKV where value is string?

LebedevRI: In principle, all json strings should be properly escaped.

LebedevRI: I suspect when the proper json support is here, possibly it will be able to handle all this transparently.

BaaMeow: This is true, but I think it doesn't matter for everything else since we know we're not dumping out escape-necessary characters in any other place.

Of course, @LebedevRI is right: proper json support with some library would likely take care of it for us.

LebedevRI: Since we support custom statistics, and custom counters,
it is totally possible that users could be passing strings that need escaping,
but i think it can be done separately later.

dominichamon:
Thanks!

test/reporter_output_test.cc
```diff
@@ -23,7 +23,8 @@ static int AddContextCases() {
            {{"^\\{", MR_Default},
             {"\"context\":", MR_Next},
             {"\"date\": \"", MR_Next},
-            {"\"executable\": \".*/reporter_output_test(\\.exe)?\",", MR_Next},
+            {"\"executable\": \".*(/|\\\\)reporter_output_test(\\.exe)?\",",
```
LebedevRI: I was thinking of something more like:
```c++
#ifndef _WIN
  {"\"executable\": \".*/reporter_output_test\",", MR_Next},
#else
  {"\"executable\": \".*\\\\reporter_output_test\\.exe\",", MR_Next},
#endif
```

dominichamon:
Oops, sorry. Looks like I merged it a bit too quickly. Feel free to send a PR with just the test update if you want.

xorz57:
Nice!

PR #653, opened 2018-08-17
Make tests pass on 1-core VMs

bmwiedemann:
found while working on reproducible builds for openSUSE

To reproduce there
osc checkout openSUSE:Factory/benchmark && cd $_
osc build -j1 --vm-type=kvm


bmwiedemann:
I don't care. I could even live with a program telling me that I have "1 CPUs". Keeping code simple is always a plus for future maintainability.

dominichamon:
LGTM

PR #656, opened 2018-08-18
[NFC] Prefix "report(_)?mode" with Aggregation.

LebedevRI:
This only specifically represents handling of reporting of aggregates.
Not of anything else. Making it more specific makes the name less generic.

This is an issue because i want to add "iteration report mode",
so the naming would be conflicting.

I have split this off, since nobody likes huge commits :)

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -424,14 +424,14 @@ namespace internal {
 class ThreadTimer;
 class ThreadManager;
 
-enum ReportMode
+enum AggregationReportMode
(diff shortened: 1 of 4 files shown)


LebedevRI:
To address the obvious question - yes, i suppose i could turn this into a bitfield, and squeeze 'iteration report mode' too. But i will also need to add '*display* aggregates only' into `AggregationReportMode`..
So this would be more messy with bitfield. I think?

LebedevRI:
Ping.

dominichamon:
LGTM

LebedevRI:
Thank you for the review.

PR #654, opened 2018-08-18
Mark Set{Items,Bytes}Processed()/{items,bytes}_processed() as deprecated.

LebedevRI:
Mark Set{Items,Bytes}Processed()/{items,bytes}_processed() as deprecated.

They are basically proto-version of custom user counters.
It does not seem that they do anything that custom user counters
don't do. And having two similar entities is not good for generalization.

Migration plan:
* ```
  SetItemsProcessed(<val>)
    =>
  state.counters.insert({
    {"<Name>", benchmark::Counter(<val>, benchmark::Counter::kIsRate)},
    ...
  });
  ```
* ```
  SetBytesProcessed(<val>)
    =>
  state.counters.insert({
    {"<Name>", benchmark::Counter(<val>, benchmark::Counter::kIsRate,
                                  benchmark::Counter::OneK::kIs1024)},
    ...
  });
  ```
* ```
  <Name>_processed()
    =>
  state.counters["<Name>"]
  ```

One thing the custom user counters miss is better support
for units of measurement.

Refs. https://github.com/google/benchmark/issues/627

PR #657, opened 2018-08-19
Counter(): add 'one thousand' param.

LebedevRI:
Needed for https://github.com/google/benchmark/pull/654

Custom user counters are quite custom. It is not guaranteed
that the user *always* expects for these to have 1k == 1000.
If the counter represents bytes/memory/etc, 1k should be 1024.

Some bikeshedding points:
1. Is this sufficient, or do we really want to go full on
   into custom types with names?
   I think just the '1000' is sufficient for now.
2. Should there be a helper benchmark::Counter::Counter{1000,1024}()
   static 'constructor' functions, since these two, by far,
   will be the most used?
3. In the future, we should be somehow encoding this info into JSON.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -373,7 +373,7 @@ class Counter {
     kAvgIterationsRate = kIsRate | kAvgIterations
   };
 
-  enum One_K {
+  enum OneK {
     // 1'000 items per 1k
     kIs1000 = 1000,
     // 1'024 items per 1k
@@ -382,10 +382,10 @@ class Counter {
 
   double value;
   Flags flags;
-  One_K oneK;
+  OneK oneK;
 
   BENCHMARK_ALWAYS_INLINE
-  Counter(double v = 0., Flags f = kDefaults, One_K k = kIs1000)
+  Counter(double v = 0., Flags f = kDefaults, OneK k = kIs1000)
       : value(v), flags(f), oneK(k) {}
 
   BENCHMARK_ALWAYS_INLINE operator double const&() const { return value; }
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -375,10 +375,11 @@ class Counter {
 
   double value;
   Flags flags;
-  int thousand;
+  std::pair<int /*Num*/, int /*Denom*/> thousand;
 
   BENCHMARK_ALWAYS_INLINE
-  Counter(double v = 0., Flags f = kDefaults, int t = 1000)
+  Counter(double v = 0., Flags f = kDefaults,
+          std::pair<int /*Num*/, int /*Denom*/> t = std::make_pair(1000, 1))
       : value(v), flags(f), thousand(t) {}
 
   BENCHMARK_ALWAYS_INLINE operator double const&() const { return value; }
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -375,9 +375,11 @@ class Counter {
 
   double value;
   Flags flags;
+  int thousand;
 
   BENCHMARK_ALWAYS_INLINE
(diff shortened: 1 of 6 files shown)


EricWF:
I haven't looked at this, but my preference would be to encode this info in the type system in some way. Sort of like std::ratio.

So maybe make the parameter something similar instead of just taking an integer?

LebedevRI:
I'm not sure how that would look. Just take `std::pair<int /*num*/, int /*den*/>` ?

LebedevRI:
Ping.

src/console_reporter.cc
```diff
@@ -150,7 +150,9 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   for (auto& c : result.counters) {
     const std::size_t cNameLen = std::max(std::string::size_type(10),
                                           c.first.length());
-    auto const& s = HumanReadableNumber(c.second.value, 1000);
+    const double one_k =
+        double(c.second.thousand.first) / double(c.second.thousand.second);
```
dominichamon: why is this specified as a pair instead of just a single number (1000 vs 1024)? or even an enum!

dominichamon: Ah, I think that is suggesting either having an enum with values 1000 and 1024, or to extend Counter to have MegaCounter and MebiCounter variants, which implicitly manage the one_k definition without it being specified.

LebedevRI: Right now i don't have/is not aware of any real use for the one_k not being 1000 or 1024, so enum _could_ work.
It is just that storing enum will take just as much space as storing an int, or a pair of shorts:
https://godbolt.org/z/9Q1gVS
So it isn't obvious to me why we want to avoid doing/allowing the more general thing at the **same** cost.

include/benchmark/benchmark.h
```diff
@@ -375,9 +375,12 @@ class Counter {
 
   double value;
   Flags flags;
+  std::pair<int /*Num*/, int /*Denom*/> thousand;
```
LebedevRI: Hmm, if it to be an enum, should i just add value `k1024` to the `Flags` enum?

dominichamon: I think it's an orthogonal feature so i'd add a OneK enum with two values.

LebedevRI: Ok, sounds good.

src/console_reporter.cc
```diff
@@ -150,8 +150,15 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   for (auto& c : result.counters) {
     const std::size_t cNameLen = std::max(std::string::size_type(10),
                                           c.first.length());
-    const double one_k =
-        double(c.second.thousand.first) / double(c.second.thousand.second);
+    const double one_k = [](benchmark::Counter::One_K k) -> double {
```
dominichamon: why not give the enums the values '1000' and '1024' and pass them through (with any necessary cast) to the HumanReadableNumber call?

LebedevRI: There is no _real_ reason not to do it.
I didn't do that because `0` and `1` (which is most likely the current values of these values)
only need one bit, while `1024` needs 11 bits. So _if_ we wanted to pack the size of `Counter`,
it would _likely_ be simpler with smaller enum values.
I don't think we care about that here, so let's avoid this premature optimization :)

include/benchmark/benchmark.h
```diff
@@ -373,11 +373,20 @@ class Counter {
     kAvgIterationsRate = kIsRate | kAvgIterations
   };
 
+  enum One_K {
```
dominichamon: nit: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

OneK is a better name.

LebedevRI:
Thank you for the review!
Awaiting CI completion..

PR #658, opened 2018-08-20
Track 'type' of the run - is it an actual measurement, or an aggregate.

LebedevRI:
This is *only* exposed in the JSON. Not in CSV, which is deprecated.

This *only* supposed to track these two states.
An additional field could later track which aggregate this is,
specifically (statistic name, rms, bigo, ...) (https://github.com/google/benchmark/pull/616)

The motivation is that we already have ReportAggregatesOnly,
but it affects the entire reports, both the display,
and the reporters (json files), which isn't ideal.

It would be very useful to have a 'display aggregates only' option,
both in the library's console reporter, and the python tooling,
This will be especially needed for the 'store separate iterations'.

I'm iffy on the actual names, so please bikeshed away :)

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1276,8 +1276,11 @@ class BenchmarkReporter {
   };
 
   struct Run {
+    enum RunType { RT_Iteration, RT_Aggregate };
+
     Run()
-        : error_occurred(false),
+        : run_type(RT_Iteration),
+          error_occurred(false),
           iterations(1),
           time_unit(kNanosecond),
           real_accumulated_time(0),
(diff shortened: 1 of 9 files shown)


LebedevRI:
Ping.

dominichamon:
I'm a little fuzzy on the usefulness, but it also doesn't affect much, so LGTM.

LebedevRI:
> I'm a little fuzzy on the usefulness

I hoped i explained it in the description..
Basically, i want a `--benchmark_report_aggregates_only={true|false}`,
but for the `tools/compare.py`, and this is the step in that direction.

> but it also doesn't affect much, so LGTM.

Thank you for the review!
Care to actually 'Review changes'  -> 'Approve'? Or textual form is sufficient?

PR #663, opened 2018-08-28
[NFC] s/console_reporter/display_reporter/

LebedevRI:
There are two destinations:
* display (console, terminal) and
* file.

And each of the destinations can be poplulated with one of the reporters:
* console - human-friendly table-like display
* json
* csv (deprecated)

So using the name console_reporter is confusing.
Is it talking about the console reporter in the sense of
table-like reporter, or in the sense of display destination?


LebedevRI:
Thank you for the review!

Issue #664, opened 2018-08-29
[RFC] *Display* aggregates only.


LebedevRI:
There is a flag 
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/src/benchmark.cc#L75-L78
and a call
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/include/benchmark/benchmark.h#L837-L840
But that affects everything, every reporter, destination:
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/src/benchmark.cc#L316

It would be quite useful to have an ability to be more picky.

More specifically, i would like to be able to only see the aggregates in the on-screen output,
but for the file output to still contain everything. The former is useful in case of a lot of repetition
(or even more so if every iteration is reported separately), while the former is **great** for tooling.

Now the problem. I'm not sure how best to do it.
The most straight-forward solution i can think of is to filter the results here:
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/src/benchmark.cc#L466-L471
So the file reporter will still receive the full `std::vector<BenchmarkReporter::Run> reports`, while
`display_reporter` may (depending on the configuration) only receive the aggregates.

Alternative solutions could include modifying the API of `BenchmarkReporter`,
and every implementation, to  skip non-aggregates, depending on some additional config params.

Thoughts?

PR #665, opened 2018-09-02
*Display* aggregates only.

LebedevRI:
There is a flag 
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/src/benchmark.cc#L75-L78
and a call
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/include/benchmark/benchmark.h#L837-L840
But that affects everything, every reporter, destination:
https://github.com/google/benchmark/blob/d9cab612e40017af10bddaa5b60c7067032a9e1c/src/benchmark.cc#L316

It would be quite useful to have an ability to be more picky.

More specifically, i would like to be able to only see the aggregates in the on-screen output,
but for the file output to still contain everything. The former is useful in case of a lot of repetition
(or even more so if every iteration is reported separately), while the former is **great** for tooling.

This does work, and is even less intrusive than i thought it will be.
But there is a slight test coverage issue :)
These tests are *all* testing the first, display, reporter.
So we don't actually check what happens with the second, file, reporter...
This problem isn't a new one, but here i would say it is more severe.
I'll see if i can easily fix it for this particular case..

Fixes https://github.com/google/benchmark/issues/664


LebedevRI:
After digging a bit more, this is the problematic bit:
https://github.com/google/benchmark/blob/5159967520f45d2bb31b226b810f4f8dc03df1c4/src/benchmark.cc#L566-L567
**If** we wanted to test that, we'd need to add some internal debug option to
disable this destination rebinding. Reading back from file seems rather fragile..

include/benchmark/benchmark.h
```diff
@@ -438,9 +438,12 @@ enum AggregationReportMode
     : unsigned
 #else
 #endif
-{ ARM_Unspecified,  // The mode has not been manually specified
-  ARM_Default,      // The mode is user-specified as default.
-  ARM_ReportAggregatesOnly };
+{ ARM_Unspecified,           // The mode has not been manually specified
+  ARM_Default,               // The mode is user-specified as default.
+  ARM_ReportAggregatesOnly,  // All reporters only display aggregates.
```
dominichamon: does the PR get even simpler if these are renamed to:

ARM_FileReportAggregatesOnly,
ARM_DisplayReportAggregatesOnly,

and then the existing flag sets them both to true, and the new flag sets only the Display one to true?

src/benchmark.cc
```diff
@@ -193,9 +203,11 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
   manager->NotifyThreadComplete();
 }
 
-std::vector<BenchmarkReporter::Run> RunBenchmark(
-    const benchmark::internal::Benchmark::Instance& b,
-    std::vector<BenchmarkReporter::Run>* complexity_reports) {
+std::pair<std::vector<BenchmarkReporter::Run> /*everything*/,
```
dominichamon: is it worth returning a custom struct to avoid having the comments tracking which member of the pair is which?

src/benchmark.cc
```diff
@@ -463,10 +488,18 @@ void RunBenchmarks(const std::vector<Benchmark::Instance>& benchmarks,
...
-      display_reporter->ReportRuns(reports);
-      if (file_reporter) file_reporter->ReportRuns(reports);
+      bool display_aggregates_only;
+      std::pair<std::vector<BenchmarkReporter::Run> /*everything*/,
+                std::vector<BenchmarkReporter::Run> /*aggregates_only*/>
+          reports = RunBenchmark(benchmark, &complexity_reports,
+                                 display_aggregates_only);
+
+      const std::vector<BenchmarkReporter::Run>& report_for_display =
+          display_aggregates_only ? reports.second /*aggregates_only*/
```
dominichamon: would this also be simpler if the return type was 'non-aggregate', 'aggregate' instead of having the first be everything?

LebedevRI: Hmm, actually it might be.
I didn't do so initially because i thought it was important to print the whole benchmark at once
(for header/context pre-print), but maybe it is too pessimistic..

src/benchmark.cc
LebedevRI: Makes sense.

include/benchmark/benchmark.h
LebedevRI: Hmm, kinda.
It gets a bit messier due to the command-line flags.

src/benchmark.cc
```diff
@@ -193,22 +203,36 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
...
+      repeats != 1 && (b.aggregation_report_mode == internal::ARM_Unspecified
+                           ? FLAGS_benchmark_report_aggregates_only ||
+                                 FLAGS_benchmark_display_aggregates_only
+                           : b.aggregation_report_mode &
+                                 internal::ARM_DisplayReportAggregatesOnly);
+  run_results.file_report_aggregates_only =
+      repeats != 1 && (b.aggregation_report_mode == internal::ARM_Unspecified
+                           ? FLAGS_benchmark_report_aggregates_only
+                           : b.aggregation_report_mode &
+                                 internal::ARM_FileReportAggregatesOnly);
```
dominichamon:
```c++
if (repeats != 1) {
  run_results.display_report_aggregates_only = (
      FLAGS_benchmark_report_aggregates_only || FLAGS_benchmark_display_aggregates_only);
  run_results.file_report_aggregates_only = FLAGS_benchmark_report_aggregates_only;
  if (b.aggregation_report_mode != internal::ARM_Unspecified) {
    run_results.display_report_aggregates_only = (
      b.aggregation_report_mode & internal::ARM_DisplayReportAggregatesOnly);
    run_results.file_report_aggregates_only = (
      b.aggregation_report_mode & internal::ARM_FileReportAggregatesOnly);
  }
}
```

a bit clearer than nested ternary operators, i think...

src/benchmark.cc
```diff
@@ -193,22 +203,36 @@ void RunInThread(const benchmark::internal::Benchmark::Instance* b,
...
   std::unique_ptr<internal::ThreadManager> manager;
   std::vector<std::thread> pool(b.threads - 1);
   const int repeats =
       b.repetitions != 0 ? b.repetitions : FLAGS_benchmark_repetitions;
-  const bool report_aggregates_only =
-      repeats != 1 &&
-      (b.aggregation_report_mode == internal::ARM_Unspecified
-           ? FLAGS_benchmark_report_aggregates_only
-           : b.aggregation_report_mode == internal::ARM_ReportAggregatesOnly);
+  run_results.display_report_aggregates_only =
```
dominichamon: why do you need to return this? you should be able to (as it was) have a local constant and decide to add to the vector or not based on the value.

src/benchmark.cc
LebedevRI: Oh duh, I've meant to clean this but overlooked :)
Thanks!

src/benchmark.cc
LebedevRI: Then i don't understand the previous suggestion, https://github.com/google/benchmark/pull/665#discussion_r214663892.

LebedevRI: But to attempt to answer, `display_report_aggregates_only` is needed,
because i need to somehow convey back which result (all the reports, or aggregates only)
to report, for both of the reporters. I need to do so because i actually look at the
`benchmark::internal::Benchmark::Instance::aggregation_report_mode`,
which is per-instance, not just `FLAGS_benchmark_report_aggregates_only` /
`FLAGS_benchmark_display_aggregates_only`, which is global.

dominichamon: Ah right, sorry.

LebedevRI:
Any further thoughts here? :)
The testing issue is unresolved, but i'm not seeing any clear/clean solution for it..

dominichamon:
I have nothing further. I've been unable to come up with a satisfactory testing approach without radical replumbing of the file reporter.

LebedevRI:
Makes sense..
Would it be unacceptably ugly to simply read back that file and check that?
I guess the troubling part is coming up with the tmp filename.

dominichamon:
```c++
#include <cstdio>
#include <filesystem>

std::filesystem::read_symlink(
    std::filesystem::path("/proc/self/fd") / std::to_string(fileno(tmpfile()))
);
```

would at least work on linux :P

LebedevRI:
Hmm, actually now that i think about it [`std::tmpnam()`](https://en.cppreference.com/w/cpp/io/c/tmpnam) might work, too,
and be more cross-platform. (ignoring the fact that `std::filesystem` is C++17)

LebedevRI:
This is so horrible ^^

test/report_aggregates_only_test.cc
```diff
@@ -0,0 +1,67 @@
...
+int SubstrCnt(const std::string& haystack, const std::string& pat) {
+  if (pat.length() == 0) return 0;
+  int count = 0;
+  for (size_t offset = haystack.find(pat); offset != std::string::npos;
+       offset = haystack.find(pat, offset + pat.length()))
+    ++count;
+  return count;
+}
+
+int main(int argc, char* argv[]) {
```
dominichamon: Is it worth putting some of this logic in the OutputTestHelper or a new helper that can be shared between the two tests?

test/display_aggregates_only_test.cc
```diff
@@ -0,0 +1,71 @@
...
+  new_argv.emplace_back(const_cast<char*>(tmp.c_str()));
+
+  argc = int(new_argv.size());
+
+  benchmark::Initialize(&argc, new_argv.data());
+  benchmark::RunSpecifiedBenchmarks();
+
+  // Read the output back from the file, and delete the file.
+  std::ifstream tmp_stream(tmp_file_name);
+  std::string JsonOutput((std::istreambuf_iterator<char>(tmp_stream)),
```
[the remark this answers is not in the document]
LebedevRI: Yes, they are necessary.
https://godbolt.org/z/V6ASzJ

test/report_aggregates_only_test.cc
LebedevRI: Hmm, maybe i can wrap some of this support code into RAII class..

LebedevRI:
@dominichamon  Thank you for taking a look!

test/display_aggregates_only_test.cc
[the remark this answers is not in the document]
LebedevRI: This should do it :)

test/report_aggregates_only_test.cc
LebedevRI: Yep, looks better now.

test/file_reporter_test_helper.h
```diff
@@ -0,0 +1,41 @@
...
+#include <cstdio>
+#include <fstream>
+#include <streambuf>
+#include <string>
+#include <vector>
+
+#include "benchmark/benchmark.h"
+#include "output_test.h"
+
+class TestBenchmarkFileReporter {
```
dominichamon: why isn't this a static method that returns the string?

test/output_test.h
```diff
@@ -201,6 +204,40 @@ namespace {
 
 const char* const dec_re = "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?";
 
+inline int SubstrCnt(const std::string& haystack, const std::string& pat) {
```
dominichamon: personally, i'd declare it here and implement it in a source file. but that's not essential.

LebedevRI: Why not.

LebedevRI:
@dominichamon thank you for the review! I will merge this soon-ish.

PR #666, opened 2018-09-03
add missing closing bracket in --help message

pseyfert:
When running a benchmark application with `--help` the usage looks like this to me

```
benchmark [--benchmark_list_tests={true|false}]
          [--benchmark_filter=<regex>]
          [--benchmark_min_time=<min_time>]
          [--benchmark_repetitions=<num_repetitions>]
          [--benchmark_report_aggregates_only={true|false}
          [--benchmark_format=<console|json|csv>]
          [--benchmark_out=<filename>]
...
```

In the line for `report_aggregates_only` the closing `]` is missing, which I added here.


LebedevRI:
The appveyor build is clearly failing regardless of this change, so i think this is good to go.

PR #667, opened 2018-09-04
GetCacheSizesMacOSX(): use consistent types.

LebedevRI:
I have absolutely no way to test this, but this looks obviously-good.

This was reported by Tim Northover @TNorthover in
http://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20180903/584223.html

> I think this breaks some 32-bit configurations (well, mine at least).
> I was using Clang (from Xcode 10 beta) on macOS and got a bunch of
> errors referencing sysinfo.cc:292 and onwards:
>
> /Users/tim/llvm/llvm-project/llvm/utils/benchmark/src/sysinfo.cc:292:47:
> error: non-constant-expression cannot be narrowed from type
> 'std::__1::array<unsigned long long, 4>::value_type' (aka 'unsigned
> long long') to 'size_t' (aka 'unsigned long') in initializer list
> [-Wc++11-narrowing]
>   } Cases[] = {{"hw.l1dcachesize", "Data", 1, CacheCounts[1]},
>                                               ^~~~~~~~~~~~~~
>
> The same happens when self-hosting ToT. Unfortunately I couldn't
> reproduce the issue on Debian (Clang 6.0.1) even with libc++; I'm not
> sure what the difference is.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -288,7 +288,7 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesMacOSX() {
     std::string name;
     std::string type;
     int level;
-    size_t num_sharing;
+    uint64_t num_sharing;
   } Cases[] = {{"hw.l1dcachesize", "Data", 1, CacheCounts[1]},
                {"hw.l1icachesize", "Instruction", 1, CacheCounts[1]},
                {"hw.l2cachesize", "Unified", 2, CacheCounts[2]},


TNorthover:
Thanks Roman! I checked on my local build and this did indeed fix the issue.

dominichamon:
The test bots do a pretty good job of covering 32-bit and such, but unfortunately didn't catch this one. We must be missing a configuration.

LebedevRI:
There are no 32-bit builds for OSX as far as i can tell.
Though i'm unfamiliar with that target so i'm unsure i can help adding them..

LebedevRI:
@dominichamon 
> We must be missing a configuration.
Yep, it turned out to be trivial https://github.com/google/benchmark/pull/669
Looks good now?

PR #673, opened 2018-09-10
Backport LLVM's r341717 "Fix flags used to compile benchmark library with clang-cl"

LebedevRI:
> `MSVC` is true for clang-cl, but `"${CMAKE_CXX_COMPILER_ID}" STREQUAL
> "MSVC"` is false, so we would enable -Wall, which means -Weverything
> with clang-cl, and we get tons of undesired warnings.
> 
> Use the simpler condition to fix things.

Patch by: Reid Kleckner @rnk
https://reviews.llvm.org/rL341717
https://github.com/llvm-mirror/llvm/commit/a8ea812648fd051aa6e847735ddab5f84a1ba5f8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,7 @@ if (BENCHMARK_BUILD_32_BITS)
   add_required_cxx_compiler_flag(-m32)
 endif()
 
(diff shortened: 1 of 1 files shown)


rnk:
Looks good, thanks!

PR #675, opened 2018-09-13
Track two more details about runs - the aggregate name, and run name.

LebedevRI:
This is related to @BaaMeow's work in https://github.com/google/benchmark/pull/616 but is not based on it.

Two new fields are tracked, and dumped into JSON:
* If the run is an aggregate, the aggregate's name is stored.
  It can be RMS, BigO, mean, median, stddev, or any custom stat name.
* The aggregate-name-less run name is additionally stored.
  I.e. not some name of the benchmark function, but the actual
  name, but without the 'aggregate name' suffix.

This way one can group/filter all the runs,
and filter by the particular aggregate type.

I *might* need this for further tooling improvement.
Or maybe not.
But this is certainly worthwhile for custom tooling.


LebedevRI:
And this is why two build systems is a **horrible** idea.

src/complexity.cc
```diff
@@ -187,8 +187,10 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
 
   // Get the data from the accumulator to BenchmarkReporter::Run's.
   Run big_o;
+  big_o.run_name = benchmark_name;
   big_o.run_type = BenchmarkReporter::Run::RT_Aggregate;
-  big_o.benchmark_name = benchmark_name + "_BigO";
+  big_o.aggregate_name = "BigO";
+  big_o.benchmark_name = big_o.run_name + "_" + big_o.aggregate_name;
```
dominichamon: if this is always true, do we need to store it? or can we calculate it in a method?

```c++
std::string Run::benchmark_name() const {
  std::string name = run_name;
  if (!aggregate_name.empty()) {
    name += "_" + aggregate_name;
  }
  return name;
}
```

LebedevRI: I think it could work in theory, but i have no clue what
```c++
   std::string benchmark_name =
       reports[0].benchmark_name.substr(0, reports[0].benchmark_name.find('/'));
```
is supposed to be doing.

LebedevRI: In other words, if we can drop that, then we can create `Run::benchmark_name()`

LebedevRI: Ah i see
```
-----------------------------------------------------------------------------
Benchmark                                      Time           CPU Iterations
-----------------------------------------------------------------------------
...
BM_Complexity_O1/1                            14 ns         14 ns    1006981
BM_Complexity_O1/8                            14 ns         14 ns    1032428
BM_Complexity_O1/64                           14 ns         14 ns    1032418
BM_Complexity_O1/512                          14 ns         14 ns     998075
BM_Complexity_O1/4096                         13 ns         13 ns    1029608
BM_Complexity_O1/32768                        14 ns         14 ns    1022146
BM_Complexity_O1/262144                       14 ns         14 ns    1008279
BM_Complexity_O1/1_BigO                    13.62 (1)      13.62 (1) 
BM_Complexity_O1/1_RMS                         1 %          1 % 
```
```
BM../test/output_test_helper.cc:108: CheckCase: Check `remaining_output.eof() == false' failed. End of output reached before match for regex "^BM_Complexity_O1_BigO %bigOStr %bigOStr[ ]*$" was found
    actual regex string "^BM_Complexity_O1_BigO [ ]* [0-9]*[.]?[0-9]+([eE][-+][0-9]+)? \([0-9]+\) [ ]* [0-9]*[.]?[0-9]+([eE][-+][0-9]+)? \([0-9]+\)[ ]*$"
    started matching near: BM_Complexity_O1/1                            14 ns         14 ns    1006981
```
So i'm not sure we can do that simplification.

LebedevRI: Hm no, i was thinking about something else there it seems..

src/complexity.cc
```diff
@@ -182,15 +182,15 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_cpu = MinimalLeastSq(n, cpu_time, reports[0].complexity);
     result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
   }
-  std::string benchmark_name =
-      reports[0].benchmark_name.substr(0, reports[0].benchmark_name.find('/'));
+
+  std::string benchmark_name = reports[0].benchmark_name().substr(
```
dominichamon: this should now be `run_name` right? at least, you assign it to `run_name` below.

LebedevRI: Hmm, right, sorry.
I think this `run_name` will be kinda broken for these range-based complexity measurements..

LebedevRI:
Thank you for the review!

PR #676, opened 2018-09-13
Un-deprecate [SG]et{Item,Byte}sProcessed, reimplement as custom counters.

### LINKED ISSUE #627: V2: set*processed() should go
They should probably be deprecated in v2.
Custom Counters are more general, more flexible.

LebedevRI:
As disscussed with @dominichamon and @dbabokin, sugar is nice.
Well, maybe not for the health, but it's sweet.
Alright, enough puns.

A special care needs to be applied not to break csv reporter. UGH.
We end up shedding some code over this.
We no longer specially pretty-print them, they are printed just like the rest of custom counters.

Fixes #627.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -548,24 +548,21 @@ class State {
 
   // Set the number of bytes processed by the current benchmark
   // execution.  This routine is typically called once at the end of a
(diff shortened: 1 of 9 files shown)


LebedevRI:
Thank you for the review!

PR #678, opened 2018-09-16
[Tooling] Rewrite generate_difference_report().

### LINKED ISSUE #677: tools/gbench/report.py: generate_difference_report() is *broken*.
https://github.com/google/benchmark/blob/7d03f2df490c89b2a2055e9be4e2c36db5aedd80/tools/gbench/report.py#L174-L176
If there are repetitions (or in general, same names), for the second benchmark, we always use the first repetition/match..
This is visible in tests, too.
This means the current U test is 'incorrect', too.

LebedevRI:
My knowledge of python is not great, so this is kinda horrible.

Two things:
1. If there were repetitions, for the RHS (i.e. the new value) we were always using the first repetition,
    which naturally results in incorrect change reports for the second and following repetitions.
    And what is even worse, that completely broke U test. :(
2. A better support for different repetition count for U test was missing.
    It's important if we are to be able to report 'iteration as repetition',
    since it is rather likely that the iteration count will mismatch.

Now, the rough idea on how this is implemented now. I think this is the right solution.
1. Get all benchmark names (in order) from the lhs benchmark.
2. While preserving the order, keep the unique names
3. Get all benchmark names (in order) from the rhs benchmark.
4. While preserving the order, keep the unique names
5. Intersect `2.` and `4.`, get the list of unique benchmark names that exist on both sides.
6. Now, we want to group (partition) all the benchmarks with the same name.
   ```
   BM_FOO:
       [lhs]: BM_FOO/repetition0 BM_FOO/repetition1
       [rhs]: BM_FOO/repetition0 BM_FOO/repetition1 BM_FOO/repetition2
   ...
   ```
   We also drop mismatches in `time_unit` here.
   _(whose bright idea was it to store arbitrarily scaled timers in json **?!** )_
7. Iterate for each partition
7.1. Conditionally, diff the overlapping repetitions (the count of repetitions may be different.)
7.2. Conditionally, do the U test:
7.2.1. Get **all** the values of `"real_time"` field from the lhs benchmark
7.2.2. Get **all** the values of `"cpu_time"` field from the lhs benchmark
7.2.3. Get **all** the values of `"real_time"` field from the rhs benchmark
7.2.4. Get **all** the values of `"cpu_time"` field from the rhs benchmark
          NOTE: the repetition count may be different, but we want *all* the values!
7.2.5. Do the rest of the u test stuff
7.2.6. Print u test
8. ???
9. **PROFIT**!

Fixes #677

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -105,17 +105,12 @@ def get_unique_benchmark_names(json):
     return uniqued
 
 
-def get_shared_members(list1, list2):
+def intersect(list1, list2):
     """
     Given two lists, get a new list consisting of the elements only contained
     in *both of the input lists*, while preserving the ordering.
     """
-    seen_in_list1 = set(list1)
-    seen_in_list2 = set(list2)
-    common_set = seen_in_list1 & seen_in_list2
-    # It doesn't matter which list we will now use.
-    common = [x for x in list1 if x in common_set]
-    return common
+    return [x for x in list1 if x in list2]
 
 
 def partition_benchmarks(json1, json2):
@@ -126,7 +121,7 @@ def partition_benchmarks(json1, json2):
     """
     json1_unique_names = get_unique_benchmark_names(json1)
     json2_unique_names = get_unique_benchmark_names(json2)
-    names = get_shared_members(json1_unique_names, json2_unique_names)
+    names = intersect(json1_unique_names, json2_unique_names)
     partitions = []
     for name in names:
         # Pick the time unit from the first entry of the lhs benchmark.
@@ -162,10 +157,7 @@ def print_utest(partition, utest_alpha, first_col_width, use_color=True):
         return []
 
     def get_utest_color(pval):
-        if pval >= utest_alpha:
-            return BC_FAIL
-        else:
-            return BC_OKGREEN
+        return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
 
     time_pvalue = mannwhitneyu(
         timings_time[0], timings_time[1], alternative='two-sided').pvalue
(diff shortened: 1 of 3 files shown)


tools/compare.py
```diff
@@ -36,6 +36,17 @@ def create_parser():
     parser = ArgumentParser(
         description='versatile benchmark output compare tool')
 
+    parser.add_argument(
+        '-a',
+        '--display_aggregates_only',
+        dest='display_aggregates_only',
+        action="store_true",
+        help="If there are repetitions, by default, we display everything - the"
+             " actual runs, and the aggregates computed. Sometimes, it is "
```
dominichamon: nit: please put the space before 'actual' on the previous line.

tools/gbench/report.py
```diff
@@ -93,9 +94,111 @@ def filter_benchmark(json_orig, family, replacement=""):
     return filtered
 
 
+def get_unique_benchmark_names(json):
```
dominichamon: i don't see one, so maybe add a test just for this?

tools/gbench/report.py
```diff
@@ -93,9 +94,111 @@ def filter_benchmark(json_orig, family, replacement=""):
...
+               (seen.add(x['name']) or True)]
+    return uniqued
+
+
+def get_shared_members(list1, list2):
+    """
+    Given two lists, get a new list consisting of the elements only contained
+    in *both of the input lists*, while preserving the ordering.
+    """
+    seen_in_list1 = set(list1)
```
dominichamon:
```python
return [x for x in list1 if value in list2]
```

i think will do the same. 

```
>>> l = [3,1,2,4]
>>> m = [3,2,4,5]
>>> 
>>> [x for x in l if x in m]
[3, 2, 4]
>>> [x for x in m if x in l]
[3, 2, 4]
>>> 
```

tools/gbench/report.py
```diff
@@ -93,9 +94,111 @@ def filter_benchmark(json_orig, family, replacement=""):
...
+    While *keeping* the order, give all the unique 'names' used for benchmarks.
+    """
+    seen = set()
+    uniqued = [x['name'] for x in json['benchmarks']
+               if x['name'] not in seen and
+               (seen.add(x['name']) or True)]
+    return uniqued
+
+
+def get_shared_members(list1, list2):
```
dominichamon: maybe call this 'intersect' if that's what it's doing.

tools/gbench/report.py
```diff
@@ -93,9 +94,111 @@ def filter_benchmark(json_orig, family, replacement=""):
...
+                      len(timings_time[1]),
+                      len(timings_cpu[0]),
+                      len(timings_cpu[1]))
+
+    # Does *everything* has at least UTEST_MIN_REPETITIONS repetitions?
+    if min_rep_cnt < UTEST_MIN_REPETITIONS:
+        return []
+
+    def get_utest_color(pval):
+        if pval >= utest_alpha:
```
dominichamon:
```python
return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
```

tools/gbench/report.py
LebedevRI: Thanks!
I think i was thinking along the lines that looking up in `set()` would be better,
but i guess we don't really care here.

LebedevRI:
Further thoughts here? Moar tests?

LebedevRI:
Alright, thank you for the review!

PR #679, opened 2018-09-17
[benchmark] Lowercase windows specific includes

mstorsjo:
The windows SDK headers don't have self-consistent casing anyway, and many projects consistently use lowercase for them, in order to fix crosscompilation with mingw headers.


mstorsjo:
This change was originally written for LLVM, at https://reviews.llvm.org/D52181, but it's not approved and committed there yet.

dominichamon:
Appveyor seems unrelated. LGTM.

PR #680, opened 2018-09-18
Include sys/time.h for cycleclock.h when building on MinGW

mstorsjo:
When building for ARM, there is a fallback codepath that uses
gettimeofday, which requires sys/time.h.

The Windows SDK doesn't have this header, but MinGW does have it.
Thus, this fixes building for Windows on ARM with MinGW
headers/libraries, while Windows on ARM with the Windows SDK still
is broken.

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -41,7 +41,7 @@ extern "C" uint64_t __rdtsc();
 #pragma intrinsic(__rdtsc)
 #endif
 
-#ifndef BENCHMARK_OS_WINDOWS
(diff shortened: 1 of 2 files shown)


mstorsjo:
> How does this affect non-ARM MinGW builds?

All MinGW environments have time.h and sys/time.h, so it shouldn't matter if we include these headers there, just like it is done on all non-windows platforms.

mstorsjo:
Also just for the record; I realize this isn't the ideal full solution. I'm not extremely familiar with the lowlevel Windows timing APIs, but I think something like QueryPerformanceCounter() could be used as fallback instead of gettimeofday.

dominichamon:
works for me. i don't have windows (let alone arm) on which to test.

PR #613, opened 2018-06-06
Document and implement cmake dependency bump to 3.5.1

dmah42:
Fixes #603

PR #681, opened 2018-09-19
Mark CMake project as C++-only

janisozaur:
This will make CMake skip all the checks for C compiler

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ foreach(p
   endif()
 endforeach()
 
-project (benchmark)
+project (benchmark CXX)
 
 option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmark library." ON)


LebedevRI:
At least on the surface, with cmake before 3.4.0-rc1, the `FindThreads.cmake` completely doesn't work if there isn't C language enabled.
Not sure if this is applicable here, or whether there are other failure points.

janisozaur:
Amended with suggested changes

janisozaur:
In about half a year when Trusty's support ends, will you consider switching CMake base version to that of Xenial's, 3.5, and removing the workaround included here?

LebedevRI:
See https://github.com/google/benchmark/pull/613, and more importantly, https://github.com/google/benchmark/issues/603#issuecomment-416928007.

LebedevRI:
Looks like the travis failure is relevant.

janisozaur:
This looks like a _fun_ one. When `project()` is set to `CXX`, the compiler seems to enable pthreads by default(?)[0][1] and doesn't require `-pthread` option. When I add `C`, like in case of pre-3.4 CMake, the C compiler is interrogated and it requires `-pthread`[2] which then gets set appropriately.

[0]
```diff
@@ -257,7 +257,10 @@ if (NOT BENCHMARK_ENABLE_EXCEPTIONS AND HAVE_STD_REGEX
 endif()
 cxx_feature_check(STEADY_CLOCK)
 # Ensure we have pthreads
+set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
 find_package(Threads REQUIRED)
+message("Threads: ${CMAKE_THREAD_LIBS_INIT}")
 
 # Set up directories
 include_directories(${PROJECT_SOURCE_DIR}/include)
```

[1]
```
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- Found Threads: TRUE  
Threads: 
```

[2]
```
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
Threads: -pthread
```

I haven't completely figured out why `pthread_create` is available in C++, but not in C or how to address this.

LebedevRI:
I think you want to try to drop `set(THREADS_PREFER_PTHREAD_FLAG TRUE)`.
https://cmake.org/cmake/help/v3.1/module/FindThreads.html

> If the use of the -pthread compiler and linker flag is prefered then the caller can set
> > THREADS_PREFER_PTHREAD_FLAG

> Please note that the compiler flag can only be used with the imported target. Use of both the imported target as well as this switch is highly recommended for new code.

janisozaur:
I tried it locally before and it didn't work. I'm still looking for ways to not hardcode `-pthread` and why it happens in the first place.

CMakeLists.txt
```diff
@@ -1,6 +1,14 @@
 cmake_minimum_required (VERSION 2.8.12)
 
-project (benchmark)
+if(NOT (CMAKE_VERSION VERSION_LESS 3.4.0-rc1))
+  project(benchmark CXX)
```
LebedevRI: This single line should be sufficient now

janisozaur:
Sure, I'll rebase it later this week

janisozaur:
Rebased

dominichamon:
@LebedevRI are you happy?

LebedevRI:
@janisozaur thank you!

PR #682, opened 2018-09-20
Add variables for specifying deployment destinations

dtoubelis:
Signed-off-by: Dmitri Toubelis <dmitri.toubelis@litmusautomation.com>

We are building `benchmark` using yocto build system. Our target deployment destinations are `/usr/lib64` for 64-bit libraries and `/usr/lib` for 32-bit ones. Currently, all deployment destinations are hardcoded. The proposed patch makes them configurable leaving defaults unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,12 @@ foreach(p
   endif()
 endforeach()
 
+set(BENCHMARK_INSTALL_BINDIR "bin" CACHE STRING "Installation directory for executables")
+set(BENCHMARK_INSTALL_INCLUDEDIR "include" CACHE STRING "Installation directory for headers")
+set(BENCHMARK_INSTALL_LIBDIR "lib" CACHE STRING "Installation directory for libraries")
+set(BENCHMARK_INSTALL_CMAKEDIR "lib/cmake/${PACKAGE_NAME}" CACHE STRING "Installation directory for cmake config files")
+set(BENCHMARK_INSTALL_PKGCONFIG "lib/pkgconfig" CACHE STRING "Installation directory for pkgconfig config files")
+
 option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmark library." ON)
 option(BENCHMARK_ENABLE_LTO "Enable link time optimisation of the benchmark library." OFF)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -55,12 +55,6 @@ target_include_directories(benchmark PUBLIC
     )
 target_link_libraries(benchmark_main benchmark)
 
-set(include_install_dir "include")
-set(lib_install_dir "lib/")
-set(bin_install_dir "bin/")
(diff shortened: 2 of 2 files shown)


dtoubelis:
GNUInstallDirs would be ideal approach. This is a quick workaround made out of despair. It gets the problem solved and you can replace it with proper implementation at your convenience.

LebedevRI:
@dominichamon thanks!
@dtoubelis do you feel like bumping required cmake version up to `v3.0.0` here, and porting straight to `GNUInstallDirs`? :)

dtoubelis:
I'm perfectly fine. It would be my preferred approach because yocto supports `GNUInstallDirs` out of the box and we use cmake 3.6.x anyway. So, feel free to close this PR if you have a better solution.

LebedevRI:
I meant in the scope of this pr, since you are making such a change, why not go for the Right solution (c)
Else someone else will have to basically repeat all this work afterwards.

dtoubelis:
Oh, I see :-) I only have few days of experience with cmake but I can give it a try.

src/CMakeLists.txt
```diff
@@ -83,26 +77,26 @@ if (BENCHMARK_ENABLE_INSTALL)
   install(
```
LebedevRI: I *think* you simply need to `include(GNUInstallDirs)`
And do s/`BENCHMARK`/`CMAKE`/ here
(and drop those old `BENCHMARK_INSTALL_*` variables)

CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ foreach(p
   endif()
 endforeach()
 
+set(BENCHMARK_INSTALL_BINDIR "bin" CACHE STRING "Installation directory for executables")
+set(BENCHMARK_INSTALL_INCLUDEDIR "include" CACHE STRING "Installation directory for headers")
+set(BENCHMARK_INSTALL_LIBDIR "lib" CACHE STRING "Installation directory for libraries")
+set(BENCHMARK_INSTALL_CMAKEDIR "lib/cmake/${PACKAGE_NAME}" CACHE STRING "Installation directory for cmake config files")
+set(BENCHMARK_INSTALL_PKGCONFIG "lib/pkgconfig" CACHE STRING "Installation directory for pkgconfig config files")
```
LebedevRI: I think `BENCHMARK_INSTALL_PKGCONFIG` is `${CMAKE_INSTALL_LIBDIR}/lib/pkgconfig`,
same with `BENCHMARK_INSTALL_CMAKEDIR`

dtoubelis:
It looks pretty straightforward but I also don't want to break you windows build. Something tells me that `GNUInstallDirs` may not work on Windows ;-)

LebedevRI:
> Something tells me that GNUInstallDirs may not work on Windows ;-)

I kind of doubt it. I think it is used elsewhere, with no special-handling for _other special os'es_

dtoubelis:
The easiest would be to keep `BENCHMARK_` variables but populate them depending on the environment. But I'll figure it out.

LebedevRI:
My sample count is small, but i literally never seen any issues like that.
I expect the dumb approach of simply using those `GNUInstallDirs`-listed variables to simply work.

LebedevRI:
Superseded by #808

PR #690, opened 2018-09-25
[NFC] Refactor RunBenchmark()

LebedevRI:
Ok, so, i'm still trying to get to the state when it will be a trivial change to report all the separate iterations.
The old code (LHS of the diff) was rather convoluted i'd say.
I have tried to refactor it a bit into *small* logical chunks, with proper comments.
As far as i can tell, i preserved the intent of the code, what it was doing before.
The road forward still isn't clear, but i'm quite sure it's not with the old code :)

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -114,6 +114,13 @@ DEFINE_int32(v, 0, "The level of verbose logging to output");
 
 namespace benchmark {
 
+namespace internal {
+
+// FIXME: wouldn't LTO mess this up?
+void UseCharPointer(char const volatile*) {}
+
+}  // namespace internal
+
 State::State(size_t max_iters, const std::vector<int64_t>& ranges, int thread_i,
              int n_threads, internal::ThreadTimer* timer,
              internal::ThreadManager* manager)
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -14,6 +14,7 @@
 
 #include "benchmark/benchmark.h"
 #include "benchmark_api_internal.h"
+#include "benchmark_runner.h"
 #include "internal_macros.h"
 
 #ifndef BENCHMARK_OS_WINDOWS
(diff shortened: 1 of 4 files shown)


src/benchmark.cc
```diff
@@ -198,140 +198,225 @@ struct RunResults {
...
+           i.iters >= kMaxIterations ||  // Too many iterations already.
+           i.seconds >= min_time ||      // The elapsed time is large enough.
+           // CPU time is specified but the elapsed real time greatly exceeds
+           // the minimum time.
+           // Note that user provided timers are except from this sanity check.
+           ((i.results.real_time_used >= 5 * min_time) && !b.use_manual_time);
   }
-  for (int repetition_num = 0; repetition_num < repeats; repetition_num++) {
+
+  void doOneRepetition(bool not_in_the_first_repetition) {
```
dominichamon: inverse booleans confuse me as i'm quite slow. can this be `bool first_repetition` instead (and obviously invert the logic everywhere)?

src/benchmark.cc
```diff
@@ -198,140 +198,225 @@ struct RunResults {
...
+      auto additional_run_stats = ComputeBigO(complexity_reports);
+      run_results.aggregates_only.insert(run_results.aggregates_only.end(),
+                                         additional_run_stats.begin(),
+                                         additional_run_stats.end());
+      complexity_reports.clear();
+    }
   }
 
-  return run_results;
+  RunResults getResults() { return run_results; }
```
dominichamon: I'm torn between this, where you have a simple accessor but the constructor is complex and actually doing the work, and an alternative where the constructor merely initialises the state, and a method `Run` that runs the benchmark and returns `RunResults`.

src/benchmark.cc
LebedevRI: The deeper problem here is that one more abstraction layer is missing.
This `not_in_the_first_repetition`/`first_repetition` should simply be `has_explicit_iteration_count`.

src/benchmark.cc
LebedevRI: Yes, i don't like this, too.
The current approach (complex constructor) seems better,
because we really wouldn't want to re-run re-use
the same `BenchmarkRunner` instance more than once.

dominichamon:
This seems to work really well.  I wonder if it's worth extracting the class into a benchmark_runner.cc implementation file to help reduce the overwhelming benchmark.cc file size.

LebedevRI:
> This seems to work really well. 

Well, it _should_ be the same algo, just less, _convoluted_?

> I wonder if it's worth extracting the class into a benchmark_runner.cc implementation file to help reduce the overwhelming benchmark.cc file size.

Done, better?

src/benchmark_runner.cc
```diff
@@ -0,0 +1,354 @@
...
+    results.real_time_used += timer.real_time_used();
+    results.manual_time_used += timer.manual_time_used();
+    results.complexity_n += st.complexity_length_n();
+    internal::Increment(&results.counters, st.counters);
+  }
+  manager->NotifyThreadComplete();
+}
+
+class BenchmarkRunner {
+  RunResults run_results;
```
dominichamon: nit: can you put the public first and have an explicit private: section please?

src/benchmark_api_internal.h
```diff
@@ -2,13 +2,22 @@
...
 #include "benchmark/benchmark.h"
+#include "commandlineflags.h"
 
 #include <cmath>
 #include <iosfwd>
 #include <limits>
 #include <string>
 #include <vector>
 
+DECLARE_double(benchmark_min_time);
```
dominichamon: why here and not benchmark_runner.h?

LebedevRI: Ah, putting them into `benchmark_runner.h` works too.
But specifically into header, *not* the `.cc`!

LebedevRI:
Anything more to do here?

From my side, all the follow-up work in `benchmark_runner.cc` will be in follow-up pr,
once i'm aware what it is..

src/benchmark_runner.cc
```diff
@@ -0,0 +1,353 @@
...
+    if ((b.complexity != oNone) && b.last_benchmark_instance) {
+      auto additional_run_stats = ComputeBigO(complexity_reports);
+      run_results.aggregates_only.insert(run_results.aggregates_only.end(),
+                                         additional_run_stats.begin(),
+                                         additional_run_stats.end());
+      complexity_reports.clear();
+    }
+  }
+
+  RunResults getResults() { return run_results; }
```
dominichamon: google-style: `get_results()`

also, mark the method as const?

src/benchmark_runner.cc
```diff
@@ -0,0 +1,353 @@
...
+  size_t iters;  // preserved between repetitions!
+  // So only the first repetition has to find/calculate it,
+  // the other repetitions will just use that precomputed iteration count.
+
+  struct IterationResults {
+    internal::ThreadManager::Result results;
+    size_t iters;
+    double seconds;
+  };
+  IterationResults doNIterations() {
```
dominichamon: `DoNIterations`

src/benchmark_runner.cc
```diff
@@ -0,0 +1,353 @@
...
+    if (b.use_manual_time) {
+      i.seconds = i.results.manual_time_used;
+    } else if (b.use_real_time) {
+      i.seconds = i.results.real_time_used;
+    }
+
+    return i;
+  }
+
+  size_t predictNumItersNeeded(const IterationResults& i) const {
```
dominichamon: `PredictNumItersNeeded`

src/benchmark_runner.cc
```diff
@@ -0,0 +1,353 @@
...
+    const size_t max_next_iters =
+        0.5 + std::max(multiplier * i.iters, i.iters + 1.0);
+    // But we do have *some* sanity limits though..
+    const size_t next_iters = std::min(max_next_iters, kMaxIterations);
+
+    VLOG(3) << "Next iters: " << next_iters << ", " << multiplier << "\n";
+    return next_iters;  // round up before conversion to integer.
+  }
+
+  bool shouldReportIterationResults(const IterationResults& i) const {
```
dominichamon: `ShouldReportIterationResults`

src/benchmark_runner.cc
```diff
@@ -0,0 +1,353 @@
...
+    return i.results.has_error_ ||
+           i.iters >= kMaxIterations ||  // Too many iterations already.
+           i.seconds >= min_time ||      // The elapsed time is large enough.
+           // CPU time is specified but the elapsed real time greatly exceeds
+           // the minimum time.
+           // Note that user provided timers are except from this sanity check.
+           ((i.results.real_time_used >= 5 * min_time) && !b.use_manual_time);
+  }
+
+  void doOneRepetition(bool is_the_first_repetition) {
```
dominichamon: `DoOneRepetition`

src/benchmark_runner.cc
LebedevRI: I think even the current non-`const` variant is not what i wanted..
https://godbolt.org/z/7ALmcF

src/benchmark_runner.cc
```diff
@@ -170,7 +170,7 @@ class BenchmarkRunner {
     }
   }
 
-  RunResults getResults() { return run_results; }
+  RunResults&& get_results() { return std::move(run_results); }
```
dominichamon: nice catch :) that's much better.

LebedevRI:
Thank you for the review!

PR #693, opened 2018-09-29
Unbreak -Wthread-safety for -DCMAKE_BUILD_TYPE=Release on platforms that annotate <pthread.h>

### LINKED ISSUE #655: Build fails: error: mutex 'mutex_' is still held at the end of function
```
n file included from /usr/local/include/gtest/internal/gtest-linked_ptr.h:74:
/usr/local/include/gtest/internal/gtest-port.h:2049:3: error: mutex 'mutex_' is still held at the end of function [-Werror,-Wthread-safety-analysis]
  }
  ^
/usr/local/include/gtest/internal/gtest-port.h:2046:32: note: mutex acquired here
    GTEST_CHECK_POSIX_SUCCESS_(pthread_mutex_lock(&mutex_));
                               ^
/usr/local/include/gtest/internal/gtest-port.h:2058:32: error: releasing mutex 'mutex_' that was not held [-Werror,-Wthread-safety-analysis]
    GTEST_CHECK_POSIX_SUCCESS_(pthread_mutex_unlock(&mutex_));
                               ^
2 errors generated.
```

jbeich:
Fixes #655. `-Werror` shouldn't be forced for release builds, anyway.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -179,7 +179,7 @@ if (BENCHMARK_ENABLE_GTEST_TESTS)
       add_dependencies(${name} googletest)
     endif()
     if (GTEST_INCLUDE_DIRS)
-      target_include_directories(${name} PRIVATE ${GTEST_INCLUDE_DIRS})
+      target_include_directories(${name} SYSTEM PRIVATE ${GTEST_INCLUDE_DIRS})
     endif()
     target_link_libraries(${name} benchmark
         ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})


LebedevRI:
It only hides the issue though.
It should be fixed upstream in google test.

jbeich:
Not so simple:
- Users with old system googletest would be still affected
- `-Wthread-safety` isn't used/supported by googletest
- googletest upstream doesn't care much about FreeBSD

jbeich:
@yurivict, I'm putting this patch under public domain. Can you resubmit it and sign CLA (if not already)? CC'ing @FreeBSDFoundation in case it can act as a proxy for FreeBSD contributors that cannot sign up for Google Account due to privacy concerns.

LebedevRI:
Well, you were able to write this patch, right?
Surely, you could submit the proper fix for google test too?

jbeich:
> Surely, you could submit the proper fix for google test too?

What is the proper fix? Something like llvm-mirror/libcxx@362353aa2d01? Annotating googletest locks properly is non-trivial.

Besides, CLA is still an issue for Google (not me). I had no trouble signing Facebook CLA because it didn't require something as stupid as registering a Facebook Account... just to submit a one-off patch.

LebedevRI:
> What is the proper fix? Something like llvm-mirror/libcxx@362353a? Annotating googletest locks properly is non-trivial.

No idea what they will consider the proper fix is, it's in gtest, not this repo.

This just feels super weird.

FreeBSD went all that trouble to annotate their pthread headers,
that is paying off in terms of that failure you stated in the commit msg,
and **instead** of fixing the source of that failure, you will want to silence **every**
project out there that uses google test this way?

If you don't care about the warning, either un-annotate your pthreads,
or pass `-Wno-error=thread-safety`,
or apply this patch in the googlebenchmark package in freebsd.

jbeich:
> will want to silence every project out there that uses google test this way?

Why do you want googletest (dependency) warnings to affect benchmark (this project)? `SYSTEM` converts `-I` into `-isystem` which makes the include path act like `/usr/include` where fixing warnings is not always a viable option. Let googletest handle `-Wthread-safety` issue at their own pace.

Other projects don't force `-Werror` for release builds because there's no guarantee newer compiler versions won't introduce more warnings or if users won't try to compile on not officially supported platforms. Not to mention, `-Werror=thread-safety` only became an issue on FreeBSD because it [cannot make up its mind](https://wiki.freebsd.org/WarnerLosh/UsrLocal) about whether `/usr/local` is a system path or not. Most Linux distributions implicitly mark the path to where googletest is installed as `SYSTEM`.

LebedevRI:
> Why do you want googletest (dependency) warnings to affect benchmark (this project)? SYSTEM converts -I into -isystem which makes the include path act like /usr/include where fixing warnings is not always a viable option.

Now that is indeed a much better question.
I do know what `SYSTEM` does, and i personally do think it should be used here.
I *think* (i hope so? can't find the discussion), i did suggest that, but clearly that wasn't the outcome.. :/

LebedevRI:
This seems to have gotten stuck, and i'm not sure it can unstuck.
Let's close this for now..

PR #694, opened 2018-09-30
Fix possible loss of data warnings in MSVC

pwnall:
Warning texts:
benchmark.h(558): warning C4244: 'argument': conversion from 'int64_t' to 'double', possible loss of data
benchmark.h(564): warning C4244: 'return': conversion from 'const double' to 'int64_t', possible loss of data
benchmark.h(587): warning C4244: 'argument': conversion from 'int64_t' to 'double', possible loss of data
benchmark.h(593): warning C4244: 'return': conversion from 'const double' to 'int64_t', possible loss of data

Thank you very much for maintaining google benchmark!

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -555,13 +555,13 @@ class State {
   BENCHMARK_ALWAYS_INLINE
   void SetBytesProcessed(int64_t bytes) {
     counters["bytes_per_second"] =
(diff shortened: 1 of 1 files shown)


dominichamon:
Thank you!

pwnall:
Thank you very much for the quick review!

LebedevRI:
Maybe in 2.0 we could change the signatures of those functions to avoid casts.

PR #697, opened 2018-10-05
Fix Clang Detection

Lord-Kamina:
For several versions now, CMake by default refers to macOS’ Clang as AppleClang instead of just Clang, which would fail STREQUAL. Fixed by changing it to MATCHES.


Lord-Kamina:
CLA signed.

dominichamon:
Thanks!

PR #699, opened 2018-10-05
benchmark_color: fix auto option

### LINKED ISSUE #559: --benchmark_color=auto (the default value) always prints color
Benchmark defaults to "auto" for deciding to whether to use color or not, as part of the [following conditional:](https://github.com/google/benchmark/blob/7b03df7ff76844a39359e9233f31ca8cdb073313/src/benchmark.cc#L466-L467) 
```cpp
  if ((FLAGS_benchmark_color == "auto" && IsColorTerminal()) ||
      IsTruthyFlagValue(FLAGS_benchmark_color)) {
    // print color.
}
```

But `IsTruthyFlagValue` is permissive and returns `true` for the string "auto", causing `IsColorTerminal()` to be ignored.

iillyyaa:
As prevously written, "--benchmark_color=auto" was treated as true,
because IsTruthyFlagValue("auto") returned true.  The fix is to
rely on IsColorTerminal test only if the flag value is "auto",
and fall back to IsTruthyFlagValue otherwise.  I also integrated
force_no_color check into the same block.

(Aside for @dominichamon: I noticed that you mentioned in the ticket that a PR was in progress, but I couldn't locate it.)

Fixes #559.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -312,13 +312,18 @@ bool IsZero(double n) {
 
 ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
   int output_opts = ConsoleReporter::OO_Defaults;
-  if ((FLAGS_benchmark_color == "auto" && IsColorTerminal()) ||
-      IsTruthyFlagValue(FLAGS_benchmark_color)) {
+  auto is_benchmark_color = [force_no_color] () -> bool {
+    if (force_no_color) {
+      return false;
+    }
+    if (FLAGS_benchmark_color == "auto") {
+      return IsColorTerminal();
+    }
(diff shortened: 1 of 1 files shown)


LebedevRI:
Looks reasonable, but is there any reasonable way to test this?

iillyyaa:
Roman, I'm sure there is a way, but my cmake-fu is not up to snuff to come up with it. I'd appreciate if you accepted the fix as is. If tests must be provided, I will do my best, but can't make promises on the timing of the update.

dominichamon:
i don't think cmake is required for tests, it would be a googletest unit test of the function, but you may have to use gmock to mock out IsColorTerminal... I'm ok with that waiting for another time.

PR #701, opened 2018-10-10
Print at least three significant digits for times.

EricWF:
Some benchmarks are particularly sensitive and they run in less than
a nanosecond. In order for the console reporter to provide meaningful
output for such benchmarks it needs to be able to display the times
using more resolution than a single nanosecond.

This patch changes the console reporter to print at least three
significant digits for all results.

### DIFF
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -53,7 +53,7 @@ bool ConsoleReporter::ReportContext(const Context& context) {
 }
 
 void ConsoleReporter::PrintHeader(const Run& run) {
-  std::string str = FormatString("%-*s %13s %13s %10s", static_cast<int>(name_field_width_),
+  std::string str = FormatString("%-*s %13s %15s %12s", static_cast<int>(name_field_width_),
                                  "Benchmark", "Time", "CPU", "Iterations");
   if(!run.counters.empty()) {
     if(output_options_ & OO_Tabular) {
@@ -98,6 +98,21 @@ static void IgnoreColorPrint(std::ostream& out, LogColor, const char* fmt,
   va_end(args);
 }
 
+
+static std::string FormatTime(double time) {
+  // Align decimal places...
+  if (time < 1.0) {
+    return FormatString("%10.3f", time);
+  }
+  if (time < 10.0) {
+    return FormatString("%10.2f", time);
+  }
+  if (time < 100.0) {
+    return FormatString("%10.1f", time);
+  }
+  return FormatString("%10.0f", time);
+}
+
 void ConsoleReporter::PrintRunData(const Run& result) {
   typedef void(PrinterFn)(std::ostream&, LogColor, const char*, ...);
   auto& Out = GetOutputStream();
@@ -117,18 +132,21 @@ void ConsoleReporter::PrintRunData(const Run& result) {
 
   const double real_time = result.GetAdjustedRealTime();
   const double cpu_time = result.GetAdjustedCPUTime();
+  const std::string real_time_str = FormatTime(real_time);
+  const std::string cpu_time_str = FormatTime(cpu_time);
+
 
   if (result.report_big_o) {
     std::string big_o = GetBigOString(result.complexity);
-    printer(Out, COLOR_YELLOW, "%10.2f %s %10.2f %s ", real_time, big_o.c_str(),
+    printer(Out, COLOR_YELLOW, "%10.2f %-4s %10.2f %-4s ", real_time, big_o.c_str(),
             cpu_time, big_o.c_str());
   } else if (result.report_rms) {
-    printer(Out, COLOR_YELLOW, "%10.0f %% %10.0f %% ", real_time * 100,
-            cpu_time * 100);
+    printer(Out, COLOR_YELLOW, "%10.0f %-4s %10.0f %-4s ", real_time * 100, "%",
+            cpu_time * 100, "%");
   } else {
     const char* timeLabel = GetTimeUnitString(result.time_unit);
-    printer(Out, COLOR_YELLOW, "%10.0f %s %10.0f %s ", real_time, timeLabel,
-            cpu_time, timeLabel);
+    printer(Out, COLOR_YELLOW, "%s %-4s %s %-4s ", real_time_str.c_str(), timeLabel,
+            cpu_time_str.c_str(), timeLabel);
   }
 
   if (!result.report_big_o && !result.report_rms) {
--- a/test/reporter_output_test.cc
+++ b/test/reporter_output_test.cc
@@ -521,7 +521,7 @@ ADD_CASES(TC_ConsoleOut, {{"^BM_UserStats/iterations:5/repeats:3/manual_time [ "
                           {"^BM_UserStats/iterations:5/repeats:3/"
                            "manual_time_median [ ]* 150 ns %time [ ]*3$"},
                           {"^BM_UserStats/iterations:5/repeats:3/"
-                           "manual_time_stddev [ ]* 0 ns %time [ ]*3$"},
+                           "manual_time_stddev [ ]* 0.000 ns %time [ ]*3$"},
                           {"^BM_UserStats/iterations:5/repeats:3/manual_time_ "
                            "[ ]* 150 ns %time [ ]*3$"}});
 ADD_CASES(
--- a/test/output_test_helper.cc
+++ b/test/output_test_helper.cc
@@ -38,17 +38,18 @@ SubMap& GetSubstitutions() {
   // Don't use 'dec_re' from header because it may not yet be initialized.
   // clang-format off
   static std::string safe_dec_re = "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?";
+  static std::string time_re = "([0-9]+[.])?[0-9]+";
   static SubMap map = {
       {"%float", "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?"},
       // human-readable float
       {"%hrfloat", "[0-9]*[.]?[0-9]+([eE][-+][0-9]+)?[kMGTPEZYmunpfazy]?"},
       {"%int", "[ ]*[0-9]+"},
       {" %s ", "[ ]+"},
-      {"%time", "[ ]*[0-9]+ ns"},
-      {"%console_report", "[ ]*[0-9]+ ns [ ]*[0-9]+ ns [ ]*[0-9]+"},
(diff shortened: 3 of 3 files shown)


LebedevRI:
I wonder if it would be good to introduce something like a scientific notation "time unit" (i.e. autodetect each time)

src/console_reporter.cc
```diff
@@ -98,6 +98,20 @@ static void IgnoreColorPrint(std::ostream& out, LogColor, const char* fmt,
   va_end(args);
 }
 
+static std::string FormatTime(double time) {
+  // Align decimal places...
+  if (time < 1) {
```
LebedevRI: Nit: `1` is `int`, `time` is `double`.

src/console_reporter.cc
```diff
@@ -117,22 +131,24 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
     const char* timeLabel = GetTimeUnitString(result.time_unit);
-    printer(Out, COLOR_YELLOW, "%10.0f %s %10.0f %s ", real_time, timeLabel,
-            cpu_time, timeLabel);
+    printer(Out, COLOR_YELLOW, "%s %-4s %s %-4s ", real_time_str.c_str(), timeLabel,
+            cpu_time_str.c_str(), timeLabel);
   }
 
   if (!result.report_big_o && !result.report_rms) {
-    printer(Out, COLOR_CYAN, "%10lld", result.iterations);
+    printer(Out, COLOR_CYAN, "%12lld", result.iterations);
```
LebedevRI: Hm, why change the `iteration` print format?

src/console_reporter.cc
```diff
@@ -98,6 +98,20 @@ static void IgnoreColorPrint(std::ostream& out, LogColor, const char* fmt,
...
+  if (time < 1) {
+    return FormatString("%14.3f", time);
+  }
+  if (time < 10) {
+    return FormatString("%13.2f ", time);
+  }
+  if (time < 100) {
+    return FormatString("%12.1f  ", time);
+  }
+  return FormatString("%10.0f    ", time);
```
LebedevRI: Have you considered letting the format string figure out the necessary padding at least?
(see `man 3 vsprintf`)

src/console_reporter.cc
EricWF: Ack. Fixed.

src/console_reporter.cc
EricWF: Could you clarify a bit?

The reason for the current implementation of `FormatString` is to ensure columns with the same place value align. For example.
```
Benchmark                                                Time                CPU         Iterations
----------------------------------------------------------------------------------------------------
BM_empty                                                0.572 ns            0.572 ns     1000000000
BM_empty/threads:72                                     0.021 ns            1.50  ns      958819032
BM_spin_empty/8                                         7.47  ns            7.47  ns      180868971
BM_spin_empty/512                                     954     ns          953     ns        1466617
BM_spin_empty/8192                                  15449     ns        15447     ns          90954
BM_spin_empty/8/threads:72                              0.258 ns           18.2   ns       72560520
BM_spin_empty/512/threads:72                           22.2   ns         1564     ns         872568
BM_spin_empty/8192/threads:72                         382     ns        26233     ns          38664
```

I'm not too familiar with printf specifiers, but if there is a tool I can use to do that, please let me know.

src/console_reporter.cc
EricWF: To align it with the `Iterations` header.

LebedevRI:
I **like** this, but that padding/alignment is really bothering me :/
image
image

How important is it to do that alignment of `.` separator? Can it be avoided?

EricWF:
> How important is it to do that alignment of `.` separator? Can it be avoided?

I think it's pretty important. It's allows you to understand the magnitude of a benchmark by glancing. 

What's you're objection exactly? To readability? Do you think the space between the value and `ns` is ugly?

src/console_reporter.cc
EricWF: Yeah, I don't love the extra width either. But being able to see sub-nanosecond results is worth the hit I think.

LebedevRI:
> > How important is it to do that alignment of `.` separator? Can it be avoided?
> 
> I think it's pretty important.

>It's allows you to understand the magnitude of a benchmark by glancing.
True.
 
> What's you're objection exactly? To readability? Do you think the space between the value and `ns` is ugly?

Readability is good; but yes, i don't like that wasted space, and the yet-increased width of the user-counter-less line.

EricWF:
> > > How important is it to do that alignment of `.` separator? Can it be avoided?
> > 
> > 
> > I think it's pretty important.
> 
> > It's allows you to understand the magnitude of a benchmark by glancing.
> 
> True.
> 
> > What's you're objection exactly? To readability? Do you think the space between the value and `ns` is ugly?
> 
> Readability is good; but yes, i don't like that wasted space, and the yet-increased width of the user-counter-less line.

Arguabbly the space isn't "wasted", since I think we agree it has value in representing magnitude. Otherwise we would just remove it. But it's not ideal either.

I can get rid of a little width (4 characters) if we don't care about making BigO calculations aligned with everything else, but that's ugly and doesn't help much.

Do you think this issue should block the revision?

LebedevRI:
> Do you think this issue should block the revision?

There are two things here as far as i can see:
1. Always print at least 3 significant digits
2. Align/pad `.`

I ***like*** 1., but i'm not quite sure about 2. yet.
So i guess i'll defer to @dominichamon.

EricWF:
A couple of alternative solutions:

1. Add a flag to change the alignment to always be right aligned to the unit, which should waste less space. But IMO getting the old behavior back isn't worth the cost of another flag.

2. Use the "wasted space" explicitly so it seems "less wasted". That is, fill the wasted space with three trailing decimal digits for all outputs. I don't like this either because those digits aren't significant and make it harder to see the information that is.

3. Just right align everything all the time. This means it's a lot harder for the user to determine the magnitude of a benchmark, and ever harder yet to compare two benchmarks next to each other.

LebedevRI:
> A couple of alternative solutions:

> 1. Add a flag to change the alignment to always be right aligned to the unit, which should waste less space. But IMO getting the old behavior back isn't worth the cost of another flag.

I agree, a flag for this does not sound good.

> 2. Use the "wasted space" explicitly so it seems "less wasted". That is, fill the wasted space with three trailing decimal digits for all outputs. I don't like this either because those digits aren't significant and make it harder to see the information that is.

And what is even worse, those 'zeros' are likely straight-up a lie.

> 3. Just right align everything all the time. This means it's a lot harder for the user to determine the magnitude of a benchmark, and ever harder yet to compare two benchmarks next to each other.

I would personally go with 3., but it is also possible that i'm simply too picky here.
(Though it is undeniable that the new layout is more width-consuming.)

As i said, i'll defer to @dominichamon..

LebedevRI:
> (Though it is undeniable that the new layout is more width-consuming.)

(I.e. if one day i/someone else "finishes" support for custom timers, and a new time column can be arbitrarily added, the width will be in short supply)

dominichamon:
What's wrong with always printing 3 decimal places instead of 3 significant digits, and not having the check for `< 100`?

LebedevRI:
If the time is `52321 ms`, does one care if it is `52321.000 ms` or `52321.666 ms` ?
I'd think that goes against the reason for setting the time units in the first place..

dominichamon:
oh you updated that from .666 so i wouldn't complain about rounding :P

I think we're hitting newer use cases as this project gets broader adoption. In general, the library was expected to be used to run benchmarks with similar timescales, so you'd get ns or ms all the way down. If we're not seeing that then we need to think a bit about how to handle it.

One option is to pick a base time scale for the entire run and use that everywhere. I'm not sure that's a great idea, but it would at least make eyeballing the runs easy. When you mix timescales you need to take care that, say, 10ms and 100ns are not being mixed up by the reader.

High-level answer: I'm not a fan of the extra whitespace between the numbers and the units.

LebedevRI:
> oh you updated that from .666 so i wouldn't complain about rounding :P

:P

> High-level answer: I'm not a fan of the extra whitespace between the numbers and the units.

Any opinion on simply not having those whitespaces, will that also be confusing?

dominichamon:
I think it would be. You'd have

```
100 ns
100 us
```

or worse

```
100 ns
 10 us
```

and would have to take care if you're comparing by eye. Perhaps we should have something like: one time scale per benchmark family?

LebedevRI:
That is not what i meant. I was literally talking about the current code, but without that padding with spaces.
I.e.
```
...
100 ms
 0.1 ms
    1 ms
...
```

dominichamon:
If it doesn't change time unit i think that's ok.

EricWF:
I've updated the patch to no longer align the decimal point as requested.

src/console_reporter.cc
```diff
@@ -98,6 +98,21 @@ static void IgnoreColorPrint(std::ostream& out, LogColor, const char* fmt,
   va_end(args);
 }
 
+
+static std::string FormatTime(double time) {
+  // Align decimal places...
```
LebedevRI: Comment outdated

test/reporter_output_test.cc
```diff
@@ -521,7 +521,7 @@ ADD_CASES(TC_ConsoleOut, {{"^BM_UserStats/iterations:5/repeats:3/manual_time [ "
                           {"^BM_UserStats/iterations:5/repeats:3/"
                            "manual_time_median [ ]* 150 ns %time [ ]*3$"},
                           {"^BM_UserStats/iterations:5/repeats:3/"
-                           "manual_time_stddev [ ]* 0 ns %time [ ]*3$"},
+                           "manual_time_stddev [ ]* 0.000 ns %time [ ]*3$"},
```
LebedevRI: Should the zero be treated differently?

dominichamon:
LGTM. I ran it and master and took a diff and it looks much nicer with the extra precision.

src/console_reporter.cc
```diff
@@ -53,7 +53,7 @@ bool ConsoleReporter::ReportContext(const Context& context) {
 }
 
 void ConsoleReporter::PrintHeader(const Run& run) {
-  std::string str = FormatString("%-*s %13s %13s %10s", static_cast<int>(name_field_width_),
+  std::string str = FormatString("%-*s %13s %15s %12s", static_cast<int>(name_field_width_),
```
LebedevRI: I'm not sure, is this still needed now that no alignment happens?

EricWF: Yes. There was a bug originally where we misaligned the last couple fields of the header when complexity was involved.

PR #702, opened 2018-10-11
Apply the `real_time_used >= 5 * min_time` check to manual time as well.

timshen91:
In the commit message of 46afd8e6, I found out the following:
> However when a benchmark uses a manual timer this heuristic isn't helpful
> and likely isn't correct since we don't know what the manual timer actually
> measures.

I think the heuristic is helpful in exactly the same way for manual timer.

In my particular use case with the manual timer, the benchmarks are GPU
benchmarks, and the manual timer reports GPU time. In a certain benchmark,
the real time elapsed 257x more than the manual time. As a result,
ShouldReportIterationResults takes 257x of time to return true.

I think it's reasonable to assume that all users want a benchmark to finish
in a fixed amount of real time for good user experience, regardless of the
real_time / manual_time ratio.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -273,10 +273,8 @@ class BenchmarkRunner {
     return i.results.has_error_ ||
            i.iters >= kMaxIterations ||  // Too many iterations already.
            i.seconds >= min_time ||      // The elapsed time is large enough.
-           // CPU time is specified but the elapsed real time greatly exceeds
-           // the minimum time.
-           // Note that user provided timers are except from this sanity check.
-           ((i.results.real_time_used >= 5 * min_time) && !b.use_manual_time);
+           // The elapsed real time greatly exceeds the minimum time.
+           i.results.real_time_used >= 5 * min_time;
   }
 
   void DoOneRepetition(bool is_the_first_repetition) {


LebedevRI:
*if* the manual 'time' is actually measuring time, in the same scale/units as the `min_time`
That isn't a requirement.

LebedevRI:
Actually, more reasonable question: why can't you just adjust the `min_time` for your benchmark?
https://github.com/google/benchmark/blob/d8c0f27448dfad95b94285e612bba1f7c55c9dd0/include/benchmark/benchmark.h#L850-L853

timshen91:
> if the manual 'time' is actually measuring time, in the same scale/units as the min_time
That isn't a requirement.

Ah yes, one shouldn't always compare real_time with min_time. What I mean is to compare real_time with min_real_time. There isn't a min_real_time, but I hope that there is a way to get it.

> Actually, more reasonable question: why can't you just adjust the min_time for your benchmark?

For a single benchmark, I can do that. I can take a look at --v=3, observe that in order to run for roughly 0.5 seconds real time, I need to set MinTime(0.00194). But it doesn't scale, as for every benchmark the MinTime() argument is going to be different, and the argument may not be stable across library upgrades.

dominichamon:
@LebedevRI is there a concern you have with removing the check for manual time? This might lead to results being reported when not enough iterations have been run, but then users can increase the minimum time using the `MinTime` function. However, it will result in manually timed benchmarks completing in a more reasonable time without waiting for all the iterations to run. That seems like a win to me...

LebedevRI:
I don't have a *specific* concern, i'm only noting that this PR is a direct revert of 46afd8e69339b546526706056da9dd5009fa01f1.
That change did not state which actual problem it solved, but since it's a direct revert,
i'm guessing that whatever that problem was, it will regress back.

So either 46afd8e69339b546526706056da9dd5009fa01f1 was an incorrect fix, or this is...

timshen91:
@EricWF , do you still remember the motivation behind 46afd8e? Do you have any solutions on solving the problem described in this patch description?

dominichamon:
ping @EricWF

LebedevRI:
My take is that it isn't stictly required for manual time to be measuring time.
It may be time, but might as well be energy consumed, or parrot count.
I suppose this could be viewed as yet another problem that is due to the lack of custom "timers".

timshen91:
> My take is that it isn't stictly required for manual time to be measuring time.
> It may be time, but might as well be energy consumed, or parrot count.
> I suppose this could be viewed as yet another problem that is due to the lack of custom "timers".

That's a fair point, but the code base is still contradicting even with such assumption. Let me try to point out it:
* [1] tells us that `i.seconds` stores either manual time or cpu time.
* [2] compares `i.seconds` and `min_time`. We should assume that `i.seconds` and `min_time` are both manual time or both cpu time, otherwise the comparison is meaningless.
* [3] tells us that `min_time` may be initialized from FLAGS_benchmark_min_time, which is cpu time. This invalids [2]'s assumption when FLAGS_benchmark_min_time is used.

All three create a contradiction. The symptom is that FLAGS_benchmark_min_time is mistreated as manual time, while its value is based on our common experience with cpu time. In my specific cast, the value is too large and caused an excessively long run.

I admit that my commit is hacky and doesn't nicely break one of the three assumptions. However, I can't find the reproducing benchmark anymore, and I won't be able to test whatever better solution I come up with.

For posterity, I think the better solution would be changing [2] to `x >= min_time`, where x is always cpu time.

[1] https://github.com/google/benchmark/blob/7d856b03047e7bf09ccc9e9878e299493541b468/src/benchmark_runner.cc#L243
[2] https://github.com/google/benchmark/blob/7d856b03047e7bf09ccc9e9878e299493541b468/src/benchmark_runner.cc#L280
[3] https://github.com/google/benchmark/blob/7d856b03047e7bf09ccc9e9878e299493541b468/src/benchmark_runner.cc#L140

PR #705, opened 2018-10-13
Fix SOURCE_DIR in HandleGTest.cmake

olzhabay:
If benchmark added as cmake subproject, HandleGTest throws an error as  does return absolute source dir.
Change it to , so it will be refering to it's own source dir.


EricWF:
@olzhabay Sorry about the mixup. Thanks for the patch!

olzhabay:
@EricWF my pleasure :)

PR #706, opened 2018-10-17
Aggregates: use non-aggregate count as iteration count.

### LINKED ISSUE #586: Statistics: iteration count
Right now each statistic contains the same iteration count as the first repetition.
https://github.com/google/benchmark/blob/50ffc781b1df6d582d6a453b9942cc8cb512db69/src/statistics.cc#L156
https://github.com/google/benchmark/blob/50ffc781b1df6d582d6a453b9942cc8cb512db69/src/statistics.cc#L128
https://github.com/google/benchmark/blob/50ffc781b1df6d582d6a453b9942cc8cb512db69/src/statistics.cc#L105
Are we *sure* this is the correct value that we should be outputting?
We don't actually store each of the iteration times, but average them, and operate "on averages".
Are we sure we don't want to put the repetition count there?

I'm currently looking into finally adding t-test stuff into tools, and thus thinking about "so what is the actual number of observations"? (and i guess the answer is - the repetition count.)

LebedevRI:
It is incorrect to say that an aggregate is computed over
run's iterations, because those iterations already got averaged.
Similarly, if there are N repetitions with 1 iterations each,
an aggregate will be computed over N measurements, not 1.
Thus it is best to simply use the count of separate reports.

Fixes #586.


dominichamon:
Can you provide a screenshot/link to the resulting output before and after so it's capture in the PR?

LebedevRI:
The tests convey that i think..
image
image

LebedevRI:
Hm, that is clearly broken.

LebedevRI:
Ok, that was a fun issue :) Test added.
Proper comparison:
image
image

LebedevRI:
Yay, thank you for the review!

PR #707, opened 2018-10-17
[NFC] RunBenchmarks(): s/has_repetitions/might_have_aggregates/

LebedevRI:
That is the real purpose of that bool. A follow-up change will
make it consider something else other than repetitions.


LebedevRI:
Thank you for the review!

PR #708, opened 2018-10-17
[NFC] BenchmarkRunner: always populate *_report_aggregates_only bools.

LebedevRI:
It is better to let the RunBenchmarks(), report() decide
whether to actually *only* output aggregates or not,
depending on whether there are actually aggregates.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -267,8 +267,8 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
       auto report = [&run_results](BenchmarkReporter* reporter,
                                    bool report_aggregates_only) {
         assert(reporter);
-        assert(
-            !(report_aggregates_only && run_results.aggregates_only.empty()));
+        // If there are no aggregates, do output non-aggregates.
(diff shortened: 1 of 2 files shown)


dominichamon:
I'm missing the subtlety here. why is it ok to remove the results count check?

LebedevRI:
It's subtle indeed.

Previously, `BenchmarkRunner()` always said that "if there are no repetitions,
then you should never output only the repetitions". And the `report()` simply assumed
that the `report_aggregates_only` bool it received makes sense, and simply used it.

Now, the logic is the same, but the blame has shifted.
`BenchmarkRunner()` always propagates what those benchmarks would have wanted
to happen wrt the aggregates. And the `report()` lambda has to actually consider
both the `report_aggregates_only` bool, and it's meaningfulness.

To put it in the context of the patch series - if the repetition count was `1`,
but `*_report_aggregates_only` was set to `true`, and we capture each iteration separately,
then we will compute the aggregates, but then output everything, both the iteration,
and aggregates, despite *_report_aggregates_only` being set to `true`.

Did that make sense?

LebedevRI:
Thank you for the review!

PR #709, opened 2018-10-17
Refactor BenchmarkRunner to be somewhat more modular.

LebedevRI:
I realize this will be a bit problematic to review, since there is no second part
of the change here, which actually makes use of this refactor,
but there is really no convient way to do that on github.

The problem here is, `BenchmarkRunner` is rather monolithic.
It actually runs the iterations, it keeps the state, it predicts the iteration count, etc.
To be able to record all the iterations separately, there needs to be a way to
adjust the behavior of `BenchmarkRunner`, without fully duplicating it.

In particular, the following points/considerations exist/are adressed:
* There shall be only one `BenchmarkRunner`, copy-paste is not the solution here.
* `BenchmarkRunner` needs to be the same, but we want for it to do different things.
* Thus, there should be some support class, that will take over the bookkeeping (of the iterations)
* That class shall be injectable via templates.
* Also, current `DoNIterations()` does actually what the name suggests - it runs `n` iterations at once.
  But we want to "basically" do `n` repetitions with `1` iteration each.
  But we can't do that, because naturally, iteration count is auto-adjusted.
* Thus, `DoNIterations()` needs to be overridable, so some other logic can be implemented ontop.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "benchmark_runner.h"
+
 #include "benchmark/benchmark.h"
 #include "benchmark_api_internal.h"
 #include "internal_macros.h"
@@ -132,18 +133,68 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
   manager->NotifyThreadComplete();
 }
 
-class BenchmarkRunner {
+class BenchmarkRunner;
+
+class AbstractIterationResultCollector {
+ public:
+  AbstractIterationResultCollector() = default;
+
+  virtual ~AbstractIterationResultCollector() = default;
+
+ protected:
+  std::vector<ThreadManager::Result> results{1};
+
+ private:
+
+  friend class BenchmarkRunner;
+
+  virtual bool HasError() const = 0;
+  virtual IterationCount Iters() const = 0;
+  virtual double ManualTimeUsed() const = 0;
+  virtual double RealTimeUsed() const = 0;
+  virtual double CpuTimeUsed() const = 0;
+
+  double Seconds(const BenchmarkInstance& b) const {
+    // The benchmark may have requested manual or real time be used
+    // in decision making. Respect that.
+    if (b.use_manual_time) {
+      return ManualTimeUsed();
+    } else if (b.use_real_time) {
+      return RealTimeUsed();
+    } else {
+      return CpuTimeUsed();
+    }
+  }
+
+  double Seconds(const ThreadManager::Result* res,
+                 const BenchmarkInstance& b) const {
+    if (b.use_manual_time) {
+      return res->manual_time_used;
+    } else if (b.use_real_time) {
+      return res->real_time_used;
+    } else {
+      return res->cpu_time_used;
+    }
+  }
+
+  // This is the business logic. Implementations need to decide what this does.
+  virtual void NewResult(ThreadManager::Result result) = 0;
+};
+
+class BenchmarkRunner final {
  public:
   BenchmarkRunner(const benchmark::internal::BenchmarkInstance& b_,
+                  AbstractIterationResultCollector* c_,
                   std::vector<BenchmarkReporter::Run>* complexity_reports_)
       : b(b_),
+        c(c_),
         complexity_reports(*complexity_reports_),
         min_time(!IsZero(b.min_time) ? b.min_time : FLAGS_benchmark_min_time),
         repeats(b.repetitions != 0 ? b.repetitions
                                    : FLAGS_benchmark_repetitions),
         has_explicit_iteration_count(b.iterations != 0),
         pool(b.threads - 1),
-        iters(has_explicit_iteration_count ? b.iterations : 1) {
+        predictedItersTotal(has_explicit_iteration_count ? b.iterations : 1) {
     run_results.display_report_aggregates_only =
         (FLAGS_benchmark_report_aggregates_only ||
          FLAGS_benchmark_display_aggregates_only);
@@ -156,7 +207,10 @@ class BenchmarkRunner {
       run_results.file_report_aggregates_only =
           (b.aggregation_report_mode & internal::ARM_FileReportAggregatesOnly);
     }
+    DoWork();
+  }
 
+  void DoWork() {
     for (int repetition_num = 0; repetition_num < repeats; repetition_num++) {
       DoOneRepetition(repetition_num);
     }
@@ -180,6 +234,8 @@ class BenchmarkRunner {
   RunResults run_results;
 
   const benchmark::internal::BenchmarkInstance& b;
+  AbstractIterationResultCollector* c;
+
   std::vector<BenchmarkReporter::Run>& complexity_reports;
 
   const double min_time;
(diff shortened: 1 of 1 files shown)


dominichamon:
I can't parse this change at all, sorry. I need much more information on why this is necessary and what you're going to do with it.

However this is _much_ more complex code and I don't understand why it needs to be.

LebedevRI:
Yeah, i expected that (:
Guess i need to post the actual dependent [WIP] PR that builds ontop of this, too.

LebedevRI:
Soft ping.

src/benchmark_runner.cc
```diff
@@ -124,6 +124,57 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
...
+    result.manual_time_used /= b.threads;
+
+    // So for how long were we running?
+    real_time_used += result.real_time_used;
+    // Base decisions off of real time if requested by this benchmark.
+    if (b.use_manual_time) {
+      seconds += result.manual_time_used;
+    } else if (b.use_real_time) {
+      seconds += result.real_time_used;
+    } else
```
dominichamon: nit, {} around the else too please.

src/benchmark_runner.cc
```diff
@@ -124,6 +124,57 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
...
+    has_error |= result.has_error_;
+    iters += result.iterations;
+
+    // Adjust real/manual time stats since they were reported per thread.
+    result.real_time_used /= b.threads;
+    result.manual_time_used /= b.threads;
+
+    // So for how long were we running?
+    real_time_used += result.real_time_used;
+    // Base decisions off of real time if requested by this benchmark.
```
dominichamon: `// The benchmark may have requested manual or real time be used in decision making. Respect that.`

src/benchmark_runner.cc
```diff
@@ -124,6 +124,57 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
...
+    } else if (b.use_real_time) {
+      seconds += result.real_time_used;
+    } else
+      seconds += result.cpu_time_used;
+
+    results.emplace_back(result);
+  }
+
+  // This is the business logic. You need to decide what this does.
+  virtual void new_result(internal::ThreadManager::Result result,
```
dominichamon: nit: `NewResult`

src/benchmark_runner.cc
```diff
@@ -124,6 +124,57 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
...
+      seconds += result.manual_time_used;
+    } else if (b.use_real_time) {
+      seconds += result.real_time_used;
+    } else
+      seconds += result.cpu_time_used;
+
+    results.emplace_back(result);
+  }
+
+  // This is the business logic. You need to decide what this does.
```
dominichamon: s/You/Implementations/

src/benchmark_runner.cc
```diff
@@ -124,6 +124,57 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
...
+                          const benchmark::internal::BenchmarkInstance& b) = 0;
+};
+
+struct IterationAveragedResults final : public AbstractIterationResults {
+  // Just replace all of the existing data.
+  void new_result(internal::ThreadManager::Result result,
+                  const benchmark::internal::BenchmarkInstance& b) final {
+    IterationAveragedResults empty;
+    empty.append(result, b);
+    *this = std::move(empty);
```
dominichamon: i know this is almost certainly safe but... ew.

i think i'd prefer something a bit more explicit. like the struct being a handler with the overridden method, and a field that stores the results with an accessor.

src/benchmark_runner.cc
```diff
@@ -148,10 +199,12 @@ class BenchmarkRunner {
       run_results.file_report_aggregates_only =
           (b.aggregation_report_mode & internal::ARM_FileReportAggregatesOnly);
     }
+  }
 
+  void do_work() {  // DO NOT CALL ME FROM THE CONSTRUCTOR!!!
```
dominichamon: this is a sign of fragility. is there an alternative design that doesn't have this limitation?

LebedevRI: I can squash it into `get_results()`. You wouldn't be able to return data from ctor..
If we really don't like virtual functions here, i guess i could _try_ CRTP.

LebedevRI: CRTP it is.

src/benchmark_runner.cc
LebedevRI: I'm having a hard time imagining it..
How will it work in the target, more complex, case of #710:
```c++
struct SeparateIterationResults final : public AbstractIterationResults {
  // *Append* the new data.
  void new_result(internal::ThreadManager::Result result,
                  const benchmark::internal::BenchmarkInstance& b) final {
    append(result, b);
  }
};
```

src/benchmark_runner.cc
```diff
@@ -148,7 +193,11 @@ class BenchmarkRunner {
       run_results.file_report_aggregates_only =
           (b.aggregation_report_mode & internal::ARM_FileReportAggregatesOnly);
     }
+  }
+
+  Derived& getDerived() { return *static_cast<Derived*>(this); }
```
dominichamon: ugh. really?

i think the issue i have is that i don't really understand the problem that you're trying to solve (still). which is completely my fault for not being able to focus on this with you.

I think you're trying to get to the point where every iteration is used when we have repetitions to avoid having statistics at the repetition level that apply to _averages_ across iterations. Is that correct?

It also seems like you're trying to do this in a way that is optional. Is that correct?

LebedevRI: > ugh. really?

The previous review note was about the comment "do not call me from constructor,
since i call virtual method". Now that problem is gone, since there are no more virtual calls.

> i think the issue i have is that i don't really understand the problem that you're trying to solve (still).

I agree.

> which is completely my fault for not being able to focus on this with you.

No, probably largely a my fault, failing to *explain* what this is doing.

> I think you're trying to get to the point where every iteration is used when we have repetitions to avoid having statistics at the repetition level that apply to averages across iterations. Is that correct?

Let me try to highlight the issue i'm trying to solve:
https://docs.google.com/spreadsheets/d/11aEs8L8c9SciHAfs30XMNyea2LEBCil8enFRAVVlvzg/edit?usp=sharing
I.e. i want to avoid this implicit averaging within the each repetition,
and keep all the measurements separate, and thus e.g. get the correct stddev,

Now, this is slightly more problematic to do than that, because we don't *have* those
`N` separate measurements, we just keep the timer running over all the `N` iterations.

> It also seems like you're trying to do this in a way that is optional. Is that correct?

Correct. I really really really don't want to pessimize the existing `BenchmarkRunner`.
There are *very* good reasons for it's current behavior.
The new alternative variant (#710) by-design has more overhead.

dominichamon: I don't like either of the options here. A comment warning about virtuals from constructors, and 'roll-your-own' virtual by downcasting this are both problematic.

I feel like we should be able to get somewhere simpler.

Abstracting BenchmarkRunner to do the iteration management and running the benchmark SGTM. Now when it comes to storing and reporting, either a simple conditional on iteration mode or (if you must) a virtual method with concrete implementations for the two modes should be doable without calling that method from the constructor. Even if that means abstracting out the construction from the running of the benchmark. What am I missing?

LebedevRI:
Updated. Unless we (i) have completely mis-understood the disscussion, this should be better.
No virtual calls from constructor, no "do-it-yourself CRTP-based virtual calls".

LebedevRI:
Kinda ping. Would love to get any actionable feedback there.
Though i got the hint that there is no time, and it is kinda complicated.

dominichamon:
I finally had the time to code up my version of this.

I think it's possible to do with less magic: no templating, no *this assignment, by removing the iters and seconds fields from the IterationResults. Instead, these become methods that are overridden. Then, to avoid the memory overhead, if we're going to be averaging we can essentially throw away the individual results.

You end up with methods for "has_error" "iterations" "seconds" and "real_time_used".

makes sense?

LebedevRI:
So i have looked. I have re-cobled together this PR. While this part _seems_ to work, i really don't see
how i could avoid making `BenchmarkRunner::DoNIterations()` `virtual` in a follow-up PR.

LebedevRI:
Kinda ping, does this look more in-line with what you imagined?

LebedevRI:
kinda-ping. i guess i should actually look whether it works for the actual problem i'm trying to solve,
but still would be good to know if this is the direction you foresaw.

dominichamon:
(forgot to hit send on the comments. sorry)

src/benchmark_runner.cc
```diff
@@ -335,15 +375,46 @@ class BenchmarkRunner {
...
+
+  bool HasError() const final { return result.has_error_; };
+  IterationCount Iters() const final { return result.iterations; };
+  double ManualTimeUsed() const final { return result.manual_time_used; };
+  double RealTimeUsed() const final { return result.real_time_used; };
+  double CpuTimeUsed() const final { return result.cpu_time_used; };
+
+  const ThreadManager::Result* results_begin() const final { return &result; };
+  const ThreadManager::Result* results_end() const final {
+    return std::next(results_begin());
```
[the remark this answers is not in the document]
LebedevRI: I clue why this is the way it is. I guess i was trying to avoid allocation on the usual hotpath.

src/benchmark_runner.cc
```diff
@@ -188,106 +243,91 @@ class BenchmarkRunner {
...
+            << b.threads * res->real_time_used << "\n";
   }
 
-  IterationCount PredictNumItersNeeded(const IterationResults& i) const {
+ private:
+  IterationCount PredictNumItersNeeded() const {
     // See how much iterations should be increased by.
     // Note: Avoid division by zero with max(seconds, 1ns).
-    double multiplier = min_time * 1.4 / std::max(i.seconds, 1e-9);
+    double multiplier = min_time * 1.4 / std::max(c->Seconds(b), 1e-9);
```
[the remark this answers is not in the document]
LebedevRI: > DoNIterations could return the time taken and iteration count (instead of the full result) which could then be passed in here and used to determine the next count. That would be healthier, i think, than deferring that to the abstraction.

But that's *precisely* why it's not done that way. While it is true that it will produce
identical meaningful results for the current `IterationAveragingResultCollector` (since that
 is what currently being used); for the `NonAveragingBenchmarkRunner` `c->Seconds(b)`
would produce the total summary time, instead of some time of last run.

Am i missing the point here?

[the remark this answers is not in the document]
dominichamon: i think i am, sorry. it's fine.

dominichamon:
all good i think.

PR #710, opened 2018-10-18
[WIP] NOT FOR AN ACTUAL REVIEW: Separate iterations

LebedevRI:
The dependent diff for https://github.com/google/benchmark/pull/709, that hopefully simplifies the review of **that** PR.

For an actual diff as **compared** to that PR, see:
https://github.com/LebedevRI/benchmark/compare/LebedevRI:BenchmarkRunner-refactor...LebedevRI:separate-iterations?expand=1

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,16 +534,33 @@ order to manually set the time unit, you can specify it manually:
 BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 ```
 
-### Reporting the mean, median and standard deviation by repeated benchmarks
-By default each benchmark is run once and that single result is reported.
+### Reporting each iteration of an run/repetition separately
+It is worth noting that internally each run/repetition usually runs several
+iterations of the benchmarked code, and then the total measurements over all
+the iterations are averaged (divided by the iteration count). It makes sense,
+since normally the measurements are expected to be noisy, and besides,
+it would be quite costly to measure each iteration separately.
+
+But, sometimes this averaging is unwelcomed.
+You can use `--benchmark_report_separate_iterations={true|false}` flag or
+`ReportSeparateIterations(bool)` function can be used to control the behaviour,
+and force each iteration to be reported separately.
+Please be aware that it comes with cost, and does introduce overhead,
+linearly dependent on the iteration count.
+
+### Reporting the mean, median and standard deviation by benchmarks
+By default each benchmark is run once and that single result is reported,
+unless reporting of separate iterations was requested.
 However benchmarks are often noisy and a single result may not be representative
 of the overall behavior. For this reason it's possible to repeatedly rerun the
 benchmark.
 
 The number of runs of each benchmark is specified globally by the
 `--benchmark_repetitions` flag or on a per benchmark basis by calling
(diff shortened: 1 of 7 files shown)


LebedevRI:
Soft ping.

README.md
```diff
@@ -534,16 +534,33 @@ order to manually set the time unit, you can specify it manually:
...
+### Reporting each iteration of an run/repetition separately
+It is worth noting that internally each run/repetition usually runs several
+iterations of the benchmarked code, and then the total measurements over all
+the iterations are averaged (divided by the iteration count). It makes sense,
+since normally the measurements are expected to be noisy, and besides,
+it would be quite costly to measure each iteration separately.
+
+But, sometimes this averaging is unwelcomed.
+You can use `--benchmark_report_separate_iterations={true|false}` flag or
+`ReportSeparateIterations(bool)` function can be used to control the behaviour,
```
dominichamon: i wonder if this is useful given the user can't really control the number of iterations (at least the max). Ie, this could output thousands++ of reports if the benchmark suddenly speeds up.

LebedevRI: That is why there is `--benchmark_report_aggregates_only=`, and i have added `--benchmark_display_aggregates_only=` :)

dominichamon: Even I'm getting confused between all these options.

Can you document/explain how these all interact in a simple way?

LebedevRI: They don't interact **directly**. They are separate.
* `benchmark_report_aggregates_only` - for all the reporters, only output the aggregates
* `benchmark_display_aggregates_only` - for display reporter, only print the aggregates; file reporter will still contain both the non-aggregates, and aggregates.
* `benchmark_report_separate_iterations` - when benchmarking, don't do one huge measurement (spanning `n` iterations), and then divide it by iteration count. instead, do all these `n` iterations separately. I.e. do no implicit averaging of measurements.

dominichamon: `--benchmark_report_aggregates_only --benchmark_report_separate_iterations` does the same as `--benchmark_report_aggregates_only` ?

LebedevRI: No.
If there were repetitions, the stddev aggregate may (will) be shockingly different between `--benchmark_report_aggregates_only --benchmark_report_separate_iterations` and `--benchmark_report_aggregates_only`.

Thinking about it, maybe `--benchmark_report_separate_iterations` isn't such a great name.
`report` looks out of place.. I can clearly see that it confuses you here.
It has different meaning as compared to to `benchmark_report_aggregates_only` / `benchmark_display_aggregates_only`.

dominichamon: maybe just --benchmark_report_every_iteration or --benchmark_average_over_iterations (with the default being true) or --benchmark_one_iteration_per_run or maybe just --benchmark_max_iterations=1 ?

LebedevRI: Certainly not `--benchmark_max_iterations=1`.
It looks way too similar to `benchmark_min_time`, while having radically different meaning.

`--benchmark_report_every_iteration` still looks quite similar to the current `--benchmark_report_separate_iterations`.

So i guess `--benchmark_average_over_iterations` / `--benchmark_one_iteration_per_run`
is the closest hit.
Maybe `--benchmark_average_over_iterations`.

PR #715, opened 2018-11-02
Disable exceptions in Microsoft STL

kbobyrev:
This is the copy of patch proposed to LLVM's copy of benchmark via https://reviews.llvm.org/D52998.


kbobyrev:
I signed it!

dominichamon:
Thanks!

PR #716, opened 2018-11-09
Added default case in switch to avoid warning

amallia:
```
warning: missing return statement at end of non-void function "benchmark::GetTimeUnitString"
warning: missing return statement at end of non-void function "benchmark::GetTimeUnitMultiplier"
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2016 NVIDIA Corporation
Built on Tue_Jan_10_13:22:03_CST_2017
Cuda compilation tools, release 8.0, V8.0.61

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1518,6 +1518,7 @@ inline const char* GetTimeUnitString(TimeUnit unit) {
     case kMicrosecond:
       return "us";
     case kNanosecond:
+    default:	
       return "ns";
   }
   BENCHMARK_UNREACHABLE();
@@ -1530,6 +1531,7 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
     case kMicrosecond:
       return 1e6;
     case kNanosecond:
+    default:	
       return 1e9;
   }
   BENCHMARK_UNREACHABLE();
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1518,10 +1518,10 @@ inline const char* GetTimeUnitString(TimeUnit unit) {
     case kMicrosecond:
       return "us";
     case kNanosecond:
-    default:	
       return "ns";
   }
   BENCHMARK_UNREACHABLE();
+  return "ns";
 }
 
 inline double GetTimeUnitMultiplier(TimeUnit unit) {
@@ -1531,10 +1531,10 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
     case kMicrosecond:
       return 1e6;
     case kNanosecond:
-    default:	
       return 1e9;
   }
   BENCHMARK_UNREACHABLE();
+  return 1e9;
 }
 
 }  // namespace benchmark


amallia:
I signed it!

include/benchmark/benchmark.h
```diff
@@ -1518,6 +1518,7 @@ inline const char* GetTimeUnitString(TimeUnit unit) {
     case kMicrosecond:
       return "us";
     case kNanosecond:
+    default:	
       return "ns";
   }
   BENCHMARK_UNREACHABLE();
```
LebedevRI: This is the line that is supposed to fix it.
I think you should instead make sure it expands to some `this is unreachable` your compiler understands.

amallia: The problem is that now all the compilers understand `BENCHMARK_UNREACHABLE();`

amallia: https://github.com/google/benchmark/blob/d8c0f27448dfad95b94285e612bba1f7c55c9dd0/include/benchmark/benchmark.h#L253

LebedevRI: Have you tried changing that to
```
#if defined(__GNUC__) || __has_builtin(__builtin_unreachable) || defined(WHATEVER_NVCC_MACRO)
  #define BENCHMARK_UNREACHABLE() __builtin_unreachable()
#elif defined(_MSC_VER)
  #define BENCHMARK_UNREACHABLE() __assume(false)
#else
  #define BENCHMARK_UNREACHABLE() ((void)0)
#endif
```
?

amallia: I believe since `__has_builtin(__builtin_unreachable) ` is false, it should not make any difference...

LebedevRI: Only if `__has_builtin` wasn't expanded to `false` itself, since it is also non-standard.

amallia: This is what I am saying....

amallia: Don't you want to support compilers that are not MSC, GNU and do not have __builtin_unreachable macro?

brenoguim: Some options:
1 - Find the construct in nvcc that represents the unreachable, and update `BENCHMARK_UNREACHABLE`
2 - Make it:
```
BENCHMARK_UNREACHABLE();
return "ns";
```
But that may trigger warnings on some compiler (doesn't seem to trigger for GCC/clang though).
3 - Create a `BENCHMARK_UNREACHABLE_OR_RETURN("ns")` ?  Not great either but should be better than adding the `default`.

amallia: These options sound good. I will try them and update the pr

brenoguim: Cool! But for the record, it was not my intention to sound as the responsible for this code. I was just passing by and decided to leave my 2 cents on ways to solve this. I should've made that clearer.

EricWF: default cases are bad for code maintainability, because I want the compiler to warn me that the switch isn't fully covered when I add an enumerator. 

Maybe we should just disable the "no default" warning in the source code?

amallia: I have removed the default cases. Opted for solution #2 from @brenoguim 
I hope this one gets accepted

brenoguim: @EricWF  the warning is actually complaining about lack of return in end of non-void function, not about the lack of default. Disabling that warning on the whole code can be harmful.

amallia: Yes @brenoguim you are correct

brenoguim: Option #2 introduces warnings in MSVC :( https://ci.appveyor.com/project/google/benchmark/builds/20198554/job/ub5ulj85wb67ym3o

EricWF: Does std::abort work? Or any other `noreturn` function?

EricWF: > Disabling that warning on the whole code can be harmful.

Not on the whole of the code base, just for that function. But I'm strongly opposed to adding a default case.

We've discussed this before, and we've decided on this direction.

brenoguim: Yeah, adding the `default` should not be an option.

I like the `std::abort` suggestion. Writing the `BENCHMARK_UNREACHABLE` last fallback as `std::abort` would keep the good code generation for the targets that support it, and silence the warning for the others.

LebedevRI:
On which compiler is nvcc based? Does it really not have `__builtin_unreachable()` or some analog of it?

dominichamon:
`do {std::abort(); } while(0)` should be a good replacement for unreachable code in any compiler, i think.

PR #721, opened 2018-11-21
Fix C++17 mode compilation with Apple clang

traceon:
Fixes the following compilation error:
```
benchmark/test/output_test_helper.cc:206:28: error: no member named 'streampos' in 'std::__1::basic_ios<char>'
```


traceon:
I signed it!

EricWF:
LGTM, but I'll wait for the bots to cycle before merging  this.

dominichamon:
thank you!

PR #723, opened 2018-11-24
Refactor and document complexity calculation

ulvgard:
Hello,

In trying to derive and understand how complexity is calculated I found that the algorithm is made less readable with the factors `sigma_gn_squared` and `sigma_time_gn` which can be cancelled in expression. Also, the link to the description of the Least-Square calculation is dead. The author has a similarly named repository (https://github.com/ismaelJimenez/cpp.leastsq) that provides a description for the least-square derivation which in my opinion is difficult to follow and arrives in an less familiar expression. 

This pull request provides a hand-waving argument on how complexity is calculated. It is derived similarly with least-square minimization and the description detail the conceptual approach in text.

### DIFF
--- a/src/complexity.cc
+++ b/src/complexity.cc
@@ -73,8 +73,8 @@ std::string GetBigOString(BigO complexity) {
 //   - time          : Vector containing the times for the benchmark tests.
 //   - fitting_curve : lambda expression (e.g. [](int64_t n) {return n; };).
 
-// For a deeper explanation on the algorithm logic, look the README file at
-// http://github.com/ismaelJimenez/Minimal-Cpp-Least-Squared-Fit
+// For a deeper explanation on the algorithm logic, please refer to
+// https://en.wikipedia.org/wiki/Least_squares#Least_squares,_regression_analysis_and_statistics
 
 LeastSq MinimalLeastSq(const std::vector<int64_t>& n,
                        const std::vector<double>& time,


ulvgard:
googlebot please try again

src/complexity.cc
```diff
@@ -73,31 +73,41 @@ std::string GetBigOString(BigO complexity) {
...
     sigma_time += time[i];
-    sigma_time_gn += time[i] * gn_i;
   }
 
   LeastSq result;
   result.complexity = oLambda;
 
   // Calculate complexity.
-  result.coef = sigma_time_gn / sigma_gn_squared;
+  result.coef = sigma_time / sigma_gn;
```
LebedevRI: I'm not sure this is equivalent
https://www.wolframalpha.com/input/?i=(t0+*+n0+%2B+t1+*+n1)+%2F+(n0*n0+%2B+n1*n1)+%3D%3D+(t0+%2B+t1)+%2F+(n0+%2B+n1)

ulvgard: @LebedevRI thanks for the feedback, I appreciate someone taking a second look at this since I arrived at a different final expression.

Indeed they are not equivalent. And in my comment I should not have stated that the factors can be "cancelled". The contribution suggests to compute the coefficient using the approach detailed in the comments. I cant find a motivation why the sum of the square of the execution time and model, with `sigma_time_gn` and  `sigma_gn_squared` is used for computing the coefficient of growth?

The minimum least-square part happens in the computation of the RMS as I interpret this function?

src/complexity.cc
```diff
@@ -73,31 +73,41 @@ std::string GetBigOString(BigO complexity) {
...
-    sigma_gn_squared += gn_i * gn_i;
     sigma_time += time[i];
-    sigma_time_gn += time[i] * gn_i;
   }
 
   LeastSq result;
   result.complexity = oLambda;
 
   // Calculate complexity.
-  result.coef = sigma_time_gn / sigma_gn_squared;
```
LebedevRI: FWIW i'm somewhat sure that this is the correct (other than fp errors) implementation of ["Least squares fitting"](https://en.wikipedia.org/wiki/Least_squares#Least_squares,_regression_analysis_and_statistics)
This function is named `MinimalLeastSq`, and has a square in the denominator; in the RHS of the diff, there is no more square..

ulvgard: Strange. I re-performed the calculation again an landed in the same result that the coefficient is estimated by the proportion between sum of measurement and model. Good to see a source stating the result with squares. It is unfortunate to see that they omit the full derivation.

Having thought about it some more, I would actually claim that these methods are equivalent. The coefficient in the current case is computed as the proportion between the sum of the area of rectangles (ti*ni) and the sum of the area of squares (ni*ni). This is the same as computing the proportion between the sum of  lines (ti) and the sum of other lines (ni). This is not generally true of course, but for real positive values it holds. It is also possible to arrive at this conclusion by formulating the least-squre equation and solve for the coefficient. 

In any case, I am happy if we pursue this PR or update the reference to the calculation with a link to your Wikipedia-article or even use the old authors new repository? Also a test-case for the computation of the coefficient could be nice if it is interesting?

pleroy: The derivation of the formula is quite straightforward: take the sum of squares _S_ in the Wikipedia article, derive it with respect to _k_ and equate the derivative with 0, you get the estimated _k_.

Using a similar analysis it seems that the quantity that you are minimizing is the sum of terms like `c * t_i - c^2 * gn_i / 2`.  Look at this expression carefully: it can be negative for outliers: outliers will actually _reduce_ the error.  Put it another way: what you are minimizing is not even mathematically a norm.  Therefore your claim that your method is "equivalent" to least square has no mathematical basis.  (It is _if_ you ignore outliers, but that's not an acceptable thing to do: outliers will occur and they should not break the statistical analysis.)

I don't think that this PR should go in.  I would welcome a reference to the Wikipedia article, though.

ulvgard:
Are any actions required by me to move forward?

LebedevRI:
Looks like CLA bot is still unhappy?

ulvgard:
> Looks like CLA bot is still unhappy?

Is it because the initial commit was made with another Author and Email? I squshed that and rewrote the auther to my CLA signed user.

LebedevRI:
https://github.com/google/benchmark/commit/5d4ad0a277e1761463a227cf50d7748fcf5f1ef4.patch
```
From: Tobias Ulvgard <tobias@ulvgard.se>
```
^ i *guess* it is checking this email. Is that the one with signed CLA?

ulvgard:
> https://github.com/google/benchmark/commit/5d4ad0a277e1761463a227cf50d7748fcf5f1ef4.patch
> 
> ```
> From: Tobias Ulvgard <tobias@ulvgard.se>
> ```
> 
> ^ i _guess_ it is checking this email. Is that the one with signed CLA?

It is yes.

ulvgard:
Is is possible to merge this? From what I understand googlebot is confused because the initial commit was made by me but with another author signature.

ulvgard:
ping @LebedevRI ?

LebedevRI:
cc @dominichamon

dominichamon:
LGTM

PR #727, opened 2018-11-26
StrFormat() is a printf-like function, mark it as such, fix fallout.

### LINKED ISSUE #714: Test arguments printed with 32 bit precision
The arguments passed to benchmarks are of type `int64_t` however when they get reported they get truncated to 32 bits. The following is output where the argument keeps increasing beyond `int` precision and is printed incorrectly (but the benchmark does the right thing).
```
baseline_push_back/std::vector<int>/67108864/manual_time              33079613 ns   32812500 ns         20
baseline_push_back/std::vector<int>/134217728/manual_time             71565844 ns   71428571 ns          7
baseline_push_back/std::vector<int>/536870912/manual_time            539197109 ns  531250000 ns          1
baseline_push_back/std::vector<int>/1073741824/manual_time           914751775 ns  890625000 ns          1
baseline_push_back/std::vector<int>/-2147483648/manual_time         1806131008 ns 1796875000 ns          1
baseline_push_back/std::vector<int>/0/manual_time                   3629563138 ns 3625000000 ns          1
```
 I think the problem is that https://github.com/google/benchmark/blob/edc77a3669026eddc380721d5a3cdccd752b76cb/src/benchmark_register.cc#L185
should use the `%lld` format specifier instead.

LebedevRI:
Fixes #714.

### DIFF
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -182,14 +182,19 @@ bool BenchmarkFamilies::FindBenchmarks(
             }
           }
 
-          instance.name += StrFormat("%d", arg);
+          // we know that the args are always non-negative (see 'AddRange()'),
+          // thus print as 'unsigned'. BUT, do a cast due to the 32-bit builds.
+          instance.name += StrFormat("%lu", static_cast<unsigned long>(arg));
           ++arg_i;
         }
 
         if (!IsZero(family->min_time_))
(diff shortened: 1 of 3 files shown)


EricWF:
LGTM. Thanks.

LebedevRI:
Thank you for the review!

PR #647, opened 2018-07-25
CI: Add Conan testing, mass building and publishing

Croydon:
This
  * is adding a recipe for the Conan package manager
  * also adds a small test to the CI if Conan package creation is working 
  * updated: also adds a upload script to the CI for the Conan recipe to a Conan repository

~~Please note that for future releases of benchmark the version string needs to be updated in conanfile.py~~ (not anymore, see discussion)

_______
~~**Future work** could include~~
**Remaining to do list:**
  * [x] setting up a free repository on Bintray: https://bintray.com/dominichamon/benchmark
  * [ ] including the package in the official Conan repository [conan-center](https://bintray.com/conan/conan-center) 
   * [x] and extending the CI to upload new versions of the recipe and built packages to the Bintray repository automatically

To use the recipe after this pull request is merged it requires to clone this git repository first; when the recipe is included in conan-center every Conan client would be able to use it via one single install command.
_______

Please let me know if you have any further questions regarding supporting Conan 😄 

Fixes #635 

//cc @raulbocanegra @p-groarke @danimtb @Mikayex @mpusz @iblis-ms

PR #728, opened 2018-11-26
Add Conan support

Croydon:
This is split from #647 

It only contains the minimum changes to get initial Conan support.

With this changes you can create a Conan package by executing

`conan create . benchmark/<version-number>@google/stable`

or better: to actual test if the package works, the test_package should be executed with it:

`conan create . benchmark/<version-number>@google/stable -tf conan/test_package`

(the parameter `tf` stands for test folder).

If no error appears the created Benchmark Conan package can be consumed by projects **locally**.

_____

This does NOT contain any kind of publishing/uploading mechanism with mass building for different configurations/operating systems etc. this should be eventually done via #647 

_____

cc @memsharded @dominichamon

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,0 +1,79 @@
+from conans import ConanFile, CMake, tools
+from conans.errors import ConanInvalidConfiguration
+import shutil
+import os
+
+
+class GoogleBenchmarkConan(ConanFile):
+    name = "benchmark"
+    description = "A microbenchmark support library."
+    topics = ("conan", "benchmark", "google", "microbenchmark")
+    url = "https://github.com/google/benchmark"
+    homepage = "https://github.com/google/benchmark"
+    author = "Google Inc."
+    license = "Apache-2.0"
+    exports_sources = ["*"]
+    generators = "cmake"
+
+    settings = "arch", "build_type", "compiler", "os"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "enable_lto": [True, False],
+        "enable_exceptions": [True, False]
+    }
+    default_options = {"shared": False, "fPIC": True, "enable_lto": False, "enable_exceptions": True}
+
+    _build_subfolder = "."
+
+    def source(self):
+        # Wrap the original CMake file to call conan_basic_setup
+        shutil.move("CMakeLists.txt", "CMakeListsOriginal.txt")
+        shutil.move(os.path.join("conan", "CMakeLists.txt"), "CMakeLists.txt")
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            if self.settings.compiler == "Visual Studio" and float(self.settings.compiler.version.value) <= 12:
+                raise ConanInvalidConfiguration("{} {} does not support Visual Studio <= 12".format(self.name, self.version))
+            del self.options.fPIC
+
+    def configure(self):
+        if self.settings.os == "Windows" and self.options.shared:
+            raise ConanInvalidConfiguration("Windows shared builds are not supported right now, see issue #639")
+
+    def _configure_cmake(self):
+        cmake = CMake(self)
+
+        cmake.definitions["BENCHMARK_ENABLE_TESTING"] = "OFF"
+        cmake.definitions["BENCHMARK_ENABLE_GTEST_TESTS"] = "OFF"
+        cmake.definitions["BENCHMARK_ENABLE_LTO"] = "ON" if self.options.enable_lto else "OFF"
+        cmake.definitions["BENCHMARK_ENABLE_EXCEPTIONS"] = "ON" if self.options.enable_exceptions else "OFF"
+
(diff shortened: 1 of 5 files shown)


conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+    _build_subfolder = "."
+
+    def source(self):
+        # Wrap the original CMake file to call conan_basic_setup
+        shutil.move("CMakeLists.txt", "CMakeListsOriginal.txt")
+        shutil.move(os.path.join("conan", "CMakeLists.txt"), "CMakeLists.txt")
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            if self.settings.compiler == "Visual Studio" and float(self.settings.compiler.version.value) <= 12:
```
uilianries: there is another option to check compiler version: https://docs.conan.io/en/latest/howtos/check_conan_version_in_conanfile.html#how-to-check-the-version-of-the-conan-client-inside-a-conanfile

conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+
+class GoogleBenchmarkConan(ConanFile):
+    name = "benchmark"
+    description = "A microbenchmark support library."
+    topics = ("conan", "benchmark", "google", "microbenchmark")
+    url = "https://github.com/google/benchmark"
+    homepage = "https://github.com/google/benchmark"
+    author = "Google Inc."
+    license = "Apache-2.0"
+    exports_sources = ["*"]
```
uilianries: where is exports LICENSE?

conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+        shutil.move("CMakeLists.txt", "CMakeListsOriginal.txt")
+        shutil.move(os.path.join("conan", "CMakeLists.txt"), "CMakeLists.txt")
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            if self.settings.compiler == "Visual Studio" and float(self.settings.compiler.version.value) <= 12:
+                raise ConanInvalidConfiguration("{} {} does not support Visual Studio <= 12".format(self.name, self.version))
+            del self.options.fPIC
+
+        if tools.get_env("CONAN_RUN_TESTS") != "True":
```
uilianries: string or boolean? https://docs.conan.io/en/latest/reference/env_vars.html#conan-run-tests

conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+            cmake.definitions["BENCHMARK_USE_LIBCXX"] = "ON" if (str(self.settings.compiler.libcxx) == "libc++") else "OFF"
+        else:
+            cmake.definitions["BENCHMARK_USE_LIBCXX"] = "OFF"
+
+        cmake.configure(build_folder=self._build_subfolder)
+        return cmake
+
+    def build_requirements(self):
+        if tools.get_env("CONAN_RUN_TESTS") and self.options.enable_gtest_tests:
+            self.build_requires("gtest/1.8.0@bincrafters/stable")
```
uilianries: the latest version is 1.8.1, but I dont know if 1.8.0 is a requirement for benchmark library

conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+            cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) else "OFF"
+            cmake.definitions["BENCHMARK_USE_LIBCXX"] = "ON" if (str(self.settings.compiler.libcxx) == "libc++") else "OFF"
+        else:
+            cmake.definitions["BENCHMARK_USE_LIBCXX"] = "OFF"
+
+        cmake.configure(build_folder=self._build_subfolder)
+        return cmake
+
+    def build_requirements(self):
+        if tools.get_env("CONAN_RUN_TESTS") and self.options.enable_gtest_tests:
```
danimtb: This line looks kind of weird to me but having a look at `BENCHMARK_ENABLE_TESTING` and `BENCHMARK_ENABLE_GTEST_TESTS` it could make sense to have it separated. Anyway bot the option and env var for tests are well managed

conanfile.py
Croydon: Since this is an in-source recipe there is no additional LICENSE file for the Conan recipe. Since we are copying all source files, we also copying the LICENSE file already.

conanfile.py
Croydon: Are you seeing a major advantage in the method from the documentation? If not I would like to keep it this way.

uilianries: nope, keep float

conanfile.py
Croydon: The `enable_gtest_tests` is only available when `CONAN_RUN_TESTS` env variable is set. So this should be fine.

Croydon:
I have actually reconsidered and removed `CONAN_RUN_TESTS` and `enable_gtest_tests` for the sake of simplification. The tests for Benchmark should be continued to run via plain CMake as it is done so far.

Please review this @dominichamon

dominichamon:
Thank you all!

Issue #722, opened 2018-11-23
host computer identification


peterlauro1973:
Would be possible to enhance benchmark output json file with an information about the host name on which the test was triggered?

just simple "uname -n" on Linux or "hostname" on windows.

This kind of information can be useful in situation when CI may trigger the benchmark tests on different hosts. The post-processing script can identify, select and compare the results not only based on cpu identification but also base the hostname.

dominichamon:
That sounds completely reasonable. Patches welcome :)

tesch1:
Closeable?

LebedevRI:
Right, thanks.

PR #733, opened 2018-11-29
#722 Adding Host Name in Reporting

jatinx:
PR with CLA signed

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1293,6 +1293,7 @@ struct CPUInfo {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(CPUInfo);
 };
 
+//Adding Struct for System Information
 struct SystemInfo {
   std::string name;
   static const SystemInfo& Get();
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1293,6 +1293,14 @@ struct CPUInfo {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(CPUInfo);
 };
 
+struct SystemInformation {
+  std::string name;
+  static const SystemInformation& Get();
+ private:
+  SystemInformation();
+  BENCHMARK_DISALLOW_COPY_AND_ASSIGN(SystemInformation);
+};
+
 // Interface for custom benchmark result printers.
 // By default, benchmark reports are printed to stdout. However an application
 // can control the destination of the reports by calling
@@ -1302,6 +1310,7 @@ class BenchmarkReporter {
  public:
   struct Context {
     CPUInfo const& cpu_info;
+    SystemInformation const& sys_info;
     // The number of chars in the longest benchmark name.
     size_t name_field_width;
     static const char* executable_name;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1293,12 +1293,12 @@ struct CPUInfo {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(CPUInfo);
 };
 
-struct SystemInformation {
+struct SystemInfo {
   std::string name;
-  static const SystemInformation& Get();
+  static const SystemInfo& Get();
  private:
-  SystemInformation();
-  BENCHMARK_DISALLOW_COPY_AND_ASSIGN(SystemInformation);
+  SystemInfo();
+  BENCHMARK_DISALLOW_COPY_AND_ASSIGN(SystemInfo);
 };
 
 // Interface for custom benchmark result printers.
@@ -1310,7 +1310,7 @@ class BenchmarkReporter {
  public:
   struct Context {
     CPUInfo const& cpu_info;
-    SystemInformation const& sys_info;
+    SystemInfo const& sys_info;
     // The number of chars in the longest benchmark name.
     size_t name_field_width;
     static const char* executable_name;
--- a/src/json_reporter.cc
+++ b/src/json_reporter.cc
@@ -77,6 +77,8 @@ bool JSONReporter::ReportContext(const Context& context) {
   std::string walltime_value = LocalDateTimeString();
   out << indent << FormatKV("date", walltime_value) << ",\n";
 
+  out << indent << FormatKV("host_name", context.sys_info.name) << ",\n";
+
   if (Context::executable_name) {
     // windows uses backslash for its path separator,
     // which must be escaped in JSON otherwise it blows up conforming JSON
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -80,7 +80,7 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
 const char *BenchmarkReporter::Context::executable_name;
 
 BenchmarkReporter::Context::Context()
-    : cpu_info(CPUInfo::Get()), sys_info(SystemInformation::Get()) {}
+    : cpu_info(CPUInfo::Get()), sys_info(SystemInfo::Get()) {}
 
 std::string BenchmarkReporter::Run::benchmark_name() const {
   std::string name = run_name;
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -38,9 +38,6 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
 
   Out << LocalDateTimeString() << "\n";
 
-  const SystemInformation &sys_info = context.sys_info;
-  Out << "System Name "<< sys_info.name << "\n";
-
   if (context.executable_name)
     Out << "Running " << context.executable_name << "\n";
 
@@ -82,7 +79,8 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
 // No initializer because it's already initialized to NULL.
 const char *BenchmarkReporter::Context::executable_name;
 
-BenchmarkReporter::Context::Context() : cpu_info(CPUInfo::Get()), sys_info(SystemInformation::Get()) {}
+BenchmarkReporter::Context::Context()
+    : cpu_info(CPUInfo::Get()), sys_info(SystemInformation::Get()) {}
 
 std::string BenchmarkReporter::Run::benchmark_name() const {
   std::string name = run_name;
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -38,6 +38,9 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
 
   Out << LocalDateTimeString() << "\n";
 
+  const SystemInformation &sys_info = context.sys_info;
+  Out << "System Name "<< sys_info.name << "\n";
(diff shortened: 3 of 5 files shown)


src/reporter.cc
```diff
@@ -79,7 +82,7 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
 // No initializer because it's already initialized to NULL.
 const char *BenchmarkReporter::Context::executable_name;
 
-BenchmarkReporter::Context::Context() : cpu_info(CPUInfo::Get()) {}
+BenchmarkReporter::Context::Context() : cpu_info(CPUInfo::Get()), sys_info(SystemInformation::Get()) {}
```
LebedevRI: Linewrap

src/reporter.cc
```diff
@@ -38,6 +38,9 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
 
   Out << LocalDateTimeString() << "\n";
 
+  const SystemInformation &sys_info = context.sys_info;
+  Out << "System Name "<< sys_info.name << "\n";
```
LebedevRI: Is this printed only for the json dumper, or for all of them?
Put differently, i don't think it should be in the console output, it is getting rather crowded already.

src/sysinfo.cc
```diff
@@ -366,6 +366,30 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
+    return std::string("Unable to Get Host Name");
+#ifndef UNICODE
+  str = hostname;
+#else
+  std::wstring wStr = hostname;
+  str = std::string(wStr.begin(), wStr.end());
+#endif
+  return str;
+#else
+  const unsigned COUNT = 64;
```
LebedevRI: Just use `HOST_NAME_MAX`

src/sysinfo.cc
```diff
@@ -366,6 +366,30 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
 #endif
 }
 
+std::string GetSystemName() {
+#if defined(BENCHMARK_OS_WINDOWS)
+  std::string str;
+  const unsigned COUNT =  32767;
```
LebedevRI: Again, why invent a magic number.
`MAX_COMPUTERNAME_LENGTH+1` ?
And that being said, isn't that a rather too big to have as an array on stack?
Maybe use resized `std::string`?

test/reporter_output_test.cc
```diff
@@ -16,6 +16,7 @@ static int AddContextCases() {
   AddCases(TC_ConsoleErr,
            {
                {"%int[-/]%int[-/]%int %int:%int:%int$", MR_Default},
+               {"System Name", MR_Next},
```
LebedevRI: What is "system name"?
How is this different from "host name"?
If it is exactly the "host name", just use that term?

src/sysinfo.cc
jatinx: HOST_NAME_MAX was not present on OSx, added local initialization.
The Job on Travis failed in config, can you please have a look at it?

jatinx:
@LebedevRI  did the asked changes, can you please have a look at it?

jatinx:
@LebedevRI its been a week, is something wrong with my design, would like some comments if thats the case.

src/sysinfo.cc
```diff
@@ -366,6 +366,33 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
+  const unsigned COUNT = MAX_COMPUTERNAME_LENGTH+1;
+  TCHAR  hostname[COUNT];
+  DWORD DWCOUNT = COUNT;
+  if (!GetComputerName(hostname, &DWCOUNT))
+    return std::string("Unable to Get Host Name");
+#ifndef UNICODE
+  str = hostname;
+#else
+  std::wstring wStr = hostname;
+  str = std::string(wStr.begin(), wStr.end());
```
LebedevRI: Does this work? Is this already done elsewhere in the library?
wchar is bigger than char, what makes sure that this isn't lossy?

src/sysinfo.cc
```diff
@@ -366,6 +366,33 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
+#endif
+  return str;
+#else
+#ifdef BENCHMARK_OS_MACOSX //Mac Doesnt have a copy for Host Name in it
+#define HOST_NAME_MAX 64
+#endif
+  const unsigned COUNT = HOST_NAME_MAX;
```
LebedevRI: Can't you just use `HOST_NAME_MAX` directly, why do you need `COUNT` variable?

src/sysinfo.cc
```diff
@@ -366,6 +366,33 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
+  char hostname[COUNT];
+  int retVal = gethostname(hostname, COUNT);
+  if (retVal != 0) return std::string("Unable to Get Host Name");
+  return std::string(hostname);
+#endif
```
LebedevRI: And others? https://github.com/google/benchmark/blob/439d6b1c2a6da5cb6adc4c4dfc555af235722396/src/internal_macros.h#L39-L73

src/sysinfo.cc
```diff
@@ -366,6 +366,33 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
+std::string GetSystemName() {
+#if defined(BENCHMARK_OS_WINDOWS)
+  std::string str;
+  const unsigned COUNT = MAX_COMPUTERNAME_LENGTH+1;
+  TCHAR  hostname[COUNT];
+  DWORD DWCOUNT = COUNT;
+  if (!GetComputerName(hostname, &DWCOUNT))
+    return std::string("Unable to Get Host Name");
+#ifndef UNICODE
+  str = hostname;
```
LebedevRI: Are you 100.0% sure this is safe?
https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-getcomputernamew
> On input, specifies the size of the buffer, in TCHARs. On output, the number of TCHARs copied to the destination buffer, not including the terminating null character.

I'd recommend to use 'returned' `DWCOUNT`.

src/sysinfo.cc
LebedevRI: Also, same comment about `DWCOUNT`.

src/sysinfo.cc
jatinx: The only segregation needed here is POSIX vs non POSIX Compliance.
POSIX has `gethostname`

src/sysinfo.cc
```diff
@@ -366,6 +366,33 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
+#else
+  std::wstring wStr = hostname;
+  str = std::string(wStr.begin(), wStr.end());
+#endif
+  return str;
+#else
```
LebedevRI:
```suggestion
#else // defined(BENCHMARK_OS_WINDOWS)
// Catch-all POSIX block.
```

src/sysinfo.cc
LebedevRI:
```suggestion
#endif // Catch-all POSIX block.
```

src/sysinfo.cc
jatinx: Using Standard wstring_convert to fix this

src/sysinfo.cc
jatinx: Removing unnecessary COUNT variable references, using macro directly

src/sysinfo.cc
jatinx: Initialising TCHAR to null and only copying DWCOUNT into string.

src/json_reporter.cc
```diff
@@ -77,6 +77,8 @@ bool JSONReporter::ReportContext(const Context& context) {
   std::string walltime_value = LocalDateTimeString();
   out << indent << FormatKV("date", walltime_value) << ",\n";
 
+  out << indent << FormatKV("host_name", context.sys_info.name) << ",\n";
```
LebedevRI: 1. We are sure that this won't contain `\\`? (see `executable_name`)
2. Do we want to print `"Unable to Get Host Name"`?2
3. I wonder if this should be after the `executable_name`.

include/benchmark/benchmark.h
```diff
@@ -1293,6 +1293,14 @@ struct CPUInfo {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(CPUInfo);
 };
 
+struct SystemInformation {
```
dominichamon: nit: for consistency with CPUInfo, maybe SystemInfo.

src/sysinfo.cc
```diff
@@ -366,6 +368,35 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizes() {
...
 }
 
+std::string GetSystemName() {
+#if defined(BENCHMARK_OS_WINDOWS)
+  std::string str;
+  const unsigned COUNT = MAX_COMPUTERNAME_LENGTH+1;
+  TCHAR  hostname[COUNT] = {'\0'};
+  DWORD DWCOUNT = COUNT;
+  if (!GetComputerName(hostname, &DWCOUNT))
+    return std::string("Unable to Get Host Name");
```
dominichamon: return empty string and don't print in that case?

include/benchmark/benchmark.h
jatinx: Can be done, let me know if you want me proceed with this

dominichamon: Yes please.

src/json_reporter.cc
dominichamon: 2. No. We should check and only output if it's available.

src/sysinfo.cc
jatinx: merging requested changes

jatinx:
Sometimes, on travis-ci, the last build on MacOS fails in configure.
It gets resolved if I restart the build with a minor comment merge.
This has happened twice now.

@dominichamon @LebedevRI

src/json_reporter.cc
jatinx: 1. Yes, I checked the output, it just outputs the hostname for windows without the `\\`
3. I am not sure about this one, it can be done, but since the output being in json only, is it desirable?

include/benchmark/benchmark.h
jatinx: Done

src/json_reporter.cc
LebedevRI: Aren't we still always printing it?

LebedevRI: Looks about right other than this.

jatinx: are you talking about the 2nd point?
I have placed an empty string incase it fails to fetch the host name

dominichamon:
Thanks! I'll keep an eye on the travis build.

PR #734, opened 2018-11-29
Remove use of std::tmpnam.

EricWF:
std::tmpnam is deprecated and its use is discouraged. For our purposes
in the tests, we really just need a file name which is unlikely to
exist.

This patch converts the tests to using a dummy random file name
generator, which should hopefully avoid name conflicts.


EricWF:
@LebedevRI Thanks for the quick review. Just waiting on the bots to cycle.

Issue #737, opened 2018-12-07
Make it possible to dump benchmark diff report in JSON format


JBakamovic:
TL;DR Be able to dump diff report of two benchmarks into the JSON file and not only on stdout as it is right now. E.g. `compare.py -d bench_diff.json benchmarks bench1.json bench2.json`

Rationale behind this functionality is to be able to manipulate the performance data directly which is much more powerful than simply printing the data onto the stdout. Some other higher-level scripts could then consume this API to make use of the data in whatever way it wants to.

One example could be detection of performance regressions integrated tightly into the CI/CD pipeline. Another one would be a plotter which would be displaying how performance data has been changing through different commits/releases/etc. You get the idea ...

Current implementation of `compare.py` does not have this functionality so I've prepared a small patch which I will send shortly afterwards.

Let me know what you think about this feature.

LebedevRI:
> TL;DR Be able to dump diff report of two benchmarks into the JSON file and not only on stdout as it is right now. E.g. `compare.py -d bench_diff.json benchmarks bench1.json bench2.json`
> 
> Rationale behind this functionality is to be able to manipulate the performance data directly which is much more powerful than simply printing the data onto the stdout. Some other higher-level scripts could then consume this API to make use of the data in whatever way it wants to.

I don't dislike this in general, but right now it is a just simple script.
i'm really uneasy about making **any** API/json stability guarantees.

> One example could be detection of performance regressions integrated tightly into the CI/CD pipeline. Another one would be a plotter which would be displaying how performance data has been changing through different commits/releases/etc. You get the idea ...

This really should be done by something much more specialized, like llvm lnt..

> Current implementation of `compare.py` does not have this functionality so I've prepared a small patch which I will send shortly afterwards.
> 
> Let me know what you think about this feature.

JBakamovic:
> I don't dislike this in general, but right now it is a just simple script.

Yes it is super simple right now but why wouldn't it provide functionality that basically anyone can have benefit of? And the change is a really simple one ...

> i'm really uneasy about making any API/json stability guarantees.

But we already have a JSON format output defined when dumping the benchmark results. Outputting the diff report to JSON would be clearly built upon it.

> This really should be done by something much more specialized, like llvm lnt..

Sorry but I fail to see what does LLVM has to do with this topic?

LebedevRI:
> > I don't dislike this in general, but right now it is a just simple script.
> 
[...]

PR #738, opened 2018-12-07
#737 Make it possible to dump benchmark diff report in JSON format

JBakamovic:
This PR brings proposed solution for functionality described in #737.

Proposed format of JSON dump is as follows:
```
[
    {
	    "measurements": [ // here comes all the measurements we've got for given benchmark
	        {
                    "real_time_other": 1.7604541402884333,
	            "cpu_time": 1.7641470201511333,
		    "time": -0.0175007535861843,
		    "real_time": 1.7918122041459086,
       		    "cpu_time_other": 1.743907157894737,
		    "cpu": -0.011472888611439235
	        },
	        ... 
	        {
		    "real_time_other": 1.8875298447986333,
      		    "cpu_time": 1.7457787153652398,
		    "time": 0.06659085312291296,
		    "real_time": 1.7696850102099235,
		    "cpu_time_other": 1.8679594937343365,
		    "cpu": 0.06998640623450084
	        }
            ],
	    "utest": { // utest section will be empty when not applicable (utest not requested)
		    "cpu_pvalue": 0.4033953048926283,
		    "time_pvalue": 0.2100750407866585,
		    "stable": false
	    },
	    "name":  "benchmark test case name",
	    "time_unit": "ms"
    },
    ... // Same for other benchmarks ...
]
```


tools/gbench/report.py
```diff
@@ -345,6 +415,89 @@ def test_basic(self):
             self.assertEqual(len(parts), 7)
             self.assertEqual(expect_lines[i], parts)
 
+    def test_json_difference_report_output(self):
+        expected_output = [
+            {
+                'name': 'BM_SameTimes',
+                'measurements': [{'time': 0.0000, 'cpu': 0.0000, 'real_time': 10, 'real_time_other': 10, 'cpu_time': 10, 'cpu_time_other': 10}],
```
LebedevRI: Please format the changes.

JBakamovic: Will do so but I'm not sure what you're thinking on particularly? I tried to format these as much as I could so that the unit tests do not become too much spread out because of the test data. Can you please give an example of how you would like it to be?

LebedevRI:
```
$ autopep8 -ia tools/compare.py tools/gbench/report.py tools/gbench/util.py
```

tools/gbench/report.py
```diff
@@ -266,11 +274,75 @@ def get_color(res):
     return output_strs
 
 
-###############################################################################
-# Unit tests
+def get_json_difference_report(
```
LebedevRI: This is bad.
Now there are two different functions that *should* only differ with the output format.
Meaning, every change will need to happen in two places.
And extra attention will need to be paid to prevent them from diverging.

Instead, `get_json_difference_report()` should be the only one,
and `generate_difference_report()` should be fed the output of `get_json_difference_report()`,
and it should just pretty-print it.

tools/compare.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import unittest
```
LebedevRI: Can you please run the autopep on git master and submit it as a pr, so this pr will be formatting-change-only clean?

tools/gbench/report.py
```diff
@@ -34,6 +36,7 @@ def __format__(self, format):
 BC_BOLD = BenchmarkColor('BOLD', '\033[1m')
 BC_UNDERLINE = BenchmarkColor('UNDERLINE', '\033[4m')
 
+UTEST_DEFAULT_ALPHA = 0.05
```
LebedevRI: I agree that the old hardcoding of `0.05` is wrong, but this is not correct either.
The specified `utest_alpha` (passed to `generate_difference_report()`) should be propagated/used.

It would be really great if you could submit this as a preparatory PR, to unclutter this PR :)

tools/gbench/report.py
LebedevRI: Hmm, weird, that didn't linewrap (80 col) these lines.

tools/gbench/report.py
```diff
@@ -143,32 +146,43 @@ def extract_field(partition, field_name):
...
+    return (min_rep_cnt >= UTEST_OPTIMAL_REPETITIONS), cpu_pvalue, time_pvalue
+
+
+def print_utest(partition, utest_alpha, first_col_width, use_color=True):
+    def get_utest_color(pval):
+        return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
+
+    timings_time = extract_field(partition, 'real_time')
+    timings_cpu = extract_field(partition, 'cpu_time')
+    stable_results, cpu_pvalue, time_pvalue = calc_utest(
```
LebedevRI: What does `stable_results` mean here? Is this about the pvalue? Or repetition count?
Please s/`stable_results`/`sufficient_repetitions`/

tools/gbench/report.py
```diff
@@ -143,32 +146,43 @@ def extract_field(partition, field_name):
     return [lhs, rhs]
 
 
-def print_utest(partition, utest_alpha, first_col_width, use_color=True):
-    timings_time = extract_field(partition, 'real_time')
-    timings_cpu = extract_field(partition, 'cpu_time')
-
+def calc_utest(timings_cpu, timings_time):
```
LebedevRI: I think these `print_utest()` changes can count as a separate preparatory NFC refactoring,
could you please split this into separate PR?

tools/compare.py
JBakamovic: Done. #739

tools/gbench/report.py
JBakamovic: I had it named `optimal_repetitions` previously but changed it to `stable_results` because I thought that communicating that result**s** (both pvalues) not being stable was sufficient (generic enough) to indicate the erroneous state. That way we are avoiding leaking details of how this error condition is actually evaluated by the underlying function call: `calc_utest`.

Let me know what you think about that.

tools/gbench/report.py
JBakamovic: Done. #740

tools/gbench/report.py
JBakamovic: > I agree that the old hardcoding of 0.05 is wrong, but this is not correct either.
> The specified utest_alpha (passed to generate_difference_report()) should be propagated/used.

You're actually referring to [this line](https://github.com/google/benchmark/blob/daa23894297781fdb1192a35d60eba1b5ad728d4/tools/gbench/report.py#L243). If so, I will make a change and send it another PR if that's what you would like.

tools/gbench/report.py
```diff
@@ -232,7 +240,7 @@ def find_test(name):
             fmt_str = "{}{:<{}s}{endc}{}{:+16.4f}{endc}{}{:+16.4f}{endc}{:14.0f}{:14.0f}{endc}{:14.0f}{:14.0f}"
 
             def get_color(res):
-                if res > 0.05:
+                if res > UTEST_DEFAULT_ALPHA:
```
LebedevRI: Magic numbers are fun, aren't they? :)
Although this is also `0.05`, this is *not* `UTEST_DEFAULT_ALPHA`,
these are used to color the change of the timers, note the `elif`.

tools/gbench/report.py
```diff
@@ -192,21 +206,15 @@ def generate_difference_report(
         json2,
         display_aggregates_only=False,
         utest=False,
-        utest_alpha=0.05,
+        utest_alpha=UTEST_DEFAULT_ALPHA,
```
LebedevRI: Ok, i think this can be changed.

tools/gbench/report.py
```diff
@@ -437,7 +632,7 @@ def test_utest(self):
         ]
         json1, json2 = self.load_results()
         output_lines_with_header = generate_difference_report(
-            json1, json2, utest=True, utest_alpha=0.05, use_color=False)
+            json1, json2, utest=True, utest_alpha=UTEST_DEFAULT_ALPHA, use_color=False)
```
LebedevRI: Ah wait, I'm not sure it is a good idea to change this.
This is a test, and it is very specifically testing what happens with `utest_alpha=0.05`.
If you pass `UTEST_DEFAULT_ALPHA`, and then change it, the test will need updating.

tools/gbench/report.py
```diff
@@ -505,7 +778,7 @@ def test_utest(self):
         json1, json2 = self.load_results()
         output_lines_with_header = generate_difference_report(
             json1, json2, display_aggregates_only=True,
-            utest=True, utest_alpha=0.05, use_color=False)
+            utest=True, utest_alpha=UTEST_DEFAULT_ALPHA, use_color=False)
```
LebedevRI: Same.

tools/gbench/report.py
JBakamovic: Ok, I misunderstood this one then.

tools/gbench/report.py
JBakamovic: Alright, it makes sense. Then I think reverting the whole change makes most sense because most occurrences of it are in unit tests. Do you agree?

LebedevRI: I have re-read the diff, and i'm not sure what is cleaner.
Out of all the places where `0.05` is used, only one can be replaced with `UTEST_DEFAULT_ALPHA`.
So yes, the situation will not be great either way. I'd undo these `UTEST_DEFAULT_ALPHA` changes. :/

JBakamovic: Agreed. I will revert the change.

tools/gbench/report.py
JBakamovic: Alright, I think I'm on my way finishing this change. Please see 9021718. `generate_difference_report` is now a consumer of data fed by `get_json_difference_report`.

One more thing I've yet to be done is to re-implement the unit-tests to complement this change.

JBakamovic: One more thing I would like to note and which is related to your comment in #740 is that `print_utest` now became much less involved and looks more natural to what is supposed to actually do.

tools/gbench/report.py
```diff
@@ -215,64 +225,112 @@ def find_test(name):
...
-
-    return output_strs
-
+            have_optimal_repetitions, cpu_pvalue, time_pvalue = calc_utest(
+                extract_field(partition, 'real_time'),
+                extract_field(partition, 'cpu_time')
+            )
+            if cpu_pvalue and time_pvalue:
+                utest_results = {
+                    'stable': have_optimal_repetitions,
```
LebedevRI: I'm not sure `stable` is correct word here.
It should be `sufficient_observation_count`.
It *looks* like they say the same thing, but the meaning is slightly different.

JBakamovic: `sufficient_observation_count` seems a bit overly wordy IMO. Perhaps, `optimal_repetitions` if `stable` is not satisfying?

Newest changes are up so please have a look at them when you find it appropriate.

LebedevRI: `sufficient` has a very direct meaning based on https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mannwhitneyu.html
> Use only when the number of observation in each sample is > 20

What is the definition of `optimal`, `stable` in this case?
We don't know that having less/more repetitions would have degraded the u test meaningfulness..

JBakamovic: > What is the definition of optimal, stable in this case?

I was following the wording of existing code (e.g. `UTEST_OPTIMAL_REPETITIONS`) so if we decide that `optimal` is not good enough changes should be done to that parts of the source code as well.

> sufficient has a very direct meaning based on https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mannwhitneyu.html

Alright, makes sense. So `sufficient_observation_count` it is?

JBakamovic: @LebedevRI ping :)

fsaintjacques:
ping?

JBakamovic:
> ping?

@fsaintjacques I believe there is not enough interest from maintainers side and PR is now somewhat stale.

dominichamon:
my apologies, i missed that this was ready for further reviews. If you'd be willing to resolve the conflicts or restart the PR i'd be happy to take a look.

JBakamovic:
@dominichamon Give me few days and I'll see what I can do. It's been a while.

JBakamovic:
I am closing this PR in favor of #1042 .

PR #740, opened 2018-12-07
Refactor U-Test calculation into separate function.

JBakamovic:
And implement 'print_utest' functionality in terms of it.

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -143,11 +143,7 @@ def extract_field(partition, field_name):
     rhs = [x[field_name] for x in partition[1]]
     return [lhs, rhs]
 
-
-def print_utest(partition, utest_alpha, first_col_width, use_color=True):
-    timings_time = extract_field(partition, 'real_time')
-    timings_cpu = extract_field(partition, 'cpu_time')
-
+def calc_utest(timings_cpu, timings_time):
     min_rep_cnt = min(len(timings_time[0]),
                       len(timings_time[1]),
                       len(timings_cpu[0]),
@@ -155,21 +151,33 @@ def print_utest(partition, utest_alpha, first_col_width, use_color=True):
 
     # Does *everything* has at least UTEST_MIN_REPETITIONS repetitions?
     if min_rep_cnt < UTEST_MIN_REPETITIONS:
-        return []
-
-    def get_utest_color(pval):
-        return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
+        return False, None, None
 
     time_pvalue = mannwhitneyu(
         timings_time[0], timings_time[1], alternative='two-sided').pvalue
     cpu_pvalue = mannwhitneyu(
         timings_cpu[0], timings_cpu[1], alternative='two-sided').pvalue
 
+    return (min_rep_cnt >= UTEST_OPTIMAL_REPETITIONS), cpu_pvalue, time_pvalue
+
+def print_utest(partition, utest_alpha, first_col_width, use_color=True):
+    def get_utest_color(pval):
+        return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
+
+    timings_time = extract_field(partition, 'real_time')
+    timings_cpu = extract_field(partition, 'cpu_time')
+    optimal_repetitions, cpu_pvalue, time_pvalue = calc_utest(timings_cpu, timings_time)
+
+    # Check if we failed miserably with minimum required repetitions for utest
+    if not optimal_repetitions and cpu_pvalue is None and time_pvalue is None:
+        return []
+
     dsc = "U Test, Repetitions: {} vs {}".format(
         len(timings_cpu[0]), len(timings_cpu[1]))
     dsc_color = BC_OKGREEN
 
-    if min_rep_cnt < UTEST_OPTIMAL_REPETITIONS:
+    # We still got some results to show but issue a warning about it.
+    if not optimal_repetitions:
         dsc_color = BC_WARNING
         dsc += ". WARNING: Results unreliable! {}+ repetitions recommended.".format(
             UTEST_OPTIMAL_REPETITIONS)
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -166,10 +166,10 @@ def print_utest(partition, utest_alpha, first_col_width, use_color=True):
 
     timings_time = extract_field(partition, 'real_time')
     timings_cpu = extract_field(partition, 'cpu_time')
-    optimal_repetitions, cpu_pvalue, time_pvalue = calc_utest(timings_cpu, timings_time)
+    have_optimal_repetitions, cpu_pvalue, time_pvalue = calc_utest(timings_cpu, timings_time)
 
     # Check if we failed miserably with minimum required repetitions for utest
-    if not optimal_repetitions and cpu_pvalue is None and time_pvalue is None:
+    if not have_optimal_repetitions and cpu_pvalue is None and time_pvalue is None:
         return []
 
     dsc = "U Test, Repetitions: {} vs {}".format(
@@ -177,7 +177,7 @@ def print_utest(partition, utest_alpha, first_col_width, use_color=True):
     dsc_color = BC_OKGREEN
 
     # We still got some results to show but issue a warning about it.
-    if not optimal_repetitions:
+    if not have_optimal_repetitions:
         dsc_color = BC_WARNING
         dsc += ". WARNING: Results unreliable! {}+ repetitions recommended.".format(
             UTEST_OPTIMAL_REPETITIONS)


JBakamovic:
Can you please trigger Travis again? It looks like a flakiness not related to this change because `apt-get install` seem to timeout.

tools/gbench/report.py
```diff
@@ -142,33 +142,41 @@ def extract_field(partition, field_name):
...
 
+    return (min_rep_cnt >= UTEST_OPTIMAL_REPETITIONS), cpu_pvalue, time_pvalue
+
+def print_utest(partition, utest_alpha, first_col_width, use_color=True):
+    def get_utest_color(pval):
+        return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
+
+    timings_time = extract_field(partition, 'real_time')
+    timings_cpu = extract_field(partition, 'cpu_time')
+    optimal_repetitions, cpu_pvalue, time_pvalue = calc_utest(timings_cpu, timings_time)
```
LebedevRI: `timings_time` and `timings_cpu` are not used in this function, only the `len(timings_cpu[0]), len(timings_cpu[1])`.
How about you pass `partition` into the `calc_utest()`, and make it additionally return those lenghts?

tools/gbench/report.py
```diff
@@ -142,33 +142,41 @@ def extract_field(partition, field_name):
...
+    if not optimal_repetitions and cpu_pvalue is None and time_pvalue is None:
+        return []
+
     dsc = "U Test, Repetitions: {} vs {}".format(
         len(timings_cpu[0]), len(timings_cpu[1]))
     dsc_color = BC_OKGREEN
 
-    if min_rep_cnt < UTEST_OPTIMAL_REPETITIONS:
+    # We still got some results to show but issue a warning about it.
+    if not optimal_repetitions:
```
LebedevRI: Hm, given that there already is `UTEST_OPTIMAL_REPETITIONS`, i guess `have_optimal_repetitions` would be slightly better.

JBakamovic: Done in the 622d96c commit.

tools/gbench/report.py
JBakamovic: Hmmm, I'm not sure about this one. `calc_utest()`, as it's name implies, return a u-test result no less no more. Narrowing the scope of the functions usually results in less bugs because there are less things that you can do wrong. So, tbh, I am not sure if this change would improve on the code.

JBakamovic: I believe that this change now became sort of obsolete (or irrelevant) because IMO is that now we have much better solution with changes being done in #737. See 9021718 to get a feeling how it looks like.

dominichamon:
good to merge?

JBakamovic:
> good to merge?

While I was working on it, focus was actually on #738 which includes this change among others, but that pull-request kinda stalled because reviewer seem not to care to reply anymore.

IIRC I think this change has a subtle bug. I'll check it.

dominichamon:
i was waiting for this to land first as a refactor base on which #738 can be built. Could you fix this up to merge first and then we'll rebase #738 on top of it?

LebedevRI:
I don't want to block these tooling changes as long as they improve things overall
(no code duplication, etc).

But i am somewhat conflicted here. I'd recommend to use proper performance tracking suite,
e.g. llvm lnt / test-suite, not trying to piggy-back on these tools..

JBakamovic:
So you decided to waste my time by requesting _couple_ of non-trivial review changes, which I did in a very timely manner, and then suddenly stopped responding? That is both disrespectful to my time and effort I've put into crafting that change.

> But i am somewhat conflicted here. I'd recommend to use proper performance tracking suite,
e.g. llvm lnt / test-suite, not trying to piggy-back on these tools..

And purpose of this library is about exactly what if not measuring performance? You're fine with printing the timings on the console but not writing them down into a file ... Interesting view.

LebedevRI:
> So you decided to waste my time by requesting _couple_ of non-trivial review changes, which I did in a very timely manner, and then suddenly stopped responding? That is both disrespectful to my time and effort I've put into crafting that change.

I'm sorry, i work at this library in the exact same role you do (as in, i don't own it, i'm not responsible for maintaining it), if i have a free moment to work on it then i work on an issue of my choosing, else i don't..
Like i said, this looks reasonable, and i'm not going to block this.
 
> > But i am somewhat conflicted here. I'd recommend to use proper performance tracking suite,
> > e.g. llvm lnt / test-suite, not trying to piggy-back on these tools..
> 
> And purpose of this library is about exactly what if not measuring performance?
It does measure performance, does it not? This is about comparing measurements.
> You're fine with printing the timings on the console but not writing them down into a file ... Interesting view.

Again, the timings are nicely writeable into file via https://github.com/google/benchmark#output-files
This is about *comparing* those written measurements, and writing the *comparison*.

JBakamovic:
> This is about comparing those written measurements, and writing the comparison. 

I am having a hard time to understand what is the problem with writing the benchmark diff results to the file? Should you wish to implement a tool which, let's say plots the data from different (or consecutive runs of) benchmarks, you can't do that without grepping the stdout. That is both very inconvenient and error-prone. And probably very platform-specific.This PR is only about making this whole thing easier.

Anyways, if anyone feels that #738 work should be continued please let someone finalize the review of my latest patch and then we can continue from there.

dominichamon:
> So you decided to waste my time by requesting _couple_ of non-trivial review changes, which I did in a very timely manner, and then suddenly stopped responding? That is both disrespectful to my time and effort I've put into crafting that change.

First of all, please take the communication back to a more respectful level. All of us working on this library do it as a volunteer effort; it isn't a full-time job for any of us. As I'm sure you know, time and effort are precious commodities and we have to do what we can to prioritise where we need to.

If you don't want to "waste your time" working on it any more then I completely understand and would never ask more of any contributor than they're able to donate freely.

> 
> > But i am somewhat conflicted here. I'd recommend to use proper performance tracking suite,
> > e.g. llvm lnt / test-suite, not trying to piggy-back on these tools..
> 
> And purpose of this library is about exactly what if not measuring performance? You're fine with printing the timings on the console but not writing them down into a file ... Interesting view.

It seems fairly in line with the well-worn Unix view of the world, doesn't it? Outputting to the console allows piping to any other tool. With the JSON output from the original library, and tools such as `jq`, it's possible to do quite a lot. This particular tooling that you're working on in this PR is, as @LebedevRI has said, specifically for diffing two sets of benchmark results. _How_ that diff is then output is something we can discuss, but outputting to console and using shell tools to redirect to a file isn't a crazy idea.

Either way, this PR as it stands is about refactoring and i'll review that in isolation, then go back to #738 once this is merged.

tools/gbench/report.py
```diff
@@ -143,33 +143,41 @@ def extract_field(partition, field_name):
...
-    def get_utest_color(pval):
-        return BC_FAIL if pval >= utest_alpha else BC_OKGREEN
+        return False, None, None
 
     time_pvalue = mannwhitneyu(
         timings_time[0], timings_time[1], alternative='two-sided').pvalue
     cpu_pvalue = mannwhitneyu(
         timings_cpu[0], timings_cpu[1], alternative='two-sided').pvalue
 
+    return (min_rep_cnt >= UTEST_OPTIMAL_REPETITIONS), cpu_pvalue, time_pvalue
```
dominichamon: you can simplify this to `return True, cpu_pvalue, time_pvalue` if you want.

JBakamovic:
> First of all, please take the communication back to a more respectful level. All of us working on this library do it as a volunteer effort; it isn't a full-time job for any of us. As I'm sure you know, time and effort are precious commodities and we have to do what we can to prioritise where we need to.

I am the last person here being disrespectful after being ignored for 4.5 months and now all of the sudden `I don't want to block these tooling changes ...`. Being dissatisfied is probably a better formulation.

> If you don't want to "waste your time" working on it any more then I completely understand and would never ask more of any contributor than they're able to donate freely.

I call "wasting someone's time" when you silently choose to ignore work that has been done after substantial follow-up work has been _demanded_ from that _very same person_. That is _not_ nice whether it is on a voluntary basis or not. It is just impolite.

> It seems fairly in line with the well-worn Unix view of the world, doesn't it? Outputting to the console allows piping to any other tool. With the JSON output from the original library, and tools such as `jq`, it's possible to do quite a lot.

It is in-line with Unix way-of-thinking but this library isn't meant only to be run on Unix-oid systems AFAIK. Wouldn't you agree that writing a tool which reads from a well-established format with wide-spread JSON libraries would be trivial for anyone out there? OTOH grepping the stdout not so much ... IMO

> This particular tooling that you're working on in this PR is, as @LebedevRI has said, specifically for diffing two sets of benchmark results.

But that functionality is already there even without this PR ... I am just trying to write those diff results down to a JSON file. Nothing more, nothing less :no_idea:

> _How_ that diff is then output is something we can discuss

I am open to suggestions. Code probably evolved from the last time I worked on this PR so ...

> Either way, this PR as it stands is about refactoring and i'll review that in isolation, then go back to #738 once this is merged.

:+1:

PR #748, opened 2018-12-23
[JSON] add threads and repetitions to the json output

BaaMeow:
Following along with #616's comments to separate out the obviously mergeable things, this PR adds new output in the JSON for threads and repetitions! This helps tools organize the output benchmark data a bit more easily.


test/complexity_test.cc
```diff
@@ -28,6 +28,8 @@ int AddComplexityTest(std::string test_name, std::string big_o_test_name,
   AddCases(TC_JSONOut, {{"\"name\": \"%bigo_name\",$"},
                         {"\"run_name\": \"%name\",$", MR_Next},
                         {"\"run_type\": \"aggregate\",$", MR_Next},
+                        {"\"repetitions\": %int,$", MR_Next},
+                        {"\"threads\": %int,$", MR_Next},
```
LebedevRI: Here and elsewhere - `threads` is never(?) auto-detected, thus it is not some `%int`, but should instead be explicitly checked.

BaaMeow: Alright, done!

BaaMeow:
Well, there's no real way to correlate some top-level "this is a family of runs" block in the JSON, and the various reported runs / samples in the output. Though, if we go with the "family" id in #616 then we can create a few top-level blocks in the JSON describing each "family" along with the parameters (threads, etc.) pass to a `BENCHMARK(...)` call or a manual benchmark registration.

Other than that, I have no idea where I would stick "repetition"!

test/complexity_test.cc
LebedevRI: Thanks.
Same for `repetitions`, i 'forgot' to mention it here.

LebedevRI:
> Other than that, I have no idea where I would stick "repetition"!

Yes, i agree, this seems to be the only choice right now.

BaaMeow:
Alright, I think I changed all the tests to be directly reflective (except for the Complexity Tests. Since it's added by a function and not linked exactly to a specific BENCHMARK() creation instance, `%int` seems more appropriate since someone can add a Complexity Benchmark with repetitions, call that function, and suddenly end up with failures unrelated to any features they were adding!)

src/json_reporter.cc
```diff
@@ -178,6 +178,8 @@ void JSONReporter::PrintRunData(Run const& run) {
     }
     BENCHMARK_UNREACHABLE();
   }()) << ",\n";
+  out << indent << FormatKV("repetitions", run.repetitions) << ",\n";
```
LebedevRI: @dominichamon @EricWF any better ideas for the name of this field?
(see my previous comments in this PR)

dominichamon: `repetitions` is the right name but it shouldn't be reported at the per-run level. i think a repetition index might be the correct way to go without repetitions as you can, in tooling, determine what the max repetition index is for a given run and group them together easily.

BaaMeow:
I fixed this to report `repetition_index` on every non-aggregate, and both `repetition_index` and `repetitions` on every iteration. It's a bit coarse but until the v2 of JSON I don't think I can make a sweeping breaking change to the actual JSON structure, where I would have a dedicated `"benchmark_registrations": { ... }` block that describes every registered benchmark in a single run so that someone could get good top-level data about each benchmark registered (from name to # of aggregate stats and other things).

I think this is a good compromise until that bright, beautiful future...

dominichamon:
This does seem uncontroversial, yes. Can you rebase?

BaaMeow:
@dominichamon Tests have to run, but this one should be all set!

include/benchmark/benchmark.h
```diff
@@ -1358,13 +1360,16 @@ class BenchmarkReporter {
 
     std::string benchmark_name() const;
     BenchmarkName run_name;
-    RunType run_type;          // is this a measurement, or an aggregate?
+    RunType run_type;  // is this a measurement, or an aggregate?
```
dominichamon: no need for this comment; it's a `RunType`, so we know what it might be :)

https://google.github.io/styleguide/cppguide.html#Variable_Comments

src/benchmark_runner.cc
```diff
@@ -151,7 +155,7 @@ class BenchmarkRunner {
 
     for (int repetition_num = 0; repetition_num < repeats; repetition_num++) {
       const bool is_the_first_repetition = repetition_num == 0;
-      DoOneRepetition(is_the_first_repetition);
+      DoOneRepetition(is_the_first_repetition, repetition_num);
```
dominichamon: i believe we don't need both `is_the_first_repetition` and `repetition_num`. We can derive the first from the second in `DoOneRepetition`.

src/complexity.cc
```diff
@@ -208,11 +212,13 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
   // Only add label to mean/stddev if it is same for all runs
   Run rms;
   rms.run_name = run_name;
-  big_o.report_label = reports[0].report_label;
   rms.run_type = BenchmarkReporter::Run::RT_Aggregate;
   rms.aggregate_name = "RMS";
   rms.report_label = big_o.report_label;
   rms.iterations = 0;
+  rms.repetition_index = -1;
```
dominichamon: given you use -1 as a sentinel, you might want to document this at the declaration of `repetition_index` in `Run`.

https://google.github.io/styleguide/cppguide.html#Variable_Comments

test/complexity_test.cc
```diff
@@ -52,7 +56,7 @@ int AddComplexityTest(std::string test_name, std::string big_o_test_name,
 // --------------------------- Testing BigO O(1) --------------------------- //
 // ========================================================================= //
 
-void BM_Complexity_O1(benchmark::State& state) {
+void BM_Complexity_O1(benchmark::State &state) {
```
dominichamon: why did these &s change? we should be normalised across the project to & with the type, afaik.

BaaMeow: That's what the clang format did, but I'll change it back!

src/benchmark_runner.cc
BaaMeow: Fixed!

include/benchmark/benchmark.h
BaaMeow: Got it!

test/complexity_test.cc
dominichamon: clang format will change things to the perceived "majority" of the file. i guess this file is mostly wrong :P

.clang-format
```diff
@@ -3,3 +3,4 @@ Language:        Cpp
 BasedOnStyle:  Google
 ...
 
+PointerAlignment: Left
```
dominichamon: :+1:

dominichamon:
thank you!

PR #749, opened 2018-12-25
Custom Timer Units: Design Attempt #1

BaaMeow:
This Pull Request is a full implementation of how custom time values might integrate into the Library seamlessly and efficiently, after some thought post-discussion in #620.

The goal of this design was to make it easy to reach in and changes the meaning of the reported values without having to change the backing storage (`double` for all iteration results) and still allow the library to work as intended (e.g., it expects `seconds` from the SetIterationTime function so it can produce estimates).

The way it works is simple: we add a second overload to `BENCHMARK(SomeBench)->UseManualTime( ... );` that takes 3 arguments:

```
Benchmark* UseManualTime(
     std::string unit_name, 
     std::string abbreviation, 
     ManualTimeCostFunc* cost_function);
```

where `ManualTimeCostFunc` is `typedef double(ManualCostFunc)(double);`.

It is meant to be used like this:
```
double cycles_to_seconds( double measurement ) {
     return /* your conversion from ic to seconds here */;
}

BENCHMARK(SomeBench)->UseManualTime("instruction cycles", "ic", cycles_to_seconds);
```

Serialization into JSON is done normally: CPU time measurements are unaffected, just the `"real_time"` field and the `"time_unit"` field reporting `"instruction cycles"` instead of `"ns"`. It also produces a new field in the JSON called `"cpu_time_unit"` that gives the traditional `"ns"`/`"us"`/`"ms"` in it, if and only if `"real_time"` is associated with some custom unit.

Console output is done using the abbreviation passed in. Because abbreviations could be longer than 4 spots (which is what is allotted by the benchmark library currently for printing), I added some logic for aligning the print that seems to be working well, considering it passes the Console and JSON tests I whipped up for it!

[Here is some exemplar JSON data](https://gist.github.com/BaaMeow/f0e47a2915153ec3fc9e648443513644) of the results from some fictitious benchmarks: the ManualTime with specific suffixes is at the end of it.

And below is a (very large) screenshot of some console output testing I did for some empty benchmarks. As you can see, it even accommodates when someone specifies an obscenely long suffix/abbreviation for their type:

Console Output of Timers


src/reporter.cc
```diff
@@ -91,6 +91,11 @@ std::string BenchmarkReporter::Run::benchmark_name() const {
 }
 
 double BenchmarkReporter::Run::GetAdjustedRealTime() const {
+  if (use_manual_time && manual_time->customized_units) {
+    double new_time = real_accumulated_time;
```
LebedevRI: That doesn't look right, it will print `132414 cycles` instead of `132 kcycles` e.g?

src/benchmark_runner.cc
```diff
@@ -232,7 +234,12 @@ class BenchmarkRunner {
     // Base decisions off of real time if requested by this benchmark.
     i.seconds = i.results.cpu_time_used;
     if (b.use_manual_time) {
-      i.seconds = i.results.manual_time_used;
+      if (b.manual_time->cost_in_seconds_) {
```
LebedevRI: Here and elsewhere:
This "if custom units exist - use it, else use some hardcoded default" is kinda ugly,
and is pretty much guaranteed to be got wrong somewhere.
Instead, it should always use custom time units, which is by default just the current seconds.

Now that you put it like that, i guess i'm no longer _fully_ convinced that it is separate from the more custom timers work..

src/reporter.cc
BaaMeow: This is mostly because I'm not sure of if the nano/micro/milli prefixes will apply to whatever the type is, so it's passed through directly (and the printed time_unit value has no prefix). I don't know of a way to signal "and nano / micro / milli / centi / etc. apply here".

src/benchmark_runner.cc
BaaMeow: I could make it so that by-default we serialize "seconds", "s", pass_through_function into ManualTime with a `customized_units == false`. That way, we just always go through `manual_time->cost_in_seconds` rather than checking the difference. It will also make printing easier too, so maybe that's the direction I go in?

LebedevRI: What i meant is, there should not be `customized_units` field.
The units should always be customized, with the default matching the current output.

BaaMeow: I made some changes! Mostly, the defaults are the new `BenchmarkTime` struct defaults. There are less conditionals now too, making it easier to reason about the code and not make future mistakes.

src/reporter.cc
BaaMeow: I made some changes to allow for custom units to be applied to be applied here without checking for customized units or manual time.

src/benchmark_runner.cc
LebedevRI: Thanks. Need to try to review this..

include/benchmark/benchmark.h
```diff
@@ -429,6 +429,46 @@ struct Statistics {
       : name_(name), compute_(compute) {}
 };
 
+// Converts whatever the stored manual time is to
+// seconds, which is used to estimate iterations
+// and similar
+typedef double(BenchmarkTimeCostFunc)(double);
+
+struct BenchmarkTime {
```
dominichamon: does this need to be part of the public API?

dominichamon: it has state, a private section, a user-defined constructor: it's a `class`, not a `struct`.

include/benchmark/benchmark.h
```diff
@@ -429,6 +429,46 @@ struct Statistics {
...
+  BenchmarkTimeCostFunc* to_cost_in_seconds_;
+
+  static double SecondsCost(double measurement) { return measurement; }
```
dominichamon: this should be part of the public API, but i don't think the rest of the class needs to be.

include/benchmark/benchmark.h
```diff
@@ -429,6 +429,46 @@ struct Statistics {
       : name_(name), compute_(compute) {}
 };
 
+// Converts whatever the stored manual time is to
+// seconds, which is used to estimate iterations
+// and similar
+typedef double(BenchmarkTimeCostFunc)(double);
+
+struct BenchmarkTime {
+  BenchmarkTimeCostFunc* to_cost_in_seconds_;
```
dominichamon: if this is public, no _ at the end please.

include/benchmark/benchmark.h
```diff
@@ -788,8 +828,12 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Arg(int64_t x);
 
-  // Run this benchmark with the given time unit for the generated output report
+  // Run this benchmark with the given time unit for the generated output report,
+  // and the given cpu multiplier for the second overload.
+  // If only one is specified it is used as the time units for both.
   Benchmark* Unit(TimeUnit unit);
+  Benchmark* Unit(TimeUnit unit, TimeUnit cpu_unit);
+  Benchmark* CpuUnit(TimeUnit unit);
```
dominichamon: s/unit/cpu_unit/

include/benchmark/benchmark.h
```diff
@@ -888,7 +932,15 @@ class Benchmark {
   // SetIterationTime(seconds) to report the measured time, which will be used
   // to control how many iterations are run, and in the printing of items/second
   // or MB/second values.
+  // The second overload creates a manual time which measures in some arbitrary 
+  // value set by the user. The Cost Function returns the "cost" of a single 
+  // benchmark iteration's in seconds, to be used to determine the cost
+  // of a single run of the benchmark max iteration estimation and so on.
+  // The unit multiplier changes
```
dominichamon: incomplete comment

include/benchmark/benchmark.h
```diff
@@ -888,7 +932,15 @@ class Benchmark {
...
   Benchmark* UseManualTime();
+  Benchmark* UseManualTime(
+      std::string units, TimeUnit unit_multiplier = kSecond,
```
dominichamon: `const std::string&` please

dominichamon: is there ever a time to call `UseManualTime("myunits")`? Ie, not use the second and third parameters? Consider either removing the default parameters from here or having all the parameters default to sensible ... defaults.

src/benchmark.cc
```diff
@@ -336,6 +340,31 @@ ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
 
 }  // end namespace internal
 
+BenchmarkTime::BenchmarkTime()
+    : to_cost_in_seconds_(SecondsCost),
+      cpu_unit_multiplier_(TimeUnit::kNanosecond),
+      unit_multiplier_(TimeUnit::kNanosecond),
+      unit_name_("s"),
```
dominichamon: we're defaulting to "s" but with nanosecond multipliers? that might warrant a comment to explain it.

src/benchmark_register.cc
```diff
@@ -265,7 +265,17 @@ Benchmark* Benchmark::Arg(int64_t x) {
...
+  return Unit(unit, unit);
+}
+
+Benchmark* Benchmark::CpuUnit(TimeUnit cpu_unit) {
+  time_.SetCpuUnitString(cpu_unit);
+  return this;
+}
+
+Benchmark* Benchmark::Unit(TimeUnit unit, TimeUnit cpu_unit) {
+  CpuUnit(cpu_unit);
```
dominichamon: it's marginally more readable to not call `CpuUnit` here but instead call `time_.SetCpuUnitString(...)`. It saves a bit of cognitive load.

src/benchmark_register.cc
```diff
@@ -402,9 +412,17 @@ Benchmark* Benchmark::UseRealTime() {
...
+}
+
+Benchmark* Benchmark::UseManualTime(
+    std::string name, TimeUnit unit_multiplier,
+    BenchmarkTimeCostFunc* cost_function) {
   CHECK(!use_real_time_)
       << "Cannot set UseRealTime and UseManualTime simultaneously.";
   use_manual_time_ = true;
+  time_.SetUnitString(unit_multiplier, name);
+  time_.to_cost_in_seconds_ = cost_function ? cost_function : BenchmarkTime::SecondsCost;
```
dominichamon: you can always set this to cost_function. If it's left unset, it'll be defaulted to SecondsCost. If the user passed nullptr then they likely made a mistake and you should CHECK that.

src/benchmark_runner.cc
```diff
@@ -232,7 +232,7 @@ class BenchmarkRunner {
     // Base decisions off of real time if requested by this benchmark.
     i.seconds = i.results.cpu_time_used;
     if (b.use_manual_time) {
-      i.seconds = i.results.manual_time_used;
+      i.seconds = b.time->Cost(i.results.manual_time_used);
```
dominichamon: this line makes me think that `b.time` should instead be `b.manual_time` to make it clear it's set from 'manual time'.

BaaMeow: `b.time` is used with regular time unit measurements too, as per @LebedevRI's advice: it is used for all time measurements, so maybe I should specify `ManualCost` to be more explicit just for this one member (as that is what it is used for).

src/benchmark_register.cc
BaaMeow: I was trying to make it so if something special was done in CpuUnit or changes were made later, someone would've have to update 2 places, but I will change it!

include/benchmark/benchmark.h
dominichamon: It is to me. I want the public API to be very clear, especially when we have to define structures by necessity in the header.

include/benchmark/benchmark.h
BaaMeow: I kept the last argument default, but with a better defaulted function name. The other 2 should always be specified though, I think.

BaaMeow:
Sorry about the CLA stuff: I'm really terrible at rebasing and merging, so I think I made the bot trip over itself several times.

include/benchmark/benchmark.h
dominichamon: Thanks, but no. Let's keep this focused. I'll take care of Statistics.

src/benchmark.cc
BaaMeow: I added a comment!

BaaMeow:
I think I resolved all the issues and questions! Thanks for reviewing it, @LebedevRI and @dominichamon. I think this will really help people with custom units make a lot of headway with benchmark!

Please do let me know if anything else needs changing. (Also thanks for taking care of `Statistics` so I wouldn't have to do it here, @dominichamon 💖!)

include/benchmark/benchmark.h
```diff
@@ -430,10 +430,50 @@ struct Statistics {
...
+
 struct BenchmarkInstance;
 class ThreadTimer;
 class ThreadManager;
 
+class BenchmarkTime {
+ public:
+  BenchmarkTime();
+
+  void SetManualCostFunction(BenchmarkTimeCostFunc* to_time_cost_in_seconds);
```
LebedevRI: I'm not sure we want to spell `Manual` here.
I understand that now it only makes sense for manual time,
but if i do proceed into custom timers, it will be needed for non-manual time,
and i think we'll end up with either two identical functions with and without `Manual`,
or forever stuck with this function.

include/benchmark/benchmark.h
```diff
@@ -430,10 +430,50 @@ struct Statistics {
...
+
+  TimeUnit GetTimeUnit() const { return unit_multiplier_; }
+
+  TimeUnit GetCpuTimeUnit() const { return cpu_unit_multiplier_; }
+
+  const std::string& GetUnitString() const { return unit_; }
+
+  const std::string& GetCpuUnitString() const { return cpu_unit_; }
+
+  double ManualCost(double measurement) const {
```
LebedevRI: Same here, i'd highly prefer to avoid spelling `Manual`.

include/benchmark/benchmark.h
```diff
@@ -888,7 +932,15 @@ class Benchmark {
   // SetIterationTime(seconds) to report the measured time, which will be used
   // to control how many iterations are run, and in the printing of items/second
   // or MB/second values.
+  // The second overload creates a manual time which measures in some arbitrary
+  // value set by the user. The Cost Function returns the "cost" of a single
```
LebedevRI: But that is not really what the `Cost Function` does, isn't it?
It converts the given custom measurement to the seconds.

include/benchmark/benchmark.h
```diff
@@ -1325,7 +1379,7 @@ class BenchmarkReporter {
         : run_type(RT_Iteration),
           error_occurred(false),
           iterations(1),
-          time_unit(kNanosecond),
+          time(),
```
LebedevRI: zero-init at least maybe?

src/benchmark.cc
```diff
@@ -334,6 +338,41 @@ ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
...
+      // this plus above makes "ns",
+      // and so on and so forth when using
+      // setters for BenchmarkTime
+      unit_name_("s"),
+      unit_(GetTimeUnitPrefixString(unit_multiplier_) + unit_name_),
+      cpu_unit_(GetTimeUnitString(cpu_unit_multiplier_)) {}
+
+void BenchmarkTime::SetManualCostFunction(
+    BenchmarkTimeCostFunc* to_time_cost_in_seconds) {
+  CHECK(to_time_cost_in_seconds);
```
LebedevRI: `CHECK(to_time_cost_in_seconds != nullptr);`

include/benchmark/benchmark.h
```diff
@@ -430,10 +430,50 @@ struct Statistics {
...
+// (as they are currently done in the benchmark runner)
+inline double DefaultBenchmarkTimeCostFunc(double measurement) {
+  return measurement;
+}
+
 struct BenchmarkInstance;
 class ThreadTimer;
 class ThreadManager;
 
+class BenchmarkTime {
```
LebedevRI: @dominichamon isn't `BenchmarkTime` a little bit too vague?
How do i know if it is not the actual time, but actually a `BenchmarkTimerDsc`?
How about using the latter name?

LebedevRI: (i'll be ok with either `BenchmarkTimerDsc` or `BenchmarkTimeDsc`, just not with `BenchmarkTime`.)

include/benchmark/benchmark.h
BaaMeow: I guess I'll remove the Manual then...!

include/benchmark/benchmark.h
LebedevRI: Hmm, i think i mis-commented.
```
 The Cost Function returns the "cost" of a single
 benchmark iteration's in seconds, to be used to determine the cost
 of a single run of the benchmark max iteration estimation and so on.
```
My question was about the word `single`.
We don't have any such guarantees, do we?
I think it will always end up being called for whatever the timer has counted,
without division by iteration count, i think.

include/benchmark/benchmark.h
BaaMeow: I went with `Info` instead of `Dsc`, because it's a bit easier to read.

include/benchmark/benchmark.h
```diff
@@ -1379,7 +1379,7 @@ class BenchmarkReporter {
         : run_type(RT_Iteration),
           error_occurred(false),
           iterations(1),
-          time(),
+          time(0),
```
LebedevRI: Oh yeah, this is C++03 header :/
`time(/*nullptr*/0),` then

include/benchmark/benchmark.h
```diff
@@ -444,11 +444,11 @@ struct BenchmarkInstance;
 class ThreadTimer;
 class ThreadManager;
 
-class BenchmarkTime {
+class BenchmarkTimeInfo {
```
LebedevRI: That is better, thanks.
I really would prefer `...Dsc`, but if @dominichamon does not share this view..

dominichamon: I don't know what Dsc stands for intuitively. I don't think Info is right either and I was fine with BenchmarkTime... maybe BenchmarkTimeConverter or BenchmarkTimeOutput if you really want to pick nits.

Weak opinion, weakly held, so don't treat this as blocking.

LebedevRI: **D**e**sc**ription
BenchmarkTime (and _maaaybe_ BenchmarkTimeInfo) to me read as too vague,
i can't tell whether it's not supposed to be holding the actual time of the benchmark,
but the description of the measurements taken, how to interpret them, print them.
Hopefully that makes sense?

dominichamon: Ah fair.. BenchmarkTimeDescription or _maybe_ BenchmarkTimeDesc then?

LebedevRI: Yep, any of these will be fine by me.

README.md
```diff
@@ -487,6 +487,48 @@ static void BM_ManualTiming(benchmark::State& state) {
...
+static double compute_cost(double measurement) {
+  // convert to some representative cost
+  return measurement;
+}
+
+static void BM_ManualTimingWithUnits(benchmark::State& state) {
+  int rotations = state.range(0);
+  std::unique_ptr<int32_t[]> data(new int32_t[rotations]]);
+
+  for (auto _ : state) {
```
EricWF: I think this example is an antipattern. It should use `KeepRunningBatch`, no?
Starting and stopping the timer every iteration certainly is. We don't recommend writing benchmarks like this, we certainly shouldn't provide them as examples.

The library goes out of it's way to give you a better way to do this.

README.md
```diff
@@ -487,6 +487,48 @@ static void BM_ManualTiming(benchmark::State& state) {
...
+    auto elapsed = rtdsc() - start;
+    state.SetIterationTime(static_cast<double>(elapsed));
+  }
+}
+BENCHMARK(BM_ManualTimingWithUnits)
+  ->Range(1, 1<<9)
+  ->UseManualTime("cycles", benchmark::kSeconds, compute_cost);
+```
+
+Internally, it treats the computed cost as "seconds", but this can be made to specify any arbitrary measurement unit for the benchmark. This will affect things like `MinTime`, so choose your limits carefully when specifying manual units that are not clearly convertible to seconds.
```
EricWF: Time is a well-defined unit of measure. It is not an arbitrary measurement unit for the benchmark. We shouldn't design an API around confusing the two.

README.md
BaaMeow: I'm a bit confused. The Readme specifically calls out using the `for` loop syntax as a better KeepRunning loop. Maybe I just lack understanding of how `KeepRunningBatch` is meant to be used to make this scenario better? `Batch` and `KeepRunningBatch` don't even exist on the readme... am I missing something? Is this a v2 feature?

We specify `UseManualTime` here; this means we have to call `SetIterationTime()` each iteration, that's why I'm using the manual timing here. I wish there was a way to batch-measure, but there is no `SetRunTime(total)` that divides by the # of iterations internally like it does for the non-manual measure.

README.md
BaaMeow: I can change the description to talk about it more appropriately...

README.md
EricWF: I can see why that documentation might be confusing.

The `KeepRunningBatch` documentation is found in the header, here: https://github.com/google/benchmark/blob/master/include/benchmark/benchmark.h#L481

The `for (auto)` formulation is preferred as default. It ekes out every bit of optimization possible. When you need to do more complicated things, like batching, you need to use other interfaces.

Everything around the main benchmark loop should be as optimized as possible. Otherwise it introduces noise into the benchmark results.

README.md
EricWF: I don't think we should allow it to represent an arbitrary value to begin with.

If we do allow custom timers, I would like the timer to be provided to the library, and have the library invoke it using the same semantics it does with the other clocks. 

I would also like to move towards using strong types for times. e.g. `std::chrono::time_point` and `std::chrono::duration`. Then we don't need `BenchmarkTimeInfo`.

BaaMeow: I think that's fundamentally cross from what this PR is trying to do, which is allow custom measurements (e.g, with rdpmc and other instruction / cycle performance counters) to be printed out in benchmarking data.

While I generally agree with `std::chrono::time_point` and `std::chrono::duration` being the right choice, they're not possible to expose to the user because this is a C++03-interface library. But lifting the internals to do that is entirely possible.

include/benchmark/benchmark.h
BaaMeow: Changed to `BenchmarkTimeDescription`.

README.md
BaaMeow: I changed the benchmark to be what I think is more appropriate for a rotl measurement, using `KeepRunningBatch`. This serves to benchmark both `KeepRunningBatch` and the advanced `UseManualTime`, but it probably doesn't address your other concerns.

README.md
BaaMeow: I don't really know what to do about this, really. If the PR isn't considered well-motivated enough (providing custom units to be used and reported by the benchmark) then I don't really have a way forward.

EricWF: The benchmark loop has been optimized within an inch of its life. It's as low-cost as possible; this is super important. Otherwise running the loop itself appears in or influences the timing data. This is BAD. It's bad bad not good. 

My main objection to this patch is that it makes you write benchmarks that make the main loop costly. This leads to bad benchmarks and bad benchmarking data.

I'm not saying your PR isn't well motivated: returning arbitrary data or measurements is a reasonable goal that we should support.

My objection is how it does this.

For example, the library automatically determines how many iterations to run the benchmark for by guessing at different iteration counts, and then seeing if the timing results meet a minimum threshold (Ex. if I run the benchmark 100 iterations, does it take 1 second).

In your manual clock example, do you want your "manual time" to be used to determine the number of iterations? Ie. If I say `--benchmark_min_time=2.0`, then benchmarks using your manual clock will run until your manual timer reports `2.0` units of w/e your manual clock represents?

At this point, I think I would be OK with an approach that allowed users to specify a manual timer that acts like the other timers. That is, it the timer should only only stopped/started when the rest of the timers are, and not every iteration.

README.md
```diff
@@ -502,32 +502,34 @@ int64_t rdtsc() {
...
+	// using the KeepRunningBatch() API
+	// divide by the # of rotations
+	// since larger rotations
+	// mean smaller necessary batch sizes
+	// for good measurements!
+	size_t batch_size = static_cast<size_t>(10000000 / rotations);
+	std::unique_ptr<int32_t[]> data(new int32_t[rotations]);
+	while (state.KeepRunningBatch(batch_size)) {
+		for (size_t i = 0; i < batch_size; ++i) {
+			auto start = rdtsc();
```
EricWF: Why is it imperative that `rdtsc` is called every iteration, and not just at the start and end of the benchmark?

BaaMeow: I am building this example based on my understanding of KeepRunningBatch, which is probably very poor right now.

My impression was I needed to do batch_size number of iterations and, when I set UseManualTime (even without this PR), I am required to submit 1 measurement per iteration. If that's not the case I can submit times for batch_size iterations, or maybe just surrounding the whole thing?

I am unsure how KeepRunningBatch works in the presence of UseManualTime, I could probably spend a few more days checking to be sure....

BaaMeow:
Unless I start doing things like templating benchmark classes and functions, there's no way I can supply a chrono-alike set of clocks and their associated time units. I can't think of a design that would either avoid type erasure or create templated classes, and I am not sure I want to commit to either. One of the benefits of Benchmark is the uniform interface, and I can't imagine arbitrary units going through `std::chrono` with a C++03-interface...

Sorry. Maybe I'll try again, but don't let that stop anyone from trying a different design!

PR #752, opened 2019-01-13
Fix header lines length

berestovskyy:
Commit 17a012d7 adds a newline to the str, so the line built from
str.length() is one character longer than it should be.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,6 +9,7 @@
 # Please keep the list sorted.
 
 Albert Pretorius <pretoalb@gmail.com>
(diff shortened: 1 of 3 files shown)


LebedevRI:
Where can the problem be observed?
Can you attach screenshots before/after?

berestovskyy:
> Where can the problem be observed?

It's every horizontal line in the header of the console reporter.

Before:

screen shot 2019-01-13 at 15 15 58

After:

screen shot 2019-01-13 at 15 18 45

PR #753, opened 2019-01-15
Move Statistics struct to internal namespace

dominichamon:


dominichamon:
Probably because the header is a mess. I'd love to simplify it down to just the public stuff with forward declarations for the private bits, but i'm not sure how possible that is.

PR #755, opened 2019-01-21
Add -lpthread to pkg-config Libs.private

domob1812:
Since pthread is required at least for GCC (according to the documentation), this should be reflected by the `pkg-config` file.

The same is, for instance, also done by the gflags library:  https://github.com/gflags/gflags/blob/1005485222e8b0feff822c5723ddcaa5abadc01a/cmake/package.pc.in#L13

### DIFF
--- a/cmake/benchmark.pc.in
+++ b/cmake/benchmark.pc.in
@@ -8,4 +8,5 @@ Description: Google microbenchmark framework
 Version: @VERSION@
 
 Libs: -L${libdir} -lbenchmark
+Libs.private: -lpthread
 Cflags: -I${includedir}


LebedevRI:
> pthread is required at least for GCC (according to the documentation), 

Link?
I suspect the problem is on cmake's side.

domob1812:
> > pthread is required at least for GCC (according to the documentation),
> 
> Link?
> I suspect the problem is on cmake's side.

https://github.com/google/benchmark/blame/master/README.md#L126 says so and I get linker errors without that change.  I'm not a `cmake` expert, though, so perhaps the change should be different - it is similar to what `gflags` does, though.

Also signed the CLA.

domob1812:
@dominichamon:  Yes exactly.  `pkg-config` is supposed to figure out all the flags you have to use for using some library, so you shouldn't need to manually add `-pthread` when you use `benchmark` through that.

@LebedevRI:  As I said, I'm not an expert with `cmake`.  So if there is some other way to pass the flags through to `pkg-config`, that is of course even better.  But it seems that other projects (e.g. `gflags`) are doing it that way, and it solves the issue.

LebedevRI:
I don't want to block the bugfix since the fix seems so trivial, so LG.
But i really don't think this is the correct solution.

LebedevRI:
No further comments from anyone, let's see if merging will cause them then :)

PR #760, opened 2019-02-02
BENCHMARK_CAPTURE() and Complexity()

### LINKED ISSUE #730: BENCHMARK_CAPTURE() + Complexity() - naming problem
If i use `BENCHMARK_CAPTURE()`, pass some args, and then use `Range()` and `Complexity()`,
for the computed complexities, only the name of the original benchmark function will be used
as the name, the name passed to the `BENCHMARK_CAPTURE()` will not be used.

I think it should only ignore the `Range()`

danielharvey458:
Possible fix for naming problem with BENCHMARK_CAPTURE() and Complexity().

I wasn't entirely clear what the expected output of {{benchmark_test.cc}} was, or whether it was validated?

Fixes #730.

### DIFF
--- a/src/complexity.cc
+++ b/src/complexity.cc
@@ -27,12 +27,11 @@ namespace benchmark {
 namespace internal {
 
 // Internal function to drop last field after '/'.
-std::string drop_last_field (const std::string &str, const char sep)
-{
+std::string drop_last_field(const std::string& str, const char sep) {
   return str.substr(0, str.find_last_of(sep));
 }
 
-}
+}  // namespace internal
 
 // Internal function to calculate the different scalability forms
 BigOFunc* FittingCurve(BigO complexity) {
@@ -193,8 +192,8 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
   }
 
-  const std::string run_name
-    = internal::drop_last_field (reports[0].benchmark_name (), '/');
+  const std::string run_name =
+      internal::drop_last_field(reports[0].benchmark_name(), '/');
 
   // Get the data from the accumulator to BenchmarkReporter::Run's.
   Run big_o;
--- a/src/complexity.cc
+++ b/src/complexity.cc
@@ -24,6 +24,16 @@
 
 namespace benchmark {
 
+namespace internal {
+
+// Internal function to drop last field after '/'.
+std::string drop_last_field (const std::string &str, const char sep)
+{
+  return str.substr(0, str.find_last_of(sep));
+}
+
+}
+
 // Internal function to calculate the different scalability forms
 BigOFunc* FittingCurve(BigO complexity) {
   static const double kLog2E = 1.44269504088896340736;
@@ -183,8 +193,8 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
   }
 
-  std::string run_name = reports[0].benchmark_name().substr(
-      0, reports[0].benchmark_name().find('/'));
+  const std::string run_name
+    = internal::drop_last_field (reports[0].benchmark_name (), '/');
 
   // Get the data from the accumulator to BenchmarkReporter::Run's.
   Run big_o;
--- a/test/benchmark_gtest.cc
+++ b/test/benchmark_gtest.cc
@@ -4,6 +4,12 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+namespace benchmark { namespace internal {
+
+std::string drop_last_field (const std::string &, char);
+
+} }
+
 namespace {
 
 TEST(AddRangeTest, Simple) {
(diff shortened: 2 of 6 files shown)


src/complexity.cc
```diff
@@ -24,6 +24,16 @@
 
 namespace benchmark {
 
+namespace internal {
+
+// Internal function to drop last field after '/'.
+std::string drop_last_field (const std::string &str, const char sep)
+{
```
LebedevRI: Please use clang-format

LebedevRI:
Can you show the console output of that test case before and after this change,
so it is easier to understand what the problem is?
(Thought yes, i think this is likely going in the right direction..)

test/benchmark_test.cc
```diff
@@ -218,6 +218,15 @@ void BM_non_template_args(benchmark::State& state, int, double) {
 }
 BENCHMARK_CAPTURE(BM_non_template_args, basic_test, 0, 0);
 
+void BM_complexity_args(benchmark::State &state, int n) {
```
LebedevRI: I don't think this is the right place for these tests.
Maybe it should go to `test/complexity_test.cc`.

dharvey1986:
> Can you show the console output of that test case before and after this change,
> so it is easier to understand what the problem is?
> (Thought yes, i think this is likely going in the right direction..)

Do you mean in this conversation, or somehow in the PR? Example of console output differences:

```
< BM_ComplexityCaptureArgs_BigO                 0.00 N          0.00 N    
< BM_ComplexityCaptureArgs_RMS                     1 %            33 %    
> BM_ComplexityCaptureArgs/capture_test_BigO       0.00 N          0.00 N    
> BM_ComplexityCaptureArgs/capture_test_RMS          0 %            20 %    
```

So `test_case_name` is included in the output.

test/complexity_test.cc
```diff
@@ -176,6 +176,30 @@ ADD_COMPLEXITY_CASES(n_lg_n_test_name, big_o_n_lg_n_test_name,
...
+}
+
+BENCHMARK_CAPTURE(BM_ComplexityCaptureArgs, capture_test, 100)
+    ->Complexity(benchmark::oN)
+    ->Range(1, 2);
+
+const std::string complexity_capture_name =
+    "BM_ComplexityCaptureArgs/capture_test";
+
+ADD_COMPLEXITY_CASES(complexity_capture_name, complexity_capture_name + "_BigO",
```
LebedevRI: Hmm yeah, i think now tests do what they should.

test/complexity_test.cc
```diff
@@ -176,6 +176,30 @@ ADD_COMPLEXITY_CASES(n_lg_n_test_name, big_o_n_lg_n_test_name,
 ADD_COMPLEXITY_CASES(n_lg_n_test_name, big_o_n_lg_n_test_name,
                      rms_o_n_lg_n_test_name, lambda_big_o_n_lg_n);
 
+// ========================================================================= //
+// -------- Testing formatting of Complexity with captured args ------------ //
+// ========================================================================= //
+
+#ifdef BENCHMARK_HAS_CXX11
```
LebedevRI: I'm pretty sure C++11-capable compiler is already required to build the google benchmark, so this isn't needed.

test/benchmark_gtest.cc
```diff
@@ -4,6 +4,12 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+namespace benchmark { namespace internal {
+
+std::string drop_last_field (const std::string &, char);
```
LebedevRI: That's ... interesting.
Honestly, i'm not sure this will always work.
Can you instead stick the declaration into some header?
`src/complexity.h` perhaps?

src/complexity.cc
```diff
@@ -183,8 +192,8 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
   }
 
-  std::string run_name = reports[0].benchmark_name().substr(
-      0, reports[0].benchmark_name().find('/'));
+  const std::string run_name =
+      internal::drop_last_field(reports[0].benchmark_name(), '/');
```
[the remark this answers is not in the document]
dharvey1986: You're absolutely right - and I thought I had an easy fix!

test/benchmark_gtest.cc
LebedevRI: It _probably_ will work. I'm just conflicted about the fact that the function you want to call
in the test, and the actual implementation of the function don't have the _same_ declaration.
Maybe that is ok/allowed, don't know. Looks like a code smell at least.

src/complexity.cc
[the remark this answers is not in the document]
LebedevRI: Yeah, i was afraid to be right yet again :(
Not sure what is the best path forward for that.

I would refactor that as a class (putting [this code](https://github.com/google/benchmark/blob/c9f2693ea97e94c8afcefb57d3074c6a6236ca23/src/benchmark_register.cc#L172-L210) into a function),
add an enum with all the substring types (`arg`, `min_time`, `repeats`, ...),
make said function take a parameter (with that enum type)
and only append the substring types specified in that param.

And call this function from the original place with type `ALL`,
and from here with `ALL & ~arg`.

That would _probably_ be least fragile.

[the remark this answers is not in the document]
dharvey1986: Thanks - that sounds good to me. I'll give it a try, perhaps I'll submit a separate PR since it will be pretty different to this solution!

test/benchmark_gtest.cc
dharvey1986: I'm fairly confident that `const T` and `T` for parameters are interchangeable. I personally mark them `const` in the definition to document that I won't modify the argument in the function body, and not `const` in the header since the caller needn't care for by-value parameters.

All that said, I can't find anything in the standard from a quick glance ensuring that this is OK, and anyway I'm happy to comply with the local conventions.

src/complexity.cc
```diff
@@ -24,6 +24,15 @@
 
 namespace benchmark {
 
+namespace internal {
+
+// Internal function to drop last field after '/'.
```
dominichamon: nit: this isn't what the function does. It drops the last field after 'sep'.

test/benchmark_gtest.cc
dominichamon: This is perfectly reasonable, especially the notion of dropping the const from value types in API declarations. It's a _little_ odd to see it as an extern declaration but it works perfectly.

LebedevRI:
Fixed by #761.

PR #761, opened 2019-02-08
BENCHMARK_CAPTURE() and Complexity() - naming problem

### LINKED ISSUE #730: BENCHMARK_CAPTURE() + Complexity() - naming problem

danielharvey458:
A second attempt to fix #730, inspired by the suggestion by @LebedevRI  in https://github.com/google/benchmark/pull/760. I'm not sure it is exactly what you were suggesting (perhaps it is) but hopefully encapsulating the name in this way is a sensible thing to do.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -256,7 +256,6 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 namespace benchmark {
 class BenchmarkReporter;
 class MemoryManager;
-class BenchmarkName;
 
 void Initialize(int* argc, char** argv);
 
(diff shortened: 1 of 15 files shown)


include/benchmark/benchmark.h
```diff
@@ -1302,6 +1303,81 @@ struct SystemInfo {
...
+  BenchmarkName();
+
+  // Construct BenchmarkName with a given 'root'.
+  BenchmarkName(std::string root);
+
+  // Append a 'field_name' to a given field of the name.
+  // Each appended field_name for a given field will be
+  // separated by a '/', and additionally each field will
+  // be separated by a '/'.
+  BenchmarkName& append(Field field, const std::string& field_name);
```
LebedevRI: `field_name` should probably be `field_content`

include/benchmark/benchmark.h
```diff
@@ -1302,6 +1303,77 @@ struct SystemInfo {
...
 
+// BenchmarkName wraps a std::string containing different fields
+// of a benchmark's name (e.g. the 'root' name, the benchmark's
+// 'args' etc.).
+class BenchmarkName {
+ public:
+  // Different fields of the benchmark's name.
+  enum Field {
+    // The name of the function being invoked in the benchmark.
+    kRoot = 1U,
```
dominichamon: nit: `kName` :)

include/benchmark/benchmark.h
```diff
@@ -1302,6 +1303,77 @@ struct SystemInfo {
...
+  BenchmarkName(std::string root);
+
+  // Append 'field_content' to a given field of the name.
+  // Each appended element of content for a given field will be
+  // separated by a '/', and additionally each field will
+  // be separated by a '/'.
+  BenchmarkName& append(Field field, const std::string& field_content);
+
+  // Get the benchmark's name including all the fields set in 'fields'
+  std::string get(Field fields) const;
```
dominichamon: we shouldn't need `str` below if we have this, as `str() == get(Field::kAll)`, right?

include/benchmark/benchmark.h
```diff
@@ -1302,6 +1303,77 @@ struct SystemInfo {
...
+  bool is_set(Field field, std::size_t enumerator_index) const;
+  size_t start_offset(std::size_t enumerator_index) const;
+  size_t end_offset(std::size_t enumerator_index) const;
+
+  static const char separator = '/';
+  static const std::size_t enumerator_count = 7;
+  static const std::size_t offset_count = enumerator_count - 1;
+
+  std::string name_;
+  uint16_t offsets_[offset_count];
```
dominichamon: just `int` is fine here.

include/benchmark/benchmark.h
```diff
@@ -1302,6 +1303,77 @@ struct SystemInfo {
...
+  // Get the size of the full name of the benchmark.
+  size_t size() const { return name_.size(); }
+
+ private:
+  bool is_set(Field field, std::size_t enumerator_index) const;
+  size_t start_offset(std::size_t enumerator_index) const;
+  size_t end_offset(std::size_t enumerator_index) const;
+
+  static const char separator = '/';
+  static const std::size_t enumerator_count = 7;
```
dominichamon: this seems fragile. it might be worth allowing the enums to take their natural values and using `kAll` (which would have value 7) as the enumerator count. Then you'd just need to shift wherever you access the field in the implementation.

include/benchmark/benchmark.h
```diff
@@ -1302,6 +1303,77 @@ struct SystemInfo {
   BENCHMARK_DISALLOW_COPY_AND_ASSIGN(SystemInfo);
 };
 
+// BenchmarkName wraps a std::string containing different fields
+// of a benchmark's name (e.g. the 'root' name, the benchmark's
+// 'args' etc.).
+class BenchmarkName {
+ public:
+  // Different fields of the benchmark's name.
+  enum Field {
```
dominichamon: thinking more, and looking at the access pattern, i think this is overkill. The sections of the benchmark name are always appended in the same order. Similarly, you only care about dropping 'args' from the complexity bit (as far as i can see).

So why not just have a struct with all the strings stored as individual fields with an accessor `name` and another `NameWithoutArgs`.

Ie:

```c++
struct BenchmarkName {
  std::string name;
  std::string args;
  std::string min_time;
  std::string iterations;
  ...

  std::string name() { return absl::StrJoin({name, args, min_time, iterations...}, "/"); }  // or equivalent non-absl version
  std::string nameNoArgs() { return absl::StrJoin({name, min_time, iterations...}, "/"); }  // also non-absl.
};
```

then the benchmark name is stored (as you have) as a BenchmarkName and we have two simple accessors for getting the formatted version.

include/benchmark/benchmark.h
dharvey1986: True, although `get(Field::kAll)` would, currently, re-build the string component-wise whereas `.str()` just does a string copy. Could add a fast-path into the implementation of `get` though.

include/benchmark/benchmark.h
dharvey1986: I agree this is fragile. With modern c++ we could compute at compile time the number of set bits in `kAll`, which would do the trick, but it has to be done in the header so wouldn't work for <= c++03 clients.

I think your suggestion would make it tricky to chose a subset of name fields - composing the bits of enumerator values seems natural for that.

Of course, as you point out, only `kAll` and `kAll | ~kArgs` are currently used, so we could just add an additional enumerator for 'name without the args' request (or, as you suggest below, drop the enumeration entirely and just have two accessors for the currently-used cases).

include/benchmark/benchmark.h
dharvey1986: That certainly makes the code simpler, I didn't consider it.

I guess the downsides of that are that in the common case (accessing the whole name) you have to re-build the full name every time. The current approach keeps the full name and can just return it. It also means adding a new field to the name requires a little bit more work than just adding an enumeration value (and, er, updating `enumerator_count`).

What do you think? Happy to go for the simpler option if preferred.

include/benchmark/benchmark.h
LebedevRI: Does this have to be in the public header?
I think this is an internal functionality.

dharvey1986: It's a member of `BenchmarkReporter`, which is the only reason it's in the header. I could store a pointer to it in `BenchmarkReporter` so that it only requires a forward-declaration?

src/benchmark_name.cc
```diff
@@ -0,0 +1,57 @@
...
+// Compute the total size of a pack of std::strings
+size_t size_impl() { return 0; }
+
+template <typename Head, typename... Tail>
+size_t size_impl(const Head& head, const Tail&... tail) {
+  return head.size() + size_impl(tail...);
+}
+
+// Join a pack of std::strings using a delimiter
+void join_impl(std::string&, char) {}
```
dominichamon: can you add a comment:

// TODO: use absl::StrJoin

dominichamon:
LGTM. I'll leave @LebedevRI to do the merge honours if they're happy.

LebedevRI:
> LGTM. I'll leave @LebedevRI to do the merge honours if they're happy.

Hm, this has fallen off my radar completely. Sorry.
This looks reasonable to me.

LebedevRI:
@danielharvey458 thank you for the fix!

Issue #671, opened 2018-09-06
Timer names are weird


LebedevRI:
I'm not sure if this can be changed in v2, but i trip on it every time.

| Textual description | googlebenchmark name | common(?) name | `TIME(1)` |
| ------------------------ | ------------------------------- |------------------- |----- |
| the time that has passed on your usual [clock](https://en.wikipedia.org/wiki/Clock) | `Time` | `wall clock` | `real`, `Elapsed (wall clock)` |
| the time spent in the main thread only | `CPU` | ? | ? |
| the total amount of the time the processor[s] has spent doing something (all threads combined) | - (i miss it) | ?  | `user` |

This is really confusing, i think.

PR #763, opened 2019-02-11
Openmp compatibility

bryan-lunt:
This patch makes Google Benchmark compatible with OpenMP and other user-level thread management.
Until now, google benchmark would only report the CPU usage of the master thread if the code being benchmarked used OpenMP or otherwise spawned multiple threads internally.

This version reports the total process CPU usage if the number of google-benchmark threads is set to <= 1 , but reverts to the existing behaviour otherwise.

It may actually be preferable to report the total process CPU usage in all cases, but this is sufficient for my needs.

We have been using google benchmark in our parallel programming class, however, every term students are confused when the CPU time roughly reflects the wall-clock time for parallelized codes doing the same amount of work. This version is also advantageous because it can better demonstrate the overhead of threading, that some tasks take more total CPU time when multi-threaded, and, sometimes, tasks may actually take less overall CPU time.

If my feature patch cannot be merged, I would like to request that the maintainers implement this. It is very important to us.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,6 +428,49 @@ BENCHMARK(BM_test)->Range(8, 8<<10)->UseRealTime();
 
 Without `UseRealTime`, CPU time is used by default.
 
+## CPU timers
+By default, the CPU timer only measures the time spent by the main thread.
+If the benchmark itself uses threads internally, this measurement may not
+be what you are looking for. Instead, there is a way to measure the total
+CPU usage of the process, by all the threads.
+
+```c++
+void callee(int i);
+
+static void MyMain(int size) {
+#pragma omp parallel for
+  for(int i = 0; i < size; i++)
+    callee(i);
+}
+
+static void BM_OpenMP(benchmark::State& state) {
+  for (auto _ : state)
+    MyMain(state.range(0);
+}
+
+// Measure the time spent by the main thread, use it to decide for how long to
+// run the benchmark loop. Depending on the internal implementation detail may
+// measure to anywhere from near-zero (the overhead spent before/after work
+// handoff to worker thread[s]) to the whole single-thread time.
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10);
+
+// Measure the user-visible time, the wall clock (literally, the time that
+// has passed on the clock on the wall), use it to decide for how long to
+// run the benchmark loop. This will always be meaningful, an will match the
+// time spent by the main thread in single-threaded case, in general decreasing
+// with the number of internal threads doing the work.
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->UseRealTime();
+
+// Measure the total CPU consumption, use it to decide for how long to
+// run the benchmark loop. This will always measure to no less than the
+// time spent by the main thread in single-threaded case.
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->MeasureProcessCPUTime();
+
+// A mixture of the last two. Measure the total CPU consumption, but use the
+// wall clock to decide for how long to run the benchmark loop.
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->MeasureProcessCPUTime()->UseRealTime();
+```
+
 ## Controlling timers
 Normally, the entire duration of the work loop (`for (auto _ : state) {}`)
 is measured. But sometimes, it is nessesary to do some work inside of
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -874,11 +874,18 @@ class Benchmark {
   // Same as ReportAggregatesOnly(), but applies to display reporter only.
   Benchmark* DisplayAggregatesOnly(bool value = true);
 
-  // If a particular benchmark is I/O bound, runs multiple threads internally or
-  // if for some reason CPU timings are not representative, call this method. If
-  // called, the elapsed time will be used to control how many iterations are
-  // run, and in the printing of items/second or MB/seconds values.  If not
-  // called, the cpu time used by the benchmark will be used.
+  // By default, the CPU time is measured only for the main thread, which may
+  // be unrepresentable if the benchmark uses threads internally. If called,
+  // the total CPU time spent by all the threads will be measured instead.
+  // By default, the only the main thread CPU time will be measured.
+  Benchmark* MeasureProcessCPUTime();
+
+  // If a particular benchmark should use the Wall clock instead of the CPU time
+  // (be it either the CPU time of the main thread only (default), or the
+  // total CPU usage of the benchmark), call this method. If called, the elapsed
+  // (wall) time will be used to control how many iterations are run, and in the
+  // printing of items/second or MB/seconds values.
+  // If not called, the CPU time used by the benchmark will be used.
   Benchmark* UseRealTime();
 
   // If a benchmark must measure time manually (e.g. if GPU execution time is
@@ -952,6 +959,7 @@ class Benchmark {
   double min_time_;
   size_t iterations_;
   int repetitions_;
+  bool measure_process_cpu_time_;
   bool use_real_time_;
   bool use_manual_time_;
   BigO complexity_;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -875,7 +875,7 @@ class Benchmark {
   Benchmark* DisplayAggregatesOnly(bool value = true);
 
   // By default, the CPU time is measured only for the main thread, which may
-  // be unrepresentable if the benchmark uses threads internally. If called,
+  // be unrepresentative if the benchmark uses threads internally. If called,
   // the total CPU time spent by all the threads will be measured instead.
   // By default, the only the main thread CPU time will be measured.
   Benchmark* MeasureProcessCPUTime();
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -86,11 +86,7 @@ BenchmarkReporter::Run CreateRunReport(
     } else {
       report.real_accumulated_time = results.real_time_used;
     }
-    if (b.threads <= 1) {
-      report.cpu_accumulated_time = results.cpu_time_used;
-    } else {
-      report.cpu_accumulated_time = results.thread_cpu_time_used;
-    }
+    report.cpu_accumulated_time = results.cpu_time_used;
     report.complexity_n = results.complexity_n;
     report.complexity = b.complexity;
     report.complexity_lambda = b.complexity_lambda;
@@ -115,7 +111,7 @@ BenchmarkReporter::Run CreateRunReport(
 // Adds the stats collected for the thread into *total.
 void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
                  ThreadManager* manager) {
-  internal::ThreadTimer timer;
+  internal::ThreadTimer timer(b->measure_process_cpu_time);
   State st = b->Run(iters, thread_id, &timer, manager);
   CHECK(st.iterations() >= st.max_iterations)
       << "Benchmark returned before State::KeepRunning() returned false!";
@@ -123,7 +119,6 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
     MutexLock l(manager->GetBenchmarkMutex());
     internal::ThreadManager::Result& results = manager->results;
     results.iterations += st.iterations();
-    results.thread_cpu_time_used += timer.thread_cpu_time_used();
     results.cpu_time_used += timer.cpu_time_used();
     results.real_time_used += timer.real_time_used();
     results.manual_time_used += timer.manual_time_used();
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -86,7 +86,11 @@ BenchmarkReporter::Run CreateRunReport(
     } else {
       report.real_accumulated_time = results.real_time_used;
     }
-    report.cpu_accumulated_time = results.cpu_time_used;
+	if(b.threads <= 1){
+		report.cpu_accumulated_time = results.cpu_time_used;
+	}else{
+    	report.cpu_accumulated_time = results.thread_cpu_time_used;
+	}
     report.complexity_n = results.complexity_n;
     report.complexity = b.complexity;
     report.complexity_lambda = b.complexity_lambda;
@@ -119,6 +123,7 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
     MutexLock l(manager->GetBenchmarkMutex());
     internal::ThreadManager::Result& results = manager->results;
     results.iterations += st.iterations();
+	results.thread_cpu_time_used += timer.thread_cpu_time_used();
     results.cpu_time_used += timer.cpu_time_used();
     results.real_time_used += timer.real_time_used();
     results.manual_time_used += timer.manual_time_used();
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -111,7 +111,10 @@ BenchmarkReporter::Run CreateRunReport(
 // Adds the stats collected for the thread into *total.
 void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
                  ThreadManager* manager) {
-  internal::ThreadTimer timer(b->measure_process_cpu_time);
+  internal::ThreadTimer timer(
+      b->measure_process_cpu_time
+          ? internal::ThreadTimer::CreateProcessCpuTime()
+          : internal::ThreadTimer::Create());
   State st = b->Run(iters, thread_id, &timer, manager);
   CHECK(st.iterations() >= st.max_iterations)
       << "Benchmark returned before State::KeepRunning() returned false!";
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -86,11 +86,11 @@ BenchmarkReporter::Run CreateRunReport(
     } else {
       report.real_accumulated_time = results.real_time_used;
     }
-	if(b.threads <= 1){
-		report.cpu_accumulated_time = results.cpu_time_used;
-	}else{
-    	report.cpu_accumulated_time = results.thread_cpu_time_used;
-	}
+    if (b.threads <= 1) {
+      report.cpu_accumulated_time = results.cpu_time_used;
+    } else {
+      report.cpu_accumulated_time = results.thread_cpu_time_used;
+    }
     report.complexity_n = results.complexity_n;
     report.complexity = b.complexity;
     report.complexity_lambda = b.complexity_lambda;
@@ -123,7 +123,7 @@ void RunInThread(const BenchmarkInstance* b, size_t iters, int thread_id,
     MutexLock l(manager->GetBenchmarkMutex());
     internal::ThreadManager::Result& results = manager->results;
     results.iterations += st.iterations();
-	results.thread_cpu_time_used += timer.thread_cpu_time_used();
+    results.thread_cpu_time_used += timer.thread_cpu_time_used();
     results.cpu_time_used += timer.cpu_time_used();
     results.real_time_used += timer.real_time_used();
     results.manual_time_used += timer.manual_time_used();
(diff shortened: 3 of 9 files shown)


bryan-lunt:
I signed it! (Signed CLA.)

LebedevRI:
> But before going any further, i have a question:
> @dominichamon @EricWF do we agree that for the normal case of google benchmark threads,
> `ThreadCPUUsage()` would be equivalent of the sum of per-thread `ThreadCPUUsage()`?

As in, are we afraid some other user-created thread will mess the measurements up?
Or that the results won't be equivalent due to some implementation detail of timers themselves?
(in kernel/etc)

dominichamon:
The patch seems reasonable to me. I ran it side-by-side with the master version and didn't see much difference in the basic tests (which use threads) so i might be missing where the benefit is. Could you demonstrate the problem with a code snippet or some output?


```
test/user_counters_test.cc:241: CheckAvgThreadsRate: Check `std::fabs((e.GetCounterAs< double >("foo")) - ((1. / e.DurationCPUTime()))) < ((0.001) * (1. / e.DurationCPUTime()))    ' failed.·                                                                                                                                                                      
  1 test/user_counters_test.cc:241: BM_Counters_AvgThreadsRate/threads:2:
  2 test/user_counters_test.cc:241: expected (double)foo=88.0367 to be EQ to 133.142
  3 test/user_counters_test.cc:241: with tolerance of 0.133142 (0.1%), but delta was -45.1053 (-33.8776%)
```

Looks like some tests need updating :)

bryan-lunt:
@dominichamon 

I created a project that demonstrates the need for this patch:

https://github.com/bryan-lunt-supercomputing/gbench_threads_patch_need_demo

The reason you see nothing different is that you are benchmarking things that are threaded by Google Benchmark itself. But google Benchmark does not give correct results when you benchmark code that spins up and shuts down its own threads, for example, any OpenMP codes.

bryan-lunt:
If you tried to use the Current GoogleBenchmark built-in threads to benchmark an OpenMP code, you'd get something much worse than just nonsense CPU times, you'd get that each GB thread calls the function which internally starts up its own threadpool.
So you'd have N_OMP_THREADS*N_GB_THREADS threads at once.

I've been using my own parameter "num_threads" but if wishes were fishes, it would be nice to be able to ask the total number of threads from within a benchmark, and to do something like "set_user_managed_threads" which results in only one GoogleBenchmark thread, but then inside we can ask the number of threads that were requested.

That way, we could have one and only one "threads" parameter, which I think would make interpreting the final output easier.

LebedevRI:
> If you tried to use the Current GoogleBenchmark built-in threads to benchmark an OpenMP code, you'd get something much worse than just nonsense CPU times, you'd get that each GB thread calls the function which internally starts up its own threadpool.

Yep, that would make no sense in general case.

The current question remains (i have not heard back from @dominichamon / @EricWF) - what
//should// CPU timer be reporting, the whole time spent (i.e. by all the threads) by the benchmark
function, or what it currently does - the time spent by the "main" thread. It might be that there is no
The Right solution, and a switch like `->UseRealTime()` (`->UseProcessorTime()`?) should be added.

> ...
> That way, we could have one and only one "threads" parameter, which I think would make interpreting the final output easier.

No. Nothing like that should be done.
If user code uses it's own threads, there is no sane way to generalize that and display
in report. There is simply can be no guarantee of anything, unless you want to put arbitrary
restrictions on the user threading. E.g. part of the benchmark function may be limited 4 threads,
and then second part is not limited by the thread count.

bryan-lunt:
> No. Nothing like that should be done.
> If user code uses it's own threads, there is no sane way to generalize that and display
> in report. There is simply can be no guarantee of anything, unless you want to put arbitrary
> restrictions on the user threading. E.g. part of the benchmark function may be limited 4 threads,
> and then second part is not limited by the thread count.

Good point.

dominichamon:
Thanks for the motivating example, that helps.

The current code supports the following concept:
* user code can be run on a single thread, or multiple threads in paralllel _by the benchmark framework_
* the benefit (and scaling limitations) of threading is then observed by the thread-0 (main thread) processing being ~constant until we hit CPU pressure, but the items per second processing increasing.

Your motivation, and the more general user thread issue, isn't something we've considered. We already have a 'manual timer' that tells the library to get out of the way, which sets a precedent for having a 'this is user threaded' setting, but that's clunky.

I think @LebedevRI is on to something already with #671 and other discussions about sorting out our timing so we are explicit about per-(benchmark)-thread time vs total process time.

LebedevRI:
> Your motivation, and the more general user thread issue, isn't something we've considered. We already have a 'manual timer' that tells the library to get out of the way, which sets a precedent for having a 'this is user threaded' setting, but that's clunky.

To give an example: (from IRC)
```
<LebedevRI> here, and example in the commit msg: https://github.com/darktable-org/rawspeed/commit/eda7972db1b469548b9bfdd054b5c58f5830e230#diff-e03d58ca0e0ab7c09683945bc5c6e825
<dhamon> in benchmark-threaded cases they also wouldn't be the same.
<LebedevRI> "Time" is wallclock, "CPU" is the time spent by main thread, "CPUTime,s" is the actual cpu time spent by all threads
<LebedevRI> iirc
<dhamon> and cpu/wallclock is, as expected with 8 threads, ~8.
``` 

> I think @LebedevRI is on to something already with #671 and other discussions about sorting out our timing so we are explicit about per-(benchmark)-thread time vs total process time.

I currently _personally_ think it would be possible to solve this by simply measuring this third timer,
and using it instead of the "time of main thread" for the "CPU" column, much like the same way
as with `->UseRealTime()`, `->UseManualTime()` - if user requested so.

I do **not** think it is worth it adding a third column to the console reporter.

pleroy:
I have wanted this feature for years.  I have code where the main thread does nothing but forking a bunch of threads that do CPU-intensive work and synchronize at the end.  At the moment the output looks like this (the second parameter of the benchmark is the number of threads used):
```
------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                       Time         CPU Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------
BM_EphemerisMultithreadingBenchmark/3/1 102495353 ns        0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/2  77841192 ns   156001 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/3  49749029 ns        0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/4  46776591 ns        0 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
BM_EphemerisMultithreadingBenchmark/3/5  46512723 ns   156001 ns        100 +9.98765338665172644e+06 m +1.99941207398748249e+07 m +2.99993670217528194e+07 m
```
The CPU time reported here is completely useless.  I would really like to have a proper display of the total CPU time consumed by all threads, as that would give me a sense of the overhead of the multithreading.

bryan-lunt:
Oh God. I was unaware that UseRealTime() got the total CPU time. So, for my needs, it should be UseManualTime to get the walltime, and my patch to get the total CPU time.

bryan-lunt:
It seems like there is a choice and a choice of reduction here.

- Choice: Does an individual GB thread get its own CPU time, or the CPU time of itself and all subthreads? (Or something else?)
- Reduction choice: When putting together the final report, does the CPU time show the smallest, largest, mean, or sum of all the individual GB thread CPU values?

LebedevRI:
> I think @LebedevRI is on to something already with #671 and other discussions about sorting out our timing so we are explicit about per-(benchmark)-thread time vs total process time.

I have now looked, and while we could rename the "Time" column in reports to "Wall", 
i do not think this would universally improve things, because the "manual time"
is displayed in that "Time" column, not "CPU" column.
So i think this will, again, have to wait for custom timers.

> Now I'm confused, does UseRealTime() get wallclock time into the "Time" column, or not?

`UseRealTime()` does not change anything about how which timer is displayed in which column.
`UseRealTime()` only tells the library to use the wallclock (instead of the "CPU",
the time of the main thread) to decide for how long to run the benchmark.

> In general, for parallel codes (by which I mean user-managed parallelism within the one function, as in my example) we are primarily interested in the wallclock time and secondarily interested in the CPU time.

Yes, i understand the needs, i have one such case myself.

> It seems like there is a choice and a choice of reduction here.
> ...

It looks like i will need to help with this code.

LebedevRI:
Updated with all the stuff!!1 Looks like cla bot loves such pushes :S
Now here is an interesting question, now that i have worked on this code, can i still review the PR? :D
I'd guess not really, perhaps @dominichamon might have a moment.

bryan-lunt:
So who has to sign off on the CLA now?

LebedevRI:
> So who has to sign off on the CLA now?

Mine's signed, it's a 'known' glitch i believe.

dominichamon:
It's up to me to check the commits are all by one of the signed folks (CLA bot doesn't like PRs from multiple people).

It looks like this satisfies the original request, but at the increase in complexity of the API. Is there a way we can change the default behaviour to be something sensible and expected without adding another boolean option and benchmark method? (the answer might be no, but i should ask).

LebedevRI:
> It looks like this satisfies the original request,

Yes.

> but at the increase in complexity of the API.

Yes :(

> Is there a way we can change the default behaviour to be something sensible and expected without adding another boolean option and benchmark method? (the answer might be no, but i should ask).

The alternative would be to always use whole-process cpu clock instead of the
main-thread-only cpu clock when the `->Threads()` wasn't called.
Pros: 
* no need for all that extra API
* no `/process_time` suffix in benchmark name.

Cons:
* Such change will change the timer used by benchmarks that do not use `->Threads()`.
  If any of these benchmarks used threads internally, the reported time will naturally be different.
  Do we know that it is what everyone affected actually wanted?
  Maybe they **actually** used the current timing method.
  Also, there will be no way to switch back to the previous (current) timer method.

The other alternative is likely again the old good 'custom timers'. But that feature is nowhere near.

LebedevRI:
> > > It looks like this satisfies the original request,
> > 
> > 
> > Yes.
> 
> Hooray!
> 
> > > but at the increase in complexity of the API.
> > 
> > 
> > Yes :(
> 
> Awww.
> 
> > > Is there a way we can change the default behaviour to be something sensible and expected without adding another boolean option and benchmark method? (the answer might be no, but i should ask).
> > 
> > 
> > The alternative would be to always use whole-process cpu clock instead of the
> > main-thread-only cpu clock when the `->Threads()` wasn't called.
> 
> ok...
> 
> > Pros:
> > 
> > * no need for all that extra API
> > * no `/process_time` suffix in benchmark name.
> 
> I like these...
> 
> > Cons:
> > 
> > * Such change will change the timer used by benchmarks that do not use `->Threads()`.
> 
> Yes, but will it actually be different for the majority of users?

... i have no clue, i have no statistical info on how which features of the library are used.
It won't be 100%, maybe around 10% of `BENCHMARK_MAIN()`'s will be affected.

> > If any of these benchmarks used threads internally, the reported time will naturally be different.
> 
> It will, and more accurate (if we believe the premise of this PR, which i think we do).
> 
> > Do we know that it is what everyone affected actually wanted?
> 
> Maybe not, but what it's doing today is a bit silly.
> 
> > Maybe they **actually** used the current timing method.
> > Also, there will be no way to switch back to the previous (current) timer method.
> 
> I might be ok with this as long as only people who are using custom threads inside their benchmark are affected and we can make the case this PR does that this is a better measurement.

I'm sure this is better for those who want to measure the total toll some computation inflicts on
the CPU. I'm just saying that if someone wants to measure how much overhead there is
(i.e. if all work is being done on worker threads, how much of the work is not being offloaded
from main thread), how much CPU time is only spent on the main thread, then those measurements
will no longer be possible. I'm unable to guesstimate how many of the current internally-threaded
benchmarks out there (as previously guesstimated 10% out of all benchmarks) want the current
existing timing, and how many consider it to be a bug.

But okay, let's consider this as a bug and avoid API for now. (Will update PR.)
It can be added later if someone complains.

src/benchmark_runner.cc
```diff
@@ -82,7 +82,11 @@ BenchmarkReporter::Run CreateRunReport(
     } else {
       report.real_accumulated_time = results.real_time_used;
     }
-    report.cpu_accumulated_time = results.cpu_time_used;
+	if(b.threads <= 1){
```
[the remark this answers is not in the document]
LebedevRI: (You are looking at an old diff)

LebedevRI:
As advised by @dominichamon, posted https://groups.google.com/forum/#!topic/benchmark-discuss/97hVnMfGYgE
Please raise concerns **NOW**.

dominichamon:
I didn't see any concerns. The one thing, then, is to make sure we call out how the timing works and what's being measured in every case. Perhaps it's worth splitting this documentation into a separate 'timing.md' document that can be referenced from the README.md?

LebedevRI:
> I didn't see any concerns. 

There was one reply, and it did raise a concern, as it can be seen from that link.

The cheap way would be to now always append `/wholeprocess_time`, but that is ugly.
The simple choice is still the same, only add a switch, don't change the default.
I'm not seeing any non-ugly alternatives for console reporter.
For json reporter, we could add an always-there field telling which timer was used for which time.
But i'm not sure what would be the most forward-compatible way.

TLDR: @dominichamon do tell which path should be taken :)

dominichamon:
Here's my proposal then:

* Go with this PR as it is (with the extra API overhead). Do a release with them in.
* Create another PR that simplifies things by making the default whatever we think the "right thing" is (i believe this might be whole-process-clock) and post notice everywhere that it's changing.
* Create a release with that change.

That way, anyone that wants the current behaviour can get any release up to the next one, and if they want to move past that release they have to embrace the new behaviour. I'm comfortable with this because i think the new behaviour is more correct (or less unexpected?) in a broad range of scenarios.

Sound good?

LebedevRI:
> Here's my proposal then:

>  * Go with this PR as it is (with the extra API overhead). Do a release with them in.

This step sounds ok (other than obvious API changes)

> Create another PR that simplifies things by making the default whatever we think the "right thing" is (i believe this might be whole-process-clock) and post notice everywhere that it's changing.

Can you specify, will it also drop `/process_time` suffix, and add `/thread_time` suffix instead?
Or do you mean to just revert this PR then, rip out all this new API, and simply flip the switch?

dominichamon:
> > Here's my proposal then:
> 
> > * Go with this PR as it is (with the extra API overhead). Do a release with them in.
> 
> This step sounds ok (other than obvious API changes)

Not sure what these obvious changes are, but "getting ready for submit" sgtm.

> 
> > Create another PR that simplifies things by making the default whatever we think the "right thing" is (i believe this might be whole-process-clock) and post notice everywhere that it's changing.
> 
> Can you specify, will it also drop `/process_time` suffix, and add `/thread_time` suffix instead?
> Or do you mean to just revert this PR then, rip out all this new API, and simply flip the switch?

i'm ok with either. It would be nicer for users i think if they could choose between the existing behaviour and the new behaviour and we just flipped the default between releases, but further up in this PR you suggest that reverting to the old behaviour once we introduce process timing is not possible.

LebedevRI:
> Not sure what these obvious changes are, but "getting ready for submit" sgtm.

None.
I'm just saying that this will introduce some new public knobs, that we plan(?) on dropping the very next release.

> i'm ok with either. It would be nicer for users i think if they could choose between the existing behaviour and the new behaviour and we just flipped the default between releases, but further up in this PR you suggest that reverting to the old behaviour once we introduce process timing is not possible.

I'm guessing you are talking about https://github.com/google/benchmark/pull/763#issuecomment-464730735 ?
That only talks specifically about dynamically deciding which cpu clock to use depending on whether `->Threads()` was used or not, with no API functionality to switch between them.

So i'm still not sure what will be the step after this PR. I'll try to elaborate to the possibilities i see:
* Keep everything as-is in this PR
* Reverse the polarity.
  * Tame variant:
    * By default measure process clock.
    * Switch json/csv field name.
    * Don't add `/process_time` suffix by default.
    * If user asked to measure main-thread time, then add `/thread_time` suffix. (and write it into old csv/json field?)
  * Wild variant:
    * Revert this PR
    * *simply* switch the clock (from thread to process -wide)
    * Switch json/csv field name.
* ???

LebedevRI:
Rebased, should be good to go if it looks good to @dominichamon.

dominichamon:
> > Not sure what these obvious changes are, but "getting ready for submit" sgtm.
> 
> None.
> I'm just saying that this will introduce some new public knobs, that we plan(?) on dropping the very next release.
> 
> > i'm ok with either. It would be nicer for users i think if they could choose between the existing behaviour and the new behaviour and we just flipped the default between releases, but further up in this PR you suggest that reverting to the old behaviour once we introduce process timing is not possible.
> 
> I'm guessing you are talking about [#763 (comment)](https://github.com/google/benchmark/pull/763#issuecomment-464730735) ?
> That only talks specifically about dynamically deciding which cpu clock to use depending on whether `->Threads()` was used or not, with no API functionality to switch between them.
> 
> So i'm still not sure what will be the step after this PR. I'll try to elaborate to the possibilities i see:
> 
> * Keep everything as-is in this PR
> * Reverse the polarity.
>   
>   * Tame variant:
>     
>     * By default measure process clock.
>     * Switch json/csv field name.
>     * Don't add `/process_time` suffix by default.
>     * If user asked to measure main-thread time, then add `/thread_time` suffix. (and write it into old csv/json field?)

After the next release, do the Tame variant above (which would require a command-line/benchmark flag).

>   * Wild variant:
>     
>     * Revert this PR
>     * _simply_ switch the clock (from thread to process -wide)
>     * Switch json/csv field name.
> * ???

I wish we had tracking for flags... if we saw the flag in Tame variant went unused i'd push for this the release after.

README.md
```diff
@@ -428,6 +428,49 @@ BENCHMARK(BM_test)->Range(8, 8<<10)->UseRealTime();
...
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->UseRealTime();
+
+// Measure the total CPU consumption, use it to decide for how long to
+// run the benchmark loop. This will always measure to no less than the
+// time spent by the main thread in single-threaded case.
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->MeasureProcessCPUTime();
+
+// A mixture of the last two. Measure the total CPU consumption, but use the
+// wall clock to decide for how long to run the benchmark loop.
+BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->MeasureProcessCPUTime()->UseRealTime();
```
dominichamon: :D

include/benchmark/benchmark.h
```diff
@@ -874,11 +874,18 @@ class Benchmark {
   // Same as ReportAggregatesOnly(), but applies to display reporter only.
   Benchmark* DisplayAggregatesOnly(bool value = true);
 
-  // If a particular benchmark is I/O bound, runs multiple threads internally or
-  // if for some reason CPU timings are not representative, call this method. If
-  // called, the elapsed time will be used to control how many iterations are
-  // run, and in the printing of items/second or MB/seconds values.  If not
-  // called, the cpu time used by the benchmark will be used.
+  // By default, the CPU time is measured only for the main thread, which may
+  // be unrepresentable if the benchmark uses threads internally. If called,
```
dominichamon: unrepresentative, i think.

src/thread_timer.h
```diff
@@ -9,13 +9,14 @@ namespace internal {
 
 class ThreadTimer {
  public:
-  ThreadTimer() = default;
+  ThreadTimer(bool measure_process_cpu_time_)
```
dominichamon: make this explicit please (to avoid an accidental implicit conversion).

https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

dominichamon: https://abseil.io/tips/94, which discusses the non-readability of callsites when using boolean parameters.

Consider two factory methods (`Create()` and `CreateProcessCpuTime()`) that can hide the boolean. The enum example in the above absl tip is likely overkill for this.

README.md
LebedevRI: This is likely what everyone who may be using threads internally will use, to be noted.

dominichamon:
Thanks!

LebedevRI:
YAY

bryan-lunt:
WOOHOO!
Thanks everyone, especially @LebedevRI and @dominichamon .

PR #770, opened 2019-02-18
[#766] add x-compile support for QNX SDP7

### LINKED ISSUE #766: add x-compile support for QNX SDP
Since googletest already supports x-compilation for QNX, it is nice to have google benchmark support it too.

jilinzhou:
Since googletest already supports x-compilation for QNX, it is nice to
 have google benchmark support it too.
Fixes #766

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -389,6 +389,8 @@ std::string GetSystemName() {
 #else // defined(BENCHMARK_OS_WINDOWS)
 #ifdef BENCHMARK_OS_MACOSX //Mac Doesnt have HOST_NAME_MAX defined
 #define HOST_NAME_MAX 64
+#elif defined(BENCHMARK_OS_QNX)
+#define HOST_NAME_MAX 154
 #endif
   char hostname[HOST_NAME_MAX];
   int retVal = gethostname(hostname, HOST_NAME_MAX);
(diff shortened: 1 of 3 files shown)


src/sysinfo.cc
```diff
@@ -389,6 +389,8 @@ std::string GetSystemName() {
 #else // defined(BENCHMARK_OS_WINDOWS)
 #ifdef BENCHMARK_OS_MACOSX //Mac Doesnt have HOST_NAME_MAX defined
 #define HOST_NAME_MAX 64
+#elif defined(BENCHMARK_OS_QNX)
+#define HOST_NAME_MAX 154
```
dominichamon: i think this is defined in limits.h for qnx (http://www.qnx.co.uk/developers/docs/6.5.0/index.jsp?topic=%2Fcom.qnx.doc.neutrino_prog%2Fposix_conformance.html)

jilinzhou: Here is what in QNX SDP 7:

```
>>grep -r "HOST_NAME_MAX" ./usr/include
./usr/include/confname.h:#define _SC_HOST_NAME_MAX					154
./usr/include/limits.h:#undef	_POSIX_HOST_NAME_MAX
./usr/include/limits.h:#define	_POSIX_HOST_NAME_MAX	255
```

dominichamon: thanks for checking!

PR #771, opened 2019-02-19
Prefer -pthread to -lpthread flag in INTERFACE_LINK_LIBS for better compatibility when cross-compiling

wesm:
Linking with `-lpthread` instead of `-pthread` can cause failures of the sort like

```
/home/travis/build/apache/arrow/cpp-toolchain/bin/ld: /home/travis/build/apache/arrow/cpp-toolchain/lib/libbenchmark.a(benchmark.cc.o): in function `benchmark::RunSpecifiedBenchmarks(benchmark::BenchmarkReporter*, benchmark::BenchmarkReporter*)':
benchmark.cc:(.text+0x2235): undefined reference to `std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)())'
/home/travis/build/apache/arrow/cpp-toolchain/bin/ld: /home/travis/build/apache/arrow/cpp-toolchain/lib/libbenchmark.a(benchmark.cc.o): in function `std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(benchmark::internal::Benchmark::Instance const*, unsigned long, int, benchmark::internal::ThreadManager*), benchmark::internal::Benchmark::Instance const*, unsigned long, int, benchmark::internal::ThreadManager*> > >::~_State_impl()':
benchmark.cc:(.text._ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPKN9benchmark8internal9Benchmark8InstanceEmiPNS4_13ThreadManagerEES8_miSA_EEEEED2Ev[_ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPKN9benchmark8internal9Benchmark8InstanceEmiPNS4_13ThreadManagerEES8_miSA_EEEEED5Ev]+0x10): undefined reference to `std::thread::_State::~_State()'
/home/travis/build/apache/arrow/cpp-toolchain/bin/ld: /home/travis/build/apache/arrow/cpp-toolchain/lib/libbenchmark.a(benchmark.cc.o): in function `std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(benchmark::internal::Benchmark::Instance const*, unsigned long, int, benchmark::internal::ThreadManager*), benchmark::internal::Benchmark::Instance const*, unsigned long, int, benchmark::internal::ThreadManager*> > >::~_State_impl()':
benchmark.cc:(.text._ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPKN9benchmark8internal9Benchmark8InstanceEmiPNS4_13ThreadManagerEES8_miSA_EEEEED0Ev[_ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPKN9benchmark8internal9Benchmark8InstanceEmiPNS4_13ThreadManagerEES8_miSA_EEEEED5Ev]+0x14): undefined reference to `std::thread::_State::~_State()'
/home/travis/build/apache/arrow/cpp-toolchain/bin/ld: /home/travis/build/apache/arrow/cpp-toolchain/lib/libbenchmark.a(benchmark.cc.o):(.data.rel.ro._ZTINSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPKN9benchmark8internal9Benchmark8InstanceEmiPNS4_13ThreadManagerEES8_miSA_EEEEEE[_ZTINSt6thread11_State_implINS_8_InvokerISt5tupleIJPFvPKN9benchmark8internal9Benchmark8InstanceEmiPNS4_13ThreadManagerEES8_miSA_EEEEEE]+0x10): undefined reference to `typeinfo for std::thread::_State'
collect2: error: ld returned 1 exit status
```

when using libraries cross-compiled on one host for use on another. CMake version 3.1 and higher has the `THREADS_PREFER_PTHREAD_FLAG` flag which toggles the behavior of `${CMAKE_THREAD_LIBS_INIT}`. 

See https://cmake.org/cmake/help/v3.1/module/FindThreads.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,6 +251,7 @@ if (NOT BENCHMARK_ENABLE_EXCEPTIONS AND HAVE_STD_REGEX
 endif()
 cxx_feature_check(STEADY_CLOCK)
 # Ensure we have pthreads
+set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
 # Set up directories


dominichamon:
Thanks!

LebedevRI:
Is this actually correct/sufficient though?
https://cmake.org/cmake/help/v3.1/module/FindThreads.html
```
If the use of the -pthread compiler and linker flag is prefered then the caller can set
  THREADS_PREFER_PTHREAD_FLAG
Please note that the compiler flag can only be used with the imported target. Use of both the imported target as well as this switch is highly recommended for new code.
```
And yet benchmark uses `CMAKE_THREAD_LIBS_INIT`, not the `Threads::Threads` import target,
as far as i can see:
https://github.com/google/benchmark/blob/505be96ab23056580a3a2315abba048f4428b04e/src/CMakeLists.txt#L30

wesm:
@LebedevRI if you use `Threads::Threads` then this is what appears in `benchmarkTargets.cmake` instead of `-pthread`. So consumers of the installed library would need to also use FindThreads in order to use the CMake config

wesm:
I haven't tested this extensively on other platforms so would be interested to get feedback from experts who understand the nuances more deeply

jcfr:
(Background: @wesm I am commenting here following up you last [tweet](https://twitter.com/wesmckinn/status/1097890908153004032).)

> So consumers of the installed library would need to also use FindThreads in order to use the CMake config

The key here is to update [cmake/Config.cmake.in](https://github.com/google/benchmark/blob/505be96ab23056580a3a2315abba048f4428b04e/cmake/Config.cmake.in) so that it has:

```
include(CMakeFindDependencyMacro)
set(THREADS_PREFER_PTHREAD_FLAG 1)
find_dependency(Threads)
```

Doing so would ensure that consumer do not have to worry about the private dependencies of the benchmark  project and would allow to use `Thread::Thread` target.

Note that [CMakeLists.txt](https://github.com/google/benchmark/blob/505be96ab23056580a3a2315abba048f4428b04e/CMakeLists.txt#L238-L239) would have to be updated to also set `THREADS_PREFER_PTHREAD_FLAG`

wesm:
Seems reasonable. @jcfr are you able to submit a PR for this?

PR #783, opened 2019-03-17
Travis-ci: attempt to fix clang+libc++ build

LebedevRI:
It broke because the libc++ is being built as part of *this*
build, with old gcc+libstdc++ (4.8?), but LLVM is preparing
to switch to C++14, and gcc+libstdc++ <5 are soft-deprecated.


LebedevRI:
Hurray.

PR #785, opened 2019-03-19
Correct-erer JSON output for inf/nan. Fixes #784

### LINKED ISSUE #784: json output writes inf and nan as inf and nan
According to https://docs.python.org/3.4/library/json.html, the (un)standard way of representing inf and nan in JSON is 'NaN', 'Infinity', and '-Infinity'.  Currently the json output puts the boring 'inf' and 'nan', even for counters, which might legitimately contain one of these values, depending on what's being measured.

tesch1:
Output javascript nan/inf values into JSON instead of c-style nan/inf.  Neither is technically "correct JSON", but many JSON parsers support the javascript symbols.

### DIFF
--- a/src/json_reporter.cc
+++ b/src/json_reporter.cc
@@ -23,6 +23,7 @@
 #include <string>
 #include <tuple>
 #include <vector>
+#include <cmath>
 
 #include "string_util.h"
 #include "timers.h"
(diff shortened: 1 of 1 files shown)


tesch1:
I signed it!

LebedevRI:
nagging: "fix for ", "output is broken" and other things people write as commit messages are useless.
The commit title needs to explain the fix/change, not just reference some bug somewhere.

dominichamon:
appveyor looks like a false negative; all tests passed then it failed on result upload.

tesch1:
@LebedevRI the new title is a work of unmitigated beauty ;o)  I will try harder next time

Issue #762, opened 2019-02-09
Negative inputs have unexpected behavior


NAThompson:
I have a question about negative arguments to ranges. Here's my code:

```cpp
#include <benchmark/benchmark.h>
#include <array>
#include <random>
#include <iostream>
#include <algorithm>
#include <boost/math/special_functions/hypergeometric_0F1.hpp>

template<class Real>
void BM_1F0(benchmark::State& state)
{
    using boost::math::hypergeometric_0F1;

    std::default_random_engine gen;
    std::uniform_real_distribution<Real> x_dis(state.range(0), 2*state.range(0));
    Real b = 7.2;

    for (auto _ : state)
    {
        benchmark::DoNotOptimize(hypergeometric_0F1(b, x_dis(gen)));
    }
}

BENCHMARK_TEMPLATE(BM_1F0, double)->RangeMultiplier(2)->Range(-8, 8);

BENCHMARK_MAIN();
```

My expectation is that I would get `state.range(0)` to take values in the set `{-8, -4, -2, -1, 0, 1, 2, 4, 8}`. However, only the first negative argument is selected, and then we obtain positive arguments:

```
------------------------------------------------------------------
Benchmark                           Time           CPU Iterations
------------------------------------------------------------------
BM_1F0<double>/-8                 297 ns        296 ns    2362295
BM_1F0<double>/1                   83 ns         83 ns    8470167
BM_1F0<double>/2                   92 ns         92 ns    7639503
BM_1F0<double>/4                  104 ns        104 ns    6663176
BM_1F0<double>/8                  118 ns        118 ns    5918060
```

What is the idiomatic way to step through a range of negative arguments?

dominichamon:
I'd expect that too, but i think that negative ranges haven't come up often. `range` is an `int64_t` so it should work in that respect, but my guess is the interpolation code doesn't actually support it.. let's see!

https://github.com/google/benchmark/blob/d8584bda67320937334b038e736f5bffcd438875/src/benchmark_register.h#L10

we have a CHECK that `lo` is `>= 0` which isn't encouraging, and then we insert `lo` followed by a loop starting at 1. So we don't actually support negative ranges.

I'm really surprised this hasn't come up before...

pleroy:
I have run into this with `DenseRange`, [fixed it in my fork](https://github.com/mockingbirdnest/benchmark/commit/7d8abdbd89807198e0013af7704abeeba3ce3099), never upstreamed it (shame).

[...]

PR #787, opened 2019-03-24
Negative ranges #762

danielharvey458:
Add FIXME in multiple_ranges_test.cc 36c7640 -- this looks like a bug to me?

Improve handling of large bounds in AddRange 02a67f0 -- minor tweak to squeeze in an extra power of 'mult' in AddRanges

Enable negative values for Range argument 2ddf958 -- main proposed commit

### DIFF
--- a/src/benchmark_register.h
+++ b/src/benchmark_register.h
@@ -17,11 +17,13 @@ void AddRange(std::vector<T>* dst, T lo, T hi, int mult) {
   static const T kmax = std::numeric_limits<T>::max();
 
   // Now space out the benchmarks in multiples of "mult"
-  for (T i = 1; i < kmax / mult; i *= mult) {
-    if (i >= hi) break;
+  for (T i = 1; i < hi; i *= mult) {
     if (i > lo) {
       dst->push_back(i);
     }
+    // Break the loop here since multiplying by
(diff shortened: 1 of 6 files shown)


src/benchmark_register.h
```diff
@@ -5,29 +5,103 @@
 
 #include "check.h"
 
+namespace benchmark {
+namespace internal {
```
danielharvey458: I presumed 'AddRanges' wasn't intended for global scope, please correct me if I'm wrong.

test/multiple_ranges_test.cc
```diff
@@ -40,6 +40,8 @@ class MultipleRangesFixture : public ::benchmark::Fixture {
   // NOTE: This is not TearDown as we want to check after _all_ runs are
   // complete.
   virtual ~MultipleRangesFixture() {
+     // FIXME Should the 'if' below be checking for equality
+     // of the values in the containers, rather than their size?
```
dominichamon: it should probably do both. The assert is no longer necessary as we're using for range loops (we used to loop using the index and it would crash) and instead we should be looping over both and comparing, then asserting/aborting with some sort of error.

dominichamon:
Thank you!

PR #792, opened 2019-04-11
fix master branch on *BSD

h-2:
Current master is broken on the BSDs. The little patch below extends the macOS exception to the BSDs.

I have signed the CLA. Should I also add myself to AUTHORS/CONTRIBUTORS?


dominichamon:
If you want. Let me know either way (we're waiting for the travis builds anyway) and i'll either merge as is or merge with your attribution :)

dominichamon:
Thanks! Merging post testing.

h-2:
Ok, I have added it to CONTRIBUTORS, but not to AUTHORS as the change is so minor it probably doesn't even qualify as a copyrightable work.

h-2:
Thanks for the quick response and review! Ci seems to pass.

PR #793, opened 2019-04-11
Add command line flags tests

dominichamon:
Also clean up the command line flags implementation to remove unused code, move internal methods to the implementation file, and add an unnamed namespace to protect those symbols from cross-translation-unit references.

### DIFF
--- a/src/commandlineflags.cc
+++ b/src/commandlineflags.cc
@@ -21,8 +21,8 @@
 #include <limits>
 
 namespace benchmark {
-
 namespace {
+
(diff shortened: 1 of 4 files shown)


src/commandlineflags.cc
```diff
@@ -21,6 +21,8 @@
 #include <limits>
 
 namespace benchmark {
+
+namespace {
```
LebedevRI: Nit:
```suggestion
namespace {

```

PR #798, opened 2019-04-15
Add CPack

efernandez:
This allows to create a `.deb` (`benchmark-1.4.0-Linux.deb`) by running: `make package`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,3 +272,12 @@ if (BENCHMARK_ENABLE_TESTING)
   endif()
   add_subdirectory(test)
 endif()
+
+set(CPACK_GENERATOR "DEB")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Google Inc.")
+set(CPACK_PACKAGE_VERSION ${VERSION})
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Google Benchmark library.")
+set(CPACK_PACKAGE_DESCRIPTION "A library to support the benchmarking of functions, similar to unit-tests.")
+set(CPACK_DEBIAN_PACKAGE_SECTION "Development")
+
+include(CPack)


LebedevRI:
Is cpack useful outside of windows ecosystem (think installers)?
Normally pretty much every package building system will use it's own config,
not cpack, since it's kinda chicken-egg problem.

CMakeLists.txt
```diff
@@ -272,3 +272,12 @@ if (BENCHMARK_ENABLE_TESTING)
...
 endif()
+
+set(CPACK_GENERATOR "DEB")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Google Inc.")
+set(CPACK_PACKAGE_VERSION ${VERSION})
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Google Benchmark library.")
+set(CPACK_PACKAGE_DESCRIPTION "A library to support the benchmarking of functions, similar to unit-tests.")
+set(CPACK_DEBIAN_PACKAGE_SECTION "Development")
+
+include(CPack)
```
dominichamon: is this the cmake 3.0 version of cpack? we require 2.8.12 and the cpack documentation for that looks different: https://cmake.org/cmake/help/v2.8.12/cpack.html

efernandez:
@LebedevRI I'm actually using this on Ubuntu Xenial.

TBH I went this path b/c it was the easiest, although I had to make these changes.

CMakeLists.txt
efernandez: I honestly didn't look into that.

I've done a quick search and the only variable I can't find in the documentation is `CPACK_PACKAGE_DESCRIPTION`, which looks like it should be either removed or replaced with `CPACK_PACKAGE_DESCRIPTION_FILE` + a text file withe the long description.

In this file I see `cmake_minimum_required (VERSION 2.8.12)`, and my version of `cmake` is 3.5.1.
Is there any way on your end you could tried with 2.8.12?
Otherwise, I could try building https://gitlab.kitware.com/cmake/cmake/tree/v2.8.12.2 and giving it a try.

dominichamon: I'd use CPACK_DESCRIPTION_SUMMARY.

We could also update cmake to 3.5.1: #801

efernandez: @dominichamon Great to hear you're bumping up to cmake 3!

I already use both `CPACK_DESCRIPTION_SUMMARY` and `CPACK_PACKAGE_DESCRIPTION`, for the short and long description.

This is how things look like in `aptitude`:
benchmark_aptitude

It looks indeed that the `CPACK_PACKAGE_DESCRIPTION` isn't showing up. :thinking: 

Same with:
``` bash
$ apt-cache show benchmark
Package: benchmark
Version: 1.4.0
Section: Development
Priority: optional
Architecture: amd64
Installed-Size: 916
Maintainer: Google Inc.
Description: Google Benchmark library.
Description-md5: 3f3fbd5479e6d7d7b1ab982eeb7ee786
Filename: pool/main/benchmark-1.4.0-Linux.deb
SHA1: 0b2e4c623651fbd1480e9df15090c9b7f18c35a6
SHA256: 12fa22510f985e01b8a9ca238937ef771762322f59fa091e28b92fb2c3200abd
Size: 228010
```

Not a big deal, since it's optional. Let me know if you want me to remove it for now, since it's doing nothing, or explore how to fix it.

LebedevRI: (https://packages.ubuntu.com/search?keywords=benchmark&searchon=names says it's packaged already)

efernandez: Good call. I think I missed that b/c I'm on Ubuntu 16.04.

I think this contribution is still useful, but I could certainly consider taking the debian from 18.04 or 18.10. It seems that they packaged benchmark 1.3.0 though, so this contribution at least makes it easier to generate a debian locally for the latest version.

Anyway, it's up to you if you want this close, or if you want me to do something else before merging. :smiley:

CMakeLists.txt
```diff
@@ -272,3 +272,12 @@ if (BENCHMARK_ENABLE_TESTING)
   endif()
   add_subdirectory(test)
 endif()
+
+set(CPACK_GENERATOR "DEB")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Google Inc.")
+set(CPACK_PACKAGE_VERSION ${VERSION})
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Google Benchmark library.")
+set(CPACK_PACKAGE_DESCRIPTION "A library to support the benchmarking of functions, similar to unit-tests.")
+set(CPACK_DEBIAN_PACKAGE_SECTION "Development")
```
LebedevRI: Should this have any of other DEB magic from https://salsa.debian.org/science-team/benchmark/blob/master/debian/control ?

CMakeLists.txt
LebedevRI: I don't have any strong opinion on this, seems not too intrusive,
although not a full-fledged package. I'll leave this up to @dominichamon.

CMakeLists.txt
efernandez: With the CPack approach we're missing that part.

https://salsa.debian.org/science-team/benchmark/blob/master/debian/control is part of the debianize process, which generates a deb file and some additional stuff so the result is compliant with the official Debian rules for packaging. Meaning that things can go into the official repos.

dominichamon: I don't know enough about debian and packaging rules. Does this PR as it is allow us to trivially create deb packages and have them in the repo?

LebedevRI: (i personally hope not, that should be done by actual maintainers, not projects themselves)

efernandez: This allows to trivially create a deb.

It doesn't create a deb that can be used in Debian/Ubuntu apt repos directly though. Neither uploaded to a Launchpad PPA. The reason being that the deb is not enough for those. They need some additional files, e.g. a `.changes` file with Launchpad.

I've tried to run `debuild -us -uc` on the debianized `benchmark` repo from https://salsa.debian.org/science-team/benchmark in Ubuntu 16.04. I had to relax the `debhelper` dependency from `>= 10` to `>= 9` and remove (hacky) the dependency on `libgmock-dev`, which isn't available in Ubuntu 16.04. I still have some errors:
``` bash
$ debuild -us -uc
...
CMake Error at src/CMakeLists.txt:14 (list):
  list does not recognize sub-command FILTER


CMake Error at /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find GTest (missing: GTEST_LIBRARY GTEST_MAIN_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.5/Modules/FindGTest.cmake:205 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/HandleGTest.cmake:102 (find_package)
  CMakeLists.txt:248 (include)
...
```

And I have `libgtest-dev` version `1.7.0-4ubuntu1` installed in my system.

I guess I could find out how to solve these issues with more time. It's just not worth it for me anymore since with CPack I got a deb we can put in our apt repos and it just works. And once we move to Ubuntu >= 18.04, we can just used the deb officially shipped with it.

So probably, you could just close this PR. :thinking:

Issue #745, opened 2018-12-14
Strings in JSON output are not escaped.


ZbigniewRA:
Function FormatKV() in json_reporter.cc does not escape strings that are being serialized.
As a result calling for example:
```
state.SkipWithError("File not found c:\\hello.txt.");
```
results in json with syntax error "Invalid escape \h":
```
    {
      "name": "BM_Calc_Big1/extensions_pgpu_zscore",
      "error_occurred": true,
      "error_message": "File not found c:\hello.txt.",
      "iterations": 1,
      "real_time": 0.0000000000000000e+00,
      "cpu_time": 0.0000000000000000e+00,
      "time_unit": "ns"
    },
```

This problem is in both 1.4.1 and v2 versions.

tesch1:
Can this be closed now?

LebedevRI:
Looks like github decided to be picky and not auto-closed this.

PR #802, opened 2019-04-16
escape special chars in csv and json output.

tesch1:
- escape \b,\f,\n,\r,\t,\," from strings before dumping
  them to json or csv.
- also faithfully reproduce the sign of nan in json.
this fixes github issue #745.

### DIFF
--- a/src/csv_reporter.cc
+++ b/src/csv_reporter.cc
@@ -39,6 +39,7 @@ std::vector<std::string> elements = {
 
 std::string CsvEscape(const std::string & s) {
   std::string tmp;
+  tmp.reserve(s.size() + 2);
   for (char c : s) {
     switch (c) {
     case '"' : tmp += "\"\""; break;
--- a/src/csv_reporter.cc
+++ b/src/csv_reporter.cc
@@ -37,6 +37,17 @@ std::vector<std::string> elements = {
     "error_occurred", "error_message"};
 }  // namespace
 
+std::string CsvEscape(const std::string & s) {
+  std::string tmp;
+  for (char c : s) {
+    switch (c) {
+    case '"' : tmp += "\"\""; break;
+    default  : tmp += c; break;
+    }
+  }
+  return '"' + tmp + '"';
+}
+
 bool CSVReporter::ReportContext(const Context& context) {
   PrintBasicContext(&GetErrorStream(), context);
   return true;
@@ -89,11 +100,11 @@ void CSVReporter::ReportRuns(const std::vector<Run>& reports) {
 
 void CSVReporter::PrintRunData(const Run& run) {
   std::ostream& Out = GetOutputStream();
-  Out << '"' << StrEscape(run.benchmark_name()) << "\",";
+  Out << CsvEscape(run.benchmark_name()) << ",";
   if (run.error_occurred) {
     Out << std::string(elements.size() - 3, ',');
     Out << "true,";
-    Out << '"' << StrEscape(run.error_message) << "\"\n";
+    Out << CsvEscape(run.error_message) << "\n";
     return;
   }
 
@@ -123,7 +134,7 @@ void CSVReporter::PrintRunData(const Run& run) {
   }
   Out << ",";
   if (!run.report_label.empty()) {
-    Out << "\"" << StrEscape(run.report_label) << "\"";
+    Out << CsvEscape(run.report_label);
   }
   Out << ",,";  // for error_occurred and error_message
 
--- a/src/csv_reporter.cc
+++ b/src/csv_reporter.cc
@@ -89,18 +89,11 @@ void CSVReporter::ReportRuns(const std::vector<Run>& reports) {
 
 void CSVReporter::PrintRunData(const Run& run) {
   std::ostream& Out = GetOutputStream();
-
-  // Field with embedded double-quote characters must be doubled and the field
-  // delimited with double-quotes.
-  std::string name = run.benchmark_name();
-  ReplaceAll(&name, "\"", "\"\"");
-  Out << '"' << name << "\",";
+  Out << '"' << StrEscape(run.benchmark_name()) << "\",";
   if (run.error_occurred) {
     Out << std::string(elements.size() - 3, ',');
     Out << "true,";
-    std::string msg = run.error_message;
-    ReplaceAll(&msg, "\"", "\"\"");
-    Out << '"' << msg << "\"\n";
+    Out << '"' << StrEscape(run.error_message) << "\"\n";
     return;
   }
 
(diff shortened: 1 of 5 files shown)


src/string_util.h
```diff
@@ -37,8 +37,9 @@ inline std::string StrCat(Args&&... args) {
   return ss.str();
 }
 
-void ReplaceAll(std::string* str, const std::string& from,
-                const std::string& to);
+struct StrEscape : std::string {
```
LebedevRI: Uhm.
/me doesn't like

dominichamon: ooh i didn't spot that. yeah, that should likely just be a method that takes a string and returns a string.

tesch1: /what /does /you /prefer /?

tesch1: how about this...

src/string_util.cc
```diff
@@ -160,13 +160,19 @@ std::string StrFormat(const char* format, ...) {
...
 
-void ReplaceAll(std::string* str, const std::string& from,
-                const std::string& to) {
-  std::size_t start = 0;
-  while ((start = str->find(from, start)) != std::string::npos) {
-    str->replace(start, from.length(), to);
-    start += to.length();
-  }
+std::string StrEscape(const std::string & s) {
+  std::string tmp;
```
LebedevRI: tmp.reserve(s.size());

tesch1: feel free to take over the PR

src/csv_reporter.cc
```diff
@@ -89,18 +89,11 @@ void CSVReporter::ReportRuns(const std::vector<Run>& reports) {
 
 void CSVReporter::PrintRunData(const Run& run) {
   std::ostream& Out = GetOutputStream();
-
-  // Field with embedded double-quote characters must be doubled and the field
-  // delimited with double-quotes.
-  std::string name = run.benchmark_name();
-  ReplaceAll(&name, "\"", "\"\"");
-  Out << '"' << name << "\",";
+  Out << '"' << StrEscape(run.benchmark_name()) << "\",";
```
LebedevRI: Can you please link where it's documented what should be escaped in CSV?
I'm probably looking in wrong places, https://tools.ietf.org/html/rfc4180 only says to replace `"` with `""`
?

src/string_util.cc
LebedevRI: I mean, it's your PR.

tesch1: my use case isn't going to notice the 2 ns (maybe?) speed improvement, if you need it feel free to add it though.

src/csv_reporter.cc
tesch1: is there even a standard for csv?   at least \r and \n need to be escaped, \b and \f probably should be, \t is a matter of taste- the tests dont check for it in any case.  and anyway, isn't csv support being deprecated #500 ?

src/string_util.cc
LebedevRI: It's a review comment, it is good to address them, so that there is forward progress on proposed changes.
Not sure how these github suggestions work:
```suggestion
  std::string tmp;
  tmp.reserve(s.size());
```

src/csv_reporter.cc
LebedevRI: > at least \r and \n need to be escaped

I'm not seeing that in a quick test with loading
> "test","my
> test"

into libreoffice:
image

And replacing `\n` with "\n":
> "test","my\ntest"

breaks it: 
image
Thus yes, i'm curious as to motivation/documentation.



> and anyway, isn't csv support being deprecated #500 ?

Yep

src/string_util.cc
tesch1: you do realize that std::string is already designed for this kind of concatenation? this is trading 0-1-2 reallocs (depending on the length of the key) with a guaranteed two calls (strlen,realloc), not sure if there's a real win here, unless it's an issue of coding style?

LebedevRI: We *know* the escaped string will be at least not shorter than the input string.
clang/gcc will not magically insert that pre-allocation.
Therefore, if we don't do it explicitly, we are relying on the sane behavior
of the `std::string` implementation in the C++ std library that is used.
* We will likely always have some allocs if the string after escaping does not fit within small-size-optimization.
* Without pre-allocation, if the source string does not fit within small-size-optimization,
and there won't be extra escape symbols, we will have allocations.
* WITH pre-allocation, if the source string does not fit within small-size-optimization,
and there won't be extra escape symbols, we will *NOT* have allocations. WIN.

It is well-known that `reserve()` is good, and it is quite obvious it won't hurt here.
I do not understand your point.

src/csv_reporter.cc
tesch1: you're right, \t and \\ should proabaly not be escaped for csv.  the other ones I think the "right thing" is to escape them.  \b because if you're lookign at csv, you typically want to see it on a terminal, and any \b would end up hidden, and they're more often than not in there by mistake.

LebedevRI: I agree that JSON changes are good.
The issue does not mention anything about CSV; are you affected by CSV side of things?

How about a middle ground solution then?
1. Click apply on that `reserve()` suggestion.
2. Drop CSV changes, thus avoiding all the questions here as to what should and should not be escaped.

src/csv_reporter.cc
```diff
@@ -37,6 +37,17 @@ std::vector<std::string> elements = {
     "error_occurred", "error_message"};
 }  // namespace
 
+std::string CsvEscape(const std::string & s) {
+  std::string tmp;
```
LebedevRI:
```suggestion
  std::string tmp;
  tmp.reserve(s.size() + 2);
```

src/json_reporter.cc
```diff
@@ -32,30 +32,47 @@ namespace benchmark {
 
 namespace {
 
+std::string StrEscape(const std::string & s) {
+  std::string tmp;
```
LebedevRI:
```suggestion
  std::string tmp;
  tmp.reserve(s.size());
```

src/csv_reporter.cc
tesch1: we should add up all the nanoseconds ever saved by this change and compare with the amount of time we spent arguing about it.

LebedevRI:
Blargh, of course github tries to be smart, and use the author of suggestions for commits,
and multi-committer PR's are too smart for CLI bot :/ I should have guessed.

tesch1:
:D googlebot removed the cla check when I applied the suggested changes.  nice.

dominichamon:
I can still merge. I will once builds go green.

LebedevRI:
> I can still merge. I will once builds go green.

Looks green to me
(that was supposed to be a word play on "LGTM" that can be read both as "G"ood/"G"reen)

dominichamon:
Thanks!

PR #801, opened 2019-04-16
Bump required cmake version from 2.x to 3.x

dominichamon:

PR #808, opened 2019-04-29
Add support for GNU Install Dirs from GNU Coding Standards. Fixes #807

### LINKED ISSUE #807: Allow global installation of the library into /usr/lib64/
Hello, everyone.

I'd be very happy if the cmake file allowed not just setting the prefix (/usr), but also the libdir suffix, that is, in my case, "64", that is, I'd like to install the library into /usr/lib64/, because that's how multilib is organized in my distribution.

Would it be possible?

lockywolf:
* src/CMakeLists.txt: Added support for setting the standard variables,
                      such as CMAKE_INSTALL_BINDIR.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,5 @@
 # Allow the source files to find headers in src/
+include(GNUInstallDirs)
 include_directories(${PROJECT_SOURCE_DIR}/src)
 
 if (DEFINED BENCHMARK_CXX_LINKER_FLAGS)
@@ -55,11 +56,12 @@ target_include_directories(benchmark PUBLIC
     )
 target_link_libraries(benchmark_main benchmark)
 
-set(include_install_dir "include")
-set(lib_install_dir "lib/")
-set(bin_install_dir "bin/")
-set(config_install_dir "lib/cmake/${PROJECT_NAME}")
-set(pkgconfig_install_dir "lib/pkgconfig")
+set(include_install_dir "${CMAKE_INSTALL_INCLUDEDIR}")
+set(lib_install_dir "${CMAKE_INSTALL_LIBDIR}")
+set(bin_install_dir "${CMAKE_INSTALL_BINDIR}")
+set(config_install_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
+set(pkgconfig_install_dir "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
+
 
 set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -56,12 +56,6 @@ target_include_directories(benchmark PUBLIC
     )
 target_link_libraries(benchmark_main benchmark)
 
-#set(include_install_dir "${CMAKE_INSTALL_INCLUDEDIR}")
-#set(lib_install_dir "${CMAKE_INSTALL_LIBDIR}")
-#set(bin_install_dir "${CMAKE_INSTALL_BINDIR}")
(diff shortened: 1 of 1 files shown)


lockywolf:
Updated.

src/CMakeLists.txt
```diff
@@ -55,11 +56,12 @@ target_include_directories(benchmark PUBLIC
...
 
-set(include_install_dir "include")
-set(lib_install_dir "lib/")
-set(bin_install_dir "bin/")
-set(config_install_dir "lib/cmake/${PROJECT_NAME}")
-set(pkgconfig_install_dir "lib/pkgconfig")
+#set(include_install_dir "${CMAKE_INSTALL_INCLUDEDIR}")
+#set(lib_install_dir "${CMAKE_INSTALL_LIBDIR}")
+#set(bin_install_dir "${CMAKE_INSTALL_BINDIR}")
+set(config_install_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
```
LebedevRI: Please just completely drop all these 5 original variables, and replace them with whatever their value would be

dominichamon: do these not end up being sensibly defaulted to the CMAKE_ versions in this case?

LebedevRI: There is no point in having these immediate variables, they are just confusing.

lockywolf:
There is no "GNU Standard" for cmake files dir and pkgconfig dir, therefore I left them as variables.

The user may want to set them differently if he actually cares to keep all pkg-config or cmake files in one place. Bindir and libdir I have replaced .

LebedevRI:
> The user may want to set them differently if he actually cares to keep all pkg-config or cmake files in one place. Bindir and libdir I have replaced .

Except that won't work, which is the whole reason for this change.
`set(name value)` will always override the `name` variable in the **current** scope,
hiding any previous variable by that name set in upper scope, or in cache.

So no, these intermediate variables **MUST** go away.

LebedevRI:
https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable
```
set(<variable> <value>... [PARENT_SCOPE])
Sets the given <variable> in the current function or directory scope.
```
https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-variables
```
Directory Scope
Each of the Directories in a source tree has its own variable bindings. Before processing the CMakeLists.txt file for a directory, CMake copies all variable bindings currently defined in the parent directory, if any, to initialize the new directory scope. CMake Scripts, when processed with cmake -P, bind variables in one “directory” scope.
```
```
When evaluating Variable References, CMake first searches the function call stack, if any, for a binding and then falls back to the binding in the current directory scope, if any. If a “set” binding is found, its value is used. If an “unset” binding is found, or no binding is found, CMake then searches for a cache entry. If a cache entry is found, its value is used. Otherwise, the variable reference evaluates to an empty string. The $CACHE{VAR} syntax can be used to do direct cache entry lookups.
```
So please, do drop those intermediate variables. This looks good otherwise.

lockywolf:
Век живи -- век учись. 

I'll try to understand how to set them properly, delete the comments and update the pull request.

LebedevRI:
Unless https://github.com/google/benchmark/issues/807 was *also* about being able to specify the pkgconfig/cmake install paths, just dropping intermediate variables is the solution.
(if it was, maybe do that in a follow-up)

lockywolf:
I did some Binging, and found that there is indeed no default place for .cmake and .pc files. 

This is a bit of a disgrace, but alas. I updated the pull request.

When is the next release planned? (If soon, I'll defer building a package for my distro till the release date comes. Otherwise I'll build from HEAD.

dominichamon:
@lockywolf With #801 and this landing we may be close to a new release. I will try to create in the next couple of weeks or so, time allowing.

PR #811, opened 2019-05-01
Don't read CMAKE_BUILD_TYPE if it is not there. Fixes #810

### LINKED ISSUE #810: Cmake error on fresh build
Cmake error while configuring a fresh build following the [installation instructions](https://github.com/google/benchmark/blob/master/README.md#installation).

```
-- Performing Test HAVE_POSIX_REGEX -- success
-- Performing Test HAVE_STEADY_CLOCK
-- Performing Test HAVE_STEADY_CLOCK
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
CMake Error at src/CMakeLists.txt:37 (string):
  string no output variable specified


-- Looking for Google Test sources

```
It is similar to #124 and goes away when I set `CMAKE_BUILD_TYPE`.

The variable is referenced without checking [here](https://github.com/google/benchmark/blob/823d24630db0f340914bb011f1ba31657348e510/src/CMakeLists.txt#L37).

Seems like CMAKE_BUILD_TYPE doesn't have any default value as mentioned [here](https://github.com/google/benchmark/blob/master/README.md#debug-vs-release). 

Platform: linux
Compiler: GCC 7.8
Cmake: 3.14.2

LebedevRI:
This *seems* like the right solution.
If none of those build types is wanted, `"None"` _should_ work.


CMakeLists.txt
```diff
@@ -14,6 +14,15 @@ endforeach()
 
 project (benchmark)
 
+# Add a sensible build type default and warning because empty means no optimization and no debug info.
```
dominichamon: this isn't the right approach as it forces a default. the correct approach is to guard the access to CMAKE_BUILD_TYPE in the pthread check with an `if(CMAKE_BUILD_TYPE)`.

LebedevRI: Interesting, this is the first time i'm seeing "it is okay to have no CMAKE_BUILD_TYPE".

dominichamon: Well, cmake allows you to not specify CMAKE_BUILD_TYPE and, in my experience, it defaults to the equivalent of 'debug'. At least locally for me (and afai remember our build bots).

PR #817, opened 2019-05-11
Iteration counts should be `uint64_t` globally.

LebedevRI:
This is a shameless rip-off of https://github.com/google/benchmark/pull/646
I did promise to look into why that proposed PR was producing
so much worse assembly, and so i finally did.

The reason is - that diff changes `size_t` (unsigned) to `int64_t` (signed).

There is this nice little `assert`:
https://github.com/google/benchmark/blob/7a1c37028359ca9d386d719a6ad527743cf1b753/include/benchmark/benchmark.h#L744
It ensures that we didn't magically decide to advance our iterator
when we should have finished benchmarking.

When `cached_` was unsigned, the `assert` was `cached_ UGT 0`.
But we only ever get to that `assert` if `cached_ NE 0`,
and naturally if `cached_` is not `0`, then it is bigger than `0`,
so the `assert` is tautological, and gets folded away.

But now that `cached_` became signed, the assert became `cached_ SGT 0`.
And we still only know that `cached_ NE 0`, so the assert can't be
optimized out, or at least it doesn't currently.

Regardless of whether or not that is a bug in itself,
that particular diff would have regressed the normal 64-bit systems,
by halving the maximal iteration space (since we go from unsigned counter
to signed one, of the same bit-width), which seems like a bug.
And just so it happens, fixing *this* bug, fixes the other bug.

This produces fully (bit-by-bit) identical state_assembly_test.s
The filecheck change is actually needed regarless of this patch,
else this test does not pass for me even without this diff.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1375,7 +1375,7 @@ class BenchmarkReporter {
     bool error_occurred;
     std::string error_message;
 
-    int64_t iterations;
+    IterationT iterations;
     int64_t threads;
     int64_t repetition_index;
     int64_t repetitions;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -411,11 +411,11 @@ enum TimeUnit { kNanosecond, kMicrosecond, kMillisecond };
 // calculated automatically to the best fit.
 enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
 
-typedef uint64_t IterationT;
+typedef uint64_t IterationCount;
 
 // BigOFunc is passed to a benchmark in order to specify the asymptotic
 // computational complexity for the benchmark.
-typedef double(BigOFunc)(IterationT);
+typedef double(BigOFunc)(IterationCount);
 
 // StatisticsFunc is passed to a benchmark in order to compute some descriptive
 // statistics over all the measurements of some type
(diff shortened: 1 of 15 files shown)


include/benchmark/benchmark.h
```diff
@@ -413,9 +411,11 @@ enum TimeUnit { kNanosecond, kMicrosecond, kMillisecond };
 // calculated automatically to the best fit.
 enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
 
+typedef uint64_t IterationT;
```
dominichamon: nit: Should be `Iterations` or `IterationCount`.

include/benchmark/benchmark.h
```diff
@@ -667,14 +667,14 @@ class State {
   const int threads;
 
  private:
-  State(size_t max_iters, const std::vector<int64_t>& ranges, int thread_i,
+  State(IterationT max_iters, const std::vector<int64_t>& ranges, int thread_i,
```
dominichamon: should we also define `Arg` or `Range` as a type instead of `int64_t` ?

src/benchmark_runner.cc
```diff
@@ -59,13 +59,13 @@ MemoryManager* memory_manager = nullptr;
 
 namespace {
 
-static const size_t kMaxIterations = 1000000000;
+static const IterationT kMaxIterations = 1000000000;
```
dominichamon: while you're here: constexpr?

include/benchmark/benchmark.h
LebedevRI: Let's go with `IterationCount`

include/benchmark/benchmark.h
LebedevRI: Hmm, possibly, but certainly not as `IterationCount`.

dominichamon:
Waiting for AppVeyor, then merge at will.

Issue #820, opened 2019-05-15
Use the IEC standard ("i" suffix) when dealing with OneK::kIs1024


MuAlphaOmegaEpsilon:
Maybe using the [IEC Standard](https://en.wikipedia.org/wiki/Kibibyte) for multiples of 2 units might make clearer the unit used to display results.

Issue #819, opened 2019-05-15
The "B" bytes suffix isn't really printed


MuAlphaOmegaEpsilon:
From the README.md file:
```
Benchmark                               Time(ns)    CPU(ns) Iterations
----------------------------------------------------------------------
BM_SetInsert/1024/1                        28928      29349      23853  133.097kB/s   33.2742k items/s
BM_SetInsert/1024/8                        32065      32913      21375  949.487kB/s   237.372k items/s
BM_SetInsert/1024/10                       33157      33648      21431  1.13369MB/s   290.225k items/s

```
The `B` suffix of bytes isn't printed.

aespondac:
This part of [README.md](https://github.com/google/benchmark/commit/201b981abd3a0bc63d2c2f5efd6871bdaceb2ccd#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L339-L345) code no longer exists, has been deleted at #1211.

@dmah42 , this Issue should be still open?

LebedevRI:
It's still in `docs/user_guide.md` though, isn't it?

Issue #818, opened 2019-05-13
Introduce scientific notation for counters


MuAlphaOmegaEpsilon:
Sometimes it becomes difficult to grasp immediately the differences between benchmarked functions that have a wide variance in a given counter.
The aforementioned readability problem is due to the usage of the SI standard by default.
This one below is an example where I am dealing with different functions to approximate sine, look at the absolute error calculated for each one: 
![Screenshot_20190513_185727](https://user-images.githubusercontent.com/26225010/57644787-27e37e80-75bd-11e9-9e94-240e0e4af5f5.png)
It's not instantly obvious how much more precise the first function is, but it would be using scientific notation:
![Screenshot_20190513_202852](https://user-images.githubusercontent.com/26225010/57645061-c7a10c80-75bd-11e9-948d-a86f488662f7.png)

The usage might be as easy as:
```cpp
state.counters["Foo"] = Counter(fooVal, benchmark::Counter::kScientificNotation);
```

dominichamon:
i don't have a strong opinion either way, but seeing a PR (and the effect it has on some test cases) might be illuminating.

LebedevRI:
The print format does not really fit within the current counter flags.
This is a yet another highlight that there is some missing extra abstraction for counters (just as timers) - the unit should be specifiable, including the printing specification and num/den strings.

MuAlphaOmegaEpsilon:
@dominichamon As soon as I have have some spare time I will upload a draft implementation.

@LebedevRI Yeah I know about the fitting problem, I had to retouch it a little bit to make the notation fit.
Right know I kept the decimal precision to 5 digits just to be on par with the maximum mantissa digits shown by the `HumanReadableFormat`.

LebedevRI:
No i meant API-vise, not printing

PR #821, opened 2019-05-15
Introduce ScientificNotation and little format updates and fixes

MuAlphaOmegaEpsilon:
This PR tries to solve this ISSUES:
* #818: Introduce scientific notation for counters in console
* #819: The "B" bytes suffix isn't really printed
* #820: Use the IEC standard ("i" suffix) when dealing with OneK::kIs1024

I tried to be as clean as possible when building commits.
I suggest to go through the commits one-by-one to follow the various branch development stages.

Any secondary commit (least fundamental ones) should be atomically droppable.

### DIFF
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -79,9 +79,8 @@ void ConsoleReporter::PrintHeader(const Run& run) {
                                  "Benchmark", "Time", "CPU", "Iterations");
   if(!run.counters.empty()) {
     if(output_options_ & OO_Tabular) {
-      for(auto const& c : run.counters) {
-        str += FormatString(" %10s", c.first.c_str());
-      }
+      for(auto const& c : run.counters)
+        str += FormatString(" %1$*2$s", c.first.c_str(), ColumnWidth(c));
     } else {
       str += " UserCounters...";
     }
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -70,7 +70,7 @@ static uint ColumnWidth(const std::pair <std::string, Counter>& pair) {
   const Counter& c = pair.second;
   const std::string& name = pair.first;
 
-  uint width = (c.format == Counter::kSI_BaseUnit) ? widths::SI : widths::SN;
+  uint width = (c.format == Counter::kSIUnits) ? widths::SI : widths::SN;
   width += (c.oneK == Counter::kIs1024) ? 1U : 0U;
   width += (c.flags & Counter::kIsRate) ? 2U : 0U;
   return widths::Padding + std::max(width, (uint) name.length());
@@ -178,7 +178,7 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   const bool tabular = (output_options_ & OO_Tabular);
   const char* fmt = tabular ? " %*s" : " %3$s=%2$*1$s";
   for (auto const& c : result.counters) { 
-    std::string value = (c.second.format == Counter::kSI_BaseUnit) ?
+    std::string value = (c.second.format == Counter::kSIUnits) ?
                         HumanReadableNumber(c.second.value, c.second.oneK) :
                         FormatString("%.*E", widths::DecimalPrecision, c.second.value);
     if (c.second.oneK == Counter::kIs1024) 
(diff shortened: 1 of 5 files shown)


src/console_reporter.cc
```diff
@@ -152,21 +175,19 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
-      } else {
-        printer(Out, COLOR_DEFAULT, " %*s", cNameLen, s.c_str());
-      }
-    } else {
-      const char* unit = (c.second.flags & Counter::kIsRate) ? "/s" : "";
-      printer(Out, COLOR_DEFAULT, " %s=%s%s", c.first.c_str(), s.c_str(),
-              unit);
-    }
+  const bool tabular = (output_options_ & OO_Tabular);
+  const char* fmt = tabular ? " %*s" : " %3$s=%2$*1$s";
```
LebedevRI: No. Now the compiler will not be able to check that the arguments make sense for the format string.

src/console_reporter.cc
```diff
@@ -152,21 +175,19 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
+  for (auto const& c : result.counters) { 
+    std::string value = (c.second.format == Counter::kSIUnits) ?
```
LebedevRI: I don't like how we are doing memory allocations for every counter printed

src/console_reporter.cc
```diff
@@ -152,21 +175,19 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
+                        HumanReadableNumber(c.second.value, c.second.oneK) :
+                        FormatString("%.*E", widths::DecimalPrecision, c.second.value);
+    if (c.second.oneK == Counter::kIs1024) 
+      value.append("iB");
```
LebedevRI: .. and then reallocations

LebedevRI:
(Also, i would *really* recommend to submit this as three separate PRs: #819 first, then #820 and then #818.)

src/console_reporter.cc
MuAlphaOmegaEpsilon: I previously did a local version using a char[] preallocated on the stack, but wasn't quite sure if doing the same thing with strcat and strcpy onto the buffer was as readable as the append calls.
Would you prefer the preallocated char buffer?

src/console_reporter.cc
MuAlphaOmegaEpsilon: Oh ok, totally missed this.

src/console_reporter.cc
LebedevRI: I would prefer the ugliest approach - just a lot of `printer` calls.
You will have to do that to make the format string a non-pointer anyway.

MuAlphaOmegaEpsilon:
I'll break this into three PRs then.
I'll also need to dig a little more on tests, since I must have missed that part while reading the contributing document.

src/console_reporter.cc
MuAlphaOmegaEpsilon: Got it. I expect to end up with only 2 calls to `printer` since 2 are the different format strings. 
Does it make sense for you?

src/console_reporter.cc
```diff
@@ -152,21 +175,19 @@ void ConsoleReporter::PrintRunData(const Run& result) {
     printer(Out, COLOR_CYAN, "%10lld", result.iterations);
   }
 
-  for (auto& c : result.counters) {
-    const std::size_t cNameLen = std::max(std::string::size_type(10),
-                                          c.first.length());
-    auto const& s = HumanReadableNumber(c.second.value, c.second.oneK);
```
LebedevRI: Ah, hm, this is already a string (too much `auto`!).
But still, appending to it is wrong.

src/console_reporter.cc
LebedevRI: Sounds plausible i suppose.

LebedevRI:
@MuAlphaOmegaEpsilon are you planning on rebasing this?
The formatting change might be a little bit more simple after https://github.com/google/benchmark/pull/850

MuAlphaOmegaEpsilon:
Sure, thanks for pinging me, I'll do this as soon as I find some spare time.
Do you still think it would be better to have this split in 3 PRs?

LebedevRI:
> Do you still think it would be better to have this split in 3 PRs?

Smaller patches are easier to review, so ideally yes please.
Maybe even split that `// Namespace containing constants related to counter values formatting.` cleanup into a preparatory NFC patch.

LebedevRI:
Reverse-ping.

LebedevRI:
@MuAlphaOmegaEpsilon do you plan on revisiting this?

dominichamon:
I'm going to close this. if anyone wants to take up the three PRs mentioned above i'd still like to take them.

PR #832, opened 2019-06-26
Make some functions const

orgads:
and ThreadManager ctor explicit.

Reported by CppCheck.

### DIFF
--- a/src/thread_manager.h
+++ b/src/thread_manager.h
@@ -11,7 +11,7 @@ namespace internal {
 
 class ThreadManager {
  public:
-  ThreadManager(int num_threads)
+  explicit ThreadManager(int num_threads)
       : alive_threads_(num_threads), start_stop_barrier_(num_threads) {}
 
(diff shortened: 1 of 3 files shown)


src/thread_manager.h
```diff
@@ -11,7 +11,7 @@ namespace internal {
 
 class ThreadManager {
  public:
-  ThreadManager(int num_threads)
+  explicit ThreadManager(int num_threads)
```
EricWF: I disagree with this change. A `ThreadManager` shouldn't be implicitly construct from an int.

orgads: But that's exactly what the explicit keyword states!

EricWF: Oh, I read the diff backwards somehow.

thank you for this fix.

dominichamon:
thank you!

orgads:
My pleasure :)

EricWF:
Thanks Dominic.

PR #833, opened 2019-07-02
Add RISC-V support in cycleclock::Now

lenary:
The RISC-V implementation of `cycleclock::Now` uses the user-space
`rdcycle` instruction to query how many cycles have happened since the
core started.

The only complexity here is on 32-bit RISC-V, where `rdcycle` can only
read the lower 32 bits of the 64-bit hardware counter. In this case,
`rdcycleh` reads the higher 32 bits of the counter. We match the powerpc
implementation to detect and correct for overflow in the high bits.

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -164,6 +164,21 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   uint64_t tsc;
   asm("stck %0" : "=Q"(tsc) : : "cc");
   return tsc;
+#elif defined(__riscv) // RISC-V
+  // Use RDCYCLE (and RDCYCLEH on riscv32)
+#if __riscv_xlen == 32
+  uint64_t cycles_low, cycles_hi0, cycles_hi1;
+  asm("rdcycleh %0" : "=r"(cycles_hi0));
+  asm("rdcycle %0" : "=r"(cycles_lo));
+  asm("rdcycleh %0" : "=r"(cycles_hi1));
+  // This matches the PowerPC overflow detection, above
+  cycles_lo &= -static_cast<int64_t>(cycles_hi0 == cycles_hi1);
+  return (cycles_hi1 << 32) | cycles_lo;
+#else
+  uint64_t cycles;
+  asm("rdcycle %0" : "=r"(cycles));
+  return cycles;
+#endif
 #else
 // The soft failover to a generic implementation is automatic only for ARM.
 // For other platforms the developer is expected to make an attempt to create


src/cycleclock.h
```diff
@@ -164,6 +164,21 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   uint64_t tsc;
```
LebedevRI: Not directly related to this patch, but this preprocessor hell is such a maze :/
This needs refactoring into separate functions.

src/cycleclock.h
```diff
@@ -164,6 +164,21 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
...
   return tsc;
+#elif defined(__riscv) // RISC-V
+  // Use RDCYCLE (and RDCYCLEH on riscv32)
+#if __riscv_xlen == 32
+  uint64_t cycles_low, cycles_hi0, cycles_hi1;
+  asm("rdcycleh %0" : "=r"(cycles_hi0));
+  asm("rdcycle %0" : "=r"(cycles_lo));
+  asm("rdcycleh %0" : "=r"(cycles_hi1));
+  // This matches the PowerPC overflow detection, above
+  cycles_lo &= -static_cast<int64_t>(cycles_hi0 == cycles_hi1);
```
LebedevRI: Now i'm curious, why is this different from the [compiler lowering](https://reviews.llvm.org/D64125) for `READCYCLECOUNTER`?

lenary: This hack matches the powerpc implementation higher in the file.

As far as @asb and I can tell, this clears all the low bits if the upper bits changed between the two reads, which will be approximately correct (you know when the upper bits changed, you overflowed, and the lower bits passed through zero). This avoids branching and looping, which might cause your cycle count to change more than this hack. Both the loop and this hack compromise accuracy, it's not clear which compromises it less, but at least the hack executes in constant time regardless of overflow. 

In the LLVM patch, I again copied the PPC lowering, which does use a loop, because that's what people will be expecting, and you know the returned low word is a bit pattern that was present in the `cycle` CSR (which you don't know with the hack, think of `rdcycle` taking two cycles and starting on an `UINT32_MAX`).

LebedevRI:
Looks ok but the CLA issue needs to be resolved.

lenary:
Yeah, it's going through the bureaucracy at work as we speak. Updates when I hear something.

asb:
> Yeah, it's going through the bureaucracy at work as we speak. Updates when I hear something.

Reluctant bureaucrat checking in :) I've docusigned the corporate CLA on behalf of lowRISC CIC, I think I now need to await a counter-signed version and perhaps explicitly list contributors afterwards.

lenary:
I signed it!

dominichamon:
thank you!

PR #812, opened 2019-05-02
Fix win32 link on case-sensitive fs

jschueller:
This fixes cross-builds with MinGW from case-sensitive file-systems (on Linux)

PR #840, opened 2019-07-21
CMake; windows: link to lowercase 'shlwapi' - consistent with headers

LebedevRI:
The filenames are consistently inconsistent in windows world, might
have something to do with default file system being case-insensitive.
While the native MinGW buils were fixed in 52613079824ac58d06c070aa9fbbb186a5859e2c
that only addressed the headers, but not libraries.
The problem remains when one tries to do a MinGW cross-build from
case-sensitive filesystem.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 
 # We need extra libraries on Windows
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
-  target_link_libraries(benchmark Shlwapi)
+  target_link_libraries(benchmark shlwapi)
 endif()
 
 # We need extra libraries on Solaris


LebedevRI:
Context: the patch was originally submitted by @jschueller here: https://reviews.llvm.org/D61371,
where i "suggested" to upstream it first.
So @jschueller submitted https://github.com/google/benchmark/pull/812, but it got stuck due to CLA
issues, and the LLVM patch got stuck too (most other bundled libraries there never backport
stuff upstream, so blocking patch with "please upstream first" is, not ideal.)
Then i suggested that if @jschueller is explicitly okay with someone "stealing" the patch and submitting
it upstream, this deadlock could be sidestepped.
So as per https://reviews.llvm.org/D61371#1586232 here we are.

dominichamon:
thanks, and thanks to @jschueller.

PR #846, opened 2019-07-29
corrected cmake commands

blackliner:


blackliner:
@googlebot I signed it!

LebedevRI:
Please can you explain what correction does this perform?
Why is the old one wrong?

blackliner:
Of course:
when i just tried out to cmake and make benchmark as it was described, i got errors. Which makes sense, because when you are in --/benchmark/build and do a cmake ../benchmark, there is no such folder --/benchmark/benchmark. I thought it was a spelling and acutally meant ../googletest, which makes sense, because you need to get the googletest targets first. After that you need to build benchmark via [cmake .. .]. You do not need to add the last dot to specify the current directory as output directory, but i think its more clear this way.
Cheers!

README.md
```diff
@@ -69,7 +69,10 @@ $ git clone https://github.com/google/googletest.git benchmark/googletest
 $ mkdir build && cd build
 # Generate a Makefile with cmake.
 # Use cmake -G <generator> to generate a different file type.
-$ cmake ../benchmark
```
LebedevRI:
```suggestion
$ cmake ../
```

LebedevRI:
> Of course:
> when i just tried out to cmake and make benchmark as it was described, i got errors. Which makes sense, because when you are in --/benchmark/build and do a cmake ../benchmark, there is no such folder --/benchmark/benchmark.

Yep, that's wrong.

> I thought it was a spelling and acutally meant ../googletest, which makes sense, because you need to get the googletest targets first.

Not, this is differently wrong.

>After that you need to build benchmark via [cmake .. .]. You do not need to add the last dot to specify the current directory as output directory, but i think its more clear this way.
> Cheers!

README.md
blackliner: I assumed one needs to cmake gtest first, but this is already enough. cmake finds googletest sources and does its thing.

blackliner:
Your last change should actually be enough. Changing it...

blackliner:
done

LebedevRI:
Now this looks ok.
I'm not sure about `CONTRIBUTORS`/`AUTHORS` files - when should they be updated.

blackliner:
> I'm not sure about `CONTRIBUTORS`/`AUTHORS` files - when should they be updated.

I have no idea, and there is no need to rush anything for deleting a few characters 😄

LebedevRI:
LG

Issue #856, opened 2019-08-21
Remove unused `doc` argument from `DEFINE_<data_type>(name, default_val, doc)` macros.


aerosayan:
As per the following discussion with Dominic Hamon in benchmark's google group : 

[LINK TO THE POST](https://groups.google.com/d/topic/benchmark-discuss/lTCY7GE6pd4/discussion)

it was understood that the `doc` argument in the `DEFINE_bool`, `DEFINE_int32`, `DEFINE_string` etc macros in `commandlineflags.h` can be removed and hence make the code much more clean.

Thus I will remove them and submit a pull request.
Thanks.

PR #857, opened 2019-08-21
Remove unused `doc` argument from `DEFINE_` macros.

aerosayan:
- Adresses : #856 

  - The unused `doc` argument was removed from the `DEFINE_` macros in
    `commandlineflags.h`

  - Converted all the previous `doc` strings passed to the `DEFINE_`
    macros to multiline comments.


dominichamon:
Thanks!

PR #861, opened 2019-08-23
Perform IWYU on library sources

LebedevRI:
Was looking at time distribution clang spends on my projects
that builds benchmark in-tree, and to my surprise
benchmark_register.cc + benchmark.cc + benchmark_runner.cc +
statistics.cc + benchmark.h were near the top.

There are two (well three) solutions here - to include less
stuff in benchmark's sources, to further partition benchmark's
sources (and to not build it in-tree :S).

This performs the first solution, since it's most straight-forward
and least controversial. The diff looks scary, and indeed i will be
quite surprized if this doesn't break build on some platform.
But I don't foresee any way to avoid that risk.

Repeating the measurements show that this either helps
(- ~2s clang front-end time), or the initial measurement was noisy.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -167,21 +167,23 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #define BENCHMARK_HAS_CXX11
 #endif
 
-#include <stdint.h>
+#include <stdint.h>  // IWYU pragma: keep
 
-#include <algorithm>
-#include <cassert>
-#include <cstddef>
-#include <iosfwd>
-#include <map>
-#include <set>
-#include <string>
-#include <vector>
+#include <algorithm>  // for max
+#include <cassert>    // for assert
+#include <cstddef>    // for size_t, ptrdiff_t
+#include <iosfwd>     // for ostream
+#include <iterator>   // for forward_iterator_tag
+#include <map>        // for map
+#include <set>        // for set
+#include <string>     // for string
+#include <vector>     // for vector
+// IWYU pragma: no_include <cstdint>
 
 #if defined(BENCHMARK_HAS_CXX11)
-#include <initializer_list>
-#include <type_traits>
-#include <utility>
+#include <initializer_list>  // IWYU pragma: keep
+#include <type_traits>       // for decay
+#include <utility>           // for pair, make_pair, forward
 #endif
 
 #if defined(_MSC_VER)
@@ -284,8 +286,6 @@ void RegisterMemoryManager(MemoryManager* memory_manager);
 
 namespace internal {
 class Benchmark;
-class BenchmarkImp;
-class BenchmarkFamilies;
 
 void UseCharPointer(char const volatile*);
 
--- a/src/benchmark_api_internal.h
+++ b/src/benchmark_api_internal.h
@@ -1,15 +1,15 @@
 #ifndef BENCHMARK_API_INTERNAL_H
 #define BENCHMARK_API_INTERNAL_H
 
-#include "benchmark/benchmark.h"
-#include "commandlineflags.h"
+#include <stdint.h>
 
-#include <cmath>
 #include <iosfwd>
-#include <limits>
-#include <memory>
 #include <string>
 #include <vector>
+
+#include "benchmark/benchmark.h"
+#include "thread_manager.h"  // IWYU pragma: keep
+#include "thread_timer.h"    // IWYU pragma: keep
 
 namespace benchmark {
 namespace internal {
--- a/src/benchmark_runner.h
+++ b/src/benchmark_runner.h
@@ -15,8 +15,11 @@
 #ifndef BENCHMARK_RUNNER_H_
 #define BENCHMARK_RUNNER_H_
 
-#include "benchmark_api_internal.h"
-#include "internal_macros.h"
+#include <vector>
+
+#include "benchmark/benchmark.h"
+#include "benchmark_api_internal.h"  // IWYU pragma: keep
+#include "commandlineflags.h"
 
 DECLARE_double(benchmark_min_time);
 
--- a/src/colorprint.cc
+++ b/src/colorprint.cc
@@ -22,14 +22,14 @@
 #include <string>
 
 #include "check.h"
-#include "internal_macros.h"
+#include "internal_macros.h"  // IWYU pragma: keep
 
 #ifdef BENCHMARK_OS_WINDOWS
 #include <windows.h>
 #include <io.h>
-#else
+#else  // ifdef BENCHMARK_OS_WINDOWS
 #include <unistd.h>
-#endif  // BENCHMARK_OS_WINDOWS
+#endif  // ifndef BENCHMARK_OS_WINDOWS
 
 namespace benchmark {
 namespace {
--- a/src/statistics.cc
+++ b/src/statistics.cc
@@ -13,15 +13,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "benchmark/benchmark.h"
+#include "statistics.h"
 
 #include <algorithm>
 #include <cmath>
+#include <iosfwd>
+#include <map>
 #include <numeric>
 #include <string>
+#include <utility>
 #include <vector>
+// IWYU pragma: no_include <memory>
+
+#include "benchmark/benchmark.h"
 #include "check.h"
-#include "statistics.h"
 
 namespace benchmark {
 
--- a/src/string_util.h
+++ b/src/string_util.h
@@ -4,7 +4,9 @@
 #include <sstream>
 #include <string>
 #include <utility>
-#include "internal_macros.h"
+
+#include "benchmark/benchmark.h"
+#include "internal_macros.h"  // IWYU pragma: keep
 
 namespace benchmark {
 
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -12,58 +12,64 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "internal_macros.h"
+#include "benchmark/benchmark.h"
+#include "internal_macros.h"  // IWYU pragma: keep
 
 #ifdef BENCHMARK_OS_WINDOWS
 #include <shlwapi.h>
 #undef StrCat  // Don't let StrCat in string_util.h be renamed to lstrcatA
 #include <versionhelpers.h>
 #include <windows.h>
+
 #include <codecvt>
-#else
-#include <fcntl.h>
-#ifndef BENCHMARK_OS_FUCHSIA
-#include <sys/resource.h>
-#endif
-#include <sys/time.h>
-#include <sys/types.h>  // this header must be included before 'sys/sysctl.h' to avoid compilation error on FreeBSD
+#include <memory>
+#else  // ifdef BENCHMARK_OS_WINDOWS
 #include <unistd.h>
+// IWYU pragma: no_include <bits/local_lim.h>
+#endif  // ifndef BENCHMARK_OS_WINDOWS
+
 #if defined BENCHMARK_OS_FREEBSD || defined BENCHMARK_OS_MACOSX || \
     defined BENCHMARK_OS_NETBSD || defined BENCHMARK_OS_OPENBSD
+#include <sys/types.h>  // this header must be included before 'sys/sysctl.h' to avoid compilation error on FreeBSD
 #define BENCHMARK_HAS_SYSCTL
 #include <sys/sysctl.h>
-#endif
-#endif
+#endif  //  defined BENCHMARK_OS_FREEBSD || defined BENCHMARK_OS_MACOSX ||
+        //  defined BENCHMARK_OS_NETBSD || defined BENCHMARK_OS_OPENBSD
+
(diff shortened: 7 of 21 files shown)


src/statistics.cc
```diff
@@ -13,16 +13,21 @@
...
-#include <numeric>
-#include <string>
-#include <vector>
-#include "check.h"
 #include "statistics.h"
 
+#include <algorithm>  // for max, nth_element, count_if
+#include <cmath>      // for sqrt
+#include <iosfwd>     // for size_t
+#include <map>        // for map, _Rb_tree_iterator, operator==
```
EricWF: `_RB_tree_iterator, operator==` are bad tool generated comments and should be removed.

src/statistics.cc
```diff
@@ -13,16 +13,21 @@
...
+#include <memory>     // for allocator_traits<>::value_type
```
EricWF: Bad comment.

src/statistics.cc
```diff
@@ -13,16 +13,21 @@
...
+#include <numeric>    // for accumulate, inner_product
+#include <string>     // for string, basic_string, operator!=
```
EricWF: bad comment: operator!=

src/statistics.cc
```diff
@@ -13,16 +13,21 @@
...
+#include <utility>    // for pair
+#include <vector>     // for vector
```
EricWF: Bod comment.

src/sysinfo.cc
```diff
@@ -12,58 +12,53 @@
...
+#include <cctype>     // for tolower
+#include <climits>    // for CHAR_BIT
+#include <cstdint>    // for uintptr_t
+#include <cstdio>     // for fprintf, size_t, stderr
+#include <cstdlib>    // for exit, getloadavg, EXIT_FAILURE
+#include <fstream>    // IWYU pragma: keep
+#include <iostream>  // for operator<<, ifstream, ostream, basi...
+#include <string>    // for string, basic_string, operator<<
+#include <utility>   // for forward
+#include <vector>    // for vector
```
EricWF: Bad comment.

include/benchmark/benchmark.h
```diff
@@ -167,21 +167,23 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
...
+#include <algorithm>  // for max
+#include <cassert>    // for assert
+#include <cstddef>    // for size_t, ptrdiff_t
+#include <iosfwd>     // for ostream
+#include <iterator>   // for forward_iterator_tag
+#include <map>        // for operator!=, map<>::mapped_type, map<>::const_...
+#include <set>        // for set
+#include <string>     // for string
+#include <vector>     // for vector
+// IWYU pragma: no_include <cstdint>
```
EricWF: What's this here for?

src/sysinfo.cc
```diff
@@ -12,58 +12,53 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "internal_macros.h"
-
 #ifdef BENCHMARK_OS_WINDOWS
```
EricWF: Where is the definition of `BENCHMARK_OS_WINDOWS` coming from if not `internal_macros.h`?

src/string_util.h
```diff
@@ -1,10 +1,11 @@
 #ifndef BENCHMARK_STRING_UTIL_H_
 #define BENCHMARK_STRING_UTIL_H_
 
-#include <sstream>
-#include <string>
-#include <utility>
-#include "internal_macros.h"
```
EricWF: I'm pretty sure `internal_macros.h` is required to configure correctly.

LebedevRI: Yeah, i was just looking into that

include/benchmark/benchmark.h
LebedevRI: `// IWYU pragma: no_include .*` tell IWYU that it is wrong and it should not suggest to include said `.*` here.
`// IWYU pragma: keep` is the other way around, that it should not suggest to delete it.
I could drop these annotations, but they will be needed the very next time anyone will
want to run IWYU again, it would be much harder to keep things correct without them.

LebedevRI:
> How were you measuring compilation times? Clang recently added the `-ftime-trace` flag which generates Chromium tracing reports.

Yes, that.

> They're wonderful for measuring this sort of thing.

> What's causing you to re-build the library so often? I would have assumed that since the sources aren't changing, the build targets are never out of date. I could understand a slowdown caused by `benchmark.h` while compiling your benchmarks though.

I was actually initially looking to perform this hygene cleanup without having any timing info,
it just so happened that i got to "measure" something before doing it, and the diff wasn't null,
thus i included it into the commit msg.

I don't really have a problem with how long it takes to build benchmark,
but that being said i don't like that there are timing outliers.

> IWYU has a bunch of flaws. The comments it generates aren't useful and are often outright wrong. It also misses plenty of required includes, and in doing so can break or cause bugs in code. I've pointed out a few instances where this happens in inline comments.

Yes, i have seen :/ It's rather far from LLVM tool quality.
But that being said, a tool is a tool. I don't really buy that it is better to not use tool at all,
just because it requires some manual intervention afterwards, it's not **that** bad.

> In general, I'm pretty skeptical this is an improvement.

include/benchmark/benchmark.h
```diff
@@ -167,21 +167,23 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
...
-#include <cassert>
-#include <cstddef>
-#include <iosfwd>
-#include <map>
-#include <set>
-#include <string>
-#include <vector>
+#include <stdint.h>  // IWYU pragma: keep
+
+#include <algorithm>  // for max
```
dominichamon: does IWYU add the "// for ..." comments?

src/benchmark_api_internal.cc
```diff
@@ -1,5 +1,16 @@
 #include "benchmark_api_internal.h"
 
+namespace benchmark {
```
dominichamon: did IWYU do this?

first, fold the two namespaces together.

Second, forward declarations aren't completely safe as they can hide a dependency. I'd rather take the hit on compilation time by including the relevant headers than have incomplete dependency trees.

src/benchmark_api_internal.h
```diff
@@ -1,15 +1,24 @@
...
-#include <vector>
+#include <stdint.h>  // for int64_t
+
+#include <iosfwd>  // for ostream
+#include <string>  // for string
+#include <vector>  // for vector
+
+#include "benchmark/benchmark.h"  // for IterationCount, AggregationReportMode
+
+namespace benchmark {
```
dominichamon: ditto re forward declarations

src/timers.cc
```diff
@@ -12,50 +12,44 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "timers.h"
```
dominichamon: the related header should always be included first. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

include/benchmark/benchmark.h
LebedevRI: By default, yes. I can drop those `// for ..` if wanted.

dominichamon: please do. it's adding noise to the diff and doesn't actually add anything. If we trust IWYU then we trust it; we don't need to document its decision-making.

src/benchmark_api_internal.cc
LebedevRI: I don't disagree that forward declarations aren't always completely safe,
especially if they are banned, like e.g. for STL.

But this is deep within the library, and does not manifest any issues thus far.
Please can you explain what kind of danger do you foresee here?

dominichamon: https://google.github.io/styleguide/cppguide.html#Forward_Declarations

LebedevRI:
CI is finally green, though it is limited to linux/osx/windows.
I think i replied/resolved all inline comments.

include/benchmark/benchmark.h
dominichamon: i'm still seeing the comments.

src/benchmark_runner.h
```diff
@@ -30,6 +32,8 @@ namespace benchmark {
 
 namespace internal {
 
+struct BenchmarkInstance;
```
dominichamon: just include benchmark_api_internal.h

src/colorprint.cc
```diff
@@ -22,14 +22,16 @@
...
-#include "internal_macros.h"
+#include "internal_macros.h"  // IWYU pragma: keep
 
 #ifdef BENCHMARK_OS_WINDOWS
 #include <windows.h>
 #include <io.h>
-#else
+#endif  // ifdef BENCHMARK_OS_WINDOWS
+
+#ifndef BENCHMARK_OS_WINDOWS
```
dominichamon: why is this better than #else ?

src/sysinfo.cc
```diff
@@ -12,59 +12,67 @@
...
+#include <cstring>  // IWYU pragma: keep
+#include <fstream>  // IWYU pragma: keep
 #include <iostream>
-#include <iterator>
-#include <limits>
+#include <string>
+#include <utility>
+#include <vector>
+
+#ifdef BENCHMARK_HAS_SYSCTL
```
dominichamon: this shouldn't need to be guarded...

src/sysinfo.cc
```diff
@@ -12,59 +12,67 @@
...
 
-#include "check.h"
 #include "cycleclock.h"
-#include "internal_macros.h"
-#include "log.h"
 #include "sleep.h"
 #include "string_util.h"
 
+#ifdef BENCHMARK_HAS_SYSCTL
+#include "check.h"
```
dominichamon: same here.. why is check guarded by sysctl?

src/timers.cc
```diff
@@ -13,49 +13,43 @@
...
+#include "internal_macros.h"  // IWYU pragma: keep
 
 #ifdef BENCHMARK_OS_WINDOWS
 #include <shlwapi.h>
 #undef StrCat  // Don't let StrCat in string_util.h be renamed to lstrcatA
 #include <versionhelpers.h>
 #include <windows.h>
-#else
-#include <fcntl.h>
-#ifndef BENCHMARK_OS_FUCHSIA
```
dominichamon: you've lost these FUCHSIA checks.

i'd strongly advise against messing with the OS_ checks at all. they're prone to breakage.

src/colorprint.cc
LebedevRI: It's indifferent in this place, but elsewhere it was closer to 
```
#if one
...
#elseif
#if two
...
#endif
#endif
```
and i think you can agree that is getting problematic to read, as compared with
```
#if one
..
#endif

#if !one && two
...
#endif
```

src/sysinfo.cc
LebedevRI: The functions provided by that header are only used within `#ifdef BENCHMARK_HAS_SYSCTL` block, i believe

src/sysinfo.cc
LebedevRI: Same as with `<memory>`, `CHECK*` here are only used within `#ifdef BENCHMARK_HAS_SYSCTL` block.

src/colorprint.cc
dominichamon: the latter is fine, but here i think the else is clearer. maybe we can eshrine this as "avoid #else .. #if ?

src/sysinfo.cc
dominichamon: i think that's too hard to know, and hard to maintain. including memory and check when they're not used isn't a significant enough cost to harm maintainability of the code.

LebedevRI: Can you be more specific please, are you worried about NOT including the headers needed (that will result in compilation error), or including unneeded headers?

dominichamon: i'm worried about the first bit: that we end up not including <memory> or "check" and they end up being needed either by the existing code or by a future change. it's easier to just include them.

LebedevRI: Well, and if future change requires boost includes so what then, add them now?
If include is needed it will need to be added, or else the code won't compile, it's impossible to miss :)

dominichamon: that's a fallacy. memory and check _are_ needed today for some configurations. it's easier to add them now globally than to remember to add them later (and it's not impossible to miss, depending on configuration and platform).

LebedevRI: > memory and check _are_ needed today for some configurations.

I fully agree, the key word being *some*.

> it's easier to add them now globally than to remember to add them later

> (and it's not impossible to miss, depending on configuration and platform).

I believe i'm missing some reasoning here. How can that happen?
The functionality of any C++ STL header is mandated by the standard,
If it provides too much, then we won't have any issues,
if it doesn't provide what it should then it's simply an non-compliant STL implementation.

dominichamon: if we don't cover a configuration in travis/appveyor then we rely on the developer checking their build. it wouldn't be great if they didn't, but we can't be sure they did.

i'd just rather not conditionally include c++ standard headers.

dominichamon:
i believe this PR may be too old to resurrect. wdyt, @LebedevRI ?

Issue #862, opened 2019-08-24
Add gitignore rule to ignore CMakeLists.txt.user file created by Qt-Creator.


aerosayan:
Using Qt-Creator to load the benchmark project using the `CMakeLists.txt` creates an XML file called `CMakeLists.txt.user` which pollutes `git status`.

Here is a snippet of the file extracted using `head` command.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE QtCreatorProject>
<!-- Written by QtCreator 4.3.0, 2019-08-24T07:07:21. -->
<qtcreator>
 <data>
  <variable>EnvironmentId</variable>
  <value type="QByteArray">{9168f244-edbf-4147-a8e1-44e12895c44b}</value>
 </data>
 <data>
  <variable>ProjectExplorer.Project.ActiveTarget</variable>
```

A rule is to be added to `.gitignore` to prevent the pollution of `git status`.

LebedevRI:
I was actually holding off on doing this because then it would be getting deleted during  `git clean -f -d -x`, which isn't what i'd want to happen/

aerosayan:
> I was actually holding off on doing this because then it would be getting deleted during `git clean -f -d -x`, which isn't what i'd want to happen.

I tried your command and I found that the file was getting deleted even when the rule was not added.
PS : I don't know much about `git clean`

LebedevRI:
That's because i meant `clean -d -f -X` apparently (it's an aliased as full-clean for me, so i don't recall)

PR #863, opened 2019-08-24
Add gitignore rule to ignore CMakeLists.txt.user

aerosayan:
- Addresses : #862

  - `CMakeLists.txt.user` is a XML file created by Qt-Creator if the
    benchmark project is loaded into Qt-Creator using the `CMakeLists.txt`
    file.

  - `CMakeLists.txt.user` pollutes `git status` thus the new rule is added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ __pycache__
 CMakeCache.txt
 CMakeFiles/
 cmake_install.cmake
+CMakeLists.txt.user
 
 # makefiles.
 Makefile


LebedevRI:
The thing to note here is that as opposed to every single(?) other thing there this isn't a tmp file
that can be trivially automatically reconstructed, it is very much a persistent config file,
it can only be recreated 'manually' by user.
QtCreator has made an extremely short-sighted decision storing it next to the `CMakeLists.txt`..

aerosayan:
> The thing to note here is that as opposed to every single(?) other thing there this isn't a tmp file
> that can be trivially automatically reconstructed, it is very much a persistent config file,
> it can only be recreated 'manually' by user.
> QtCreator has made an extremely short-sighted decision storing it next to the `CMakeLists.txt`..

Yes. I also observed the same behavior.

If the `CMakeLists.txt.user` file is deleted though, then the user can recover it by just restarting Qt Creator and re importing the project.

And it was observed that, since benchmark is a very light weight library the re import and re configuration time required is around 2 seconds for my `i5-4570` system.

LebedevRI:
> > The thing to note here is that as opposed to every single(?) other thing there this isn't a tmp file
> > that can be trivially automatically reconstructed, it is very much a persistent config file,
> > it can only be recreated 'manually' by user.
> > QtCreator has made an extremely short-sighted decision storing it next to the `CMakeLists.txt`..
> 
> Yes. I also observed the same behavior.
> 
> If the `CMakeLists.txt.user` file is deleted though, then the user can recover it by just restarting Qt Creator and re importing the project.
That's not true.

> And it was observed that, since benchmark is a very light weight library the re import and re configuration time required is around 2 seconds for my `i5-4570` system.

You also need to consider the time it takes to re-configure e.g. all the various cmake options, etc.

LebedevRI:
> Qt-Creator being one of the most frequently used full fledged IDE will probably be used by current and/or future contributors.

'will probably'?

The thing is, there is a large number of editors out there, and each one of them can have it's own
list of such temporary files. It's plain impossible to list *all* of them, but rather the cases that matter
should be listed.

Now, i'm using the editor in question, and i **know** this change will negatively impact me
for the reasons stated above. Is this change beneficial for your workflow since you are
using that editor?

dominichamon:
we should probably clean up the `.gitignore` if we think there's too much in there already. :)

Issue #869, opened 2019-09-18
Add support for hardware performance counters


wahez:
A benchmark can show you if some code is fast or slow. Having performance counters running automatically during a benchmark can also tell you why it is slow.

LebedevRI:
> Having performance counters running automatically during a benchmark can also tell you why it is slow.

Can you expand that thought please?
Right away i can't recall which of those counters would always tell *why* it is slow.

wahez:
Good point. Let me rephrase that: It will help you find out why it is slow (by allowing you to measure cache misses, instruction count, etc).

dominichamon:
is this something you could do with custom user counters? it might be worth a prototype that does it in "user space" (ie, in the benchmark) to show the value and then we could promote it into the library.

dominichamon:
now available thanks to #1114

PR #870, opened 2019-09-18
Added support for hardware performance counters via PAPI library (#869)

wahez:
This adds support for performance counters, so you can see why a
benchmark is slow. Currently only works for linux and if the papi
library and headers are found. The output is generated as
UserCounters.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -48,6 +48,7 @@ Sayan Bhattacharjee <aero.sayan@gmail.com>
 Shuo Chen <chenshuo@chenshuo.com>
 Steinar H. Gunderson <sgunderson@bigfoot.com>
 Stripe, Inc.
+Walther Zwart <walther.zwart@gmail.com>
 Yixuan Qiu <yixuanq@gmail.com>
 Yusuke Suzuki <utatane.tea@gmail.com>
 Zbigniew Skowron <zbychs@gmail.com>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,12 @@ cxx_feature_check(STEADY_CLOCK)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+# Check if we have libpapi headers
+find_package(Papi)
+if (Papi_FOUND)
+  add_definitions(-DBENCHMARK_HAS_PAPI)
+endif()
+
 # Set up directories
 include_directories(${PROJECT_SOURCE_DIR}/include)
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -69,6 +69,7 @@ Sayan Bhattacharjee <aero.sayan@gmail.com>
 Shuo Chen <chenshuo@chenshuo.com>
 Tobias Ulvgård <tobias.ulvgard@dirac.se>
 Tom Madams <tom.ej.madams@gmail.com> <tmadams@google.com>
+Walther Zwart <walther.zwart@gmail.com>
 Yixuan Qiu <yixuanq@gmail.com>
 Yusuke Suzuki <utatane.tea@gmail.com>
 Zbigniew Skowron <zbychs@gmail.com>
--- a/cmake/Modules/FindPapi.cmake
+++ b/cmake/Modules/FindPapi.cmake
@@ -1,0 +1,14 @@
+find_path(
+  Papi_INCLUDE_DIR
+  NAMES papi.h)
+
+find_library(Papi_LIBRARIES papi)
+
+include(FindPackageHandleStandardArgs)
+
+find_package_handle_standard_args(
+  Papi
+  DEFAULT_MSG
+  Papi_LIBRARIES Papi_INCLUDE_DIR)
+
+mark_as_advanced(Papi_INCLUDE_DIR Papi_LIBRARIES)
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -435,6 +435,7 @@ struct Statistics {
 
 struct BenchmarkInstance;
 class ThreadTimer;
+class PerformanceCounter;
 class ThreadManager;
 
 enum AggregationReportMode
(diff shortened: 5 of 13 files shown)


LebedevRI:
Two main observations:
1. This simply integrates into custom user counters,
    there is no crippling need for this to be in the library itself, correct?
2. This yet again shows that we really need custom "timers", to simplify adding such stuff,
    and make library more flexible and thus avoid pulling everything into the library itself :(

wahez:
Re 1. The output indeed integrates with user counters (which is not great, because it's not really a user counter), but unlike user counters, it doesn't need any change to the code for the benchmark. You can enable any performance counter for any existing benchmark by changing the commandline. I did it outside of the library before this, but it is inconvenient (you need to enable it explicitly per test in the code and you need to pause/resume in all the right places). This is what triggered me to integrate it into the library.

Re 2. I guess it could be kept outside of the library if the library supported custom timers. You would lose the commandline configuration and it would need to be enabled per test in the code, but at least it would automatically pause/resume. I understand that you don't want to pull every conceivable feature into the library, but in my opinion microbenchmarking and performance counters are a very natural fit.

biojppm:
I agree with @wahez regarding the natural fit between user counters and hardware performance counters. In other words, providing HW counter facilities is something that really makes sense for this library. 

In fact, PAPI and other HW counters were exactly my main motivation when I added user counters to this library, see eg #240 and #262 . 

Having said this, putting HW counters in every benchmark as this PR proposes is not a good idea; that approach is something that IMO should be revised.

First of all, not all counters are available in every platform. For example, AFAIK PAPI is Linux only. There's also NVIDIA counters for GPUs, etc. So it stands to reason that this library should not always look for everything that might be of interest.

That is, the user should explicitly request these facilities. So if HW counters of any type are to be added, something like this should be the approach:

```cmake
# in the main cmake file
option(ENABLE_PAPI_COUNTERS
    "Enable facilities for setting PAPI hardware counters"
    OFF)

# ...
if(ENABLE_PAPI_COUNTERS)
    # find papi includes and libs
    # ...
    target_compile_definitions(benchmark
         PUBLIC -DBENCHMARK_HAVE_PAPI)
endif()
```

Then, providing this in the library could be done through functions, and maybe a glue fixture. Something like this:

```c++
// in a specific header file, eg, benchmark/papi.h.
// Maybe this could even be in a separate library bundled with benchmark, eg benchmark_papi

#ifdef BENCHMARK_HAVE_PAPI
// bit flags for individual counters
typedef enum : uint32_t {
    // data cache misses
    L1_DCM = (1 << 0),
    L2_DCM = (1 << 1),
    L3_DCM = (1 << 2),
    // instruction cache misses
    L1_ICM = (1 << 0),
    L2_ICM = (1 << 1),
    L3_ICM = (1 << 2),
    // provide aggregate bitmasks to make it easier for the user
    DCM = (L1_DCM|L2_DCM|L3_DCM),
    ICM = (L1_ICM|L2_ICM|L3_ICM),
    // etc; same approach for remaining possible counters
    //...
}  PAPICounters;

//...

// function to enable counters on the state; to be called before entering the bm loop
void startPAPICounters(State &st, PAPICounters counters)
{
    // go through the requested flags, and start the counters
    if(counters && L1_DCM) {
        // call papi to start counting PAPI_L1_DCM
        // ....
    } 
}

void stopPAPICounters(benchmark::State &st, PAPICounters counters)
{
    if(counters && L1_DCM) {
        auto l1dcm = /*get value from PAPI*/;
        st.counters["L1_DCM"] = Counter(l1dcm, Counter::kIsRate);
    } 
    // etc...
}

// assume arg(0) is where we get our counter specs
class PAPIFixture : public Fixture {
public:
  void SetUp(const State& st) {
    startPAPICounters(st, static_cast<PAPICounters>(st.Arg(0)));
  }
  void TearDown(const State& state) {
    stopPAPICounters(st, static_cast<PAPICounters>(st.Arg(0)));
  }
};
#endif // BENCHMARK_HAVE_PAPI
```

With the above, client code would be something like:

```c++
#include <benchmark/papi.h>
BENCHMARK_DEFINE_F(PAPIFixture, BarTest)(benchmark::State& st) {
   for (auto _ : st) {
     ...
  }
}
BENCHMARK_REGISTER_F(PAPIFixture, BarTest)->Arg(DCM|ICM);
// measure all data+instruction cache misses 
```

The macro call above is bulky and stands to be improved; we could ideally achieve something as easy for the user  as this:

```c++
#include <benchmark/papi.h>
BENCHMARK_REGISTER_PAPI_F(BarTest, DCM|ICM) {
   for (auto _ : st) {
     ...
  }
}
```

HTH.

wahez:
@biojppm, thanks for your input.
I like the cmake change you propose. Also I can see that it would be useful to enable the counters in the code.
What I don't agree with is to enable the counters during Setup. The reason why I would like this in the library is to have the performance counters pause when the timer pauses.
Also I'm not sure it's a good idea to copy all the papi counters into the library. That would be a lot of repetition and the available counters might depend on the papi version. That would give extra maintenance especially if you do the same for windows and nvidia gpus.

biojppm:
> What I don't agree with is to enable the counters during Setup. The reason why I would like this in the library is to have the performance counters pause when the timer pauses.

This is a valid point you make. But it is not necessarily an argument against using a fixture. It's just that as it currently stands, the fixture class version does not accommodate use cases such as these, where we want to hook into when the timer starts/stops and not just SetUp()/TearDown(). I see counters more as a fixture-thing than a timer-thing, but I reckon it could be argued either way. If there is agreement that this sort of activity should happen at the fixture level, then we could propose extending the Fixture interface to accommodate this use case.

I also accept your point on integration of the PAPI counters. But please note that you are doing exactly that in your ```internal::PerformanceCounter``` class, and with a tighter coupling.

dominichamon:
This appears to be abandoned. It's a shame as i think that there is a place for hardware counters if we can find the right approach.

wahez:
I would still like to have this in the library, but given the comments I
don't know what the right approach is. If there is a consensus, I'm happy
to make the changes.

Walther

dominichamon:
#1114 is, i think, a cleaner approach to the same issue as it is opt-in both at compile and runtime.

dominichamon:
landed version using libpfm in #1114

PR #874, opened 2019-09-19
Guard ASSERT_THROWS checks with BENCHMARK_HAS_NO_EXCEPTIONS

### LINKED ISSUE #872: string_util_gtest.cc fails with exceptions turned off
This makes use of ASSERT_THROWS, but fails when exceptions are turned off. Should be pretty simple to guard those test cases with BENCHMARK_HAS_NO_EXCEPTIONS

I'll send a PR

GMNGeoffrey:
This allows the test be run with exceptions turned off

Also add myself to CONTRIBUTORS, as this is my first PR.
I don't need to be added to AUTHORS, as I am a Google employee.

This closes #872


dominichamon:
Oh great.. We must miss this case in the travis config. Thanks for catching it!

PR #875, opened 2019-09-20
Guard definition of __STDC_FORMAT_MACROS in ifndef

### LINKED ISSUE #873: benchmark_register.cc redefines __STDC_FORMAT_MACROS
This causes build errors if __STDC_FORMAT_MACROS is already defined. I think the definition can just be guarded by an ifndef. I'll send a PR

GMNGeoffrey:
This macro is sometimes already defined and redefining it results
in build errors.

This closes #873

### DIFF
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -34,7 +34,9 @@
 #include <sstream>
 #include <thread>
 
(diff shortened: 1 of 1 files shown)


dominichamon:
thanks!

PR #876, opened 2019-09-20
Define HOST_NAME_MAX for NaCl and RTEMS

### LINKED ISSUE #871: HOST_NAME_MAX undefined for NaCl, Movidius Myriad, (others?)
The sysinfo.cc [GetSystemName function](https://github.com/google/benchmark/blob/b988639f316ed0aca69297effbf67015c3bf0640/src/sysinfo.cc#L413) uses HOST_NAME_MAX, but this is not defined for all systems. In particular, I'm currently getting build errors on NaCl and Movidious Myriad. The former has a BENCHMARK_OS_* macro already defined, but the latter does not, I think.

Also note that NaCl actually does have a HOST_NAME_MAX defined in local_lim.h, but that only ends up getting included in limits.h if __USE_POSIX is defined (http://stackoverflow.com/q/30084116). So unconditionally defining HOST_NAME_MAX there seems inadvisable, since with posix set it will error on redefinition.

Also since this is just a debug function for getting the system name, it seems better if this is not a build-stopping error. We could assume some default value for HOST_NAME_MAX, return the empty string if it's not set, or something else. Either way, could issue a #warning if HOST_NAME_MAX is not set.

If the host name is too big to fit in the length passed to gethostname, it's [undefined](http://man7.org/linux/man-pages/man2/gethostname.2.html) whether there's an error and whether the buffer is null terminated. I think since we're constructing the std::string from an array, it will work either way though, so we should be fine just using some default for HOST_NAME_MAX.

I'm thinking something like:

```
std::string GetSystemName() {
// ...
#ifndef HOST_NAME_MAX
#ifdef BENCHMARK_HAS_SYSCTL // BSD/Mac Doesnt have HOST_NAME_MAX defined
#define HOST_NAME_MAX 64
#elif defined(BENCHMARK_OS_NACL)
#define HOST_NAME_MAX 64
#elif defined(BENCHMARK_OS_QNX)
#define HOST_NAME_MAX 154
#else
#warning "HOST_NAME_MAX not defined. using 64"
#define HOST_NAME_MAX 64
#endif
#endif // def HOST_NAME_MAX
  char hostname[HOST_NAME_MAX];
  int retVal = gethostname(hostname, HOST_NAME_MAX);
  if (retVal != 0) return std::string("");
  return std::string(hostname);
#endif // Catch-all POSIX block.
}
```

Thoughts?

GMNGeoffrey:
These OS's don't always have HOST_NAME_MAX defined, resulting in
build errors.

A few related changes as well:
* Only define HOST_NAME_MAX if it's not already defined. There are
  some cases where this is already defined, e.g. with NaCl if
  __USE_POSIX is set. To avoid all of these, only define it if it's
  not already defined.
* Default HOST_NAME_MAX to 64 and issue a #warning. Having the wrong
  max length is pretty harmless. The name just ends up getting
  truncated and this is only for printing debug info. Because we're
  constructing a std::string from a char[] (so defined length), we
  don't need to worry about gethostname's undefined behavior for
  whether the truncation is null-terminated when the hostname
  doesn't fit in HOST_NAME_MAX. Of course, this doesn't help people
  who have -Werror set, since they'll still get a warning.

This closes #871

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -429,11 +429,20 @@ std::string GetSystemName() {
 #endif
   return str;
 #else // defined(BENCHMARK_OS_WINDOWS)
+#ifndef HOST_NAME_MAX
 #ifdef BENCHMARK_HAS_SYSCTL // BSD/Mac Doesnt have HOST_NAME_MAX defined
+#define HOST_NAME_MAX 64
+#elif defined(BENCHMARK_OS_NACL)
 #define HOST_NAME_MAX 64
 #elif defined(BENCHMARK_OS_QNX)
 #define HOST_NAME_MAX 154
+#elif defined(BENCHMARK_OS_RTEMS)
+#define HOST_NAME_MAX 256
+#else
+#warning "HOST_NAME_MAX not defined. using 64"
+#define HOST_NAME_MAX 64
 #endif
+#endif // def HOST_NAME_MAX
   char hostname[HOST_NAME_MAX];
   int retVal = gethostname(hostname, HOST_NAME_MAX);
   if (retVal != 0) return std::string("");


GMNGeoffrey:
I don't really understand the travis failures. A couple of a failures are failed apt-get commands, which seem very unlikely to be my change (https://travis-ci.org/google/benchmark/jobs/587582970, https://travis-ci.org/google/benchmark/jobs/587582971)

The other one looks like the linker failed for cmake, but not for bazel. It gives no information about the failure though... https://travis-ci.org/google/benchmark/jobs/587582972

dominichamon:
restarting travis build just to check. agree it's odd but we'll see post-restart.

dominichamon:
great. thanks!

PR #878, opened 2019-09-24
fixed the param order in g++ command and fixed the path for -L

cpei-avalara:
The old parameter order in g++ command gives error when compiles. And the path after -L in that same line is not consistent with the one used above that section.


cpei-avalara:
@googlebot I signed it!

cpei-avalara:
@googlebot I signed it!

PR #883, opened 2019-09-24
Read options from environment (Fixes #881)

### LINKED ISSUE #881: Read options from environment
Benchmark execution options can be specified using the CLI (eg. `--benchmark_out=<filename>` for the results file location). Would be nice to also be able to read these options from the execution environment:  setting `BENCHMARK_OUT_FORMAT=<filename>` would be equivent of `--benchmark_out_format=json`. Google test has a [similar mechanism](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#controlling-test-output).

It seems like Benchmark [already has some helper code for this](https://github.com/google/benchmark/blob/1d41de84633bf40ddf0ad50165d77200e3d8b5fd/src/commandlineflags.cc#L80). Activating it would help integrating benchmarks execution in CI pipelines driven by env. variables for example.

t-chaik:
Initialize option flags from environment variables values if they are defined, eg. `BENCHMARK_OUT=<filename>` for `--benchmark_out=<filename>`. Command line flag value always prevails.

Fixes https://github.com/google/benchmark/issues/881.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -73,14 +73,14 @@ DEFINE_int32(benchmark_repetitions, 1);
 // Report the result of each benchmark repetitions. When 'true' is specified
 // only the mean, standard deviation, and other statistics are reported for
 // repeated benchmarks. Affects all reporters.
-DEFINE_bool( benchmark_report_aggregates_only, false);
+DEFINE_bool(benchmark_report_aggregates_only, false);
 
 // Display the result of each benchmark repetitions. When 'true' is specified
 // only the mean, standard deviation, and other statistics are displayed for
 // repeated benchmarks. Unlike benchmark_report_aggregates_only, only affects
 // the display reporter, but  *NOT* file reporter, which will still contain
 // all the output.
-DEFINE_bool( benchmark_display_aggregates_only, false);
+DEFINE_bool(benchmark_display_aggregates_only, false);
 
 // The format to use for console output.
 // Valid values are 'console', 'json', or 'csv'.
@@ -142,7 +142,7 @@ State::State(IterationCount max_iters, const std::vector<int64_t>& ranges,
   // which must be suppressed.
 #if defined(__INTEL_COMPILER)
 #pragma warning push
-#pragma warning(disable:1875)
+#pragma warning(disable : 1875)
 #elif defined(__GNUC__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Winvalid-offsetof"
@@ -309,7 +309,6 @@ std::unique_ptr<BenchmarkReporter> CreateReporter(
 #pragma GCC diagnostic pop
 #endif
 
-
 }  // end namespace
 
 bool IsZero(double n) {
@@ -318,7 +317,7 @@ bool IsZero(double n) {
 
 ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
   int output_opts = ConsoleReporter::OO_Defaults;
-  auto is_benchmark_color = [force_no_color] () -> bool {
+  auto is_benchmark_color = [force_no_color]() -> bool {
     if (force_no_color) {
       return false;
     }
--- a/src/commandlineflags.cc
+++ b/src/commandlineflags.cc
@@ -14,6 +14,7 @@
 
 #include "commandlineflags.h"
 
+#include <algorithm>
 #include <cctype>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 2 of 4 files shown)


src/benchmark.cc
```diff
@@ -52,56 +52,66 @@
 #include "thread_timer.h"
 
 // Print a list of benchmarks. This option overrides all other options.
-DEFINE_bool(benchmark_list_tests, false);
+DEFINE_bool(benchmark_list_tests,
+            benchmark::BoolFromEnv("benchmark_list_tests", false));
```
LebedevRI: This looks too intrusive, it should be handled on some other level of abstraction probably

LebedevRI: To be more specific, i think the e.g. `DEFINE_bool` itself should be modified.

test/commandlineflags_gtest.cc
```diff
@@ -34,14 +34,30 @@ TEST(BoolFromEnv, False) {
...
 }
 
 TEST(BoolFromEnv, True) {
   ASSERT_EQ(setenv("BENCHMARK_IN_ENV", "1", 1), 0);
   EXPECT_EQ(BoolFromEnv("in_env", false), true);
   unsetenv("BENCHMARK_IN_ENV");
 
-  ASSERT_EQ(setenv("BENCHMARK_IN_ENV", "foo", 1), 0);
+  ASSERT_EQ(setenv("BENCHMARK_IN_ENV", "bar", 1), 0);
   EXPECT_EQ(BoolFromEnv("in_env", false), true);
```
t-chaik: Fun: `"foo"` starts with an `f` which evaluates to `false` for `IsTruthyFlagValue()`... Would it be an acceptable breakage to change `IsTruthyFlagValue()`'s logic for something more strict? Any other suggestion welcome of course.

src/commandlineflags.cc
```diff
@@ -92,44 +92,50 @@ static std::string FlagToEnvVar(const char* flag) {
...
+bool BoolFromEnv(const char* flag, bool default_val) {
   const std::string env_var = FlagToEnvVar(flag);
-  const char* const string_value = getenv(env_var.c_str());
-  return string_value == nullptr ? default_value
-                                 : strcmp(string_value, "0") != 0;
+  const char* const value_str = getenv(env_var.c_str());
+  return value_str == nullptr ? default_val : IsTruthyFlagValue(value_str);
+}
+
+int32_t Int32FromEnv(const char* flag, int32_t default_val) {
```
dominichamon: not for this change: should this be `int` instead of `int32_t`?

src/commandlineflags.cc
```diff
@@ -92,44 +92,50 @@ static std::string FlagToEnvVar(const char* flag) {
...
+  if (value_str == nullptr) {
+    // The environment variable is not set.
+    return default_val;
+  }
+
+  int32_t value = default_val;
+  if (!ParseInt32(std::string("Environment variable ") + env_var, value_str,
+                  &value)) {
+    return default_val;
+  }
```
dominichamon: this can be simplified to:

```c++
const char* const value_str = getenv(env_var.c_str());
int32_t value = default_val;
if (value_str == nullptr ||
    !ParseInt32(std::string("Environment variable ") + env_var, value_str, &value)) {
  return default_val;
}
return value;
```

test/commandlineflags_gtest.cc
dominichamon: i think as long as it supports:

T, t, yes, true, True, TRUE as true and
F, f, no, false, False, FALSE as false

we can accept it.

dominichamon:
appveyor error looks real:

```
541 C:\projects\benchmark\test\commandlineflags_gtest.cc(117): error: Expected equality of these values:
542  BoolFromEnv("in_env", false)
543    Which is: false
544  true
545[  FAILED  ] BoolFromEnv.True (1 ms)
546
```

mablanchard:
@LebedevRI, @dominichamon: does that patch look better to you now? Anything else to adjust/change?

src/commandlineflags.cc
```diff
@@ -92,44 +92,40 @@ static std::string FlagToEnvVar(const char* flag) {
...
-                                 : strcmp(string_value, "0") != 0;
+  const char* const value_str = getenv(env_var.c_str());
+  return value_str == nullptr ? default_val : IsTruthyFlagValue(value_str);
 }
 
-// Reads and returns a 32-bit integer stored in the environment
-// variable corresponding to the given flag; if it isn't set or
-// doesn't represent a valid 32-bit integer, returns default_value.
-int32_t Int32FromEnv(const char* flag, int32_t default_value) {
+int Int32FromEnv(const char* flag, int default_val) {
```
dominichamon: this should still return and take an `int32_t`.

src/commandlineflags.cc
```diff
@@ -214,9 +210,15 @@ bool IsFlag(const char* str, const char* flag) {
...
-  if (value.empty()) return true;
-  char ch = value[0];
-  return isalnum(ch) &&
-         !(ch == '0' || ch == 'f' || ch == 'F' || ch == 'n' || ch == 'N');
+  if (value.size() == 1) {
+    char v = value[0];
+    return isalnum(v) &&
+           !(v == '0' || v == 'f' || v == 'F' || v == 'n' || v == 'N');
+  } else if (!value.empty()) {
+    return !(value == "false" || value == "False" || value == "FALSE" ||
```
dominichamon: what about fAlSe? :)

maybe transform value to lower case before testing?

LebedevRI: I'd suggest also accepting `off`

src/commandlineflags.h
```diff
@@ -10,22 +10,51 @@
 // Macros for declaring flags.
 #define DECLARE_bool(name) extern bool FLAG(name)
 #define DECLARE_int32(name) extern int32_t FLAG(name)
-#define DECLARE_int64(name) extern int64_t FLAG(name)
```
LebedevRI: Because it's not used?

LebedevRI:
I agree, this is the correct direction, and the PR is almost there. Thank you for working on this!
Last few comments.

src/commandlineflags.h
t-chaik: Yes, I can keep it around if you prefer. We are lacking `ParseInt64Flag`/`Int64FromEnv` though.

LebedevRI: No, i think that okay.

LebedevRI:
Hm, one last thought - we probably want something about this in docs?
But that can come in next pr.

LebedevRI:
@t-chaik thank you for the contribution!

t-chaik:
@LebedevRI , @dominichamon: thanks for the reviews! Any idea when the next release (including this) will be cut?

LebedevRI:
I don't think there's strict release schedule, so somewhere in the next 7 months i guess?

t-chaik:
@LebedevRI: Ok, thanks again!

dominichamon:
@LebedevRI ouch, that hurts :stuck_out_tongue: 

@t-chaik I keep an eye on git log since the last release and make a call when enough fixes/big pieces land. So some time in the next couple of months probably.

LebedevRI:
Yeah, i guess it wouldn't be bad to try to cut it every 6 months.

Issue #885, opened 2019-10-03
Cross compiling with toolchain file fails to configure on second run of CMake


KyleFromKitware:
When configuring with:

```
cmake .. -GNinja -DCMAKE_TOOLCHAIN_FILE=/path/to/arm/toolchain.cmake
```

The first time around I see the following messages:

```
-- Performing Test HAVE_STD_REGEX
CMake Warning at third_party/benchmark/cmake/CXXFeatureCheck.cmake:38 (message):
  If you see build failures due to cross compilation, try setting
  HAVE_STD_REGEX to 0
Call Stack (most recent call first):
  third_party/benchmark/CMakeLists.txt:247 (cxx_feature_check)


-- Performing Test HAVE_STD_REGEX -- success
-- Performing Test HAVE_GNU_POSIX_REGEX
-- Performing Test HAVE_GNU_POSIX_REGEX -- failed to compile
-- Performing Test HAVE_POSIX_REGEX
CMake Warning at third_party/benchmark/cmake/CXXFeatureCheck.cmake:38 (message):
  If you see build failures due to cross compilation, try setting
  HAVE_POSIX_REGEX to 0
Call Stack (most recent call first):
  third_party/benchmark/CMakeLists.txt:249 (cxx_feature_check)


-- Performing Test HAVE_POSIX_REGEX -- success
-- Performing Test HAVE_STEADY_CLOCK
CMake Warning at third_party/benchmark/cmake/CXXFeatureCheck.cmake:38 (message):
  If you see build failures due to cross compilation, try setting
  HAVE_STEADY_CLOCK to 0
Call Stack (most recent call first):
  third_party/benchmark/CMakeLists.txt:257 (cxx_feature_check)


-- Performing Test HAVE_STEADY_CLOCK -- success
```

This looks good, as it is warning me about the failure to run due to cross-compiling. However, when I run CMake a second time, I see the following:

```
-- Performing Test HAVE_STD_REGEX -- compiled but failed to run
-- Performing Test HAVE_GNU_POSIX_REGEX -- failed to compile
-- Performing Test HAVE_POSIX_REGEX -- compiled but failed to run
CMake Error at third_party/benchmark/CMakeLists.txt:251 (message):
  Failed to determine the source files for the regular expression backend


-- Configuring incomplete, errors occurred!
```

This is because of a variable in the CMake script that isn't being cached. I am currently preparing a patch to fix this issue.

PR #886, opened 2019-10-03
Cache RUN_${FEATURE} variable in CXXFeatureCheck.cmake

KyleFromKitware:
When cross-compiling, this variable was not set on the second run
of CMake, resulting in the next check failing even though it shouldn't
be run in the first place. Fix this by caching the variable.

Fixes: #885


KyleFromKitware:
@googlebot I signed it!

dominichamon:
Thanks!

PR #887, opened 2019-10-08
Switch to Starlark for C++ rules.

junyer:
While I'm here, update to google/googletest v1.10.0 and also
format all of the files that I touched.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -8,6 +8,8 @@ config_setting(
     visibility = [":__subpackages__"],
 )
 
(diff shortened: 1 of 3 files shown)


junyer:
Tested with `bazel build //...`. PTAL, @dominichamon. :)

dominichamon:
do you mind separating out the googletest bump from the rules change? just in case one or the other causes issues for some platform or some users.

otherwise, LGTM. and thanks!

junyer:
Sure, done. :)

Issue #888, opened 2019-10-10
Building on Windows with VS2017 and BUILD_SHARED_LIBS=ON doesn't create .lib files


KyleFromKitware:
When I configure on Windows with:

```
cmake .. -GNinja -DBUILD_SHARED_LIBS=ON
```

and build, no .lib file is generated for the .dlls. Furthermore, running `dumpbin` on one of the .dlls:

```
dumpbin /exports src\benchmark.dll
```

confirms that no symbols are exported in the .dll. I suspect they are not given the proper `__declspec(dllexport)` directive.

ura1286:
can you share the steps of building on windows with Visual 2017?

KyleFromKitware:
```
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Tools\VsDevCmd.bat"
cmake .. -GNinja -DBUILD_SHARED_LIBS=ON -DBENCHMARK_ENABLE_GTEST_TESTS=OFF
ninja
```

I disabled tests so I wouldn't have to deal with building GTest.

PR #889, opened 2019-10-10
Set CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS

KyleFromKitware:
When building on Windows with BUILD_SHARED_LIBS=ON, the symbols were
not being properly exported in the DLL. Fix this by setting
CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS.

Fixes: #888

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ option(BENCHMARK_DOWNLOAD_DEPENDENCIES "Allow the downloading and in-tree buildi
 # in cases where it is not possible to build or find a valid version of gtest.
 option(BENCHMARK_ENABLE_GTEST_TESTS "Enable building the unit tests which depend on gtest" ON)
 
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 set(ENABLE_ASSEMBLY_TESTS_DEFAULT OFF)
 function(should_enable_assembly_tests)
   if(CMAKE_BUILD_TYPE)


CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ option(BENCHMARK_DOWNLOAD_DEPENDENCIES "Allow the downloading and in-tree buildi
 # in cases where it is not possible to build or find a valid version of gtest.
 option(BENCHMARK_ENABLE_GTEST_TESTS "Enable building the unit tests which depend on gtest" ON)
 
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
LebedevRI: Move it higher, and make it a cache variable:
```
set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ${BUILD_SHARED_LIBS} CACHE BOOL "")
```

KyleFromKitware: Why a cache variable? This only takes effect when building with `BUILD_SHARED_LIBS` on Windows - otherwise it does nothing. See [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html).

LebedevRI: I think it may be best to force users hand if the `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` was already specified?

KyleFromKitware: This variable is meant to be set by the project, not by the user. It's not intended to be a user-toggleable feature, but merely a convenience to avoid having to prefix everything with `__declspec(dllexport)`.

dominichamon:
travis failure is unrelated (apt timeout/failure)

CMakeLists.txt
zackgalbreath: I'm also not seeing the benefit of making this a user-settable cache variable. I don't think you would ever want to build gRPC DLLs that export no symbols, would you?

LebedevRI:
I think this is okay as a stop-gap measure.
But i really think a proper fix (with `-fvisibility=hidden` default, and a white-list) is needed.
I tried looking into that previously, but got sidetracked.

KyleFromKitware:
> But i really think a proper fix (with -fvisibility=hidden default, and a white-list) is needed.

I agree, but that will involve sifting through a lot of code. This at least gets it working.

PR #896, opened 2019-10-24
Remove superfluous scaling in JSON reporter. Fixes #884

### LINKED ISSUE #884: Cache size unit in JSON output
I see overscaled cache sizes in the json output, like
```
"caches": [
      {
        "type": "Data",
        "level": 1,
        "size": 32000000,
        "num_sharing": 2
      } ...
```
while the console output is ok.
```
CPU Caches:
  L1 Data 32K (x6)
```
The problem seems to be that the size is multiplied by 1000 again [here](https://github.com/google/benchmark/blob/master/src/json_reporter.cc#L138), though it's already in bytes from [here](https://github.com/google/benchmark/blob/b8bce0c7ed4e02426fa4de21fec385fbaa84a7b2/src/sysinfo.cc#L273).

leanil:
Fixes https://github.com/google/benchmark/issues/884.

### DIFF
--- a/src/json_reporter.cc
+++ b/src/json_reporter.cc
@@ -135,7 +135,7 @@ bool JSONReporter::ReportContext(const Context& context) {
     out << cache_indent << FormatKV("level", static_cast<int64_t>(CI.level))
         << ",\n";
     out << cache_indent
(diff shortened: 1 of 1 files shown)


leanil:
@googlebot I signed it!

PR #901, opened 2019-11-05
CTest must use proper paths to executables

gjasny:
With the following syntax:

```
  add_test(NAME <name> COMMAND <command> [<arg>...])
```

if `<command>` specifies an executable target it will automatically
be replaced by the location of the executable created at build time.

This is important if a `<Configuration>_POSTFIX` like `_d` is used.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,7 +41,7 @@ build_script:
   - cmake --build . --config %configuration%
 
 test_script:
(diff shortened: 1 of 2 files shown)


LebedevRI:
Hm, indeed.

LebedevRI:
appveyor failure seems related

LebedevRI:
LG, thank you for the fix.

PR #902, opened 2019-11-18
Add method Explicit(), taking a list of Args, to be used one at a time.

jorgbrown:
Old code: BENCHMARK(BM_SortedArrayLinearSearch)->Arg(3)->Arg(7)->Arg(15)->Arg(31)->Arg(63);
New code: BENCHMARK(BM_SortedArrayLinearSearch)->Explicit({3, 7, 15, 31, 63});

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -788,6 +788,7 @@ class Benchmark {
 
   // Run this benchmark once with "x" as the extra argument passed
   // to the function.
+  // Note: Rather than call Arg() repeatedly, use Explicit() below.
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Arg(int64_t x);
 
@@ -804,9 +805,11 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* DenseRange(int64_t start, int64_t limit, int step = 1);
 
+#if defined(BENCHMARK_HAS_CXX11)
   // Run this benchmark once for every value in list.
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* Explicit(std::initializer_list<int64_t> list);
+#endif
 
   // Run this benchmark once with "args" as the extra arguments passed
   // to the function.
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -804,6 +804,10 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* DenseRange(int64_t start, int64_t limit, int step = 1);
 
+  // Run this benchmark once for every value in list.
+  // REQUIRES: The function passed to the constructor must accept an arg1.
+  Benchmark* Explicit(std::initializer_list<int64_t> list);
+
   // Run this benchmark once with "args" as the extra arguments passed
   // to the function.
   // REQUIRES: The function passed to the constructor must accept arg1, arg2 ...
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -283,6 +283,14 @@ Benchmark* Benchmark::Arg(int64_t x) {
   return this;
 }
 
+Benchmark* Benchmark::Explicit(std::initializer_list<int64_t> arglist) {
+  CHECK(ArgsCnt() == -1 || ArgsCnt() == 1);
+  for (int64_t i : arglist) {
+    args_.push_back({i});
+  }
+  return this;
+}
+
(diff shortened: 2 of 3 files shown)


test/benchmark_test.cc
```diff
@@ -176,7 +176,7 @@ static void BM_ParallelMemset(benchmark::State& state) {
     delete test_vector;
   }
 }
-BENCHMARK(BM_ParallelMemset)->Arg(10 << 20)->ThreadRange(1, 4);
+BENCHMARK(BM_ParallelMemset)->Explicit({10 << 20})->ThreadRange(1, 4);
```
LebedevRI: No, please add standalone test.
Also, you may want to mention it in docs next to `Arg()`

LebedevRI:
CI failures seem relevant.

test/benchmark_test.cc
```diff
@@ -218,6 +218,36 @@ void BM_non_template_args(benchmark::State& state, int, double) {
 }
 BENCHMARK_CAPTURE(BM_non_template_args, basic_test, 0, 0);
 
+static void BM_ExplicitlySpecifiedArgs(benchmark::State& state) {
+  // The only purpose of this benchmark is to verify that the incoming
+  // Arg values (aka state.range(0)) are the ones we expect, in the
+  // order we expect them in.
+  static int last_arg = 0;
```
LebedevRI: s/int/int64_t/

LebedevRI:
@jorgbrown ping :) this is so close to being ready

jorgbrown:
Sorry for the delay Roman; I've applied your suggested change.

I'd prefer to make the change to README.md separately, if you don't mind.

test/benchmark_test.cc
```diff
@@ -218,6 +218,36 @@ void BM_non_template_args(benchmark::State& state, int, double) {
...
+      default:
+        assert(false);  // unexpected arg
+        break;
+    }
+  }
+  for (auto _ : state) {
+    last_arg = state.range(0);
+  }
+}
+BENCHMARK(BM_ExplicitlySpecifiedArgs)->Explicit({6, 9, 4, 2});
```
LebedevRI, line 249: We don't check that the benchmark actually runs though.
But the same is true for BM_DenseThreadRanges, yes.

src/benchmark_register.cc
```diff
@@ -283,6 +283,16 @@ Benchmark* Benchmark::Arg(int64_t x) {
   return this;
 }
 
+#if defined(BENCHMARK_HAS_CXX11)
+Benchmark* Benchmark::Explicit(std::initializer_list<int64_t> arglist) {
+  CHECK(ArgsCnt() == -1 || ArgsCnt() == 1);
+  for (int64_t i : arglist) {
+    args_.push_back({i});
+  }
```
LebedevRI, line 291:
```suggestion
args_.insert(args_.end(), arglist.begin(), arglist.end());
```

include/benchmark/benchmark.h
```diff
@@ -804,6 +805,12 @@ class Benchmark {
   // REQUIRES: The function passed to the constructor must accept an arg1.
   Benchmark* DenseRange(int64_t start, int64_t limit, int step = 1);
 
+#if defined(BENCHMARK_HAS_CXX11)
+  // Run this benchmark once for every value in list.
+  // REQUIRES: The function passed to the constructor must accept an arg1.
+  Benchmark* Explicit(std::initializer_list<int64_t> list);
```
LebedevRI, line 811: Uhm, are we not crippling ourselves by using `std::initializer_list` here?
I agree that it should be ArrayRef/Span, but elsewhere here we seem to take `std::vector`.

LebedevRI:
> Sorry for the delay Roman; I've applied your suggested change.
No problem, but after taking another look at this i have some more comments.
I'm not sure about `std::initializer_list` vs `std::vector` one,
@dominichamon thoughts?

> I'd prefer to make the change to README.md separately, if you don't mind.
If that PR will follow immediately then why not,
but not separating docs from code is generally good.

include/benchmark/benchmark.h
dominichamon: yes. given the `initializer_list` requires brace initialization and a `vector` has an implicit constructor that takes an `initializer_list`, we should err on taking a `vector` here.

dominichamon:
I'd also prefer if we can keep the documentation up to date (README and any changes in the benchmark.h header comment) in this PR.

dominichamon:
@jorgbrown are you still planning on coming back to this?

LebedevRI:
reverse-ping, thanks

dominichamon:
I think this now has been implemented as `ArgsProduct` which zips vectors together but should also work with a single vector.

Issue #903, opened 2019-11-21
bug and potential fix for using long double with DoNotOptimize and gcc


KlausLeppkes:
Hi,

1.) I discovered that g++ has a problem with +m inside DoNotOptimize. 
As Jakub Jelinek stated in my gcc bugreport, +g works instead of +m (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92597).
2.) I attached the small example (test.cpp+makefile), which gives wrong values using g++[9.2] with +m.
Also, if you define -DFIX inside the makefile, gcc is working fine. Clang and Intel seem to not having this problem, but are also checked.

For reproduction, g++ is called to compile with +m:
$>make 
...
 FMA: 0; TYPE: double; ASM Volatile: 1 => y=6.9529e-305
 FMA: 1; TYPE: long double; ASM Volatile: 1 => y=-nan
make: *** [makefile:17: diff_gcc] Error 1

If you uncomment the -DFIX inside the CFLAGS, gcc produces the correct numbers.

I don't know if this fixes the problem entirely, but I wanted to let you know that there is a problem and a workaround for gcc.

Cheers,
Klaus

[googlebench_bugreport.zip](https://github.com/google/benchmark/files/3874038/googlebench_bugreport.zip)

dominichamon:
@KlausLeppkes thanks for the comprehensive bug report!

@EricWF this is your area of the code and you know it best so over to you :)

KlausLeppkes:
I search for some doc and found
"g" : Any register, memory or immediate integer operand is allowed, except for registers that are not general registers." (x86 specified!).
vs.
 "m" : A memory operand is allowed, with any kind of address that the machine supports in general.
source: https://www.ibiblio.org/gferg/ldp/GCC-Inline-Assembly-HOWTO.html

So for x86 (and only x86!), m is to restrictive and "g" should be used in my opinion.

dominichamon:
closed the PR but i'd like to understand if there's something that we need to do here after all.

MatejKafka:
GCC maintainers are pretty adamant that this is not a compiler bug, and that the `"+m,r"` is incorrect. Not that it is documented anywhere, so we're gonna have to take their word for it. :)

The issue is that the in/out operand is internally split into 2 separate operands: `__asm__ __volatile__("" : "=m,r"(value) : "m,0"(value) : "memory");`. When `DoNotOptimize` is used to obfuscate a fixed input to a benchmark, this results in `value` never being initialized:

[...]

PR #904, opened 2019-11-22
Fix for gcc with asm volatile. replaced m with g specifier

KlausLeppkes:
... to also allow registers to be covered.
Corresponding Issue: https://github.com/google/benchmark/issues/903
GCC discussion:  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92597

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,6 +35,7 @@ Joao Paulo Magalhaes <joaoppmagalhaes@gmail.com>
 Jussi Knuuttila <jussi.knuuttila@gmail.com>
 Kaito Udagawa <umireon@gmail.com>
 Kishan Kumar <kumar.kishan@outlook.com>
+Klaus Leppkes <klaus.leppkes@gmail.com>
 Lei Xu <eddyxu@gmail.com>
 Matt Clarkson <mattyclarkson@gmail.com>
 Maxim Vafin <maxvafin@gmail.com>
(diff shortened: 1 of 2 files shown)


LebedevRI:
First of all, i think you want to also file at least a bugreport to clang.
Maybe it should diagnose it, or maybe we'll learn why this isn't correct.

LebedevRI:
Also, this really looks like GCC bug, i'm not sure it should be applied for other compilers?

LebedevRI:
https://gcc.gnu.org/onlinedocs/gcc/Simple-Constraints.html
> ‘m’ A memory operand is allowed, with any kind of address that the machine supports in general. Note that the letter used for the general memory constraint can be re-defined by a back end using the TARGET_MEM_CONSTRAINT macro.
> ‘g’ Any register, memory or immediate integer operand is allowed, except for registers that are not general registers.

KlausLeppkes:
1. CI is failing with g instead of m. So you can obviously reject my merge request.
2. Nevertheless, from my understanding of the doc is the set of operand types of m a strict subset of (operand types) of g; "m" is only memory and g is more "general" (memory, gp-registers, even immediate values). 
The compiler obviously disagrees, so it's very likely that I am simply wrong. 
What am I missing here? Any enlightenment is very welcome.

dominichamon:
clang agrees with you, it's just gcc that doesn't. I'll wait for @EricWF to comment as the resident compiler expert :)

dominichamon:
i'm going to close this just to reduce our backlog of PRs. I'd also like to understand more about why this might be broken though (@EricWF )

Issue #899, opened 2019-10-31
Cache size kB binary or SI?


leanil:
Shouldn't the scaling factor [here](https://github.com/google/benchmark/blob/cf446a18bf37d5cc9b116f59cf9e6a9be89e58cc/src/sysinfo.cc#L273) be `1024` instead of `1000`?
I know SI says 1kB = 1000 byte, but I think cache sizes are usually binary multiples (at least those that I could check).

https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu
https://en.wikipedia.org/wiki/Kilobyte

dominichamon:
then it would be kiB <https://en.wikipedia.org/wiki/Kibibyte> or maybe KB.
We could support that if it makes more sense to report in kiB for cache
sizes.

LebedevRI:
`lscpu` agrees:
```
$ cat /sys/devices/system/cpu/cpu0/cache/index*/size 
16K
64K
2048K
8192K
$ lscpu | grep "cache"
L1d cache:                       64 KiB
L1i cache:                       256 KiB
L2 cache:                        8 MiB
L3 cache:                        8 MiB
```

leanil:
> We could support that if it makes more sense to report in kiB for cache sizes.

I wouldn't change the reporting (K in console, bytes in json), the question seems to be if you want to store the correct number of bytes internally.

So what's the conclusion? Can I make another 1 line PR changing that scale factor to 1024?
It seems there are no related tests.

dominichamon:
If you did want to change it to binary K, then the unit in the reporting
should also change to kiB.

LebedevRI:
This isn't just an UI issue though.
Those cache sizes are exposed via `CPUInfo` (or whats-it-called),
and they can be used by benchmarks. So this really is a correctness issue.
Right now i //think// the `*1000` is incorrect.

LebedevRI:
git blame says that was introduced in https://github.com/google/benchmark/commit/27e0b439cf0f9be329f617056795b4e5c3612f34#diff-374e7bbe55f23379d7a8ac1d378a9d2dR211 and not modified since.
@EricWF comments?

LebedevRI:
So, i've briefly chatted with @craig.topper, and while there doesn't seem to be
a definitive guaranteed answer, if we look at the raw data
```
<ctopper> lebedevri: looks like all the low level pieces are also in /sys/devices/system/cpu/cpu0/cache/index*/
[...]

PR #911, opened 2019-12-01
CPU caches are binary units, not SI.

LebedevRI:
As disscussed in https://github.com/google/benchmark/issues/899,
it is all but certain that the multiplier should be 1024, not 1000.

Fixes https://github.com/google/benchmark/issues/899

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -270,7 +270,7 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesFromKVFS() {
       else if (f && suffix != "K")
         PrintErrorAndDie("Invalid cache size format: Expected bytes ", suffix);
       else if (suffix == "K")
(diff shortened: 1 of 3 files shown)


src/sysinfo.cc
```diff
@@ -270,7 +270,7 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesFromKVFS() {
       else if (f && suffix != "K")
```
LebedevRI: I wonder if we should be proactive and also support `"M"` ?

dominichamon:
thanks for also updating the unit

dominichamon:
travis looks unrelated... something around brew changed?

LebedevRI:
Thank you for the review!

As for brew, maybe updating it to use gcc@8 will help..

PR #918, opened 2019-12-15
Travis-CI: try to update XCode/GCC versions to unbreak CI

LebedevRI:


LebedevRI:
And now OS X+GCC CI entry fails with `Failed to determine the source files for the regular expression backend`, which means none of the regex engines worked.
Not sure what's going on there.

PR #923, opened 2019-12-28
Add d postfix to Debug libraries

### LINKED ISSUE #922: Cannot install different configurations
Both the release and debug versions share the same `benchmark` and `benchmark_main` names, making it impossible to install them under the same location.

This caused overwrites when installing and later linker errors when trying to use the libraries.

pixxon:
Resolves #922, works same as googletest.


dominichamon:
failure looks unrelated. lgtm

LebedevRI:
(As noted in https://github.com/google/benchmark/issues/928, this broke pkgconfig files)

PR #926, opened 2020-01-14
Alias CMake Targets. Fixes #921

### LINKED ISSUE #921: Provide benchmark:: Aliased CMake Library
It's common to provide an aliased version of libraries that include the import prefix to make switching between ```find_package``` and ```add_subdirectory``` calls effortless. Because the ```benchmark::benchmark``` and ```benchmark::benchmark_main``` targets do not have aliased targets, users must switch between ```benchmark::benchmark``` and ```benchmark``` in ```target_link_libraries```.
Adding an aliased library with CMake is as easy as:
```cmake
add_library(benchmark::benchmark ALIAS benchmark)
```

jwillikers:
This adds two CMake alias targets `benchmark::benchmark` and `benchmark::benchmark_main` to facilitate more convenient / less error-prone linking within CMake. See issue #921  

I updated targets to link to the aliased targets internally, so as to take advantaged of the added safety CMake provides when linking against namespaced targets.

Additionally, I added a section according to the suggestion from @keith-bennett-gbg describing how to use 'find_package' and 'add_subdirectory' to incorporate the CMake project.

Fixes https://github.com/google/benchmark/issues/921

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,9 +177,9 @@ BENCHMARK(BM_StringCopy);
 BENCHMARK_MAIN();
 ```
 
-To run the benchmark, compile and link against the `benchmark` library
-(libbenchmark.a/.so). If you followed the build steps above, this
-library will be under the build directory you created.
+To run the benchmark, compile and link against the `benchmark` library (libbenchmark.a/.so).
+When using CMake, it is recommended to link against the provided `benchmark::benchmark` target.
+If you followed the build steps above, this library will be under the build directory you created.
 
 ```bash
 # Example on linux after running the build steps above. Assumes the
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -194,6 +193,21 @@ Alternatively, link against the `benchmark_main` library and remove
...
+```
+Alternatively, ```add_subdirectory``` will incorporate the library directly in to one's CMake project.
+```cmake
+add_subdirectory(benchmark)
+```
+Either way, link to the library as follows.
+```cmake
+target_link_libraries(MyTarget benchmark::benchmark)
+```
+
```
LebedevRI: 1. Newline after the header
2. I think some docs lines aren't properly linewrapped

keith-bennett-gbg:
:+1: builds on my machine
:tada: README addition

README.md
jwillikers: @LebedevRI I have added a newline after the header, ensured the lines I modified wrapped to a maximum length of 80 characters, and also moved the *Usage with CMake* up one header-level to make it a direct descendant of the *Usage* section.

jwillikers:
Your welcome. Thanks for the review 😁

PR #938, opened 2020-02-14
Relax CHECK condition in benchmark_runner.cc

### LINKED ISSUE #937: Assert failure because of SkipWithError()
If `SkipWithError()` is used before the benchmark loop, and we also early exit before the benchmark loop starts the assert failure is triggered:

```
src/benchmark_runner.cc:120: RunInThread: Check `st.iterations() >= st.max_iterations' failed. Benchmark returned before State::KeepRunning() returned false!
```

at https://github.com/google/benchmark/blob/master/src/benchmark_runner.cc#L120.

The example unit test:
```cpp
void BM_error_no_running(benchmark::State& state) {
  state.SkipWithError("error message");
}
BENCHMARK(BM_error_no_running);
ADD_CASES("BM_error_before_running", {{"", true, "error message"}});
```

I was thinking about fixing the issue by modifying the assert condition allow it pass if there was a failure in the benchmark. But there is no API to read this information from a `State` object.

chfast:
If the benchmark state contains an error, do not expect any iterations has been run.
This allows using SkipWithError() and return early from the benchmark function.

Fixes #937.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1184,7 +1184,9 @@ Users must explicitly exit the loop, otherwise all iterations will be performed.
 Users may explicitly return to exit the benchmark immediately.
 
 The `SkipWithError(...)` function may be used at any point within the benchmark,
-including before and after the benchmark loop.
+including before and after the benchmark loop. Moreover, if `SkipWithError(...)`
+has been used, it is not required to reach the benchmark loop and one may return
+from the benchmark function early.
 
 For example:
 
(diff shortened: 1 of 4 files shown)


LebedevRI:
@chfast ping :)

chfast:
> And a negative test, if there isn't one already (if there is one, point me to it?),

I was not able to implement this. Are there any examples of tests expecting `CHECK()` macro failures?

LebedevRI:
> > And a negative test, if there isn't one already (if there is one, point me to it?),
> 
> I was not able to implement this. Are there any examples of tests expecting `CHECK()` macro failures?

Hmm. I was thinking of using googletest's death tests, but since there clearly isn't a test
that would test the current assertion, adding one would be cumbersome given how
it would need to be called.

So lets not, lets just add the docs :)

PR #940, opened 2020-02-18
Use `benchmark::IterationCount` rather than `int64_t` in lambda to co…

NAThompson:
…mplexity


LebedevRI:
Right, i guess missed that.
Thank you!

Issue #769, opened 2019-02-17
Timer/counter adjustment by the ->Threads() could use a verification


LebedevRI:
This could be nothing, but the adjustment
https://github.com/google/benchmark/blob/99d1356c041f49dccfc55104e79e282f815f6849/src/benchmark_runner.cc#L223-L225
and
https://github.com/google/benchmark/blob/99d1356c041f49dccfc55104e79e282f815f6849/src/reporter.cc#L83-L93
looks weird. I'm not sure if that is fully correct?
Don't we end up double-correcting the results, since we don't divide the iteration count by thread num,
and thus we basically divide by thread num twice?
E.g.:
```
$ ./test/internal_threading_test 
Running benchmarks...
-------------------------------
Testing ConsoleReporter Output
-------------------------------
2019-02-17 13:54:25
Running ./test/internal_threading_test
Run on (8 X 4000 MHz CPU s)
CPU Caches:
  L1 Data 16K (x8)
  L1 Instruction 64K (x4)
  L2 Unified 2048K (x4)
  L3 Unified 8192K (x1)
Load Average: 0.13, 0.23, 0.30
***WARNING*** Library was built as DEBUG. Timings may be affected.
---------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                             Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------------------------------------------------------
BM_MainThread/iterations:1/threads:1                                           50009433 ns     50007079 ns            1 invtime=19.9972/s
BM_MainThread/iterations:1/real_time/threads:1                                 50003061 ns     50003352 ns            1 invtime=19.9988/s
BM_MainThread/iterations:1/manual_time/threads:1                               50000000 ns     49979580 ns            1 invtime=20/s
BM_MainThread/iterations:1/process_time/threads:1                              50003885 ns     49976581 ns            1 invtime=20.0094/s
BM_MainThread/iterations:1/process_time/real_time/threads:1                    50002967 ns     49990535 ns            1 invtime=19.9988/s
BM_MainThread/iterations:1/process_time/manual_time/threads:1                  50000000 ns     50003141 ns            1 invtime=20/s
BM_MainThread/iterations:1/threads:2                                           25002051 ns     50003801 ns            2 invtime=19.9985/s
BM_MainThread/iterations:1/real_time/threads:2                                 25001916 ns     50003783 ns            2 invtime=39.9969/s
[...]

PR #946, opened 2020-03-15
Fix reported time for multithreaded benchmarks

koct9i:
In report real and manual per iteration is divided by thread count
because total time is adjusted to per-thread while iterations counter is not.
See #769 

Process CPU time is special - each thread reports time spent by all threads
thus it actually must be divided by thread count once to get total cpu time and
second time to get per-thread CPU time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -910,9 +910,12 @@ Without `UseRealTime`, CPU time is used by default.
 ### CPU Timers
 
 By default, the CPU timer only measures the time spent by the main thread.
+For multithreaded benchmarks timer measures average among working threads.
+
 If the benchmark itself uses threads internally, this measurement may not
 be what you are looking for. Instead, there is a way to measure the total
-CPU usage of the process, by all the threads.
+CPU usage of the process, by all the threads. Resulting total CPU time
+also will be divided among working threads.
 
 ```c++
 void callee(int i);
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -82,6 +82,7 @@ BenchmarkReporter::Run CreateRunReport(
   report.repetitions = b.repetitions;
 
   if (!report.error_occurred) {
+    // This is the total time across all threads.
     if (b.use_manual_time) {
       report.real_accumulated_time = results.manual_time_used;
     } else {
@@ -227,10 +228,8 @@ class BenchmarkRunner {
     // And get rid of the manager.
     manager.reset();
 
-    // Adjust real/manual time stats since they were reported per thread.
-    i.results.real_time_used /= b.threads;
-    i.results.manual_time_used /= b.threads;
-    // If we were measuring whole-process CPU usage, adjust the CPU time too.
+    // If we were measuring whole-process CPU usage then each thread reports
+    // total CPU time of all threads. Divide by threads to get real value.
     if (b.measure_process_cpu_time) i.results.cpu_time_used /= b.threads;
 
     VLOG(2) << "Ran in " << i.results.cpu_time_used << "/"
@@ -245,6 +244,9 @@ class BenchmarkRunner {
     } else if (b.use_real_time) {
       i.seconds = i.results.real_time_used;
     }
+
+    // Adjust time stats to average since they were reported by all threads.
+    i.seconds /= b.threads;
 
     return i;
   }
--- a/AUTHORS
+++ b/AUTHORS
@@ -51,6 +51,7 @@ Sayan Bhattacharjee <aero.sayan@gmail.com>
 Shuo Chen <chenshuo@chenshuo.com>
 Steinar H. Gunderson <sgunderson@bigfoot.com>
 Stripe, Inc.
+Yandex LLC
 Yixuan Qiu <yixuanq@gmail.com>
 Yusuke Suzuki <utatane.tea@gmail.com>
 Zbigniew Skowron <zbychs@gmail.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -55,6 +55,7 @@ Jussi Knuuttila <jussi.knuuttila@gmail.com>
 Kai Wolf <kai.wolf@gmail.com>
 Kaito Udagawa <umireon@gmail.com>
 Kishan Kumar <kumar.kishan@outlook.com>
+Konstantin Khlebnikov <khlebnikov@yandex-team.ru>
 Lei Xu <eddyxu@gmail.com>
 Matt Clarkson <mattyclarkson@gmail.com>
 Maxim Vafin <maxvafin@gmail.com>
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1381,11 +1381,15 @@ class BenchmarkReporter {
     bool error_occurred;
     std::string error_message;
 
+    // Total iterations across all threads.
     IterationCount iterations;
+
     int64_t threads;
     int64_t repetition_index;
     int64_t repetitions;
     TimeUnit time_unit;
+
+    // Total time across all threads.
     double real_accumulated_time;
     double cpu_accumulated_time;
 
--- a/test/internal_threading_test.cc
+++ b/test/internal_threading_test.cc
@@ -4,20 +4,27 @@
 #include <chrono>
 #include <thread>
 #include "../src/timers.h"
+#include "../src/check.h"
 #include "benchmark/benchmark.h"
-#include "output_test.h"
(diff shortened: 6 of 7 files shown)


koct9i:
@googlebot I signed it!

LebedevRI:
Thank you for looking into this. With multi-threaded benchmarks, It is hard to tell whether
the currently-reported numbers are correct or not, as noted in mentioned #769.

koct9i:
Bug in current reports is pretty obvious for trivial benchmark which spins or sleep fixed amount of time.
It seems present test/internal_threading_test.cc could be used for that. But it needs validation.
But writing checks for it using "output_test" is a pain =) I'll play with custom reported, this should easier.

src/benchmark_runner.cc
```diff
@@ -227,10 +228,8 @@ class BenchmarkRunner {
     // And get rid of the manager.
     manager.reset();
 
-    // Adjust real/manual time stats since they were reported per thread.
-    i.results.real_time_used /= b.threads;
-    i.results.manual_time_used /= b.threads;
-    // If we were measuring whole-process CPU usage, adjust the CPU time too.
+    // If we were measuring whole-process CPU usage then each thread reports
+    // total CPU time of all threads. Divide by threads to get real value.
     if (b.measure_process_cpu_time) i.results.cpu_time_used /= b.threads;
```
LebedevRI: Shouldn't this also go away, don't we now divide it twice?

koct9i: Nope. Each thread reports total process time individually and they are summed i.e. multiplied by thread count.

koct9i:
Oh, I get it - spin-wait in test watching into real time. :man_facepalming:

koct9i:
Bump. Ready for review/merge.

README.md
```diff
@@ -910,9 +910,12 @@ Without `UseRealTime`, CPU time is used by default.
 ### CPU Timers
 
 By default, the CPU timer only measures the time spent by the main thread.
+For multithreaded benchmarks timer measures average among working threads.
+
 If the benchmark itself uses threads internally, this measurement may not
 be what you are looking for. Instead, there is a way to measure the total
-CPU usage of the process, by all the threads.
+CPU usage of the process, by all the threads. Resulting total CPU time
+also will be divided among working threads.
```
LebedevRI, line 918: I wonder if this is too vague. Maybe this should mention that it is specifically about https://github.com/google/benchmark#multithreaded-benchmarks, not the threads inside of the workload itself.

dominichamon:
the code seems reasonable. The threads division has changed back and forth a few times (as i mentioned in #769).

i'm just running a diff between the head version and this change to see what the output looks like for tests.

dominichamon:
This is the before

```
----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
BM_spin_empty/8/threads:4                           23.0 ns         90.2 ns      7967324
BM_spin_empty/512/threads:4                         1367 ns         4908 ns       158756
BM_spin_empty/8192/threads:4                       22378 ns        79529 ns         9364
```

This is the after
```
----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
BM_spin_empty/8/threads:4                            168 ns         95.0 ns     33131036
BM_spin_empty/512/threads:4                         9492 ns         4960 ns       512840
BM_spin_empty/8192/threads:4                      155608 ns        80357 ns        31952
```

Before this change, it's clear that the time is the end-to-end time given the parallelisation (23 ~= 90/4, 1300 ~= 4900/4, etc). In the latter case i find it hard to understand why we're ~doubling the reported time.

another example:

before
```
BM_pause_during                                      804 ns          796 ns       798884
BM_pause_during/threads:4                            355 ns         1221 ns       568668
BM_pause_during/real_time                            972 ns          970 ns       848250
BM_pause_during/real_time/threads:4                  361 ns         1250 ns      1890216
```

after
```
BM_pause_during                                     1248 ns         1188 ns       594011
BM_pause_during/threads:4                           2873 ns         1297 ns      2099924
BM_pause_during/real_time                           1326 ns         1196 ns       552044
BM_pause_during/real_time/threads:4                 2357 ns         1300 ns       975240
```

again i find it hard to understand why the threading appears to cause the measured time to double. the former is what i'd expect: ~4x improvement by using 4 threads.

koct9i:
My patch normalize all times to 1 iteration in 1 thread. So, if threads runs all time then "real" time should be equal to 'CPU" time or be slightly more. For perfect parallelism times should stay the same regardless of count of threads. 

In your case 'real' time much bigger then 'cpu', this means threads were preempted. I guess you are using osx? Probably these artifacts were caused by it's scheduler.

For Linux:

Before
```
BM_spin_empty/8                                     43.2 ns         43.2 ns     16180998
BM_spin_empty/512                                   2448 ns         2448 ns       285791
BM_spin_empty/8192                                 39142 ns        39138 ns        17843
BM_spin_empty/8/threads:8                           11.8 ns         94.0 ns      7782976
BM_spin_empty/512/threads:8                          708 ns         5638 ns       125112
BM_spin_empty/8192/threads:8                       11666 ns        92589 ns         7576
```

After
```
BM_spin_empty/8                                     46.8 ns         46.8 ns     14851377
BM_spin_empty/512                                   2638 ns         2638 ns       263970
BM_spin_empty/8192                                 42235 ns        42222 ns        16626
BM_spin_empty/8/threads:8                            109 ns          109 ns     54453880
BM_spin_empty/512/threads:8                         6212 ns         6201 ns       911128
BM_spin_empty/8192/threads:8                      101850 ns       101688 ns        55528
```

As you can see in 'Before' 'real' time was erroneously  divided by 8.

Performance for per-cpu threads is around twice lower because this is 4 core cpu with hyper-threading and turbo-boost cannot reach top frequency when all cores are active.

dominichamon:
I think this change is correct. The wall-time (Time) and CPU time should match in terms of per-thread vs global. The benefit of threading should be visible when the real and CPU times don't change but the amount of processing (items per second, etc) increases. As noted, if threading _does_ lead to a change in timing it indicates where the parallelism breaks down.

krzikalla:
Since it is related, I want to add another problem to this issue: the real time currently measures the average wall-clock time  of every single thread. It should rather measures the wall-clock time of the thread pack. That is, the actual real time is the time, when the last thread enters FinishKeepRunning. Only by that means you can spot balancing issues.
Example: if one thread runs 1 sec and 7 threads run 1 ms each, then currently 0,125 ms real time is reported. This is not the wall-clock time! The wall-clock time is 1 sec.
Solution: one could move the barrier in FinishKeepRunning before PauseTiming. But this would tamper the timing. Instead don't sum up the real times, but use the maximum one at https://github.com/google/benchmark/blob/1302d2ce094a9753b0f81a81ea74c0fa71fae582/src/benchmark_runner.cc#L127. Of course you must not divide real time by num_threads then.

dominichamon:
@krzikalla would it be ok to defer the wall-clock time change to a separate PR or do you want to solve that as part of this change?

koct9i:
@krzikalla 
> Example: if one thread runs 1 sec and 7 threads run 1 ms each, then currently 0,125 sec real time is reported. This is not the wall-clock time! The wall-clock time is 1 sec.

This seems odd. If each thread did same single iteration then this means execution speed is unpredictable.
Instead of forging "sane" metrics benchmark should measure and report deviation among iterations.

Present design of benchmark loop could easily estimate deviation among averages for each thread.
To catch variations within thread it have to measure time periodically or compare progress.

krzikalla:
> would it be ok to defer the wall-clock time change to a separate PR or do you want to solve that as part of this change?

It can be a separate PR. I just wanted to raise awareness for this issue and thought, it would be better to add it to this discussion here instead of creating a new one.

> This seems odd. If each thread did same single iteration then this means execution speed is unpredictable.

Not necessarily. Other possible scenarios are poor balancing or a hit by Amdahl's law.

> Instead of forging "sane" metrics benchmark should measure and report deviation among iterations.
> 
> Present design of benchmark loop could easily estimate deviation among averages for each thread.
> To catch variations within thread it have to measure time periodically or compare progress.

Normally, you don't report deviations. You report wall clock time, since that affect the usability of the code.

dominichamon:
I'm returning to this, sorry for the break. Do you think there's more to do here or are you happy with it?

LebedevRI:
This was fixed at the source in https://github.com/google/benchmark/pull/1836.

Issue #928, opened 2020-01-21
pkg-config reports incorrect library with Debug build: gcc/Linux


ambitslix:
pkg-config --libs reports benchmark and benchmark_main when in reality Debug builds end up as "libbenchmarkd.a" &  "libbenchmark_maind.a". Solution is to configure benchmark.pc according to build type or just have the same name. Probably different build types should reside in different locations to avoid name collisions.

LebedevRI:
Hmm, interesting. I guess googletests pkgconfig files are also broken in that configuration.
In principle i guess this may be solvable by asking cmake "what is the filename of the library",
and using the variable you stored that answer in in `.pc.in` instead of hardcoded value.
If you'd like to take a look, i could review i suppose.

ambitslix:
Hello, yes  common problem encountered with other libraries as well. I can take a look at it on the weekend, as you said I would configure .pc according to actual file names from cmake.

LebedevRI:
@dominichamon to make progress here we mainly need to decide *what* is the solution
* do nothing
* revert original problematic pr
* drop pkg-config support
* merge this as-is. i mean, this does solve the problem.. (best course of action?)
* bump cmake to 3.15, and fix this properly (i advise against this.)
* ???

this is more of an executive decision, not for me to make it here.

ambitslix:
Drop pkg-config support is my number one vote, then second would be merge this fix.

dominichamon:
revert the original PR i think. it solved a relatively niche problem but has clearly opened up more issues so needs more thought.

PR #947, opened 2020-03-16
Revert "Add d postfix to Debug libraries"

LebedevRI:
This reverts commit 5ce2429af7a8481581896afaa480552cc7584808.

Reverts https://github.com/google/benchmark/pull/923
Reopens https://github.com/google/benchmark/issues/922
Fixes https://github.com/google/benchmark/issues/928
Closes https://github.com/google/benchmark/pull/930

See discussion in https://github.com/google/benchmark/issues/928
this broke pkg-config support, since there we don't account
for the suffix, nor is it trivial to do so.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -23,7 +23,6 @@ set_target_properties(benchmark PROPERTIES
   OUTPUT_NAME "benchmark"
   VERSION ${GENERIC_LIB_VERSION}
   SOVERSION ${GENERIC_LIB_SOVERSION}
-  DEBUG_POSTFIX "d"
 )
 target_include_directories(benchmark PUBLIC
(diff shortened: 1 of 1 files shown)


AUTHORS
```diff
@@ -26,7 +26,6 @@ Eugene Zhuk <eugene.zhuk@gmail.com>
 Evgeny Safronov <division494@gmail.com>
 Federico Ficarelli <federico.ficarelli@gmail.com>
 Felix Homann <linuxaudio@showlabor.de>
-Gergő Szitár <szitar.gergo@gmail.com>
```
dominichamon: you can leave this in, i think. we should still credit the original contribution.

LebedevRI: Oh, this isn't intentional, i only used github builtin "revert" feature.
Let be fix this up..

PR #955, opened 2020-04-08
Fix cycleclock::Now for RISC-V and PPC

luismarques:
Fixes the following issues with the implementation of `cycleclock::Now`:

- The RISC-V implementation wouldn't compile due to a typo;

- Both the PPC and RISC-V implementation's asm statements lacked the
  volatile keyword. This resulted in the repeated read of the counter's
  high part being optimized away, so overflow wasn't handled at all.
  Multiple counter reads could also be misoptimized, especially in LTO
  scenarios.

- Relied on the zero/sign-extension of inline asm operands, which isn't
  guaranteed to occur and differs between compilers, namely GCC and Clang.

The PowerPC64 implementation was improved to do a single 64-bit read of
the time-base counter.

The RISC-V implementation was improved to do the overflow handing in
assembly, since Clang would generate a branch, defeating the purpose
of the non-branching counter reading approach.

I dug into the history of the project and the API usage, and it's not quite 
clear that branching would actually be problematic. For instance, the gperftools 
repo inherited the same code and eventually changed the PPC implementation to a 
branching one (before dropping the code completely). Still, for the considered 
use case the chosen approach seems sensible (if more complex and error-prone to 
implement), so I have kept non-branching implementations.

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -84,13 +84,21 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   return (high << 32) | low;
 #elif defined(__powerpc__) || defined(__ppc__)
   // This returns a time-base, which is not always precisely a cycle-count.
-  int64_t tbl, tbu0, tbu1;
-  asm("mftbu %0" : "=r"(tbu0));
-  asm("mftb  %0" : "=r"(tbl));
-  asm("mftbu %0" : "=r"(tbu1));
-  tbl &= -static_cast<int64_t>(tbu0 == tbu1);
-  // high 32 bits in tbu1; low 32 bits in tbl  (tbu0 is garbage)
-  return (tbu1 << 32) | tbl;
+#if defined(__powerpc64__) || defined(__ppc64__)
+  int64_t tb;
+  asm volatile("mfspr %0, 268" : "=r" (tb));
+  return tb;
+#else
+  uint32_t tbl, tbu0, tbu1;
+  asm volatile(
+    "mftbu %0\n"
+    "mftbl %1\n"
+    "mftbu %2"
+    : "=r"(tbu0), "=r"(tbl), "=r"(tbu1));
+  tbl &= -static_cast<int32_t>(tbu0 == tbu1);
+  // high 32 bits in tbu1; low 32 bits in tbl  (tbu0 is no longer needed)
+  return (static_cast<uint64_t>(tbu1) << 32) | tbl;
+#endif
 #elif defined(__sparc__)
   int64_t tick;
   asm(".byte 0x83, 0x41, 0x00, 0x00");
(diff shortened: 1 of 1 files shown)


LebedevRI:
@lenary @asb thoughts on RISC-V changes?

src/cycleclock.h
```diff
@@ -167,16 +175,22 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
...
-  uint64_t cycles_low, cycles_hi0, cycles_hi1;
-  asm("rdcycleh %0" : "=r"(cycles_hi0));
-  asm("rdcycle %0" : "=r"(cycles_lo));
-  asm("rdcycleh %0" : "=r"(cycles_hi1));
-  // This matches the PowerPC overflow detection, above
-  cycles_lo &= -static_cast<int64_t>(cycles_hi0 == cycles_hi1);
-  return (cycles_hi1 << 32) | cycles_lo;
+  uint32_t cycles_lo, cycles_hi0, cycles_hi1;
+  // This asm also includes the PowerPC overflow handling strategy, as above.
+  // Implemented in assembly because Clang insisted on branching.
```
LebedevRI, line 180: Can you show the snippet that results in branching?
I see no branching here: https://godbolt.org/z/GNcPy7

LebedevRI, line 180: I really wonder if this is clang risc-v backend being too aggressive in lowering that into branches.

src/cycleclock.h
```diff
@@ -84,13 +84,21 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
...
+  return tb;
+#else
+  uint32_t tbl, tbu0, tbu1;
+  asm volatile(
+    "mftbu %0\n"
+    "mftbl %1\n"
+    "mftbu %2"
+    : "=r"(tbu0), "=r"(tbl), "=r"(tbu1));
+  tbl &= -static_cast<int32_t>(tbu0 == tbu1);
   // high 32 bits in tbu1; low 32 bits in tbl  (tbu0 is garbage)
```
LebedevRI, line 99: tbu0 isn't *really* garbage though, it served it's purpose of overflow checking

dominichamon:
This seems overall decent to me, though I wouldn't mind seeing some ppc/risc-v benchmark output before and after the change. i don't have access to either platform. maybe you do @luismarques ?

src/cycleclock.h
luismarques: I think here the original writer meant `tbu0 is now garbage`, and you're interpreting it more in the sense of always being garbage, which indeed would be wrong. Would adding the "now" address this concern sufficiently?

src/cycleclock.h
luismarques: It really depends on your scheduling model, assumptions on which branch is more likely, etc. But I guess from the point of view of the compiler it's like "what do you mean I shouldn't branch? I didn't write anything anywhere to even suggest that I shouldn't do that!" :)
Curiously, it's actually more typical to be the other way around, since LLVM's IR more natively relies on the notion of conditional moves, while GCC seems to be more branch oriented.

lenary:
@luismarques showed me why my contribution was both wrong (variable naming), and bad (GCC and clang can eliminate some of the assembly if you don't have it in a single ASM block), leading to miscompiles, as he has explained.

I haven't had a moment to check or benchmark this on a risc-v system, and don't have powerpc systems to check this on.

I agree there are almost certainly improvements needing to be made to the risc-v LLVM backend around tuning sequences like this to avoid branches so we can avoid the inline assembly here, but reasonably this change improves things for versions of clang that have already been released, and also for RISC-V gcc.

With the caveat that I haven't managed to compile and test this, I'm happy for it to be committed.

luismarques:
> @luismarques showed me why my contribution was both wrong (variable naming), and bad (GCC and clang can eliminate some of the assembly if you don't have it in a single ASM block), leading to miscompiles, as he has explained.

Well, no need to be so negative about your contribution :)
You based your implementation on the PPC one, so it was fairly reasonable to assume that the things you hadn't changed were OK to rely on. The broader issue is that the software ecosystem was still bootstrapping, so it was difficult to test the changes.

@dominichamon 
> This seems overall decent to me, though I wouldn't mind seeing some ppc/risc-v benchmark output before and after the change. i don't have access to either platform. maybe you do @luismarques ?

I have two QEMU VMs, 32- and a 64-bit RISC-V ones. But they are quite finicky, especially the 32-bit one. Just compiling software and their dependencies without running out of memory on the RV32 VM is extremely difficult. This code is going to be vended into the LLVM copy, so that might provide some testing. But my plan was to get this approved upstream before porting the changes to LLVM. Perhaps the manual inspection might suffice for now, unless somebody can test the PPC changes?

src/cycleclock.h
LebedevRI: `tbu0 is no longer needed` i think

src/cycleclock.h
LebedevRI: Yeah, it is indeed a balance between "what's the branch misprediction cost" vs "is this branch predictable" vs "is this actually perf-sensitive" vs "how much more instrs are we willing to do to avoid the branch"
The other thing is that as far as i understand for RISC-V, conditional move instruction
will only be available in `B`it manipulation ISA, and i guess currently such `select`
is simply expanded into a branch instead of some branch-less and-of-sub sequence.

luismarques: Yup, that's exactly right. In fact, one of the first things I worked on in LLVM's RISC-V backend was improving the code generation when you `select`ed a value that had to be moved in two steps, as LLVM would mangle that up into a less than ideal branch lowering mess.

src/cycleclock.h
luismarques: Oh yeah, that makes sense. I was confusing it with the RV implementation, where `%0` is overwritten with the mask.

asb:
I've not tested this, but at least by inspection it seems correct. LGTM.

dominichamon:
Thank you all!

PR #956, opened 2020-04-10
Remove warnings for internal use of CSVReporter

KeithMoyer:
In a previous commit[1], diagnostic pragmas were used to avoid this
warning. However, the incorrect warning flag was indicated, leaving the
warning in place. -Wdeprecated is for deprecated features while
-Wdeprecated-declarations for deprecated functions, variables, and
types[2].

Tested via `bazel build //:benchmark` with (no warning) and without (got warning) this change.

[1] c408461983dd3adf49d450d7db926fc46f1d99a0
[2] https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html


dominichamon:
Thank you!

Issue #954, opened 2020-04-08
[FR] Python bindings


chr1sj0nes:
We have created some simple Python bindings for google/benchmark that we would like to contribute.

The intent is to allow the following usage:
```python
import benchmark

@benchmark.register       # Equivalent to `BENCHMARK(f)` in C++.
def bench_index(state):   # state is a shim for `benchmark::State`.
  l = list(range(10000))  # Setup code is not benchmarked (as with C++).
  while state:            # Equivalent to `for (auto _ : state)` in C++.
    l.index(5000)         # Like C++ only code inside the loop is timed.

if __name__ == '__main__':
  benchmark.main()
```

It doesn't require much to get a simple example like this working. Namely:
1. Have a (private) pybind11 extension to expose a wrapper around:
    - `benchmark::Initialize`
    - `benchmark::RegisterBenchmark`
    - `benchmark::RunSpecifiedBenchmarks`
    - `benchmark::State` (a shim only supporting `KeepRunning` renamed to `__bool__`).
2. Create a Python API exposing:
    - `register` (calls RegisterBenchmark)
    - `main` (calls Initialize, RunSpecifiedBenchmarks)

chr1sj0nes:
Where would be the best place to put such bindings? `//bindings/python/`?

dominichamon:
Great!

I think bindings/python makes sense (or bindings/py if you want to be hip and groovy). This also plays nicely with #776 which is offering fortran/rust bindings.

dominichamon:
Should have resolved this when #957 was merged :)

PR #957, opened 2020-04-14
Add Python bindings.

chr1sj0nes:
#954


LebedevRI:
This is missing licence header blobs and cmake stuff.

dominichamon:
Great! I have two requests:

1. Can we turn the example into a py_test? I know it doesn't actually test anything, but if we can then run it on travis we get some level of protection against someone introducing a bug that affects the bindings.

2. is there any documentation we can add (to https://github.com/google/benchmark/tree/master/docs, linked from the main README.md)?

setup.py
```diff
@@ -0,0 +1,124 @@
...
+    version=_get_version(),
+    url='https://github.com/google/benchmark',
+    description='A library to benchmark code snippets.',
+    author='Google',
+    author_email='benchmark-py@google.com',
+    # Contained modules and scripts.
+    package_dir={'': 'bindings/python'},
+    packages=setuptools.find_packages('bindings/python'),
+    install_requires=_parse_requirements('bindings/python/requirements.txt'),
+    cmdclass=dict(build_ext=BuildBazelExtension),
```
LebedevRI, line 108: Hmm.
Reading up on this, and i think either the cmake build system needs to be explicitly deprecated,
or this should be cmake-based.

dominichamon: We could have the Python bindings requiring bazel (given they require absl which doesn't have a good cmake solution). No?

chr1sj0nes:
> This is missing licence header blobs and cmake stuff.

Ah... I checked and saw that many files didn't have the copyright / license header, but I've just seen that the public headers do. Which files require it? The published files (the Python in this case), or all of them?

What do you mean by CMake stuff? Is is not okay to rely on Bazel for building the Python extension?

chr1sj0nes:
> This is missing licence header blobs and cmake stuff.

I've added the license headers to the two Python files that will be distributed. Are they needed anywhere else?

dominichamon:
If someone wants the python bindings,  they need to use blaze, right? i think that's ok for now and if someone needs cmake we/they can always add that.

chr1sj0nes:
> If someone wants the python bindings, they need to use bazel, right? i think that's ok for now and if someone needs cmake we/they can always add that.

Yes, that's correct

LebedevRI:
On Wed, Apr 15, 2020 at 2:30 PM Chris Jones <notifications@github.com>
wrote:

> This is missing licence header blobs and cmake stuff.
>
> Ah... I checked and saw that many files didn't have the copyright /
> license header, but I've just seen that the public headers do. Which files
> require it? The published files (the Python in this case), or all of them?
>
Every file, IMO

> What do you mean by CMake stuff? Is is not okay to rely on Bazel for
> building the Python extension?
>
I guess i'll just refrain from the review then :)

chr1sj0nes:
Hi Dominic. Apologies for my procrastination on this. I've changed the example to be a `py_test`. Is there anything else that needs fixing before this can go in?

dominichamon:
i think it's great. maybe a markdown doc linked from the main README to complete the documentation story would be ideal if you can, but if you want to do that as a followup let me know and i'll land this.

chr1sj0nes:
It would be great to get this in. If I can do the documentation in a follow-up, that would be preferable to me.

PR #960, opened 2020-04-23
Add missing <cerrno> header

NewProggie:
This commit fixes a current build error on Android where 'errno' is an unidentified
symbol due to a missing header

### DIFF
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -1,6 +1,7 @@
 #include "string_util.h"
 
 #include <array>
+#include <cerrno>
 #include <cmath>
 #include <cstdarg>
 #include <cstdio>
--- a/src/string_util.cc
+++ b/src/string_util.cc
@@ -1,7 +1,9 @@
 #include "string_util.h"
 
 #include <array>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Makes sense, but i wonder if it should be guarded under `BENCHMARK_STL_ANDROID_GNUSTL` macro.

NewProggie:
> Makes sense, but i wonder if it should be guarded under `BENCHMARK_STL_ANDROID_GNUSTL` macro.

It just so happened that I noticed this while doing a cross compilation for Android. However, this header should be added in any case as you're using a symbol from this header file anyways, I think?

LebedevRI:
Sure, but as far as i can tell, currently `errno` is only used in that file if `BENCHMARK_STL_ANDROID_GNUSTL` set.

NewProggie:
> Sure, but as far as i can tell, currently `errno` is only used in that file if `BENCHMARK_STL_ANDROID_GNUSTL` set.

Sorry, didn't notice this before. Thank you, I will update my PR

PR #966, opened 2020-05-22
Drop unused mingw.py

cclauss:
Avoid Syntax Warning on Python 3.8: `3 is 3`  # —> Syntax Warning


LebedevRI:
Is that script even used?
`$ grep -r mingw.py` suggests it's not.

dominichamon:
Added in https://github.com/google/benchmark/commit/60910b3de7aa85ff28da14f34c7084711f99ac74. Idea was likely to use it in appveyor but it never was.

LebedevRI:
Should we just drop that script?

dominichamon:
i think so? unless it is useful for building an appveyor matrix.

LebedevRI:
Thank you for bringing this up!

PR #975, opened 2020-06-08
Add some simple docs for installing the Python bindings from source.

chr1sj0nes:


bindings/python/benchmark/example.py
```diff
@@ -11,7 +11,14 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Example of Python using C++ benchmark framework."""
+"""Example of Python using C++ benchmark framework.
+
+To run this example, you must first install the `benchmark` Python package.
+
+To install using `setup.py`, download and extract the `benchmark` source.
+In the extracted directory, execute:
+  python setup.py install
```
dominichamon, line 20: this happens as part of `bazel test //...`. is there a way to make the bazel test work without this installation?

chr1sj0nes: I'm not sure, I'm afraid

dominichamon:
FTR, I needed to use `python3 setup.py install` but i think that's my local environment.

also, it still doesn't pass the test. I think i'd like to get the //bindings/... test passing before committing this in case there's more to the docs.

here's what i have:

```
$ echo $PYTHONPATH
/usr/local/buildtools/current/sitecustomize:/home/dominic/python
$ ls ~/python/
absl_py-0.9.0-py3.7.egg  easy-install.pth  google_benchmark-0.1.0-py3.7-linux-x86_64.egg  __pycache__  site.py
$ ls ~/python/google_benchmark-0.1.0-py3.7-linux-x86_64.egg/
benchmark  EGG-INFO
$ bazel test //bindings/...
INFO: Analyzed 5 targets (0 packages loaded, 0 targets configured).
INFO: Found 4 targets and 1 test target...
FAIL: //bindings/python/benchmark:example (see /home/dominic/.cache/bazel/_bazel_dominic/20306993ad39534bfa0a5ec8520b7ad0/execroot/com_github_google_benchmark/bazel-out/k8-fastbuild/testlogs/bindings/python/benchmark/example/test.log)
INFO: Elapsed time: 0.284s, Critical Path: 0.15s
INFO: 2 processes: 2 linux-sandbox.
INFO: Build completed, 1 test FAILED, 2 total actions
//bindings/python/benchmark:example                                      FAILED in 0.1s
  /home/dominic/.cache/bazel/_bazel_dominic/20306993ad39534bfa0a5ec8520b7ad0/execroot/com_github_google_benchmark/bazel-out/k8-fastbuild/testlogs/bindings/python/benchmark/example/test.log

INFO: Build completed, 1 test FAILED, 2 total actions
$ cat bazel-testlogs/bindings/python/benchmark/example/test.log 
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //bindings/python/benchmark:example
-----------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/dominic/.cache/bazel/_bazel_dominic/20306993ad39534bfa0a5ec8520b7ad0/sandbox/linux-sandbox/178/execroot/com_github_google_benchmark/bazel-out/k8-fastbuild/bin/bindings/python/benchmark/example.runfiles/com_github_google_benchmark/bindings/python/benchmark/example.py", line 16, in <module>
    import benchmark
ModuleNotFoundError: No module named 'benchmark'
```

is the installation to `google_benchmark` the issue?

chr1sj0nes:
> is the installation to `google_benchmark` the issue?

No, that's fine.

This works for me - the only difference is that I used a virtual Python environment. I'm sure how bazel decides which interpreter to use. I did the following:
```
python3 -m venv test-env
source test-env/bin/activate
python setup.py install
```

dominichamon:
within the venv i have python and python3 both reporting the same version (good). however, i get the same result from running bazel test (bad).

running the script directly (`python bindings/python/benchmark/example.py`) works fine, as does running the script in `bazel-bin`. so bazel itself seems to be running with a different environment.

I found https://github.com/bazelbuild/bazel/issues/7653 which looked related but setting legacy_create_init didn't help.

Which version of bazel are you using?

chr1sj0nes:
bazel 3.2.0

dominichamon:
hm, same.

well this looks fine as documentation and i'll see if i can figure out how to run the tests. it seems like my runfiles aren't available for some reason.

PR #980, opened 2020-06-15
timers: silence format overflow warning

brianwolfe:
A previous PR https://github.com/google/benchmark/pull/965 created an overflow warning in some compilers because the hour field for the timezone offset is not explicitly limited to the range [0, 99]. Expanded the buffer size to avoid all related compiler warnings. I don't think this runs where we would be constrained on stack size, so this should be fine.

@bdice @jrhemstad @LebedevRI

### DIFF
--- a/src/timers.cc
+++ b/src/timers.cc
@@ -189,8 +189,9 @@ std::string LocalDateTimeString() {
   std::size_t timestamp_len;
   long int offset_minutes;
   char tz_offset_sign = '+';
-  char tz_offset[kTzOffsetLen + 1];
-  char storage[kTimestampLen + kTzOffsetLen + 1];
+  // Long enough buffers to avoid format-overflow warnings
+  char tz_offset[128];
+  char storage[128];
 
 #if defined(BENCHMARK_OS_WINDOWS)
   std::tm *timeinfo_p = ::localtime(&now);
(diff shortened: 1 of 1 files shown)


brianwolfe:
I'm confused, I think `sizeof(tz_offset) == 7`, am I missing something?

LebedevRI:
> I'm confused, I think `sizeof(tz_offset) == 7`, am I missing something?

In `int k[1]` you can only write to `k[0]` while `k[1]` is one-past-the-end.

brianwolfe:
Sure, but shouldn't sprintf be writing to `tz_offset[0]` to `tz_offset[5]` for the 6 characters of the tz offset, then writing the null character to `tz_offset[6]`?

brianwolfe:
Actually, the right way to handle this is probably just to use `snprintf` instead of `sprintf`. WDYT?

davidwendt:
You can force the error to be shown by setting `-Wformat-overflow=2 -Werror`. 

The warning is about how the max size of a `long int offset_minutes` could overflow the `tzn_offset` var.

This should fix it though:
```
char tz_offset[kTzOffsetLen + 37];
```
There is other code that ensures the size remains at length 6:
https://github.com/google/benchmark/blob/99c52f141472a87c3016bb479fa3ef0e67789950/src/timers.cc#L217

Changing to `snprintf` just changes the error to `format-truncation` instead

brianwolfe:
ack, thanks for the help with reproducing.

brianwolfe:
@davidwendt what compiler version are you using? I wasn't able to reproduce locally. Sorry for the noise on this rather silly simple change.

davidwendt:
> @davidwendt what compiler version are you using? I wasn't able to reproduce locally. Sorry for the noise on this rather silly simple change.

I'm using gcc 7.5.0 on Linux 18.04. It's interesting because I was able to see the error with the following flags:
```
-std=c++11  -Wall  -Wextra  -Wshadow  -fstrict-aliasing  -Wno-deprecated-declarations  -Wstrict-aliasing -O3
```
Removing the `-O3` seemed suppress the error for some reason.

brianwolfe:
😞  looks like I broke more than one build. 🤦

brianwolfe:
Thanks. At least this commit passes all the CI checks & your checks again.

brianwolfe:
@LebedevRI @dominichamon sorry for the issue with my previous change. This change should fix https://github.com/google/benchmark/issues/981. PTAL.

dominichamon:
Thanks for the quick fix. I should have spotted the lack of `snprintf`.

PR #982, opened 2020-06-16
bazel support for tools

al-tu:
tools subpackage made with py_library and py_binary bazel rules
usage example:
bazel run //tools:compare -- benchmark <benchmark_baseline> <benchmark_contender>


al-tu:
@googlebot I signed it!

WORKSPACE
```diff
@@ -35,3 +35,26 @@ new_local_repository(
     path = "/usr/include/python3.6",  # May be overwritten by setup.py.
 )
 
+http_archive(
```
dominichamon, line 38: i'm curious why we don't need to bring in rules_python for the //bindings/python/benchmark/... package that also uses py_library and py_binary. Any idea?

tools/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+load("@py_deps//:requirements.bzl", "requirement")
+
+load("@rules_python//python:defs.bzl", "py_binary")
+load("@rules_python//python:defs.bzl", "py_library")
+
+py_library(
+    name = "gbench",
+    srcs = glob(["gbench/**/*.py"]),
```
dominichamon, line 8: i think this can be "gbench/*.py"

WORKSPACE
al-tu: right! these are basic python rules, no need to load anything

tools/BUILD.bazel
al-tu: fixed

WORKSPACE
al-tu: just to be clear: rules_python are used for pip requirements management. can't see any other (simpler) ways of listing numpy and scipy dependencies for gbench

dominichamon: that's fine. i'm trying to make sure we don't end up in some weird state where the requirements for the bindings and the requirements for the tools are somehow different and contradictory.

dominichamon:
This LGTM but i could do with someone who knows more about python distribution and ideally bazel to let me know it's reasonable to have pip stuff in here, and what the ramifications of that would be.

dominichamon:
I just merged #994 which solves the requirements bit without bringing bazel into tools. does that conflict with this? i prefered the smaller change to make things easier while i understand what the ramifications of requiring bazel for this would be.

al-tu:
well, the only reason to list "rules_python" in the WORKSPACE is to be able to automatically manage dependencies for gbench (numpy, scipy). There would be no conflict if is just remove those lines from tools/BUILD.bazel and do not require "rules_python". I guess it's ok if we agree on the fact that notes in docs/tools.md are sufficient for anyone trying to run those tools with bazel (approach similar to that in [#994](https://github.com/google/benchmark/pull/994)). From the other side of things - automatic deps management is probably the main reason to bring those targets into the project in the first place.
Thanks for the review, i'm ready to edit this MR or close it if it seems like an unreasonable feature addition.

dominichamon:
Still hoping that someone out there can help me understand if this is reasonable or is going to break in unexpected ways for some users.

al-tu:
Hello! It seems like no one's going to come and shed some light on it, unless we ask for it :) Is there any chance we can ask someone from bazel team to have a look at this? I suppose it's a matter of a quick look to reason about whether this approach is appropriate for this particular project or not.

dominichamon:
I think we can resolve the conflicts and i'll just get it landed. if it breaks for someone we can fix it (it looks mostly reasonable to me).

dominichamon:
looks like the python bindings are failing with this change.

al-tu:
python bindings finally made it to successful build with latest numpy & scipy versions

dominichamon:
thanks for all the attempts and recommits. i'll take a look.

dominichamon:
Thanks!

PR #985, opened 2020-06-18
add googletest as a submodule

DapengFeng:
add googletest as a submodule of benchmark.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,9 +51,6 @@ build/
 _build/
 build*/
 
-# in-source dependencies
-/googletest/
-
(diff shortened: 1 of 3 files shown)


DapengFeng:
> @googlebot I signed it!

dominichamon:
we've explicitly not done this for a reason.

some people already have googletest installed globally. Others are running benchmark alongside googletest within an existing project. having it as a submodule leads to projects and developers having multiple versions of googletest that can conflict, or at the very least, cause confusion.

as @LebedevRI has said, if there's a specific problem this solves that is more serious than the issues it would create, then we should discuss it.

PR #988, opened 2020-06-19
Add support for changing cpu affinity

koct9i:
CPU affinity mask allows to specify set of processors for each thread.
This could be used in seveal ways:

* limit cpu power to show how benchmark scales with machine size
* bind threads to processors to eliminate noise from cpu scheduler

Add mehods Processors(), ProcessorRange(), DenseProcessorRange() which
define count of available processors similar to count of threads.

Method BindThreadsToProcessors() pins each thread to exactly one processor.
When count of threads bigger than count of processors then processors are
distributed in round-robin order.

Signed-off-by: Konstantin Khlebnikov <khlebnikov@yandex-team.ru>

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -686,6 +686,69 @@ std::vector<double> GetLoadAvg() {
 
 }  // end namespace
 
+namespace internal {
+
+std::vector<int> GetThreadAffinity() {
+  std::vector<int> cpus;
+
+#ifdef BENCHMARK_OS_WINDOWS
+  DWORD_PTR procmask, sysmask;
+￼￼
+  // Get process cpu mask because actual thread cpu mask is hard to get.
+  // This is used only at start to get initial cpu affinity mask.
+ ￼if (!GetProcessAffinityMask(GetCurrentProcess(), &procmask, &sysmask))
+    PrintErrorAndDie("Failed during call to GetProcessAffinityMask: ",
+                     GetLastError());
+
+  cpus.resize(0);
+  for (unsigned int cpu = 0; cpu < sizeof(procmask) * 8; cpu++)
+    if ((procmask >> cpu) & 1)
+      cpus.push_bash(cpu);
+￼
+#else  // BENCHMARK_OS_WINDOWS
+  cpu_set_t cpuset;
+  CPU_ZERO(&cpuset);
+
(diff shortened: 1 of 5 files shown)


src/sysinfo.cc
```diff
@@ -686,13 +686,77 @@ std::vector<double> GetLoadAvg() {
...
 }  // end namespace
 
+namespace internal {
+
+std::vector<int> GetThreadAffinity() {
+  std::vector<int> cpus;
+
+#ifdef BENCHMARK_OS_WINDOWS
+  DWORD_PTR procmask, sysmask;
+￼￼
```
LebedevRI, line 696: These symbols look out of place here and elsewhere

dominichamon:
looks like `cpu_set_t` isn't available on osx. maybe [this](https://developer.apple.com/library/archive/releasenotes/Performance/RN-AffinityAPI/#//apple_ref/doc/uid/TP40006635-CH1-DontLinkElementID_2)?

our CI doesn't cover android or ppc or riscv, so we may need some followups to support these platforms. It might be worth being more explicit in your `#ifdef` blocks about which platforms are expected to be supported and then making it clear to users if they try to call the method on an unsupported platform.

lkleen:
> 
> looks like `cpu_set_t` isn't available on osx. maybe [this](https://developer.apple.com/library/archive/releasenotes/Performance/RN-AffinityAPI/#//apple_ref/doc/uid/TP40006635-CH1-DontLinkElementID_2)?
> 
yes, this code sets the cpu affinity on mac

`
// note: this should be called before thread was started!
thread_affinity_policy ap;
ap.affinity_tag = cpu + 1; // tag == 0 -> no affinity
int ret = thread_policy_set (mach_thread_self (), THREAD_AFFINITY_POLICY, (integer_t*) &ap, THREAD_AFFINITY_POLICY_COUNT);
`

koct9i:
Yes, macos has no API for true cpu-affinity. Only binding each thread to one cpu.
So, fist use case (emulating smaller machine by limiting count of cpus) cannot be implemented here.

Maybe this feature should be dropped in favor of equal functionality for all platforms.
But benchmark should still check cpu-affinity before starting to not disturb externally set limitation.
(i.e. # taskset -c 0-3 ./benchmark)

Or we could keep it for platforms where it works. Macos machines are not so big in comparison to usual linux servers.

Issue #944, opened 2020-03-10
[FR] More time units


ujos:
In order to evaluate my function performance regardless of CPU frequency, I'd like to be able to measure time in CPU cycles.

Let's assume I have a function which on CPU `A` takes 100 CPU cycles to finish. Then, I run this function on the same but overclocked CPU `B` and again it takes 100 CPU cycles. What does it mean? It means that code is good, memory is not a bottleneck. 

What if on CPU `B` it takes 150 cycles. It means that most probably some CPU cycles are stale on waiting for something. Probably, that requires investigation. 

Or some another perspective. What if on `CPU A` it takes 100 cycles, while on `CPU B` (similar frequency) it takes 200 cycles? Well... that is interesting and worthes investigation.

Basically, I know what CPUs I'am running and I can convert nanoseconds returned from Google Benchmark into CPU cycles, but it requires extra step. However, I'm not really sure that `time * frequency` will return accurate value.

From other side, I can use `perf` tool to investigate performance in details. Unfortunately, on Windows there is no `perf`. Visual Studio also does not work for me (it crashes, if I enable CPU cache misses counter). Moreover, on production I cannot install Visual Studio. Intel tools are not for free.

Therefore, what I'd like to ask You is to add CPU cycles time unit. It can be printed along with time or instead of time (depends how can it be implemented). 

To obtain the CPU cycles You can use either `__rdtsc` or `__rdtscp` intrinsic. Last one is better as it is serializeable instruction and guaranteed that CPU does not execute any instruction of the `for (auto _ : state)` loop after `__rdtscp` is called.

damageboy:
Would like to see this as an option too.
While being a problematic area, adding cycles has a normalizing effect, at least on more modern hardware with invariant rdtsc supprot.

dominichamon:
in regard `__rdtsc` or `__rdtscp`, we have `benchmark::cycleclock::Now()` that returns the number of cycles since power-on and uses `rdtsc` internally (or `mfspr`, `clock_gettime` etc, etc).

This could be implemented, then, by adding a new state to `ThreadTimer` that tracks clock cycles alongside the existing `start_real_time_` and `start_cpu_time_`.

dominichamon:
see #989 for a proof of concept.

dominichamon:
This is now resolved i think thanks to #1114.

PR #989, opened 2020-06-22
Add cycles to benchmark output

dominichamon:
Not expected to pass tests yet, but a proof of concept for #944


dominichamon:
Test fixes added. This should be good for testing, if anyone wants to. @damageboy @ujos who both requested this in #944: you might want to patch this in and see if it works for you.

dominichamon:
an example:

`BM_CalculatePiRange` has a parameter `x` that governs how much work to do (how many iterations to use to calculate pi). In the below results we see that the CPU time increases linearly with `x` and the number of cycles is constant.

```json
 {                               
    "name": "BM_CalculatePiRange/1",                                          
    "cycles": 2283850203,
    "cpu_time": 9.960771796194326,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/8",
    "cycles": 1924822988,
    "cpu_time": 45.427452928869954,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/64",
    "cycles": 2271234956,
    "cpu_time": 550.63497396762,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/512",
    "cycles": 1861170539,
    "cpu_time": 4153.484606791196,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/4096",
    "cycles": 2050949685,
    "cpu_time": 33583.06278647821,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/32768",
    "cycles": 2036071505,
    "cpu_time": 268493.11315688526,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/262144",
    "cycles": 2040215893,
    "cpu_time": 2146894.726993865,
    "unit": "ns"
  },
  {
    "name": "BM_CalculatePiRange/1048576",
    "cycles": 2066375310,
    "cpu_time": 8586935.804878065,
    "unit": "ns"
  },
```

ujos:
@dominichamon Does it work for JSON only? CSV and Console do not report cycles

ujos:
> and the number of cycles is constant

I'm not sure I understand, why CPU time increases and cycles are const?

dominichamon:
@ujos i only added json reporting for the first pass.

I'm also not sure how to interpret the constant cycles against an increasing CPU time. it _might_ be related to the multi-core issue.. ie, if the work is being done on core 2 but i'm getting the cycle clocks from core 1, then i won't measure the right thing.

perhaps this requires some affinity setting.

ujos:
> I'm also not sure how to interpret the constant cycles against an increasing CPU time.

I just do not understand, why cycles are cost while CPU time is increasing... More iterations `CalculatePi` is doing, more CPU cycles this function has to consume. So, that was not clear why `BM_CalculatePiRange/1048576` and `BM_CalculatePiRange/1` reported same cycles.

> perhaps this requires some affinity setting.

One more additional parameter which is required if user wants to report cycles?

dominichamon:
dropping in favour of #1104

PR #993, opened 2020-06-26
Expose methods for custom main functions in Python.

skye:


skye:
cc @chr1sj0nes

bindings/python/benchmark/__init__.py
```diff
@@ -60,3 +60,8 @@ def _run_benchmarks(argv):
 
 def main(argv=None):
   return app.run(_run_benchmarks, argv=argv, flags_parser=_flags_parser)
+
+
+# Methods for use with custom main function.
+initialize = _benchmark.Initialize
+run_benchmarks = _benchmark.RunSpecifiedBenchmarks
```
dominichamon, line 67: `run_benchmarks = _run_benchmarks` ? Then you get the common usage error check on argv.

skye: This is for use with custom main functions, so there may be `argv` entries, or it might not be convenient to pass in `argv` at all.

dominichamon:
thanks!

PR #994, opened 2020-06-29
Add requirements.txt for python dependencies

ottojo:
This adds a `requirements.txt` file to the python tools, which makes it possible to install its dependencies (currently only scipy) using `pip`. This also makes it possible to add additional dependencies in the future without any changes to the installation process.


ottojo:
I am just now seeing that #982 already exists, which seems to address the same issue.

dominichamon:
LGTM. I'll merge when CI passes.

dominichamon:
thank you!

PR #997, opened 2020-07-02
Remove "BENCHMARK_" prefix from env var version of CLI flags

### LINKED ISSUE #995: [BUG] BENCHMARK_OUT_FORMAT and BENCHMARK_OUT env variables are not working
**Describe the bug**

According to the documentation

> Write benchmark results to a file with the --benchmark_out=<filename> option (or set BENCHMARK_OUT). Specify the output format with --benchmark_out_format={json|console|csv} (or set BENCHMARK_OUT_FORMAT={json|console|csv}). Note that specifying --benchmark_out does not suppress the console output.

However setting those environment variables does not work.
This is because the command argument names already start by `benchmark` so the function `FlagToEnvVar` adds the `BENCHMARK_` prefix again
https://github.com/google/benchmark/blob/8982e1ee6aef31e48170400b7d1dc9969b156e5e/src/commandlineflags.cc#L91

So the actual variables that the code checks are `BENCHMARK_BENCHMARK_OUT_FORMAT` and `BENCHMARK_BENCHMARK_OUT`.

**System**
  - OS: Any
  - Compiler and version: Any

**To reproduce**
1. sync to tag v1.5.1
2. cmake/bazel...
3. make ...
4. Try any example with `BENCHMARK_OUT` set to `test_out.json` and `BENCHMARK_OUT_FORMAT` to `json` 

**Expected behavior**

`BENCHMARK_OUT` and `BENCHMARK_OUT_FORMAT` should have the same effect as the flags `--benchmark_out` and `--benchmark_out_format` respectively, or the documentation should give the correct environment variables.

dominichamon:
Fixes #995

### DIFF
--- a/src/commandlineflags.cc
+++ b/src/commandlineflags.cc
@@ -88,7 +88,7 @@ static std::string FlagToEnvVar(const char* flag) {
   for (size_t i = 0; i != flag_str.length(); ++i)
     env_var += static_cast<char>(::toupper(flag_str.c_str()[i]));
 
-  return "BENCHMARK_" + env_var;
+  return env_var;
 }
 
 }  // namespace
--- a/test/commandlineflags_gtest.cc
+++ b/test/commandlineflags_gtest.cc
@@ -26,175 +26,175 @@ int unsetenv(const char* name) {
 #endif  // BENCHMARK_OS_WINDOWS
 
 TEST(BoolFromEnv, Default) {
-  ASSERT_EQ(unsetenv("BENCHMARK_NOT_IN_ENV"), 0);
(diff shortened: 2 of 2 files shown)


LebedevRI:
So the problem is that we end up with double `BENCHMARK_` prefix?
Can we at least assert that the prefix is already there?

dominichamon:
> So the problem is that we end up with double `BENCHMARK_` prefix?
> Can we at least assert that the prefix is already there?

there's one flag that doesn't have a benchmark prefix (and shouldn't): `v`. I don't see a need to assert on something that is a 'best practice' rather than an invariant.

PR #1007, opened 2020-07-21
Fix RegisterBenchmark example

Yannic:
This change fixes the `RegisterBenchmark` example in the Readme file to return the result of `RunSpecifiedBenchmarks()`.
If this return is omitted, the program will always exit with status 0, even if one of the benchmarks failed (e.g. by using `state.SkipWithError(...)`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1167,7 +1167,7 @@ int main(int argc, char** argv) {
   for (auto& test_input : { /* ... */ })
       benchmark::RegisterBenchmark(test_input.name(), BM_test, test_input);
   benchmark::Initialize(&argc, argv);
-  benchmark::RunSpecifiedBenchmarks();
+  return benchmark::RunSpecifiedBenchmarks();
 }
 ```
 


LebedevRI:
Hm, but then, `BENCHMARK_MAIN() ` is also wrong.

dominichamon:
> Hm, but then, `BENCHMARK_MAIN() ` is also wrong.

This PR seems to misunderstand `RunSpecifiedBenchmarks` return value. It will return 0 if no benchmarks match the filter. This is not 'success' so shouldn't return as 0 from `main`. If it returns a non-zero number (which would be returned from `main` as an error with this PR) the non-zero number is the number of matched benchmarks to run. It doesn't take into account any skip with error runs anyway.

dominichamon:
If you want to pursue this PR, @Yannic, can you make sure you're actually returning a meaningful status from `BENCHMARK_MAIN` (likely `RunSpecifiedBenchmarks(...) == 0 ? 1 : 0` or similar) as well as updating the docs?

if you don't want to then please close this.

Yannic:
I did more tests and you're right, no idea what I was testing (or thinking) when opening this PR. I'll look into how to achieve that the program exits with non-zero status when there are any errors and then open another PR.

Let me close this. Thank you for the explanation!

Issue #1005, opened 2020-07-20
[BUG] JSON reporter should not report cpu_scaling_enabled if library can not detect it


Lectem:
**Describe the bug**

On any non linux platform (or in case of error), google benchmark will report cpu_scaling as false. (See [sysinfo.cc#L212](https://github.com/google/benchmark/blob/master/src/sysinfo.cc#L212))

I think it should instead return an enum (enabled, disabled, unknown) or at least provide a way to know if the value is correct.

Right now, the [json reporter](https://github.com/google/benchmark/blob/master/src/json_reporter.cc#L125) will always report `"cpu_scaling_enabled": false,` on non-linux platforms which is wrong.

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Anything but linux
  - Compiler and version: any

**To reproduce**
1. sync to commit ...
2. cmake/bazel...
3. make ...
4. Run any benchmark with a json reporter on a non-linux platform

**Expected behavior**

The json output file should either:

- not contain the `cpu_scaling_enabled`
- not be a boolean (report `"unknown"`)
- have the correct value

PR #1008, opened 2020-07-28
JSONReporter: don't report on scaling if we didn't get it (#1005)

Turim:
1. It should have been std::optional, but it seems it would slightly increment C++ standard version to 17, so it's the pair then.
2. That JSONReporter::Context is very limited in terms of 'create different context in order to test them (dependency injection thing)', so I were unsuccessful to create meaningful test (thoughts were: stringstream->JSONReporter->Report->find strings in the output)

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -176,6 +176,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <map>
 #include <set>
 #include <string>
+#include <utility>
 #include <vector>
 
 #if defined(BENCHMARK_HAS_CXX11)
@@ -1297,7 +1298,7 @@ struct CPUInfo {
   int num_cpus;
   double cycles_per_second;
   std::vector<CacheInfo> caches;
-  bool scaling_enabled;
+  std::pair<bool /*validity*/, bool /*value*/> scaling_enabled;
   std::vector<double> load_avg;
 
   static const CPUInfo& Get();
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1295,7 +1295,7 @@ struct CPUInfo {
     int num_sharing;
   };
 
-  enum class Scaling {
+  enum Scaling {
     UNKNOWN,
     ENABLED,
     DISABLED
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1295,10 +1295,16 @@ struct CPUInfo {
     int num_sharing;
   };
 
+  enum class Scaling {
+    UNKNOWN,
+    ENABLED,
+    DISABLED
+  };
+
   int num_cpus;
   double cycles_per_second;
   std::vector<CacheInfo> caches;
-  std::pair<bool /*validity*/, bool /*value*/> scaling_enabled;
+  Scaling scaling;
   std::vector<double> load_avg;
 
   static const CPUInfo& Get();
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -214,7 +214,7 @@ CPUInfo::Scaling CpuScaling(int num_cpus) {
   // We don't have a valid CPU count, so don't even bother.
   if (num_cpus <= 0) return CPUInfo::Scaling::UNKNOWN;
 #ifdef BENCHMARK_OS_QNX
-  return CPUInfo::Scaling::DISABLED;
+  return CPUInfo::Scaling::UNKNOWN;
 #endif
 #ifndef BENCHMARK_OS_WINDOWS
   // On Linux, the CPUfreq subsystem exposes CPU information as files on the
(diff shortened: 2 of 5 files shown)


LebedevRI:
Thinking out loud, if we don't know, should we err on the safe side and instead say that it is enabled?

include/benchmark/benchmark.h
```diff
@@ -1297,7 +1298,7 @@ struct CPUInfo {
   int num_cpus;
   double cycles_per_second;
   std::vector<CacheInfo> caches;
-  bool scaling_enabled;
+  std::pair<bool /*validity*/, bool /*value*/> scaling_enabled;
```
dominichamon, line 1301: i imagine an enum is more natural here. 

```c++
enum Scaling {
  UNKNOWN,
  ENABLED,
  DISABLED
};
```

dominichamon:
> Thinking out loud, if we don't know, should we err on the safe side and instead say that it is enabled?

JSON standard is to not have the key if the value would be "null". so i think this PR is correct in that it only sets the key if we know the value.

LebedevRI:
> > Thinking out loud, if we don't know, should we err on the safe side and instead say that it is enabled?
> 
> JSON standard is to not have the key if the value would be "null". so i think this PR is correct in that it only sets the key if we know the value.

It wouldn't be null. I was suggesting that, when we don't know the answer, instead of saying "`false`, scaling is not enabled", "`true` we don't know that it is not enabled".

Turim:
I'm not sure if the information is useful (like _we know nothing about scaling_ -- imho looks like a junk)

include/benchmark/benchmark.h
Turim: I agree, could be more clear

src/sysinfo.cc
```diff
@@ -210,11 +210,11 @@ bool ReadFromFile(std::string const& fname, ArgT* arg) {
...
 }
 
-std::pair<bool /*validity*/, bool /*value*/> CpuScalingEnabled(int num_cpus) {
+CPUInfo::Scaling CpuScaling(int num_cpus) {
   // We don't have a valid CPU count, so don't even bother.
-  if (num_cpus <= 0) return std::make_pair(false, false);
+  if (num_cpus <= 0) return CPUInfo::Scaling::UNKNOWN;
 #ifdef BENCHMARK_OS_QNX
-  return std::make_pair(true, false);
+  return CPUInfo::Scaling::DISABLED;
```
dominichamon, line 217: this should be unknown now i think.

src/sysinfo.cc
```diff
@@ -224,10 +224,10 @@ std::pair<bool /*validity*/, bool /*value*/> CpuScalingEnabled(int num_cpus) {
   for (int cpu = 0; cpu < num_cpus; ++cpu) {
     std::string governor_file =
         StrCat("/sys/devices/system/cpu/cpu", cpu, "/cpufreq/scaling_governor");
-    if (ReadFromFile(governor_file, &res) && res != "performance") return make_pair(true, true);
+    if (ReadFromFile(governor_file, &res) && res != "performance") return CPUInfo::Scaling::ENABLED;
   }
 #endif
-  return std::make_pair(true, false);
+  return CPUInfo::Scaling::DISABLED;
```
dominichamon, line 230: should this be disabled or unknown?

src/sysinfo.cc
Turim: So we're doing more of bugfixing also? I just got the previous algo..

src/sysinfo.cc
Turim: My thoughts were for DISABLED as for the fallback for line 227 ?
But considering line 217 - could have been fixed also

src/sysinfo.cc
dominichamon: the PR is about not reporting if we don't get scaling information. if the new contract is to report something useful when we can and not if we can't, then this isn't a bugfix. it's about aligning the code with the new contract you've introduced.

dominichamon:
Just waiting for CI to pass, then i'll run some checks locally (just in case) and then we're good. Thank you so much!

Turim:
> Just waiting for CI to pass, then i'll run some checks locally (just in case) and then we're good. Thank you so much!

https://ci.appveyor.com/project/ukai/benchmark/builds/34340295/job/93x0447l7yl7ihpb
For me it's hard to tell, is it just a t/o or an issue? (looks like the lines got swapped or?)
I believe I shouldn't have changed the output..

dominichamon:
> > Just waiting for CI to pass, then i'll run some checks locally (just in case) and then we're good. Thank you so much!
> 
> https://ci.appveyor.com/project/ukai/benchmark/builds/34340295/job/93x0447l7yl7ihpb
> For me it's hard to tell, is it just a t/o or an issue? (looks like the lines got swapped or?)
> I believe I shouldn't have changed the output..

You've changed the output so the line doesn't exist on windows any more. So the expectation would need to change to not include the scaling  key i think.

Turim:
Trying to do quickfix (VS2019 - I presume - has too many failures on tests to be representative), taking VS2017/upgrading WSL (renew gcc) to reproduce locally might take a white

include/benchmark/benchmark.h
```diff
@@ -1295,7 +1295,7 @@ struct CPUInfo {
     int num_sharing;
   };
 
-  enum class Scaling {
+  enum Scaling {
```
dominichamon, line 1298: ah yes, because it's in the header. :)

Turim: Any insights? :) Is C++03 used at present in 2020? :)

dominichamon: yes. we still support it for installations. ie, you need >c++11 to _build_ the library, but you shouldn't need it to link to the library. this will change once the base libc/compilers change for LTS OSes.

dominichamon:
thanks!

PR #1011, opened 2020-07-28
implicit fallback in switch statement QNX

### LINKED ISSUE #1010: Implicit fallback
https://github.com/google/benchmark/blob/99010118805eb12ce56122f859c3604c2a427d05/src/sysinfo.cc#L386

asobhy-qnx:
Fixes #1010

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -384,6 +384,7 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesQNX() {
       case CACHE_FLAG_UNIFIED :
         info.type = "Unified";
         info.level = 2;
+        break;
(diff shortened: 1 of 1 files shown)


src/sysinfo.cc
```diff
@@ -384,6 +384,7 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesQNX() {
       case CACHE_FLAG_UNIFIED :
         info.type = "Unified";
         info.level = 2;
+        break;
       case CACHE_FLAG_SHARED :
         info.type = "Shared";
         info.level = 3;
```
dominichamon, line 390: we should probably have one here too (though it's less of a problem).

asobhy-qnx:
addressed by #1012

PR #1013, opened 2020-07-29
Create build-and-test.yml

dominichamon:
First attempt at a non-travis/non appveyor CI solution


dominichamon:
@LebedevRI this looks ok right? i'll keep the others around for a while but i hope we can move away from them at some point.

PR #1017, opened 2020-08-05
Update build instructions to better use CMake

adambadura:
Build instructions needlessly referred to make when CMake offers a command-line interface to abstract away from the specific build system.

Furthermore, CMake offers command-line "tool mode" which performs basic filesystem operations. While the syntax is a bit more verbose than Linux commands it is platform-independent. Now the commands can be copy-pasted on both Linux and Windows and will just work.

Finally, the Release build type and link-time optimizations are included in initial commands. A natural flow for a new-comer is to read and execute the commands and only then learn that one has to go back and redo them again this time with proper parameters. Now instead the parameters are only explained later but present already in the initial commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,13 +70,13 @@ $ git clone https://github.com/google/googletest.git benchmark/googletest
 # Go to the library root directory
 $ cd benchmark
 # Make a build directory to place the build output.
-$ mkdir build && cd build
-# Generate a Makefile with cmake.
-# Use cmake -G <generator> to generate a different file type.
-$ cmake ../
+$ cmake -E make_directory "build"
+# Generate build system files with cmake.
+$ cmake -E chdir "build" cmake -DCMAKE_BUILD_TYPE=Release ../
+# or, starting with CMake 3.13, use a simpler form:
+# cmake -DCMAKE_BUILD_TYPE=Release -S . -B "build"
 # Build the library.
-# Use make -j<number_of_parallel_jobs> to speed up the build process, e.g. make -j8 .
-$ make
+$ cmake --build "build" --config Release --parallel
 ```
 This builds the `benchmark` and `benchmark_main` libraries and tests.
 On a unix system, the build directory should now look something like this:
@@ -94,13 +94,13 @@ On a unix system, the build directory should now look something like this:
 Next, you can run the tests to check the build.
 
 ```bash
-$ make test
+$ cmake --build "build" --config Release --target test
 ```
 
 If you want to install the library globally, also run:
 
 ```
-sudo make install
+sudo cmake --build "build" --config Release --target install
 ```
 
 Note that Google Benchmark requires Google Test to build and run the tests. This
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,13 +94,13 @@ On a unix system, the build directory should now look something like this:
 Next, you can run the tests to check the build.
 
 ```bash
-$ make test
+$ cmake --build "build" --config Release --target test
```
adambadura, line 97: I have checked the new commands on both Linux and Windows and all seem to work fine (well, the installation failed expectedly with the access rights but looked correct up to that point) except for running tests. I guess it works fine with Makefile (or Ninja) build systems where there is a `test` target. But Visual Studio doesn't have such a target and fails for the new command.

However, it would fail with the old one just as well obviously, so no regress here.

I'm afraid progress could be made here only by using `ctest` instead.

README.md
```diff
@@ -70,13 +70,13 @@ $ git clone https://github.com/google/googletest.git benchmark/googletest
 # Go to the library root directory
 $ cd benchmark
 # Make a build directory to place the build output.
-$ mkdir build && cd build
-# Generate a Makefile with cmake.
-# Use cmake -G <generator> to generate a different file type.
-$ cmake ../
+$ cmake -E make_directory "build"
+# Generate build system files with cmake.
+$ cmake -E chdir "build" cmake -DCMAKE_BUILD_TYPE=Release -DBENCHMARK_ENABLE_LTO=true ../
```
dominichamon, line 75: ENABLE_LTO is considered an advanced feature which is why it's not in the initial instructions.

I think there's a separate discussion about whether it should just be defaulted to `true` :)

README.md
dominichamon: why don't we suggest `ctest`?

README.md
adambadura: Well, as a new user coming to the library I didn't have this impression reading the `README.md`. Instead, I was like: "_well, why didn't you tell me this before I built it without the flag!_" :)

So, perhaps we should change the `README.md` so that it explains what is "advanced" about it and why should one consider not using it. Or we should follow with adding it by default as you mentioned.

Anyway, I'm fine with removing it here if this is what you prefer.

README.md
adambadura: Because we use `enable_testing()` instead of `include(CTest)`. I think it would be better to do the other, but it would go beyond just changing `README.md`. Perhaps as a separate change?

dominichamon: works for me :)

README.md
dominichamon: filed #1022 to consider changing the default. i think we should remove it from here until we can resolve that so the initial build instructions are as short as possible.

README.md
dominichamon: filed #1023 for this.

README.md
adambadura: OK, I will prepare an update for this pull request.

adambadura: I have fixed this in the new version.

dominichamon:
thanks!

PR #1019, opened 2020-08-10
Add support for threading via OpenMP

krzikalla:
This PR adds OpenMP as threading option. If configured with -DBENCHMARK_ENABLE_OPENMP=ON, then the library will use a openmp parallel region instead of std::thread for multithreaded benchmarks.


LebedevRI:
Please, can you explain the more general problem you are trying to solve?

I'm not yet really sure this is a good idea in general.
For example, if `benchmark` was configured with openmp,
and if the code you are benchmarking uses openmp itself,
since nested parallel regions are disablled by default,
the code-level openmp will no longer work.

krzikalla:
1. If most program parts runs in one parallel region, then you don't want to measure the creation and destruction of the openmp parallel region inside the measurement loop (usually this is only a small part of the code). 
2. Masuring different numbers of openmp threads would require complex argument sets, whereas with this PR arguments and output are in place as usual. 
3. You have to take care, that omp_get_max_threads() is already working, when dynamic the benchmark initialization runs. 
Pt. 2 and 3 can also be achieved by the benchmark writer. But openmp uses pragmas, so you can't just write a function carrying out the argument evaluation and parallelization. You have to repeat several code lines in every benchmark.

Those are the general issues. Now to the more specific ones (applying to our user scenario): our code can be called in or out of a parallel region. In the first case most of the program uses one parallel region, so you don't open and close it constantly. 
In the second case a region is opened just for a specific task (e.g. a loop). The rest of the measured code runs sequentially. This can have some performance impact. E.g. if the creation of a shared variable with copyprivate is expensive in a parallel region, but a no-op in a sequential region. I want to see such impacts.

Regarding your concern: if you have a benchmark opening a parallel region by itself, then you have to run it using Threads(OmpNoParallelRegion). Or, don't pass any Threads() arguments, since OmpNoParallelRegion==1.

LebedevRI:
> Regarding your concern: if you have a benchmark opening a parallel region by itself, then you have to run it using Threads(OmpNoParallelRegion). Or, don't pass any Threads() arguments, since OmpNoParallelRegion==1.

I think you missed my point.

krzikalla:
@googlebot I signed it!

LebedevRI:
> > Regarding your concern: if you have a benchmark opening a parallel region by itself, then you have to run it using Threads(OmpNoParallelRegion). Or, don't pass any Threads() arguments, since OmpNoParallelRegion==1.
> 
> I think you missed my point.

To reword, my question is: what if the code being benchmarked already is using OpenMP-based threading internally,
and it is being `benchmark`ed by more than one thread to see how well it scales?

I'm not against this per-se, but i think
1. We need to think about the bigger picture - are we okay with special-casing just OpenMP here,
    is this an one-time thing, or is some other API will be added next?
2. While i agree that it might be helpful to not measure the cost of `omp parallel` (i.e. thread creation),
    i'm not sure i'm okay with such compile-time switch.

krzikalla:
To paraphrase: you have OpenMP-based code and want to measure its scaling by using threads. That's currently impossible - neither with the master nor with my PR. And you want, that my PR enables that scenario too. Right?

Indeed, I have thought for a quick moment, that OmpNoParallelRegion could be a bitmask rather than a value. Then one could specify Threads(OmpNoParallelRegion+num_threads) and internally no region is opened, but just omp_num_threads is set.
This would also solve the problem, that a parallel region with one thread is currently not possible, since OmpNoParallelRegion+1 would be different from 1. I haven't implemented OmpNoParallelRegion as bitmask yet, since this requires some more changes. But it could be a possible solution to your point.

To 1: I am not, sure I understand this "special-casing" term properly. IMHO OpenMP is no special case. At least in my environment (HPC) it is rather the default threading API. TBH, I have never seen code using std::thread for data parallelism.
To 2: So you want an additional runtime flag. Sometimes like BM(...)->UseOpenMP()->Threads(1)->Threads(2)->... Seems reasonable and I can add this.

include/benchmark/benchmark.h
```diff
@@ -415,6 +415,13 @@ enum TimeUnit { kNanosecond, kMicrosecond, kMillisecond };
 // calculated automatically to the best fit.
 enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
 
+// If OpenMP threading is used, two special values can be passed to Threads():
+// OmpDefaultNumThreads: use omp_get_max_threads().
+// OmpNoParallelRegion==1: don't open a parellel region. A parallel region with one thread is not possible.
```
LebedevRI, line 420: Can you please quote where the spec says that?
`#pragma omp parallel num_threads(1)` works just fine.

src/benchmark_runner.cc
```diff
@@ -203,6 +207,17 @@ class BenchmarkRunner {
...
+#ifdef _OPENMP
+    if (b.threads == OmpNoParallelRegion) {
+      RunInThread(&b, iters, 0, manager.get());
+    }
+    else {
+    #pragma omp parallel num_threads(b.threads)
+      {
+        RunInThread(&b, iters, omp_get_thread_num(), manager.get());
+      }
+    }
```
LebedevRI, line 219: I would think this should be closer to
```c++
// Start specified thread count, so that the benchmark, if it uses threads internally, doesn't have to.
#pragma omp parallel num_threads(b.thread_count)
{
  // Do we actually want to run the same thing in each thread?
  if(!(b.threading & OmpNoParallelRegion)) {
    // Alright then, in each thread, run the benchmark.
    RunInThread(&b, iters, omp_get_thread_num(), manager.get());
  } else {
    // If not, then just do it it a single thread.
#pragma omp single
    RunInThread(&b, iters, omp_get_thread_num(), manager.get());
  }
}
```

LebedevRI:
> To paraphrase: you have OpenMP-based code and want to measure its scaling by using Threads().

Yep.

> That's currently impossible - neither with the master nor with my PR.

Hm, it is though? TBF i don't know either way right now, but i will be surprised if that is actually the case.

> To 1: I am not, sure I understand this "special-casing" term properly.

By special-casing i mean adding a second/changing backend for `Threads()`,
instead of defaulting to the most basic threading implementation (what we currently do).

> IMHO OpenMP is no special case. At least in my environment (HPC) it is rather the default threading API.

And that's the thing, it clearly differs by environment. I'm wondering if this is an unique occurrence,
or we will later on be okay with repeating that with some yet another API.

> To 2: So you want an additional runtime flag. Sometimes like BM(...)->UseOpenMP()->Threads(1)->Threads(2)->... Seems reasonable and I can add this.

I think that is the _only_ acceptable way forward, yes.

> TBH, I have never seen code using std::thread for data parallelism.

From off the top of my head, [rawspeed was doing that until recently](https://github.com/darktable-org/rawspeed/pull/158),
[darktable does that for it's worker threads](https://github.com/darktable-org/darktable/blob/1dbdbf019ea9fdcc63bc2141254fa24041bbb385/src/control/jobs.c#L634), there are likely other examples.
But now if *i'm* being honest, i've never seen OpenMP being used in this way, for abstract not data-driven parallelism.
This is really something i'd expect ptheads/`std::threads` to be used for, that is why i'm not yet really sold here..

So i would really like to see the PR's description be updated with explicit list of goals/trade-offs first.

I have more comments on the diff itself, but i'll post that later.

include/benchmark/benchmark.h
krzikalla: The comment was misleading. I just meant, that my implementation didn't support parallel region with just one thread.
In a new version the issue will be gone by design.

src/benchmark_runner.cc
krzikalla: Before I finish my changes, just let us define possibple scenarios:
1. No threading.
2. Use std::thread (as usual).
3. Open an openmp region. Threads(...) will determine the number of threads in that region.
4. Don't open an openmp region, but just set the number of threads via omp_set_num_threads(). The benchmark will open the region by itself.

Do we agree, that the list is complete regarding this PR? 
Do we agree, that the framework does not support a combination of mode 2 with any of the other modes - e.g. no one wants n std::threads with m openmp threads?

How do we expose these modes to the user? I see three options:
1. We encode it in the Threads argument(s) via some bit flags. E.g. `->Threads(OmpRegion|8)` or `->ThreadRange(OmpRegion|4, OmpRegion|8)`.
2. We have dedicated functions like `->UseOpenMP()->`ThreadRange(4, 8)`
3. The mode is a (defaulted) argument to the Threads function family. E.g.  `->Threads(8, OmpRegion)` or `->ThreadRange(4, 8, OmpRegion)`.

Out of these three options I like the third one most. Do you see another approach?

LebedevRI: Will reply to the rest later, but 

> * We have dedicated functions like `->UseOpenMP()->ThreadRange(4, 8)`

^ this

krzikalla:
> From off the top of my head, [rawspeed was doing that until recently](https://github.com/darktable-org/rawspeed/pull/158),
> [darktable does that for it's worker threads](https://github.com/darktable-org/darktable/blob/1dbdbf019ea9fdcc63bc2141254fa24041bbb385/src/control/jobs.c#L634), there are likely other examples.
> But now if _i'm_ being honest, i've never seen OpenMP being used in **this** way, for abstract not-data-driven parallelism.
> This is really something i'd expect ptheads/`std::threads` to be used for, that is why i'm not yet really sold here..

Here I don't get you. OpenMP isn't indeed used for task-based parallelism yet very often (since the necessary features were added rather recently). But for data-based parallelism it is used quiote a lot. And if I run **scalability** tests, I firstly think of data-based parallelism. So IMHO OpenMP support is rather important for such scalability tests.

src/benchmark_runner.cc
krzikalla: At a first glance this seems to be the best option. However note, that this would fix the whole benchmark family to one mode. Thus, if you like to test another threading mode too, you have to re-register the benchmark with all other arguments being the same. 

How about a fourth option: use dedicated functions, e.g. `->OmpThreads(8)` or `->OmpThreadRange(4, 8)`?
This approach isn't extensible that nicely to other parallelization APIs though.

LebedevRI: > At a first glance this seems to be the best option. However note, that this would fix the whole benchmark family to one mode. Thus, if you like to test another threading mode too, you have to re-register the benchmark with all other arguments being the same.

Yep.

> Or do you want to have an ordering semantic? E.g.
> 
> ```
> BM(...)->Threads(2)->UseOpenMP()->Threads(8);
> ```
> 
> tests BM with 2 std::threads and then with 8 OpenMP threads. I don't like this.

No, i don't want not like this.

LebedevRI:
> > To paraphrase: you have OpenMP-based code and want to measure its scaling by using Threads().
> 
> Yep.
> 
> > That's currently impossible - neither with the master nor with my PR.
> 
> Hm, it is though? TBF i don't know either way right now, but i will be surprised if that is actually the case.

So, i have just tried this:
```
#include <omp.h>
#include <thread>
#include <iostream>
#include <vector>

void foo(int outer) {
#pragma omp critical
std::cout << "foo got called, outer = "<<outer<<"\n";
  #pragma omp parallel num_threads(8)
{
  #pragma omp for
for(int i = 0; i != 8; ++i) {
#pragma omp critical
std::cout << "started " << omp_get_num_threads() << " threads, hello from thread " << omp_get_thread_num() << ", i = "<<i<<", outer = "<<outer<<"\n";
}
}
}

int main() {
  omp_set_num_threads(8);

  std::vector<std::thread> ts;
    for(int i = 0; i != 8; ++i)
    ts.emplace_back(foo, i);

  for(auto& t : ts)
    t.join();

  return 0;
}
```

Produces:
```
$ ./a.out 
foo got called, outer = 0
foo got called, outer = 1
foo got called, outer = 2
foo got called, outer = 3
started 8 threads, hello from thread 0, i = 0, outer = 0
foo got called, outer = 4
started 8 threads, hello from thread 0, i = 0, outer = 3
foo got called, outer = 5
started 8 threads, hello from thread 0, i = 0, outer = 1
started 8 threads, hello from thread 7, i = 7, outer = 0
started 8 threads, hello from thread 6, i = 6, outer = 0
started 8 threads, hello from thread 4, i = 4, outer = 0
started 8 threads, hello from thread 5, i = 5, outer = 0
started 8 threads, hello from thread 0, i = 0, outer = 4
foo got called, outer = 7
started 8 threads, hello from thread 3, i = 3, outer = 0
started 8 threads, hello from thread 2, i = 2, outer = 4
started 8 threads, hello from thread 0, i = 0, outer = 2
started 8 threads, hello from thread 3, i = 3, outer = 2
started 8 threads, hello from thread 1, i = 1, outer = 2
started 8 threads, hello from thread 2, i = 2, outer = 3
started 8 threads, hello from thread 3, i = 3, outer = 3
started 8 threads, hello from thread 4, i = 4, outer = 3
started 8 threads, hello from thread 0, i = 0, outer = 5
started 8 threads, hello from thread 4, i = 4, outer = 5
started 8 threads, hello from thread 2, i = 2, outer = 5
started 8 threads, hello from thread 4, i = 4, outer = 2
started 8 threads, hello from thread 5, i = 5, outer = 2
started 8 threads, hello from thread 7, i = 7, outer = 2
started 8 threads, hello from thread 6, i = 6, outer = 5
started 8 threads, hello from thread 7, i = 7, outer = 5
started 8 threads, hello from thread 5, i = 5, outer = 3
foo got called, outer = 6
started 8 threads, hello from thread 5, i = 5, outer = 5
started 8 threads, hello from thread 6, i = 6, outer = 2
started 8 threads, hello from thread 0, i = 0, outer = 7
started 8 threads, hello from thread 4, i = 4, outer = 4
started 8 threads, hello from thread 5, i = 5, outer = 4
started 8 threads, hello from thread 7, i = 7, outer = 4
started 8 threads, hello from thread 3, i = 3, outer = 5
started 8 threads, hello from thread 4, i = 4, outer = 7
started 8 threads, hello from thread 5, i = 5, outer = 7
started 8 threads, hello from thread 6, i = 6, outer = 4
started 8 threads, hello from thread 2, i = 2, outer = 2
started 8 threads, hello from thread 3, i = 3, outer = 4
started 8 threads, hello from thread 3, i = 3, outer = 7
started 8 threads, hello from thread 1, i = 1, outer = 5
started 8 threads, hello from thread 6, i = 6, outer = 7
started 8 threads, hello from thread 0, i = 0, outer = 6
started 8 threads, hello from thread 2, i = 2, outer = 7
started 8 threads, hello from thread 7, i = 7, outer = 3
started 8 threads, hello from thread 7, i = 7, outer = 7
started 8 threads, hello from thread 2, i = 2, outer = 1
started 8 threads, hello from thread 3, i = 3, outer = 1
started 8 threads, hello from thread 6, i = 6, outer = 3
started 8 threads, hello from thread 1, i = 1, outer = 7
started 8 threads, hello from thread 4, i = 4, outer = 6
started 8 threads, hello from thread 6, i = 6, outer = 6
started 8 threads, hello from thread 1, i = 1, outer = 1
started 8 threads, hello from thread 7, i = 7, outer = 6
started 8 threads, hello from thread 3, i = 3, outer = 6
started 8 threads, hello from thread 7, i = 7, outer = 1
started 8 threads, hello from thread 6, i = 6, outer = 1
started 8 threads, hello from thread 2, i = 2, outer = 6
started 8 threads, hello from thread 4, i = 4, outer = 1
started 8 threads, hello from thread 5, i = 5, outer = 1
started 8 threads, hello from thread 5, i = 5, outer = 6
started 8 threads, hello from thread 1, i = 1, outer = 4
started 8 threads, hello from thread 1, i = 1, outer = 6
started 8 threads, hello from thread 2, i = 2, outer = 0
started 8 threads, hello from thread 1, i = 1, outer = 3
started 8 threads, hello from thread 1, i = 1, outer = 0

```

Do you agree that it works just fine now, before the patch?
Now i'm even more confused/unsold here.

> * If most program parts runs in one parallel region, then you don't want to measure the creation and destruction of the openmp parallel region inside the measurement loop (usually only a small code snippet of the program).

Could you please be more specific here? Are you saying that OpenMP run-time library creates it's threads
each time, and merges them afterwards, so each time a thread is needed, a new one is created from scratch,
instead of reusing some previously-created thread that is spin-waiting since last use?

krzikalla:
> Do you agree that it works just fine now, before the patch?

Actually no. TBH I don't know the purpose of your example. Of course every std::thread can span openmp threads and vice versa. But this has nothing to do with scalability tests in google/benchmark. When I want to test the scaling, I write e.g.
````
BM(...)->ThreadRange(1, 8);
````
And then my code should be measured with 1 to 8 threads. Currently that's only possible with the std::thread API.

> Could you please be more specific here? Are you saying that OpenMP run-time library creates it's threads
> each time, and merges them afterwards, so each time a thread is needed, a new one is created from scratch,
> instead of reusing some previously-created thread that is spin-waiting since last use?

Nope. OpenMP runtimes normally create a thread pool at the beginning and then reuse the threads from that pool.. Still, there is some overhead. 

ANyway, I will make another PR, which tries to solve the problem in a more general way (without OpenMP specifics).

krzikalla:
Before I create a new PR, you can already get an impression of the new approach here:
https://github.com/krzikalla/benchmark/tree/threadingapi#multithreaded-benchmarks

It is completely agnostic to OpenMP or any other threading API..

PR #1021, opened 2020-08-18
Fix Clang builds on Windows

### LINKED ISSUE #974: can not configure the solution with ninja clang 10 (windows 10)
when compiling project from GitHub with google benchmark using cmake, ninja and clang 10( clang++) in windows 10. the configuration fails with the following issues.
-- Performing Test HAVE_THREAD_SAFETY_ATTRIBUTES
-- Performing Test HAVE_THREAD_SAFETY_ATTRIBUTES -- failed to compile
-- Performing Test HAVE_STD_REGEX
-- Performing Test HAVE_STD_REGEX -- failed to compile
-- Performing Test HAVE_GNU_POSIX_REGEX
-- Performing Test HAVE_POSIX_REGEX
CMake Error at deps/googlebenchmark/CMakeLists.txt:183 (message):

jeremyong:
Fixes #974. The `cxx_feature_check` now has an additional
optional argument which can be used to supply extra cmake flags
to pass to the `try_compile` command. The `CMAKE_CXX_STANDARD=14`
flag was determined to be the minimum flag necessary to correctly
compile and run the regex feature checks when compiling with Clang
under Windows (n.b. this does *not* refer to clang-cl, the frontend
to the MSVC compiler). The additional flag is not enabled for any
other compiler/platform tuple.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,11 +245,17 @@ if (BENCHMARK_USE_LIBCXX)
   endif()
 endif(BENCHMARK_USE_LIBCXX)
 
+set(EXTRA_CXX_FLAGS "")
+if (WIN32 AND "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+  # Clang on Windows fails to compile the regex feature check under C++11
+  set(EXTRA_CXX_FLAGS "-DCMAKE_CXX_STANDARD=14")
+endif()
+
 # C++ feature checks
 # Determine the correct regular expression engine to use
-cxx_feature_check(STD_REGEX)
-cxx_feature_check(GNU_POSIX_REGEX)
-cxx_feature_check(POSIX_REGEX)
+cxx_feature_check(STD_REGEX ${EXTRA_CXX_FLAGS})
+cxx_feature_check(GNU_POSIX_REGEX ${EXTRA_CXX_FLAGS})
+cxx_feature_check(POSIX_REGEX ${EXTRA_CXX_FLAGS})
 if(NOT HAVE_STD_REGEX AND NOT HAVE_GNU_POSIX_REGEX AND NOT HAVE_POSIX_REGEX)
   message(FATAL_ERROR "Failed to determine the source files for the regular expression backend")
 endif()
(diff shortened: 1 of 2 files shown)


jeremyong:
@googlebot I signed it!

CMakeLists.txt
```diff
@@ -245,18 +245,25 @@ if (BENCHMARK_USE_LIBCXX)
   endif()
 endif(BENCHMARK_USE_LIBCXX)
 
+set(EXTRA_CXX_FLAGS "")
+if (WIN32 AND "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+  # Clang on Windows fails to compile the regex feature check under C++11
+  set(EXTRA_CXX_FLAGS "-DCMAKE_CXX_STANDARD=14")
```
LebedevRI, line 251: Could you please check, is it still broken with `set(EXTRA_CXX_FLAGS "-DCMAKE_CXX_STANDARD=11")` ?

dominichamon: i expect it will as per the original bug #974.

dominichamon:
travis failure looks like an infra issue unrelated to this PR.

CMakeLists.txt
jeremyong: @dominichamon is correct

@LebedevRI 14 and up is required for the patch to fix Clang on Windows. 11 is already the default of `try_run` and `try_compile`

jeremyong:
The build failure does seem spurious (hung trying to fetch from the apt repos). Is there any way to retrigger it? I don't believe the build tested in that line of the matrix is any different from any of the others (my change only affects clang on windows)

dominichamon (approved):

jeremyong:
@LebedevRI Are there any changes requested for this PR? It'd be great to have the project building with Clang on Windows. Cheers

LebedevRI:
> @LebedevRI Are there any changes requested for this PR?

Not from me. I suspect this might only be papering over the real reason,
since this works elsewhere, but i guess i don't care enough to block this.

jeremyong:
> I suspect this might only be papering over the real reason

The "real reason" is simply a bug in the std::regex implementation of Clang's C++11 stdlib on windows, but I didn't care to chase this down given that the user can simply choose a version of the standard where this isn't an issue, should they choose to use that compiler/OS tuple. However, the defect that is being patched here is that your build system is issuing a *false negative*. That is, even if I CHOOSE C++14 or C++17 with clang on windows, your build system fails the compilation regardless of whether this should work.

LebedevRI:
> The "real reason" is simply a bug in the std::regex implementation of Clang's C++11 stdlib on windows, but I didn't care to chase this down

Aha, well, as i expected. Can you please file a bug in the proper tracker, too?

dominichamon:
CI still unrelated... travis unable to connect to launchpad apparently. PR LGTM

LebedevRI:
I still want to see a bugreport, but feel free to merge.

PR #1027, opened 2020-08-24
Add support for other multi-threading APIs

krzikalla:
This PR replaces the OpenMP PR. It adds more flexibility to multi-threaded benchmarks. 
You can perform scalability tests independent of std::thread. 
And by the means of ThreadState you can also open a parallel region before the iteration loop, which enables you to measure the asynchronous progress of multiple threads across iterations.
For more information see the section "Multithreaded Benchmarks" in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -905,6 +905,60 @@ BENCHMARK(BM_test)->Range(8, 8<<10)->UseRealTime();
 
 Without `UseRealTime`, CPU time is used by default.
 
+Google/benchmark uses `std::thread` as multithreading environment per default.
+If you want to use another multithreading environment (e.g. OpenMP), you can 
+turn off the automatic threading using the `ManualThreading` function.
+```c++
+static void BM_MultiThreaded(benchmark::State& state) {
+  // Setup code here.
+  for (auto _ : state) {
+#pragma omp parallel num_threads(state.threads)
+    // Run the multithreaded test. 
+  }
+  // Teardown code here.
+}
+
+BENCHMARK(BM_MultiThreaded)->ManualThreading()->Threads(1)->Threads(2)->Threads(4);
+```
+The above example creates a parallel region in each iteration. 
+This includes the setup and teardown of the parallel region in the time measurment and it 
+adds an implicit barrier at the end of each iteration.   
+You can avoid these effects, if you run the whole loop in parallel. 
+Then you must not use the `state` object directly, but create a `ThreadState` object in each thread.
+```c++
+static void BM_MultiThreaded(benchmark::State& state) {
+  // Setup code (shared objects) here.
+#pragma omp parallel num_threads(state.threads)
+  {
+    // Thread-local setup code here.
+    for (auto _ : benchmark::ThreadState(state)) {
+      // Run the multithreaded test. 
+    }
+  }
+  // Teardown code here.
+}
+
+BENCHMARK(BM_MultiThreaded)->ManualThreading()->Threads(1)->Threads(2)->Threads(4);
+```
+If you use the `ThreadState` object and explicitly specify the number of threads, then you must 
+use `ManualThreading` and the number of created `ThreadState` objects must match the number of specified threads. 
+However, if you use `ThreadState` without explicitly specifying the number of threads, 
+then the number of threads is determined by the number of created `ThreadState` objects.
+```c++
+static void BM_MultiThreaded(benchmark::State& state) {
+  // Setup code here.
+#pragma omp parallel 
+  for (auto _ : benchmark::ThreadState(state)) {
+    // Run the multithreaded test. 
+  }
+  // Teardown code here.
+}
+
+BENCHMARK(BM_MultiThreaded);  // measures omp_get_max_threads number of threads.
+```
+Specifying `ManualThreading` is optional in this case.
+
+
 <a name="cpu-timers" />
 
 ### CPU Timers
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -641,6 +641,9 @@ class State {
     return max_iterations - total_iterations_ + batch_leftover_;
   }
 
+  BENCHMARK_ALWAYS_INLINE
+  int GetNumThreadStates() const { return num_thread_states_; }
+
  private
      :  // items we expect on the first cache line (ie 64 bytes of the struct)
   // When total_iterations_ is 0, KeepRunning() and friends will return false.
@@ -685,8 +688,21 @@ class State {
   void FinishKeepRunning();
   internal::ThreadTimer* timer_;
   internal::ThreadManager* manager_;
+  int num_thread_states_;
 
   friend struct internal::BenchmarkInstance;
+  friend class ThreadState;
+};
+
+class ThreadState : public State
+{
+ public:
+  explicit ThreadState(State& s);
+  ~ThreadState();
+ private:
+  State* parent_;
+
+  ThreadState(const ThreadState&);
 };
 
 inline BENCHMARK_ALWAYS_INLINE bool State::KeepRunning() {
@@ -945,6 +961,9 @@ class Benchmark {
   // Equivalent to ThreadRange(NumCPUs(), NumCPUs())
   Benchmark* ThreadPerCpu();
 
+  // Don't create threads. Let the user evaluate state.threads and/or use ThreadState.
+  Benchmark* ManualThreading() { manual_threading_ = true; return this; }
+
   virtual void Run(State& state) = 0;
 
  protected:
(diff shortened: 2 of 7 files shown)


src/benchmark.cc
```diff
@@ -315,6 +343,16 @@ bool IsZero(double n) {
   return std::abs(n) < std::numeric_limits<double>::epsilon();
 }
 
+void MergeResults(State& st, ThreadTimer* timer, ThreadManager* manager)
+{
+  ThreadManager::Result& results = manager->results;
+  results.cpu_time_used += timer->cpu_time_used();
+  results.real_time_used = std::max(results.real_time_used, timer->real_time_used());
```
krzikalla, line 350: Note to myself: leftover of another fix. 
Here it still must be `results.real_time_used += timer->real_time_used();`

include/benchmark/benchmark.h
```diff
@@ -685,8 +688,21 @@ class State {
   void FinishKeepRunning();
   internal::ThreadTimer* timer_;
   internal::ThreadManager* manager_;
+  int num_thread_states_;
 
   friend struct internal::BenchmarkInstance;
+  friend class ThreadState;
```
dominichamon, line 694: avoid friends please. add a protected section with accessors if you need that.

include/benchmark/benchmark.h
```diff
@@ -685,8 +688,21 @@ class State {
...
+};
+
+class ThreadState : public State
+{
```
dominichamon, line 698: note we use google style so this (and other braces) should be on the previous line. you can use clang-format to make this easier.

src/benchmark.cc
```diff
@@ -212,6 +213,33 @@ void State::FinishKeepRunning() {
   manager_->StartStopBarrier();
 }
 
+ThreadState::ThreadState(State& s) :
+  State(s),
+  parent_(&s)
+{
+  CHECK(!s.started_) << "Don't create a ThreadState object after measurement has started";
+  timer_ = new internal::ThreadTimer(*timer_);
```
dominichamon, line 221: why is this allocated on the heap? If you're copying anyway, it could be a non-pointer member which helps avoid the explicit `delete` call in the destructor.

src/benchmark.cc
```diff
@@ -212,6 +213,33 @@ void State::FinishKeepRunning() {
...
+}
+
+ThreadState::~ThreadState()
+{
+  CHECK(error_occurred() || iterations() >= max_iterations)
+      << "Benchmark returned before ThreadState::KeepRunning() returned false!";
+  {
+    MutexLock l(manager_->GetBenchmarkMutex());
+    internal::MergeResults(*this, timer_, manager_);
+    assert(parent_->total_iterations_ == 0 || parent_->total_iterations_ == total_iterations_);
```
dominichamon, line 231: why `CHECK` above and `assert` here?

src/benchmark_runner.cc
```diff
@@ -117,17 +117,36 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
           ? internal::ThreadTimer::CreateProcessCpuTime()
           : internal::ThreadTimer::Create());
   State st = b->Run(iters, thread_id, &timer, manager);
-  CHECK(st.error_occurred() || st.iterations() >= st.max_iterations)
-      << "Benchmark returned before State::KeepRunning() returned false!";
+  assert(b->explicit_threading || b->threads == 1);
+  if (st.GetNumThreadStates() > 0)
+  {
+    CHECK((!b->explicit_threading) || b->manual_threading)
+        << "Benchmark " << b->name.str() << " run with managed threading. It must not create ThreadStates!";
```
dominichamon, line 124: i believe this line is too long. please check with clang-format.

src/benchmark_runner.cc
```diff
@@ -186,6 +206,7 @@ class BenchmarkRunner {
   const int repeats;
   const bool has_explicit_iteration_count;
 
+  const int num_managed_threads;  // number of managed threads, must be before pool
```
dominichamon, line 209: why does the order matter?

dominichamon (commented):
some style notes while i work through the ramifications.

can you provide some example output in the PR so i can see if there's any impact?
also we'll need some tests.

README.md
```diff
@@ -905,6 +905,60 @@ BENCHMARK(BM_test)->Range(8, 8<<10)->UseRealTime();
...
+      // Run the multithreaded test. 
+    }
+  }
+  // Teardown code here.
+}
+
+BENCHMARK(BM_MultiThreaded)->ManualThreading()->Threads(1)->Threads(2)->Threads(4);
+```
+If you use the `ThreadState` object and explicitly specify the number of threads, then you must 
+use `ManualThreading` and the number of created `ThreadState` objects must match the number of specified threads.
```
LebedevRI, line 944: Is this postcondition actually required? Can it be lifted?

LebedevRI (commented):
Thanks. This looks more acceptable.
I think this needs some tests.

src/benchmark_runner.cc
krzikalla: num_managed_threads is used to initialize pool, thus the logic, which initializes num_managed_threads is not duplicated.

src/benchmark.cc
krzikalla: As far as I understand, a failed CHECK is something a user (aka the benchmark programmer) did wrong (e.g. wrong API usage). A failed assert is something we (aka the framework programmers) did wrong.

src/benchmark.cc
krzikalla: This would require the ThreadTimer definition being a part of benchmark.h.

krzikalla:
> can you provide some example output in the PR so i can see if there's any impact?
What exactly do you want to see? 

> also we'll need some tests.
Can I expect OpenMP availability, so that we can test with OpenMP as the alternative threading API? (question goes also @LebedevRI )

README.md
krzikalla: Do you mean the postcondition, that number of ThreadStates == number of threads? I don't see a way to do this. 
Note, that this postcondition is only requried, if you explicitly specify the number of threads via `Threads()` and if you create at least one `ThreadState` object during the benchmark run. It would be odd, if in this case the numbers doesn't match.

src/benchmark.cc
dominichamon: rats.

fair point. if only we could get the API up to C++11 and use unique_ptr...

README.md
LebedevRI: > Do you mean the postcondition, that number of ThreadStates == number of threads?

Yes.

> I don't see a way to do this.
> Note, that this postcondition is only requried, if you explicitly specify the number of threads via `Threads()` and if you create at least one `ThreadState` object during the benchmark run. It would be odd, if in this case the numbers doesn't match.

In particular, as discussed (?) in previous PR, i would like to call `->Threads()`,
but disable `benchmark`-based threading. I.e. i'd like a way to use threads
as yet another parameter, without it actually creating all those threads,
and without dividing the measurements by the "thread count".

dominichamon:
> > can you provide some example output in the PR so i can see if there's any impact?
> 
> What exactly do you want to see?

what the output (tabular is fine) looks like when using openmp.

> 
> > also we'll need some tests.
> 
> Can I expect OpenMP availability, so that we can test with OpenMP as the alternative threading API? (question goes also @LebedevRI )

i'm pretty sure you could check for openmp availability before enabling those tests... and add them to the buildbots.

LebedevRI:
> > > can you provide some example output in the PR so i can see if there's any impact?
> > 
> > 
> > What exactly do you want to see?
> 
> what the output (tabular is fine) looks like when using openmp.
> 
> > > also we'll need some tests.
> > 
> > 
> > Can I expect OpenMP availability, so that we can test with OpenMP as the alternative threading API? (question goes also @LebedevRI )
> 
> i'm pretty sure you could check for openmp availability before enabling those tests... and add them to the buildbots.

While that should work, i would personally maybe suggest just sticking with using `std::thread` for those tests.
It won't really matter what is used in tests.

dominichamon:
> > > > can you provide some example output in the PR so i can see if there's any impact?
> > > 
> > > 
> > > What exactly do you want to see?
> > 
> > 
> > what the output (tabular is fine) looks like when using openmp.
> > > > also we'll need some tests.
> > > 
> > > 
> > > Can I expect OpenMP availability, so that we can test with OpenMP as the alternative threading API? (question goes also @LebedevRI )
> > 
> > 
> > i'm pretty sure you could check for openmp availability before enabling those tests... and add them to the buildbots.
> 
> While that should work, i would personally maybe suggest just sticking with using `std::thread` for those tests.
> It won't really matter what is used in tests.

that's true. as long as the api is being exercised and the expectations met, it shouldn't matter.

Issue #1028, opened 2020-08-25
[FR] Add ability to create benchmark arguments via cartesian product


chriswasser:
**Is your feature request related to a problem? Please describe.**
Currently, it is only possible to reduce long lists of `->Args({...})` calls via `->Ranges({...})`. While I agree that this interface offers a concise syntax for generating a large number of arguments, I often find myself needing more fine-grained control over the set of generated inputs than `->RangeMultiplier(...)` can provide. This may be due to the requirement for specific values or different range multipliers for different arguments.

**Describe the solution you'd like**
I would like to propose an extension to the public `Benchmark` interface by adding a function `->CartesianProduct({...})` which generates the cartesian product of all the supplied vectors. This would enable fine-grained control over the constructed benchmark arguments without sacrificing on the compact syntax. As a small example:
```cpp
BENCHMARK(BM_SetInsert)
    ->CartesianProduct({{1, 2, 3}, {10}, {6, 12, 24, 36}})
// would be equivalent to
BENCHMARK(BM_SetInsert)
    ->Args({1, 10, 6})
    ->Args({1, 10, 12})
    ->Args({1, 10, 24})
    ->Args({1, 10, 36})
    ->Args({2, 10, 6})
    ->Args({2, 10, 12})
    ->Args({2, 10, 24})
    ->Args({2, 10, 36})
    ->Args({3, 10, 6})
    ->Args({3, 10, 12})
    ->Args({3, 10, 24})
    ->Args({3, 10, 36})
```

**Describe alternatives you've considered**
At the moment, I have a large number of individual `->Args({...})` calls throughout my own benchmarking code, which is quite repetitive and error prone when copy and pasting with slight modifications. Other than that, I am not aware of any other way to achieve the desired functionality.

**Additional context**
I have already implemented this feature on my fork of the project and would like to merge it into the main repository if there are no objections against the proposition.

dominichamon:
this seems utterly reasonable. I think i'd nitpick the name, if you don't mind. While it is a Cartesian product of vectors, that doesn't explain to the user what it's doing. Perhaps something like `ArgsCombination` or `ArgsProduct` ? i'm terrible at naming, but something that indicates that you're setting the args would be better i think.

chriswasser:
[...]

PR #1029, opened 2020-08-25
Add CartesianProduct with associated test

chriswasser:
* Use CartesianProduct in Ranges to avoid code duplication
* Add new cartesian_product_test to CMakeLists.txt
* Update AUTHORS & CONTRIBUTORS
* Addresses #1028

Since this PR adds functionality to the public interface of the library, it might make sense to also add an example for `CartesianProduct` to the README.

After I wrote my version of `CartesianProduct`, I realized that `Ranges` already implemented this kind of functionality. As I preferred the usage of `std::accumulate` and the `do {...} while(...)` to the previous implementation, I just renamed variables to their old names where sensible and kept the rest of my reimplementation. I am not sure about the correct placement of `args` (`tmp` in the old `Ranges` code) but it seems to save some work when construction the vector / reserving the required memory once and only clearing the contents of it after each iterations.

Open to criticism and feedback. Thanks for the review in advance.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ Alex Steele <steeleal123@gmail.com>
 Andriy Berestovskyy <berestovskyy@gmail.com>
 Arne Beer <arne@twobeer.de>
 Carto
+Christian Wassermann <christian_wassermann@web.de>
 Christopher Seymour <chris.j.seymour@hotmail.com>
 Colin Braley <braley.colin@gmail.com>
 Daniel Harvey <danielharvey458@gmail.com>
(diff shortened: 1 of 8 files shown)


dominichamon:
LGTM modulo naming per FR. thanks for the test!

Also please do add some documentation to the readme. Something in [this section](https://github.com/google/benchmark#passing-arguments).

chriswasser:
Renaming applied and new example added to README. In my opinion, `ArgsProduct` offers a nice interface which lies directly between the automatic argument generation by `Ranges` and the high customizability offered by `Apply`.

dominichamon:
just waiting for CI

dominichamon:
thanks so much for this!

chriswasser:
That went smoothly. Great pleasure contributing and thanks for the quick responses 👍

PR #1030, opened 2020-08-25
Fix realtime measurement for multi-threaded benchmarks

krzikalla:
I tried to quick-fix the wall-clock time (real time) as discussed in PR #946. This is a prerequiste for adding some tests in PR #1027.
However, the changes now seriously overlap with PR #946 (esp. the change in line https://github.com/google/benchmark/blob/5c25ad3acb24802943935783aadd5e035e6edbf0/src/benchmark_runner.cc#L125).

I realized, that a lot of design questions must be answered before we can go on with #946 and this one.

1. Do we allow a different iteration count across threads. (IMHO no and I'd check this) If yes: what results are reported then? What are their meanings?
2. How is manual timing handled in multi-threaded benchmarks? IMHO the current implementation does not match the comment at https://github.com/google/benchmark/blob/5c25ad3acb24802943935783aadd5e035e6edbf0/include/benchmark/benchmark.h#L553-L554 IMHO and according to the comment it should be handled similar to real time.
3. Which values are written in which mode and what is their meaning? (default/real time/manual time * default/ProcessCPUTime)

@koct9i : Maybe you can use my tests as a template for your PR.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -122,9 +122,9 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
   {
     MutexLock l(manager->GetBenchmarkMutex());
     internal::ThreadManager::Result& results = manager->results;
-    results.iterations += st.iterations();
+    results.iterations = st.iterations();
     results.cpu_time_used += timer.cpu_time_used();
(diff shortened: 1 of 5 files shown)


LebedevRI:
@krzikalla thank you for looking into this.

Issue #1031, opened 2020-08-27
[BUG] undefined references to lround, log, pow: benchmark uses math.h/cmath but doesn't specify the `-lm` linker flag on FreeBSD/OpenBSD when using Bazel


yesudeep:
**Describe the bug**

```shell
❯ bazel build //test/...
WARNING: /usr/home/yesudeep/.cache/bazel/_bazel_yesudeep/ffbd632d168ee0cddab7ac9029ce644e/external/remote_java_tools_linux/BUILD:671:11: in hdrs attribute of cc_library rule @remote_java_tools_linux//:combiners: Artifact 'external/remote_java_tools_linux/java_tools/src/tools/singlejar/zip_headers.h' is duplicated (through '@remote_java_tools_linux//:transient_bytes' and '@remote_java_tools_linux//:zip_headers'). Since this rule was created by the macro 'cc_library', the error might have been caused by the macro implementation
INFO: Analyzed 30 targets (0 packages loaded, 0 targets configured).
INFO: Found 30 targets...
ERROR: /usr/home/yesudeep/code/yesudeep/benchmark/test/BUILD:40:12: Linking of rule '//test:args_product_test' failed (Exit 1) clang failed: error executing command /usr/bin/clang -o bazel-out/freebsd-fastbuild/bin/test/args_product_test '-Wl,-rpath,$ORIGIN/../_solib_freebsd/' -Lbazel-out/freebsd-fastbuild/bin/_solib_freebsd ... (remaining 10 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
ld: error: bazel-out/freebsd-fastbuild/bin/_solib_freebsd/liblibbenchmark.so: undefined reference to lround
ld: error: bazel-out/freebsd-fastbuild/bin/_solib_freebsd/liblibbenchmark.so: undefined reference to log
ld: error: bazel-out/freebsd-fastbuild/bin/_solib_freebsd/liblibbenchmark.so: undefined reference to pow
clang: error: linker command failed with exit code 1 (use -v to see invocation)
INFO: Elapsed time: 0.437s, Critical Path: 0.25s
INFO: 1 process: 1 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
```

**System**
Which OS, compiler, and compiler version are you using:
  - OS: FreeBSD 12.1-STABLE (GhostBSD 20.08.20 XFCE)
  - Compiler and version: clang version 12.0.0

**To reproduce**
[...]

PR #1032, opened 2020-08-27
Adds -lm linker flag for FreeBSD/OpenBSD and uses github.com/bazelbuil…

yesudeep:
…d/platforms for platform detection.

Addresses: https://github.com/google/benchmark/issues/1031

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@bazel_skylib//lib:selects.bzl", "selects")
 
 licenses(["notice"])
 
@@ -36,13 +37,9 @@ cc_library(
         exclude = ["src/benchmark_main.cc"],
     ),
     hdrs = ["include/benchmark/benchmark.h"],
-    linkopts = select({
+    linkopts = selects.with_or({
         ":windows": ["-DEFAULTLIB:shlwapi.lib"],
-        ":freebsd": [
-            "-pthread",
-            "-lm",
-        ],
-        ":openbsd": [
+        (":freebsd", ":openbsd"): [
             "-pthread",
             "-lm",
         ],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,5 +1,4 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
-load("@bazel_skylib//lib:selects.bzl", "selects")
 
 licenses(["notice"])
 
@@ -27,6 +26,11 @@ config_setting(
     visibility = [":__subpackages__"],
 )
 
+BSD_LINKOPTS = [
+    "-pthread",
+    "-lm",
+]
+
 cc_library(
     name = "benchmark",
     srcs = glob(
@@ -37,12 +41,10 @@ cc_library(
         exclude = ["src/benchmark_main.cc"],
     ),
     hdrs = ["include/benchmark/benchmark.h"],
-    linkopts = selects.with_or({
+    linkopts = select({
         ":windows": ["-DEFAULTLIB:shlwapi.lib"],
-        (":freebsd", ":openbsd"): [
-            "-pthread",
-            "-lm",
-        ],
+        ":freebsd": BSD_LINKOPTS,
+        ":openbsd": BSD_LINKOPTS,
         "//conditions:default": ["-pthread"],
     }),
     strip_include_prefix = "include",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,14 +1,30 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
 licenses(["notice"])
 
 config_setting(
-    name = "windows",
-    values = {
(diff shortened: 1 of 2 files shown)


yesudeep:
This also addresses a broken build for google/cctz on FreeBSD.

```shell
ERROR: /usr/home/yesudeep/code/yesudeep/cctz/BUILD:120:8: Linking of rule '//:cctz_benchmark' failed (Exit 1) clang failed: error executing command /usr/bin/clang -o bazel-out/freebsd-fastbuild/bin/cctz_benchmark bazel-out/freebsd-fastbuild/bin/_objs/cctz_benchmark/cctz_benchmark.pic.o bazel-out/freebsd-fastbuild/bin/libtime_zone.a ... (remaining 8 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
ld: error: undefined symbol: ceilf
>>> referenced by time_zone_impl.cc
>>>               time_zone_impl.pic.o:(ceil(float)) in archive bazel-out/freebsd-fastbuild/bin/libtime_zone.a

ld: error: undefined symbol: lround
>>> referenced by benchmark_runner.cc
>>>               benchmark_runner.pic.o:(benchmark::internal::(anonymous namespace)::BenchmarkRunner::PredictNumItersNeeded(benchmark::internal::(anonymous namespace)::BenchmarkRunner::IterationResults const&) const) in archive bazel-out/freebsd-fastbuild/bin/external/com_github_google_benchmark/libbenchmark.a
>>> referenced by json_reporter.cc
>>>               json_reporter.pic.o:(benchmark::(anonymous namespace)::RoundDouble(double)) in archive bazel-out/freebsd-fastbuild/bin/external/com_github_google_benchmark/libbenchmark.a

ld: error: undefined symbol: log
>>> referenced by complexity.cc
>>>               complexity.pic.o:(benchmark::FittingCurve(benchmark::BigO)::$_3::operator()(unsigned long) const) in archive bazel-out/freebsd-fastbuild/bin/external/com_github_google_benchmark/libbenchmark.a
>>> referenced by complexity.cc
>>>               complexity.pic.o:(benchmark::FittingCurve(benchmark::BigO)::$_4::operator()(unsigned long) const) in archive bazel-out/freebsd-fastbuild/bin/external/com_github_google_benchmark/libbenchmark.a

ld: error: undefined symbol: pow
>>> referenced by complexity.cc
>>>               complexity.pic.o:(std::__1::_MetaBase<(std::is_arithmetic<double>::value) && (std::is_arithmetic<int>::value)>::_EnableIfImpl<std::__1::__promote<double, int, void> >::type pow<double, int>(double, int)) in archive bazel-out/freebsd-fastbuild/bin/external/com_github_google_benchmark/libbenchmark.a
>>> referenced by complexity.cc
>>>               complexity.pic.o:(std::__1::_MetaBase<(std::is_arithmetic<unsigned long>::value) && (std::is_arithmetic<int>::value)>::_EnableIfImpl<std::__1::__promote<unsigned long, int, void> >::type pow<unsigned long, int>(unsigned long, int)) in archive bazel-out/freebsd-fastbuild/bin/external/com_github_google_benchmark/libbenchmark.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
INFO: Elapsed time: 0.261s, Critical Path: 0.10s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

BUILD.bazel
```diff
@@ -22,6 +38,14 @@ cc_library(
     hdrs = ["include/benchmark/benchmark.h"],
     linkopts = select({
         ":windows": ["-DEFAULTLIB:shlwapi.lib"],
+        ":freebsd": [
```
dominichamon, line 41: if we `load("@bazel_skylib//lib:selects.bzl", "selects")` then we can replace this with:

```
linkopts = selects.with_or({
    ":windows": ["-DEFAULTLIB:shlwapi.lib"],
    (":freebsd", ":openbsd"): ["-pthread", "-lm"],
})
```

which i think is going to be easier to read/maintain. assuming freebsd/openbsd don't diverge.

as an alternative, you can define:

```
selects.config_setting_group(
    name = "bsd",
    match_any = [":freebsd", ":openbsd"],
)
```

and then just use ":bsd" as the key here.

up to you :)

yesudeep: That would work within the repository, but then it requires every dependent repository to install @bazel_skylib too, I guess. It turns out that Bazel doesn't load `WORKSPACE` dependencies transitively (?), and the "@platforms" package appears to load implicitly. I'm wondering whether these Bazel utility repositories should be built-ins or be loaded by default.

I've done something simpler: defined a variable named `BSD_LINKOPTS` and shared it with both the ":freebsd" and ":openbsd" keys. Would that work for you?

yesudeep: I've filed an [issue](https://github.com/bazelbuild/bazel/issues/12014) with the Bazel team. I like the approach you've suggested, but adding a dependency here would complicate things for users, which I'm not so sure they'd like.

dominichamon: that's a great compromise. thanks!

Issue #672, opened 2018-09-07
`make` fails on 18.04 with gcc 7.3.0 + LTO


tophj-ibm:
Hi, I'm hitting some linker warning/errors when going through the `Building` steps on 18.04, I've tried with latest release, master, and v2 branches.

This looks to me like some sort of conflict with the included c++ files and then benchmark code but I'm not sure, any help would be appreciated. I've attached the full gist below, here is a snippet.

```
$  cmake .. -DCMAKE_BUILD_TYPE=RELEASE -DBENCHMARK_ENABLE_LTO=true -DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON
...
-- Build files have been written to: /benchmark/build

$ make
...
[ 33%] Linking CXX executable cxx03_test
/usr/include/c++/7/bits/stl_tree.h:677:16: note: type name 'std::_Rb_tree<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, benchmark::Counter>, std::_Select1st<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, benchmark::Counter> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, benchmark::Counter> > >::_Rb_tree_impl<std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, false>' should match type name 'std::_Rb_tree<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, benchmark::Counter>, std::_Select1st<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, benchmark::Counter> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, benchmark::Counter> > >::_Rb_tree_impl<std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, true>'
         struct _Rb_tree_impl
                ^
[...]

PR #1034, opened 2020-08-30
Set default option for "LTO" to "True"

### LINKED ISSUE #1022: [FR] LTO should default to true
BENCHMARK_ENABLE_LTO=true is recommended by the README.md. It should likely be the default if we think it is that important.

ahmedkrmn:
Closes #1022 

Set the default option for `BENCHMARK_ENABLE_LTO` to `True`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ project (benchmark CXX)
 
 option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmark library." ON)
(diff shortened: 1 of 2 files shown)


dominichamon:
looks like this has some issues in release builds.

LebedevRI:
> looks like this has some issues in release builds.

Ah yes, that's #672 i guess.
I'm not sure what is going on there.
It appears to originate from `cxx03_test`, but that doesn't make sense,
i may be misremembering, but i think i might have asked in #gcc 
whether such LTO with `-std=c++*` mismatch should work and IIRC the answer was yes.

dominichamon:
I'm going to close this but I'll leave the bug open it cause anyone wants to try again

PR #1035, opened 2020-08-30
Revert previous linker additions for FreeBSD as the problem is Bazel …

yesudeep:
…using `/usr/bin/clang` instead of `/usr/bin/clang++` to link C++ code.

The correct fix for this and several other libraries would be getting Bazel to use `/usr/bin/clang++` to link C++ code first
and then fix any remaining linker errors.

See: https://github.com/bazelbuild/bazel/issues/12023


dominichamon:
Thanks for following up and cleaning up :)

PR #1036, opened 2020-09-03
benchmark.h: make Run() implementation override

khancyr:
This prevent to trigger an error with recent GCC (gcc 9.x) with -Werror=suggest-override flag

see
````
gbenchmark/include/benchmark/benchmark.h:1005:16: error: ‘virtual void benchmark::internal::FunctionBenchmark::Run(benchmark::State&)’ can be marked override [-Werror=suggest-override]
 1005 |   virtual void Run(State& st);
      |                ^~~
compilation terminated due to -Wfatal-errors.
cc1plus: some warnings being treated as errors
````

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1008,7 +1008,7 @@ class FunctionBenchmark : public Benchmark {
   FunctionBenchmark(const char* name, Function* func)
       : Benchmark(name), func_(func) {}
 
-  virtual void Run(State& st);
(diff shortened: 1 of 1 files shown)


LebedevRI:
As bots correctly note, `override` is C++11, while the header is supposed to be C++03.
`-Wsuggest-*` aren't a warnings, really.

khancyr:
@LebedevRI haaa I miss that the lib should be usable in C++03 ... So it will close this as it won't pass... my bad

PR #1037, opened 2020-09-04
Bind more State methods/attributes to Python

AntoinePrv:
Hi,

The following PR binds more of the `State` methods in Python.
  - Bind `Counter::Flags` and bind `operator|`
  - Bind `Counter::OneK`
  - Bind `Counter`
  - Make Counter implicitly constructible from `float`, `int`
  - Bind some more `State` methods
  - Bind `State::counters` making it possible to use `state.counters["foo"] = 42` and `state.counters["foo"] = Counter(42, Counter.kIsRate)`...

However, there remains a number of questions:
  - [ ] Explicit the naming convention for function.methods? Python typically uses `snake_case`, which seems to have been done for methods but not functions
  - [ ] Define a test strategy/framework. This is hard because `State` are not constructible without registering a benchmark and running the main function.

### DIFF
--- a/bindings/python/google_benchmark/__init__.py
+++ b/bindings/python/google_benchmark/__init__.py
@@ -29,10 +29,12 @@ Example usage:
 
 from absl import app
 from google_benchmark import _benchmark
+from google_benchmark._benchmark import Counter
 
 __all__ = [
     "register",
     "main",
+    "Counter",
 ]
 
 __version__ = "0.1.0"
--- a/bindings/python/google_benchmark/__init__.py
+++ b/bindings/python/google_benchmark/__init__.py
@@ -41,27 +41,27 @@ __version__ = "0.1.0"
 
 
 def register(f=None, *, name=None):
-  if f is None:
-    return lambda f: register(f, name=name)
-  if name is None:
-    name = f.__name__
-  _benchmark.RegisterBenchmark(name, f)
-  return f
+    if f is None:
+        return lambda f: register(f, name=name)
+    if name is None:
+        name = f.__name__
+    _benchmark.RegisterBenchmark(name, f)
+    return f
 
 
 def _flags_parser(argv):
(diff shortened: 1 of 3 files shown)


AntoinePrv:
@googlebot I signed it!

dominichamon:
It's not really a testing strategy, but having these new features in `example.py` can help document them (and running that will test them in a way... at least ensure the syntax is correct everywhere).

AntoinePrv:
Thanks, I will add these in the example file. Is the file run in CI? If not it could be a good addition, but I'm not sure I'm the best person to contribute for this.

AntoinePrv:
I added some more example to the files.
Some methods are not added to the example because they are meaningless without #1038.

I also noticed that `state.range()` segfaults when there is no range. In C++ there is an `assert` but this does not work in Python.
To avoid this, we could create and expose a `State.range_size()`, and add a runtime check that the range is not empty.

dominichamon:
> Thanks, I will add these in the example file. Is the file run in CI? If not it could be a good addition, but I'm not sure I'm the best person to contribute for this.

it isn't. it likely should be :)

dominichamon:
Thanks for the extra examples, and the formatting changes. It looks like it now follows https://google.github.io/styleguide/pyguide.html#s3.4-indentation ?

AntoinePrv:
I ran [Black](https://github.com/psf/black) because the former formatting seemed far from Python convention.
In hindsight, it seems to fit most of the Google style guide. Perhaps the line length default to 88 is a bit over Google's.

dominichamon:
i added a pylint (fatal) check to our CI but i had to reformat some things. you have some conflicts now just from the formatting but i'll see if i can take care of that as it's my fault.

dominichamon:
LGTM. travis looks unrelated (not sure why `ctest` is expanding to `bazel test` instead of `make test`).

PR #1039, opened 2020-09-08
Create pylint.yml

dominichamon:


dominichamon:
with the latest version, this only fails on "fatal" issues. we should add `--error-fail` to pylint-exit to fail on errors too but we're not error clean.

PR #1040, opened 2020-09-09
Bind benchmark builder to Python

### LINKED ISSUE #1038: [FR][Python] Bind Benchmark methods / builder pattern
Missing features in Python, such as the ability to use`->MeasureProcessCPUTime()`, `->UseRealTime()`, `->Unit(benchmark::kMillisecond)`...

We currently have 
```python
@register(name="Some benchmark)
def sum_million(state):
  while state:
    sum(range(1_000_000))
```
I suggest extending the API in a similar fashion to the C++ one. So something like
```python
@register(name="Some benchmark)
    .measure_process_time()
    .use_real_time()
    .unit(benchmark.kMilisecond)
def sum_million(state):
  while state:
    sum(range(1_000_000))
```

Alternatively, we could use Python named parameters inside the `register` function
```python
@register(
    name="Some benchmark)
    measure_process_time=True,
    use_real_time=True,
    unit=benchmark.kMilisecond,
)
def sum_million(state):
  while state:
    sum(range(1_000_000))
```
One downside is that in C++, the extra arguments to `BENCHMARK_REGISTER` are passed to the benchmark function.
Doing so here may prevent us from doing this in the future.

**Additional context**

AntoinePrv:
The following PR Close #1038
 - Bind `TimeUnit` and `BigO`
 - Bind `internal::Benchmark`
 - Modify `RegisterBenchmark` to return the benchmark object
 - Bind builder methods from `internal::Benchmark`
 - Add a Python decorator `option` to use the `Benchmark` options.

Example
```python
@benchmark.register
@benchmark.option.range_multiplier(2)
@benchmark.option.range(1 << 10, 1 << 18)
@benchmark.option.complexity(benchmark.oN)
def computing_complexity(state):
    while state:
        sum(range(state.range(0)))
    state.complexity_n = state.range(0)
```

Todo
 - [x] This builds on #1037 so preferably merge the other one first
 - [x] ~the multi-threaded benchmark is currently segfaulting and needs to be fixed~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,29 +12,32 @@ from setuptools.command import build_ext
 HERE = os.path.dirname(os.path.abspath(__file__))
 
 
-IS_WINDOWS = sys.platform.startswith('win')
+IS_WINDOWS = sys.platform.startswith("win")
 
 
 def _get_version():
     """Parse the version string from __init__.py."""
-    with open(os.path.join(
-            HERE, 'bindings', 'python', 'google_benchmark', '__init__.py')) as init_file:
+    with open(
+        os.path.join(HERE, "bindings", "python", "google_benchmark", "__init__.py")
+    ) as init_file:
         try:
             version_line = next(
-                line for line in init_file if line.startswith('__version__'))
+                line for line in init_file if line.startswith("__version__")
+            )
         except StopIteration:
-            raise ValueError('__version__ not defined in __init__.py')
+            raise ValueError("__version__ not defined in __init__.py")
         else:
             namespace = {}
             exec(version_line, namespace)  # pylint: disable=exec-used
-            return namespace['__version__']
+            return namespace["__version__"]
 
 
 def _parse_requirements(path):
     with open(os.path.join(HERE, path)) as requirements:
         return [
-            line.rstrip() for line in requirements
-            if not (line.isspace() or line.startswith('#'))
+            line.rstrip()
+            for line in requirements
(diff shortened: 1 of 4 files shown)


AntoinePrv:
Ok, so indeed I cannot run Python in multiple thread without acquiring the GIL.
I tried to register the threads and acquire the GIL, but it now deadlocks.
Probably due to

> In a multithreaded test (benchmark invoked by multiple threads simultaneously), it is guaranteed that none of the threads will start until all have reached the start of the benchmark loop

Anyways, seems to be out of scope for this issue

dominichamon:
We should probably update the `__version__` in __init__.py too with this change.

AntoinePrv:
I rebased on `master` and bumped version to `0.2.0`. I also fixed an indentation bug that arose in the `setup.py`.

setup.py
```diff
@@ -58,71 +62,81 @@ def run(self):
...
         ]
 
         if IS_WINDOWS:
             # Link with python*.lib.
             for library_dir in self.library_dirs:
-                bazel_argv.append('--linkopt=/LIBPATH:' + library_dir)
+                bazel_argv.append("--linkopt=/LIBPATH:" + library_dir)
 
-            self.spawn(bazel_argv)
+        self.spawn(bazel_argv)
```
dominichamon, line 93: i fixed this same issue in #1041  :)

well spotted!

dominichamon:
LGTM but i'll wait a while in case there's any issues raised.

Issue #1053, opened 2020-10-06
[FR] Please allow to use defines as benchmark names


emmenlau:
**Is your feature request related to a problem? Please describe.**
We have a number of benchmarks that compare different implementations of an algorithm that has a standard interface. Therefore the benchmark code between these implementations is identical. Instead of duplicating the benchmark, we use defines to select the implementation and then include the actual benchmark code. This works very well, with one problem: The name of a benchmark in the macro can not itself be a define. So the following does not work:
```
#define MY_BECHMARK_NAME SomeBenchmark
BENCHMARK_DEFINE_F(MY_BECHMARK_NAME, TestSomething)(benchmark::State& aState) {
    // code goes here...
}
```

**Describe the solution you'd like**
I'm under the impression that the above example code should and could work. The problem seems to come from the macro stringification in for example `BENCHMARK_DEFINE_F`, where a separate macro invocation would be required to stringify the names. See for example the bottom example in https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

**Describe alternatives you've considered**
Instead of using a define for the benchmark name, we could use functions that are called. These functions can be shared between implementations. However the actual benchmark macros will need to be repeated, and they constitute already a significant number of lines. We want to avoid diverging implementations, and maintaining all the macro invocations for two or three otherwise identical benchmarks becomes tedious and error prone.

pleroy:
It seems to me that:
```
using MyBechmarkName = SomeBenchmark;
BENCHMARK_DEFINE_F(MyBechmarkName, TestSomething)(benchmark::State& aState) {...}
```
would do what you want without messing with macros.

emmenlau:
Dear @pleroy thanks for the very quick response! I tried your suggestion but I'm not sure that works. I get error:
```
SomeBenchmark.cc:26:62: error: ‘SomeBenchmark’ does not name a type
 using MyBechmarkName = SomeBenchmark;
                        ^~~~~~~~~~~~~~~~~~~~~~~~~
```

emmenlau:
Ok then I did it wrong! Now I tried with a fixture class name. But then it seems the macro binding is stronger than the using-statement? I get error:
```
SomeBenchmark.cc:44:20: error: ‘MyBechmarkName_TestSomething_Benchmark’ has not been declared
 BENCHMARK_DEFINE_F(MyBechmarkName, TestSomething)(benchmark::State& aState) {
                    ^
```

pleroy:
[...]

PR #1054, opened 2020-10-06
Added support for macro expansion in benchmark names

emmenlau:
This PR is a proof-of-concept, not a full implementation. It adds partial support for macro expansion in benchmark names, as outlined in issue #1053.

If this PR should be acceptable, it should be further extended to support more macros, and possible should be applied to internal class names too.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1106,6 +1106,9 @@ class Fixture : public internal::Benchmark {
   BENCHMARK_PRIVATE_CONCAT(_benchmark_, BENCHMARK_PRIVATE_UNIQUE_ID, n)
 #define BENCHMARK_PRIVATE_CONCAT(a, b, c) BENCHMARK_PRIVATE_CONCAT2(a, b, c)
 #define BENCHMARK_PRIVATE_CONCAT2(a, b, c) a##b##c
+// Helper for concatenation with macro name expansion
+#define BENCHMARK_PRIVATE_CONCAT_NAME(BaseClass, Method) \
+    BaseClass##_##Method##_Benchmark
 
 #define BENCHMARK_PRIVATE_DECLARE(n)                                 \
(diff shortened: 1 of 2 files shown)


LebedevRI:
Needs a test

pleroy:
See my comments in #1053, the year is 2020, I don't see why we would get out of our way for supporting type aliases written using macros.

LebedevRI (changes requested):
Needs test coverage

LebedevRI:
I think the seems rather self-contained, but i'd like to see a test to actually confirm what isn't working.

emmenlau:
Now I understand. I've added commit e2f9d98 `Adding test for defined names in test fixtures` that modifies `test/fixture_test.cc` to use a "defined" name. This is the same behavior as we would like to use, to build various benchmarks based on the same benchmark definition, albeit with different names.

The test compiles and works fine as long as it is used in combination with 54308f9 `include/benchmark/benchmark.h: Added support for macro expansion in benchmark names`. However when you go back one commit, removing the support for defined benchmark names, the build will fail.

Is that helpful? We can wrap all of this in a more suitable way, I'm just trying to illustrate the point.

emmenlau:
This issue is still in state "changes requested". Are the submitted changes useful or is there something else I should do to move this forward? Thanks for your consideration!

LebedevRI (approved):
This seems to have slipped through the cracks, sorry about that.
It does seem like that testcase doesn't work without the patch - https://godbolt.org/z/64YWKo.
The change itself looks like a clean improvement to me - i wouldn't imagine
the reverse change be acceptable, one dropping `BENCHMARK_PRIVATE_CONCAT_NAME`.
So this LGTM.

@dominichamon ?

dominichamon:
@pleroy until we can support building this library as if it's 2020 (including the macro API) i think it's reasonable to support enhancements, especially if they're relatively lightweight.

LGTM, sorry for the delay in reviewing.

PR #1060, opened 2020-10-15
src/benchmark_register.h: add missing <limits> inclusion

trofi:
Noticed missing header when was building llvm with gcc-11:

```
llvm-project/llvm/utils/benchmark/src/benchmark_register.h:17:30:
  error: 'numeric_limits' is not a member of 'std'
   17 |   static const T kmax = std::numeric_limits<T>::max();
      |                              ^~~~~~~~~~~~~~
```

### DIFF
--- a/src/benchmark_register.h
+++ b/src/benchmark_register.h
@@ -1,6 +1,7 @@
 #ifndef BENCHMARK_REGISTER_H
 #define BENCHMARK_REGISTER_H
 
+#include <limits>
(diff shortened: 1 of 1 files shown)


dominichamon:
looks like macos failure is unrelated... upstream gmock issues.

LebedevRI:
This is obviously good, i'm gonna merge as-is.
Thanks.

Issue #1062, opened 2020-10-21
[FR] z/OS XL compiler inline assembly support in CycleTimer


fanbo-meng:
**Is your feature request related to a problem? Please describe.**

Build is failing with z/OS XL compiler with following error message because the inline assembly code in cycleclock.h under s390 arch. z/OS XL compiler uses HLASM syntax. 
```
INFORMATIONAL: Inline assembly statements caused HLASM RC=8 with the following messages:
                                       8 stck     2256(4)
 ASMA141E Bad character in operation code - 2256(4)
 ASMA435I Record 8 in sysinfo.cc(__asm) on volume: ASMA96
                                      12 stck     2256(4)
 ASMA141E Bad character in operation code - 2256(4)
 ASMA435I Record 12 in sysinfo.cc(__asm) on volume: ASMA96
 ERROR: Inline assembly statements caused HLASM RC=8. Compilation terminated.
```
**Describe the solution you'd like**

Need to fix the inline assembly code and define new macros for the z/OS operating system and XL compiler in internal_macros.h and use them to distinguish different OS and compiler under s390 architecture. I will deliver a patch for this issue.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

dominichamon:
all yours. i don't have the means to test that combination but hopefully it'll be a relatively straightforward change.

PR #1063, opened 2020-10-21
z/OS XL compiler inline assembly support in CycleTimer

fanbo-meng:
Adding correct inline assembly code for z/OS XL compiler. Defining new macros for the z/OS operating system and XL compiler and use them to distinguish different OS and compiler under s390 architecture. This fixes https://github.com/google/benchmark/issues/1062.

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -170,7 +170,12 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
 #elif defined(__s390__)  // Covers both s390 and s390x.
   // Return the CPU clock.
   uint64_t tsc;
+#if defined(BENCHMARK_OS_ZOS) && defined(COMPILER_IBMXL)
+  // z/OS XL compiler HLASM syntax.
+  asm(" stck %0" : "=m"(tsc) : : "cc");
+#else
   asm("stck %0" : "=Q"(tsc) : : "cc");
+#endif
   return tsc;
 #elif defined(__riscv) // RISC-V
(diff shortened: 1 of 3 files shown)


src/cycleclock.h
```diff
@@ -170,7 +170,12 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
 #elif defined(__s390__)  // Covers both s390 and s390x.
   // Return the CPU clock.
   uint64_t tsc;
+#if defined(BENCHMARK_OS_ZOS) && defined(COMPILER_IBMXL)
+  // z/OS XL compiler HLASM syntax.
+  asm(" stck %0" : "=m"(tsc) : : "cc");
```
dominichamon, line 175: is the whitespace before stck important? other than that, the only difference is the register constraint as far as i can see. i don't suppose we can replace `Q` with `m` in the non-ZOS branch (still s390)?

fanbo-meng: Yes the whitespace before instruction is intentional as it is part of the inline HLASM syntax and will otherwise cause the compilation error as described in the issue. In terms of constraint I'm using `m` because `Q` isn't supported by the XL compiler on z/OS. I can't comment on whether `Q` can be replaced by `m` on non z/OS branches as I'm only developing on z/OS, so I'm not sure if there will be side effects if we replace `Q` with `m` in the original code below.

dominichamon: ok, sounds good. i can't test the other branch but maybe we can replace the whole thing with your new version longer term.

dominichamon (approved):

Issue #1066, opened 2020-10-28
[FR] z/OS does not support nanosleep and needs a workaround


abhina-sree:
**Is your feature request related to a problem? Please describe.**

The build is failing on z/OS because there is no support for nanosleep() which is called in src/sleep.cc.

**Describe the solution you'd like**

I would like to provide a workaround just for z/OS which uses sleep() and usleep().

**Describe alternatives you've considered**
 N/A

**Additional context**
N/A

abhina-sree:
This PR is merged in.

PR #1067, opened 2020-10-28
Nanosleep workaround for z/OS in sleep.cc

abhina-sree:
z/OS does not support nanosleep. Instead, we call sleep() to sleep for the full seconds and usleep() to sleep for the remaining microseconds. This should fix this [issue](https://github.com/google/benchmark/issues/1066)

### DIFF
--- a/src/sleep.cc
+++ b/src/sleep.cc
@@ -42,8 +42,8 @@ void SleepForMicroseconds(int microseconds) {
   // sleep for the remaining microseconds because usleep() will fail if its
   // argument is greater than 1000000.
   div_t sleepTime = div(microseconds, kNumMicrosPerSecond);
-  unsigned int seconds = sleepTime.quot;
-  while (seconds > 0)
+  int seconds = sleepTime.quot;
+  while (seconds != 0)
     seconds = sleep(seconds);
   while (usleep(sleepTime.rem) == -1 && errno == EINTR)
     ;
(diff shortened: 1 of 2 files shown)


src/sleep.cc
```diff
@@ -33,11 +37,23 @@ void SleepForSeconds(double seconds) {
 }
 #else   // BENCHMARK_OS_WINDOWS
 void SleepForMicroseconds(int microseconds) {
+#ifdef BENCHMARK_OS_ZOS
+  // z/OS does not support nanosleep. Instead call sleep() and then usleep() to
+  // sleep for the remaining microseconds because usleep() will fail if its
+  // argument is greater than 1000000.
+  div_t sleepTime = div(microseconds, kNumMicrosPerSecond);
+  unsigned int seconds = sleepTime.quot;
```
dominichamon, line 45: `div_t` contains two integers. you store it in `unsigned int` and then loop until it's <= 0. given it can't be <0 as it's unsigned this is not appropriate.

Please store it as an `int`.

Also as `sleep` is documented as returning explicitly `0` if there's no time left to sleep, you can check `while (seconds != 0)`.

abhina-sree: Thanks for reviewing. I've updated the PR to address your comments.

dominichamon (approved):

dominichamon:
great, thanks! Once the CI finishes i can merge it.

Issue #1068, opened 2020-11-02
[FR] Rename mftbl to mftb


steven-wan-yu:
**Is your feature request related to a problem? Please describe.**
The build is failing on AIX because `mftbl` is not recognized. 

**Describe the solution you'd like**
`mftb` and `mftbl` are equivalent, suggest renaming `mftbl` to only have `mftb`.

**Describe alternatives you've considered**
N/A

**Additional context**
Linux Kernel Mailing List reference for replacing mftbl with mftb, https://lkml.org/lkml/2020/10/1/1048.

steven-wan-yu:
I'll be working on a patch for this.

PR #1069, opened 2020-11-02
Rename 'mftbl' to 'mftb'

steven-wan-yu:
AIX does not recognize `mftbl`. Since `mftb` and `mftbl` are equivalent, rename `mftbl` to `mftb` instead. This should fix https://github.com/google/benchmark/issues/1068.


steven-wan-yu:
@googlebot I signed it!

dominichamon:
Thanks! Once this passes CI i'll merge it.

steven-wan-yu:
It looks like one of the checks has hit a timeout? My apologies if this was due to my force-pushing and/or follow-on commit during the test. Is there a way to trigger a re-run?

dominichamon:
it's likely not you, but i'm rerunning anyway.

Issue #1075, opened 2020-12-12
[FR]Add option to report benchmark time in seconds


vasil-pashov:
**Is your feature request related to a problem? Please describe.**
Benchmark currently supports nanoseconds, microseconds and milliseconds. Benchmarks which take several seconds are really hard on the eyes. Support for reporting in seconds could be added.

**Describe the solution you'd like**
Add TimeUnit::kSeconds and update the functions taking TimeUnit so that reporting in seconds could be allowed in console, json and csv

dominichamon:
Sounds reasonable to me. PR welcome.

feserr:
I have made a PR for it #1076.

PR #1076, opened 2020-12-20
Add an option to report in seconds.

feserr:
Add the option to report in seconds to solve the issue #1075.


feserr:
@googlebot I signed it!

LebedevRI:
Thanks, i guess this looks about right.
This seems to be missing test coverage - are there no tests on how the other time units are actually printed to CSV/JSON?

feserr:
> Thanks, i guess this looks about right.
> This seems to be missing test coverage - are there no tests on how the other time units are actually printed to CSV/JSON?

Almost all test are checking nanoseconds and there was another to check the CSV report in microseconds.

I have added to new test to check console, JSON and CSV output of every time option.

test/reporter_output_test.cc
```diff
@@ -712,7 +799,7 @@ ADD_CASES(
 // ========================================================================= //
 // ------------------------- Testing StrEscape JSON ------------------------ //
 // ========================================================================= //
-#if 0 // enable when csv testing code correctly handles multi-line fields
+#if 0  // enable when csv testing code correctly handles multi-line fields
```
feserr, line 802: I didn't notice this change, It have been change by the auto formatter. Should I left it as before?

LebedevRI: I'd just leave it.

LebedevRI:
Ok, thanks, i guess that is it.
Will merge once CI finishes.

LebedevRI (approved):

Issue #1077, opened 2020-12-21
[FR]Add Bazel target for Release builds


aidanwolter3:
**Is your feature request related to a problem? Please describe.**
I am using Bazel to pull in this repository and depend on `:benchmark`, but when I run the benchmarks, I find that I get the error:
```
***WARNING*** Library was built as DEBUG. Timings may be affected.
```

Afaict, there is no way to build this library in Release without using cmake, and adding a define.

**Describe the solution you'd like**
We should provide a way for Bazel clients to build the library as Release. Maybe we can add a new target `:benchmark_release`? Or if I'm completely missing some magic way to do this, let me know :D

**Additional context**

_WORKSPACE_
```
# Benchmarking framework version 1.5.2.
git_repository(
    name = "benchmark",
    commit = "73d4d5e8d6d449fc8663765a42aa8aeeee844489",
    shallow_since = "1599818118 +0100",
    remote = "https://github.com/google/benchmark.git",
)
```

_BUILD_
```
cc_binary(
    name = "benchmark",
    srcs = ["benchmark.cc"],
    deps = [ "@benchmark//:benchmark"],
)
```

aidanwolter3:
I have found a couple ways to do this, but I prefer the first one, as it is easier to use by clients.

Prefer: https://github.com/google/benchmark/pull/1078
I don't prefer this one: https://github.com/aidanwolter3/benchmark/pull/new/bazel-release-config-setting

Yannic:
Wouldn't `bazel build --compilation_mode=opt` build it in release mode (including passing `-DNDEBUG` + also enabling optimizations)? Just adding `-DNDEBUG` unconditionally is only going to fool the library into thinking it was build for release mode when it actualy wasn't.

aidanwolter3:
Oh that's nice! That's exactly the kind of thing I was looking for, but didn't find. It appears that `--compilation_mode=opt` works for me.

PR #1081, opened 2020-12-29
Revert "Add bazel target for benchmark_release"

Yannic:
Reverts google/benchmark#1078

Uncondtionally adding `-DNDEBUG` will only fool the target into thinking it's build in release mode when it actually is not.
Instead, users should pass `--compilation_mode=opt` ([docs](https://github.com/google/benchmark/issues/1077#issuecomment-750495312)) when building the benchmark to build it in release mode (including enabling all compilation optimizations).

See https://github.com/google/benchmark/issues/1077#issuecomment-750495312

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -10,8 +10,8 @@ config_setting(
     visibility = [":__subpackages__"],
 )
 
-filegroup(
-    name = "benchmark_srcs",
+cc_library(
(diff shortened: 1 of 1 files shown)


aidanwolter3 (approved):
Thanks!

dominichamon:
Thank you!

Issue #1084, opened 2021-01-08
[FR] Use of CPM.cmake for download dependencies (Google Test libraries)


Andres6936:
**Is your feature request related to a problem? Please describe.**
With the arrival of  CMake 14 now is possible to configure and download the dependencies of project (Google Test for this case) 

**Describe the solution you'd like**
With the arrival of CMake 14, a script was added that allows to get repositories from the Internet, taking into account this the user [TheLartians](https://github.com/TheLartians) decided to write a [script](https://github.com/TheLartians/CPM.cmake) for CMake that acts as a package manager. In my opinion, this marks a new step for C++ and CMake, allowing work with libraries that are downloaded and updated without direct intervention of the user.

**Additional context**
Currently the download and configuration of dependencies without direct intervention of the user not is possible, since it depends on a parameter that the user must provide to CMake **[BENCHMARK_DOWNLOAD_DEPENDENCIES]**, the which is deactivated by default.

dominichamon:
How broadly available is cmake 14? This project continues to support a broad range of platforms and OSes and so we tend to be a little slower on the uptake of new abilities than other projects might be.

Andres6936:
@dominichamon The only part that really uses CMake 14 is the testing (Fetch_Content), from there on, everything is completely valid for CMake >= 3, I have observed the release date of CMake 3.14 (Mar 14, 2019) and yes, it is relatively new and it may take time for versions greater than or equal to this to be adopted in projects that use CMake, making this change momentarily problematic.

One option I contemplate is the separation of the tests in a new CMake project, allowing to generate the compilation files for the main project and leave the tests as a CMake sub-project to be generated with CMake 14. This separation translates only to changes in the CMake scripts, leaving the project structure intact for other generation tools.

Andres6936:
Close Issue, see details #1085

PR #1085, opened 2021-01-08
Automatic download and configuration of dependencies for test [Google Test]

Andres6936:
Now the Google Test libraries are automatically configured and downloaded when the option BENCHMARK_DOWNLOAD_DEPENDENCIES is ON (default is OFF).

See #1084 for more context.


dominichamon:
I'm not sure what this is adding exactly. `BENCHMARK_DOWNLOAD_DEPENDENCIES` already triggers the download of googletest if necessary...

Andres6936:
CPM is a wrapper around Fetch_Content, you are right, there is no more sense in this PR when BENCHMARK_DOWNLOAD_DEPENDENCIES already does the job. I close PR.

Issue #892, opened 2019-10-17
Conan ARM gcc build fails due to -m32/-m64 flags not supported


michkrom:
The _conan create_ or (build as dependency) fails on ARM targets (such as RPi)  as the -m32/-m64  flags are not supported by arm gcc compiler. 

Possibly, fence setting BENCHMARK_BUILD_32_BITS when ARM target is detected (conanfile.py). E.g.:

cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) **and 'arm' not in str(self.settings.arch)** else "OFF"

However, clang seems to support -m32/64  bit so the above suggestion may not be the best.

dominichamon:
@Croydon who did the conan integration.

Croydon:
@michkrom For issues with Conan packages please fill an issue in the [Conan Center Index](https://github.com/conan-io/conan-center-index) 

Upstream maintainers might not be involved in the Conan recipes so their issue trackers are most of the time the wrong place to report issues

Thanks! 😄

Croydon:
Actually, I forgot that we have a conanfile directly in this repository too. Never mind then 

It is still an open question how to handle such a hybrid setup: https://github.com/conan-io/conan-center-index/issues/22

Croydon:
This might be issue https://github.com/conan-io/conan-center-index/issues/370

@michkrom You are cross-compiling to ARM, right?

michkrom:
Actually, I was just compiling on the target, a rpi4.

dnilosek:
Not to revive a zombie thread here, but the issue still exists even with the prior fix. The string check for "64" doesnt work for most ARM arch types in Conan (for example, `armv8`). I ran into it building for a Jetson TX2. The CMake-only solution you have to this was to pass the responsibility of adding `-m32` compiler flag to to user (https://github.com/google/benchmark/blob/master/CMakeLists.txt#L22). 

Unless Conan has a decent way of understanding when a 32-bit build needs to be made, I would recommend doing the same here and forcing the user to add the compiler flag themselves.

Croydon:
I guess this has been fixed in the Conan Center Index recipe via https://github.com/conan-io/conan-center-index/pull/4653

PR #1088, opened 2021-02-01
Fix broken conan build for arm targets (#892)

ralfholly:
This fixes issue #892.

Building via conan for an arm platform (e. g. arch=armv8) doesn't work.

When compiling on an arm platform, "arch" is typically set to "armv#" where # denotes the version. Since there is no "64" contained in "arch" (as in "x86_64" on a PC), the code in conanfile.py sets the option "BENCHMARK_BUILD_32_BITS" to ON, which later adds the compiler option "-m32". This option is not supported by gcc and the build fails accordingly.

The check for the "64" substring in "arch" makes no sense for gcc, so I made it conditional.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -51,7 +51,10 @@ class GoogleBenchmarkConan(ConanFile):
 
         # See https://github.com/google/benchmark/pull/638 for Windows 32 build explanation
         if self.settings.os != "Windows":
-            cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) else "OFF"
+            if self.settings.compiler == "gcc":
+                cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "OFF"
+            else:
+                cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) else "OFF"
             cmake.definitions["BENCHMARK_USE_LIBCXX"] = "ON" if (str(self.settings.compiler.libcxx) == "libc++") else "OFF"
         else:
             cmake.definitions["BENCHMARK_USE_LIBCXX"] = "OFF"


jhurliman:
@ralfholly will this work for clang on arm64 as well? I'm stuck right now trying to get benchmark building on an NVIDIA Jetson Xavier via the conan recipe.

ralfholly:
> @ralfholly will this work for clang on arm64 as well? I'm stuck right now trying to get benchmark building on an NVIDIA Jetson Xavier via the conan recipe.

AFAIK clang does support the '-m32' option, as opposed to gcc. I tried to obtain benchmark from ConanCenter for an AGX-based project, but we're using gcc. I thought there were no problems regarding clang. What problems are you facing with clang?

jhurliman:
@ralfholly the issue is I don't want to build 32-bit binaries on 64-bit ARM (Jetson Xavier AGX) using clang. The line:
```
cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) else "OFF"
```

Does not recognize `armv8` as a 64-bit platform.

ralfholly:
> @ralfholly the issue is I don't want to build 32-bit binaries on 64-bit ARM (Jetson Xavier AGX) using clang. The line:
> 
> ```
> cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) else "OFF"
> ```
> 
> Does not recognize `armv8` as a 64-bit platform.

@jhurliman Right, this line makes no sense at all on armv8. IMHO, this line should be removed altogether. There should be an explicit conan option to enable 32-bit builds.

ralfholly:
> > @ralfholly the issue is I don't want to build 32-bit binaries on 64-bit ARM (Jetson Xavier AGX) using clang. The line:
> > ```
> > cmake.definitions["BENCHMARK_BUILD_32_BITS"] = "ON" if "64" not in str(self.settings.arch) else "OFF"
> > ```
> > 
> > 
> > Does not recognize `armv8` as a 64-bit platform.
> 
> @jhurliman Right, this line makes no sense at all on armv8. IMHO, this line should be removed altogether. There should be an explicit conan option to enable 32-bit builds.

@Croydon According to the log, the conanfile.py was added by you. Can you please contribute to this discussion?

Croydon:
The Conan ecosystem evolved quite a lot in the meantime. By now, I would recommend to not add Conanfiles upstream anymore, but only to the [Conan Center Index](https://github.com/conan-io/conan-center-index).

Right now, benchmark is in both which doubles the maintenance cost and leads to confusion. The only real advantage having a Conan recipe also directly in this repository is to more easily use development versions via Conan too as opposed to only having release versions available via the Conan Center.

Conan Center Index is [compiling a huge matrix](https://github.com/conan-io/conan-center-index/blob/master/docs/supported_platforms_and_configurations.md#supported-platforms-and-configurations) which means that it is also better tested. 

The [benchmark recipe in the Conan Center Index](https://github.com/conan-io/conan-center-index/blob/9f31fd518de2ef8454136aa4cda1ae16b046b14e/recipes/benchmark/all/conanfile.py#L65) is also setting this flag and passes fine for Clang and GCC, so not doing it anymore for GCC looks wrong to me.

Having said that all, I'm not sure enough about this flag and armv8 to give a confident statement, but probably it should be always 64bit then?

Introducing a new option to make it explicit would be also a way.

dominichamon:
are you recommending we remove the conanfile from the repo?

Croydon:
> are you recommending we remove the conanfile from the repo?

Yes

ralfholly:
> The Conan ecosystem evolved quite a lot in the meantime. By now, I would recommend to not add Conanfiles upstream anymore, but only to the [Conan Center Index](https://github.com/conan-io/conan-center-index).
> 
> Right now, benchmark is in both which doubles the maintenance cost and leads to confusion. The only real advantage having a Conan recipe also directly in this repository is to more easily use development versions via Conan too as opposed to only having release versions available via the Conan Center.
> 
> Conan Center Index is [compiling a huge matrix](https://github.com/conan-io/conan-center-index/blob/master/docs/supported_platforms_and_configurations.md#supported-platforms-and-configurations) which means that it is also better tested.
> 
> The [benchmark recipe in the Conan Center Index](https://github.com/conan-io/conan-center-index/blob/9f31fd518de2ef8454136aa4cda1ae16b046b14e/recipes/benchmark/all/conanfile.py#L62) is also setting this flag and passes fine for Clang and GCC, so not doing it anymore for GCC looks wrong to me.
> 
> Having said that all, I'm not sure enough about this flag and armv8 to give a confident statement, but probably it should be always 64bit then?
> 
> Introducing a new option to make it explicit would be also a way.

There might be use cases for compiling for 32-bit on a 64-bit platform


> > are you recommending we remove the conanfile from the repo?
> 
> Yes

But this means that the problem remains. If someone pulls benchmark from ConanCenter, it will still not build for armv8. Where does the ConanCenter conanfile.py come from? Is there a maintainer?

Croydon:
> But this means that the problem remains. If someone pulls benchmark from ConanCenter, it will still not build for armv8. Where does the ConanCenter conanfile.py come from? Is there a maintainer?

https://github.com/conan-io/conan-center-index/tree/master/recipes/benchmark

All recipes are maintained by the community and everyone can create pull requests against it

jhurliman:
I filed https://github.com/conan-io/conan-center-index/issues/4677 so this issue can be fixed in the proper repo. Maybe best to just delete the `conanfile.py` from this repo, or wait for a fix to land in the central registry and then backport it here.

PR #1052, opened 2020-10-05
Add MSVC ARM64 support

janisozaur:
Ran all the tests on my Raspberry Pi and it works fine. Results attached.

[user_counters_thousands_test.txt](https://github.com/google/benchmark/files/5329502/user_counters_thousands_test.txt)
[user_counters_test.txt](https://github.com/google/benchmark/files/5329503/user_counters_test.txt)
[user_counters_tabular_test.txt](https://github.com/google/benchmark/files/5329504/user_counters_tabular_test.txt)
[templated_fixture_test.txt](https://github.com/google/benchmark/files/5329505/templated_fixture_test.txt)
[string_util_gtest.txt](https://github.com/google/benchmark/files/5329506/string_util_gtest.txt)
[statistics_gtest.txt](https://github.com/google/benchmark/files/5329507/statistics_gtest.txt)
[skip_with_error_test.txt](https://github.com/google/benchmark/files/5329508/skip_with_error_test.txt)
[report_aggregates_only_test.txt](https://github.com/google/benchmark/files/5329509/report_aggregates_only_test.txt)
[register_benchmark_test.txt](https://github.com/google/benchmark/files/5329510/register_benchmark_test.txt)
[options_test.txt](https://github.com/google/benchmark/files/5329511/options_test.txt)
[multiple_ranges_test.txt](https://github.com/google/benchmark/files/5329512/multiple_ranges_test.txt)
[memory_manager_test.txt](https://github.com/google/benchmark/files/5329513/memory_manager_test.txt)
[map_test.txt](https://github.com/google/benchmark/files/5329514/map_test.txt)
[link_main_test.txt](https://github.com/google/benchmark/files/5329515/link_main_test.txt)
[internal_threading_test.txt](https://github.com/google/benchmark/files/5329516/internal_threading_test.txt)
[filter_test.txt](https://github.com/google/benchmark/files/5329517/filter_test.txt)

PR #1090, opened 2021-02-08
Add MSVC ARM64 support to cmake

janisozaur:
Split from https://github.com/google/benchmark/pull/1052, adds workaround for cmake when targeting ARM64 with MSVC


dominichamon:
I think this is good (sorry for the huge delay) but i have no obvious way to test it. @janisozaur, or someone else, can you confirm it's been tested and works?  :)

janisozaur:
Yes, it was tested in the original PR #1052 and the results are still available. I did use the compiled library to build OpenRCT2 and run its benchmarks.

PR #1094, opened 2021-02-19
Fix typo in invalid file name error message.

phoenix-meadowlark:
Was confused why `'/tmp/benchmark.csv` was an invalid file name. (It was invalid because the binary is executing on android).


dominichamon:
thank you!

PR #1095, opened 2021-02-19
Update 'Output Files' section to reflect csv support.

phoenix-meadowlark:
Add note to discourage users from attempting to use the deprecated csv output format.


dominichamon:
that's a great addition, thank you.

PR #1101, opened 2021-03-05
disable bazel for now

dominichamon:
Unclear why this is failing so disabling for now. bazel build still works locally so this is a bazelisk/github workflow thing.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -11,23 +11,23 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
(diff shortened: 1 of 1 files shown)


EwoutH:
On March 5th https://github.com/actions/virtual-environments/pull/490 got merged, which added bazelisk to the default installed applications. That probably caused a conflict with the manual installation of bazelisk in the workflow itself.

dominichamon:
Huh, thanks! So just removing the bazelisk installation should work. I'll try that.

PR #1104, opened 2021-03-10
Add cycles to benchmark output

chfast:
Continuation of https://github.com/google/benchmark/pull/989.


src/json_reporter.cc
```diff
@@ -223,9 +223,9 @@ void JSONReporter::PrintRunData(Run const& run) {
   if (!run.report_big_o && !run.report_rms) {
     out << indent << FormatKV("iterations", run.iterations) << ",\n";
     out << indent << FormatKV("real_time", run.GetAdjustedRealTime()) << ",\n";
-    out << indent << FormatKV("cpu_time", run.GetAdjustedCPUTime());
-    out << ",\n"
-        << indent << FormatKV("time_unit", GetTimeUnitString(run.time_unit));
+    out << indent << FormatKV("cpu_time", run.GetAdjustedCPUTime()) << ",\n";
```
dominichamon, line 226: nit: i think this line and a few others are too long (over 80 characters).

LebedevRI:
I will need more some more convincing here.
Some initial questions:
* what cycle count are we talking about? for the main thread? for all threads?
* how precisely do you want to know the results? would it be enough to simply multiply the timers by `GetCPUCyclesPerSecond()`?
* what about the following warnings in https://github.com/google/benchmark/blob/d90321ff7ab553f97449b7017caf29ec18e74c6b/src/cycleclock.h#L7-L18 ?

chfast:
> I will need more some more convincing here.

At this point we should consider this a feasibility study. Current values does not seem right. I will spend a bit more time on this, but can't guarantee any progress.

> Some initial questions:
> 
>     * what cycle count are we talking about? for the main thread? for all threads?

I consider this a replacement for CPU time. I don't really benchmark multi-threaded applications, but I guess this would be the sum of cycles from all threads.

For many my cases CPU cycles are better than CPU time because they are resistant to variadic CPU frequency. They produce the same results with turbo, without turbo and even _if frequency is unstable_. Moreover, they are "better comparable" across different CPU architectures.

> 
>     * how precisely do you want to know the results? would it be enough to simply multiply the timers by `GetCPUCyclesPerSecond()`?

Assuming that `GetCPUCyclesPerSecond()` value is correct, this only works if the value is constant over the benchmark run. I'm not fully satisfied with this solution but would take it if there is no better option.

> 
>     * what about the following warnings in https://github.com/google/benchmark/blob/d90321ff7ab553f97449b7017caf29ec18e74c6b/src/cycleclock.h#L7-L18
>        ?

I think we are fine if we measure at the beginning and the end of the benchmark loop. Moreover, there is a way to make the read "serializing" as described in https://github.com/dterei/gotsc#reading-the-tsc.

chfast:
An update with rather bad news.

1. We cannot use `rdtsc` directly, because it counts cycles with some arbitrary CPU tsc frequency. It's difficult to get this value (I'm currently doing so by a custom linux kernel module). And even if we knew it we would still need to know the CPU frequency at which the benchmark was run on to scale number of cycles accordingly. This is also hard. E.g. on my CPU benchmarks run at ~4.1GHz although the max frequency is 4.6GHz.
2. Moreover, using `rdtsc` would produce wallclock like values what is not very useful IMHO.
3. The `GetCPUCyclesPerSecond()` approach looks much better in this context except the values it returns are not correct. E.g. for my system it returns the tsc frequency which is 1.99 GHz if available or the CPU max frequency otherwise. While my benchmarks run at ~4.1 GHz when turbo is enabled and at 1.8 GHz when turbo is disabled.

I'm open to discuss all options, also in some more synchronous discussion environment. But at this point I don't see any clear path forward.

chfast:
I created a [PoC of a method to get current CPU frequency and converting CPU time to CPU cycles](https://github.com/chfast/cycles). We could read the current CPU frequency at the end of the benchmark loop and store this information in the results. Then the reporters can use this information to convert CPU time to CPU cycles.

LebedevRI:
> I created a [PoC of a method to get current CPU frequency and converting CPU time to CPU cycles](https://github.com/chfast/cycles). We could read the current CPU frequency at the end of the benchmark loop and store this information in the results. Then the reporters can use this information to convert CPU time to CPU cycles.

Won't that still be susceptible to dynamic cpu freq scaling?

chfast:
> Won't that still be susceptible to dynamic cpu freq scaling?

Yes. It assumes the CPU freq is constant during benchmark run. In my limited experiments with this, the results vary...

dominichamon:
This seems to have stalled. I don't think there's a good way forward. Should we just close this?

chfast:
Yes, this is currently stalled.

I have tried a number of things but have not reported back.

In the end, I only have got good results when using PAPI library. This is probably quite heavy, but I have not benchmarked the overhead.

I believe the https://github.com/google/benchmark/pull/1114 has solved it in more generic way. My plan is to try it out and report how it replaces this PR.

chfast:
This is obsolete by #1114 with `--benchmark_perf_counters=cycles`.

PR #1113, opened 2021-04-18
Use fewer ramp up repetitions when KeepRunningBatch is used

matta:
Use the benchmark's reported iteration count when estimating
iterations for the next repetition, rather than the requested
iteration count.  When the benchmark uses KeepRunningBatch the actual
iteration count can be larger than the one the runner requested.

Prior to this fix the runner was underestimating the next iteration
count, sometimes significantly so.  Consider the case of a benchmark
using a batch size of 1024.  Prior to this change, the benchmark
runner would attempt iteration counts 1, 10, 100 and 1000, yet the
benchmark itself would do the same amount of work each time: a single
batch of 1024 iterations.  The discrepancy could also contribute to
estimation errors once the benchmark time reached 10% of the target.
For example, if the very first batch of 1024 iterations reached 10% of
benchmark_min_min time, the runner would attempt to scale that to 100%
from a basis of one iteration rather than 1024.

This bug was particularly noticeable in benchmarks with large batch
sizes, especially when the benchmark also had slow set up or tear down
phases.

With this fix in place it is possible to use KeepRunningBatch to
achieve a kind of "minimum iteration count" feature by using a larger
fixed batch size.  For example, a benchmark may build a map of 500K
elements and test a "find" operation.  There is no point in running
"find" just 1, 10, 100, etc., times.  The benchmark can now pick a
batch size of something like 10K, and the runner will arrive at the
final max iteration count with in noticeably fewer repetitions.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -109,7 +109,7 @@ BenchmarkReporter::Run CreateRunReport(
 }
 
 // Execute one thread of benchmark b for the specified number of iterations.
-// Adds the stats collected for the thread into *total.
+// Adds the stats collected for the thread into manager->results.
 void RunInThread(const BenchmarkInstance* b, IterationCount iters,
                  int thread_id, ThreadManager* manager) {
   internal::ThreadTimer timer(
@@ -236,8 +236,10 @@ class BenchmarkRunner {
     VLOG(2) << "Ran in " << i.results.cpu_time_used << "/"
             << i.results.real_time_used << "\n";
 
-    // So for how long were we running?
-    i.iters = iters;
+    // By using KeepRunningBatch a benchmark can iterate more times than
+    // requested, so take the iteration count from i.results.
+    i.iters = i.results.iterations / b.threads;
+
     // Base decisions off of real time if requested by this benchmark.
     i.seconds = i.results.cpu_time_used;
     if (b.use_manual_time) {
--- a/test/basic_test.cc
+++ b/test/basic_test.cc
@@ -108,15 +108,30 @@ void BM_KeepRunning(benchmark::State& state) {
 BENCHMARK(BM_KeepRunning);
 
 void BM_KeepRunningBatch(benchmark::State& state) {
-  // Choose a prime batch size to avoid evenly dividing max_iterations.
-  const benchmark::IterationCount batch_size = 101;
+  // Choose a batch size >1000 to skip the typical runs with iteration
+  // targets of 10, 100 and 1000.  If these are not actually skipped the
+  // bug would be detectable as consecutive runs with the same iteration
+  // count.  Below we assert that this does not happen.
+  const benchmark::IterationCount batch_size = 1009;
+
+  static benchmark::IterationCount prior_iter_count = 0;
   benchmark::IterationCount iter_count = 0;
   while (state.KeepRunningBatch(batch_size)) {
     iter_count += batch_size;
   }
   assert(state.iterations() == iter_count);
+
+  // Verify that the iteration count always increases across runs (see
+  // comment above).
+  assert(iter_count == batch_size            // max_iterations == 1
+         || iter_count > prior_iter_count);  // max_iterations > batch_size
+  prior_iter_count = iter_count;
 }
-BENCHMARK(BM_KeepRunningBatch);
+// Register with a fixed repetition count to establish the invariant that
+// the iteration count should always change across runs.  This overrides
+// the --benchmark_repetitions command line flag, which would otherwise
+// cause this test to fail if set > 1.
+BENCHMARK(BM_KeepRunningBatch)->Repetitions(1);
 
 void BM_RangedFor(benchmark::State& state) {
   benchmark::IterationCount iter_count = 0;


matta:
Here is an example of the improvement seen in Abseil benchmarks.  Same result, similar iteration count, but the individual benchmarks using `KeepRunningBatch` finish in about half the time wall clock time.

Running all of the Abseil `stl_set_int32_t` benchmarks drops from 96 to 67 seconds.

### With fix:
```
% time ./bazel-bin/absl/container/btree_benchmark --benchmark_filter=BM_stl_set_int32_t               
2021-04-18T09:16:01-07:00
Running ../../bazel-bin/absl/container/btree_benchmark
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 0.33, 0.60, 0.53
-------------------------------------------------------------------------------
Benchmark                                     Time             CPU   Iterations
-------------------------------------------------------------------------------
BM_stl_set_int32_t_Insert                   868 ns          867 ns       838864
BM_stl_set_int32_t_InsertSorted             298 ns          298 ns      2411734
BM_stl_set_int32_t_InsertSmall              140 ns          140 ns      4956312
BM_stl_set_int32_t_Lookup                   764 ns          764 ns       954752
BM_stl_set_int32_t_FullLookup               162 ns          162 ns      4317977
BM_stl_set_int32_t_Delete                   942 ns          942 ns       838864
BM_stl_set_int32_t_DeleteRange              107 ns          107 ns      6606054
BM_stl_set_int32_t_QueueAddRem             1997 ns         1997 ns      2097152
BM_stl_set_int32_t_MixedAddRem             2293 ns         2293 ns      4194304
BM_stl_set_int32_t_Fifo                     109 ns          109 ns      6469273
BM_stl_set_int32_t_FwdIter                  101 ns          101 ns      6946343
BM_stl_set_int32_t_InsertRangeRandom 1156528950 ns   1156460818 ns            1
BM_stl_set_int32_t_InsertRangeSorted   70176196 ns     70176992 ns           10
./bazel-bin/absl/container/btree_benchmark   67.22s user 0.61s system 99% cpu 1:07.85 total

```

### Without fix:
```
% time ./bazel-bin/absl/container/btree_benchmark --benchmark_filter=BM_stl_set_int32_t
2021-04-18T09:12:29-07:00
Running ./bazel-bin/absl/container/btree_benchmark
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 0.75, 0.67, 0.52
-------------------------------------------------------------------------------
Benchmark                                     Time             CPU   Iterations
-------------------------------------------------------------------------------
BM_stl_set_int32_t_Insert                   867 ns          867 ns       734006
BM_stl_set_int32_t_InsertSorted             297 ns          297 ns      2306876
BM_stl_set_int32_t_InsertSmall              139 ns          140 ns      5007200
BM_stl_set_int32_t_Lookup                   771 ns          771 ns       986480
BM_stl_set_int32_t_FullLookup               162 ns          162 ns      3782448
BM_stl_set_int32_t_Delete                   920 ns          920 ns       734006
BM_stl_set_int32_t_DeleteRange              108 ns          108 ns      6291480
BM_stl_set_int32_t_QueueAddRem             1973 ns         1973 ns      2097152
BM_stl_set_int32_t_MixedAddRem             2387 ns         2387 ns      4194304
BM_stl_set_int32_t_Fifo                     110 ns          110 ns      6385349
BM_stl_set_int32_t_FwdIter                  102 ns          102 ns      6878566
BM_stl_set_int32_t_InsertRangeRandom 1171164989 ns   1171161832 ns            1
BM_stl_set_int32_t_InsertRangeSorted   70163345 ns     70160951 ns           10
./bazel-bin/absl/container/btree_benchmark  2>&1  95.75s user 0.56s system 99% cpu 1:36.33 total
```

### With fix (completes in 7.4 seconds)
```
% time ./bazel-bin/absl/container/btree_benchmark --benchmark_filter=BM_stl_set_int32_t_DeleteRange --v=3
Running ./bazel-bin/absl/container/btree_benchmark
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 0.37, 0.49, 0.45
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 1
-- LOG(2): Ran in 0.00927572/0.00927806
-- LOG(3): Next iters: 1048580, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 1048580
-- LOG(2): Ran in 0.115609/0.115625
-- LOG(3): Next iters: 6349025, 6.05488
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 6349025
-- LOG(2): Ran in 0.68657/0.686669
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_stl_set_int32_t_DeleteRange        107 ns          107 ns      6396338
./bazel-bin/absl/container/btree_benchmark 7.27s user 0.06s system 99% cpu 7.348 total
```
### Without fix (completes in 13.5 seconds)
```
% ./bazel-bin/absl/container/btree_benchmark --benchmark_filter=BM_stl_set_int32_t_DeleteRange --v=3
Running ./bazel-bin/absl/container/btree_benchmark
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 0.18, 0.42, 0.43
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 1
-- LOG(2): Ran in 0.00944516/0.00944853
-- LOG(3): Next iters: 10, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 10
-- LOG(2): Ran in 0.00932258/0.00932574
-- LOG(3): Next iters: 100, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 100
-- LOG(2): Ran in 0.00948083/0.00948501
-- LOG(3): Next iters: 1000, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 1000
-- LOG(2): Ran in 0.00925071/0.00925326
-- LOG(3): Next iters: 10000, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 10000
-- LOG(2): Ran in 0.00948569/0.00948977
-- LOG(3): Next iters: 100000, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 100000
-- LOG(2): Ran in 0.0095513/0.00955367
-- LOG(3): Next iters: 1000000, 10
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 1000000
-- LOG(2): Ran in 0.113674/0.113695
-- LOG(3): Next iters: 6157970, 6.15797
-- LOG(2): Running BM_stl_set_int32_t_DeleteRange for 6157970
-- LOG(2): Ran in 0.660694/0.660775
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_stl_set_int32_t_DeleteRange        107 ns          107 ns      6186622
./bazel-bin/absl/container/btree_benchmark  --v=3  13.46s user 0.06s system 99% cpu **13.530 total**

```

src/benchmark_runner.cc
```diff
@@ -122,6 +122,12 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
   {
     MutexLock l(manager->GetBenchmarkMutex());
     internal::ThreadManager::Result& results = manager->results;
+    if (results.iterations_per_thread == 0) {
+      results.iterations_per_thread = st.iterations();
+    } else {
+      CHECK_EQ(results.iterations_per_thread, st.iterations())
```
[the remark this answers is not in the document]
dominichamon: I don't think we require it, but I also can't see where the failure case would come in. `iterations` is defined in terms of `max_iterations` (should be the same for every thread), `total_iterations_` which counts down from `max_iterations` and should be the same for every thread and `batch_leftover`, ditto.

or did i miss something?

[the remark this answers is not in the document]
LebedevRI: I think this mainly could come from different threads using different batch sizes,
but after thinking about it, indeed i don't think this should be a problem in majority of the cases.

src/benchmark_runner.cc
```diff
@@ -236,8 +242,11 @@ class BenchmarkRunner {
     VLOG(2) << "Ran in " << i.results.cpu_time_used << "/"
             << i.results.real_time_used << "\n";
 
+    // By using KeepRunningBatch a benchmark can iterate more times than
+    // requested, so take the iteration count from i.results.
+    i.iters = i.results.iterations_per_thread;
```
LebedevRI, line 247: So this is just `i.iters = i.results.iterations / b.threads;` ?

LebedevRI (commented):
Actually, why do we need this new `iterations_per_thread` if we simply can divide sum of iterations by count of threads?

src/benchmark_runner.cc
dominichamon: i think that's right :)

matta: I agree.  Shall I change this line to `i.iters = i.results.iterations / b.threads;`?

That would simplify the patch and avoid the concern that the CHECK would trip.  I could even be more conservative and use `i.iters = std::max(iters, i.results.iterations / b.threads);`

LebedevRI: Let's go for division, i guess.

matta: Sounds good, done.

src/benchmark_runner.cc
```diff
@@ -236,8 +236,11 @@ class BenchmarkRunner {
     VLOG(2) << "Ran in " << i.results.cpu_time_used << "/"
             << i.results.real_time_used << "\n";
 
+    // By using KeepRunningBatch a benchmark can iterate more times than
+    // requested, so take the iteration count from i.results.
+    i.iters = i.results.iterations / b.threads;
+
     // So for how long were we running?
```
dominichamon, line 243: this comment is now misplaced. remove or move up a couple of lines.

test/basic_test.cc
```diff
@@ -108,15 +108,30 @@ void BM_KeepRunning(benchmark::State& state) {
 BENCHMARK(BM_KeepRunning);
 
 void BM_KeepRunningBatch(benchmark::State& state) {
-  // Choose a prime batch size to avoid evenly dividing max_iterations.
-  const benchmark::IterationCount batch_size = 101;
+  // Choose a batch size >1000 to skip the typical runs with iteration
+  // targets of 10, 100 and 1000.  If these are not actually skipped the
+  // bug would be detectable as consecutive runs with the same iteration
+  // count.  Below we assert that this does not happen.
+  const benchmark::IterationCount batch_size = 1009;
```
dominichamon, line 115: can you retain the comment that this needs to be prime to avoid the even division please?

matta: I removed the comment because I don't think it applies any more. If I understand the original intent correctly, the prime batch size was to help ensure that `max_iters` was not an even multiple of the batch size.  With the new approach, `max_iters` will immediately become a multiple of the batch size, whatever it is.  E.g. with a batch size of 17 `max_iters` will go from 1 to 17 to 170, etc.  Only the final 1.4x estimate might fail to be a multiple, but does a prime batch size makes that any more likely?

With this new way of growing `max_iters`, I could not think of a way to continue to exercise the not-evenly-divisible code paths explicitly, so I removed the comment instead.  It would be nice to retain that test coverage, so I'm open to ideas.  My only idea was to crack open some of these helper functions to make them more explicitly testable, but I didn't want to start off with that big of a change.

LebedevRI:
(@dominichamon feel free to merge once ready)

dominichamon:
Thank you so much!

PR #1114, opened 2021-04-19
Support optional, user-directed collection of performance counters

mtrofin:
The patch allows an engineer wishing to drill into the root causes
of a regression, for example. Currently, only single threaded runs
are supported. The feature is a build-time opt in, and then a runtime
opt in.

The engineer may run the benchmark executable, passing a list of
performance counter names (using libpfm's naming scheme) at the
command line. The counter values will then be collected and reported
back as UserCounters.

This is different from #240 in that it is a benchmark user opt-in, and
the counter collection is transparent to the benchmark.

Currently, this is only supported on platforms where libpfm is
supported.

libpfm: http://perfmon2.sourceforge.net/

### DIFF
--- a/.github/workflows/build-and-test-perfcounters.yml
+++ b/.github/workflows/build-and-test-perfcounters.yml
@@ -1,0 +1,44 @@
+name: build-and-test-perfcounters
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  job:
+    # TODO(dominic): Extend this to include compiler and set through env: CC/CXX.
+    name: ${{ matrix.os }}.${{ matrix.build_type }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, ubuntu-16.04, ubuntu-20.04]
+        build_type: ['Release', 'Debug']
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: install libpfm
+      run: sudo apt install libpfm4-dev
+
+    - name: create build environment
+      run: cmake -E make_directory ${{ runner.workspace }}/_build
+
+    - name: configure cmake
+      shell: bash
+      working-directory: ${{ runner.workspace }}/_build
+      run: cmake  -DBENCHMARK_ENABLE_LIBPFM=1 -DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+
+    - name: build
+      shell: bash
+      working-directory: ${{ runner.workspace }}/_build
+      run: cmake --build . --config ${{ matrix.build_type }}
+
+    # Skip testing, for now. It seems perf_event_open does not succeed on the
+    # hosting machine, very likely a permissions issue.
+    # TODO(mtrofin): Enable test.
+    # - name: test
+    #   shell: bash
+    #   working-directory: ${{ runner.workspace }}/_build
+    #   run: sudo ctest -C ${{ matrix.build_type }} --rerun-failed --output-on-failure
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,6 +270,10 @@ cxx_feature_check(STEADY_CLOCK)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+if (BENCHMARK_ENABLE_LIBPFM)
+  find_package(PFM)  
+endif()
+
 # Set up directories
 include_directories(${PROJECT_SOURCE_DIR}/include)
 
--- a/README.md
+++ b/README.md
@@ -296,6 +296,8 @@ too (`-lkstat`).
 
 [Setting the Time Unit](#setting-the-time-unit)
 
+[User-Requested Performance Counters](#perf-counters)
+
 [Preventing Optimization](#preventing-optimization)
 
 [Reporting Statistics](#reporting-statistics)
@@ -1067,6 +1069,42 @@ order to manually set the time unit, you can specify it manually:
 ```c++
 BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 ```
+
+<a name="perf-counters" />
+
+### User-Requested Performance Counters
+
+When running benchmarks, the user may choose to request collection of
+performance counters. This may be useful in investigation scenarios - narrowing
+down the cause of a regression; or verifying that the underlying cause of a
+performance improvement matches expectations.
+
+This feature is available if:
+
+* The benchmark is run on an architecture featuring a Performance Monitoring
+  Unit (PMU),
+* The benchmark is compiled with support for collecting counters. Currently,
+  this requires [libpfm](http://perfmon2.sourceforge.net/) be available at build
+  time, and
+* Currently, there is a limitation that the benchmark be run on one thread.
+
+The feature does not require modifying benchmark code. Counter collection is
+handled at the boundaries where timer collection is also handled. 
+
+To opt-in:
+
+*  Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
+*  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
+
+To use, pass a comma-separated list of counter names through the
+`-benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
+they are platform specific, but some (e.g. `CYCLES` or `INSTRUCTIONS`) are
+mapped by libpfm to platform-specifics - see libpfm
+[documentation](http://perfmon2.sourceforge.net/docs.html) for more details.
+
+The counter values are reported back through the [User Counters](custom-counters)
+mechanism, meaning, they are available in all the formats (e.g. JSON) supported
+by User Counters.
 
 <a name="preventing-optimization" />
 
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ too (`-lkstat`).
 
 [Setting the Time Unit](#setting-the-time-unit)
 
-[User-Requested Performance Counters](#perf-counters)
+[User-Requested Performance Counters](docs/perf_counters.md)
 
 [Preventing Optimization](#preventing-optimization)
 
@@ -1069,42 +1069,6 @@ order to manually set the time unit, you can specify it manually:
 ```c++
 BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 ```
-
-<a name="perf-counters" />
-
-### User-Requested Performance Counters
-
-When running benchmarks, the user may choose to request collection of
-performance counters. This may be useful in investigation scenarios - narrowing
-down the cause of a regression; or verifying that the underlying cause of a
-performance improvement matches expectations.
-
-This feature is available if:
-
-* The benchmark is run on an architecture featuring a Performance Monitoring
-  Unit (PMU),
-* The benchmark is compiled with support for collecting counters. Currently,
-  this requires [libpfm](http://perfmon2.sourceforge.net/) be available at build
-  time, and
-* Currently, there is a limitation that the benchmark be run on one thread.
-
-The feature does not require modifying benchmark code. Counter collection is
-handled at the boundaries where timer collection is also handled. 
-
-To opt-in:
-
-*  Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
-*  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
-
-To use, pass a comma-separated list of counter names through the
-`-benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
-they are platform specific, but some (e.g. `CYCLES` or `INSTRUCTIONS`) are
-mapped by libpfm to platform-specifics - see libpfm
-[documentation](http://perfmon2.sourceforge.net/docs.html) for more details.
-
-The counter values are reported back through the [User Counters](custom-counters)
-mechanism, meaning, they are available in all the formats (e.g. JSON) supported
-by User Counters.
 
 <a name="preventing-optimization" />
 
--- a/docs/perf_counters.md
+++ b/docs/perf_counters.md
@@ -25,7 +25,7 @@ To opt-in:
 *  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
 
 To use, pass a comma-separated list of counter names through the
-`-benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
+`--benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
 they are platform specific, but some (e.g. `CYCLES` or `INSTRUCTIONS`) are
 mapped by libpfm to platform-specifics - see libpfm
 [documentation](http://perfmon2.sourceforge.net/docs.html) for more details.
--- a/docs/perf_counters.md
+++ b/docs/perf_counters.md
@@ -1,0 +1,35 @@
+<a name="perf-counters" />
+
+# User-Requested Performance Counters
+
+When running benchmarks, the user may choose to request collection of
+performance counters. This may be useful in investigation scenarios - narrowing
+down the cause of a regression; or verifying that the underlying cause of a
+performance improvement matches expectations.
+
+This feature is available if:
+
+* The benchmark is run on an architecture featuring a Performance Monitoring
+  Unit (PMU),
+* The benchmark is compiled with support for collecting counters. Currently,
+  this requires [libpfm](http://perfmon2.sourceforge.net/) be available at build
+  time, and
+* Currently, there is a limitation that the benchmark be run on one thread.
+
+The feature does not require modifying benchmark code. Counter collection is
+handled at the boundaries where timer collection is also handled. 
+
+To opt-in:
+
+*  Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
+*  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
+
+To use, pass a comma-separated list of counter names through the
+`-benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
+they are platform specific, but some (e.g. `CYCLES` or `INSTRUCTIONS`) are
+mapped by libpfm to platform-specifics - see libpfm
+[documentation](http://perfmon2.sourceforge.net/docs.html) for more details.
+
+The counter values are reported back through the [User Counters](../README.md#custom-counters)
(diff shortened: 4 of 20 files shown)


CMakeLists.txt
```diff
@@ -270,6 +270,13 @@ cxx_feature_check(STEADY_CLOCK)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+if (BENCHMARK_LIBPFM_LIB_DIR AND BENCHMARK_LIBPFM_INCLUDE_DIR)
```
dominichamon, line 273: should this be using `find_package` instead?

src/benchmark.cc
```diff
@@ -163,11 +170,23 @@ void State::PauseTiming() {
   // Add in time accumulated so far
   CHECK(started_ && !finished_ && !error_occurred_);
   timer_->StopTimer();
+  if (perf_counters_measurement_) {
+    auto measurements = perf_counters_measurement_->StopAndGetMeasurements();
+    for (auto& name_and_measurement : measurements) {
```
dominichamon, line 175: `const auto&` ?

src/benchmark_runner.cc
```diff
@@ -155,6 +164,10 @@ class BenchmarkRunner {
            internal::ARM_DisplayReportAggregatesOnly);
       run_results.file_report_aggregates_only =
           (b.aggregation_report_mode & internal::ARM_FileReportAggregatesOnly);
+      if (b.threads > 1 && perf_counters_measurement.IsValid()) {
```
dominichamon, line 167: given you're checking this here, maybe you don't need the check above in the initializer list and can avoid storing the `perf_counters_measurement_ptr`?

CMakeLists.txt
mtrofin: I see, so if libpfm is available, then the build is auto-enabling counters support - or should there still be a master switch?

src/benchmark_runner.cc
dominichamon, line 167: consider:

```c++
CHECK(b.threads == 1 || !perf_counters_measurement.IsValid()) <<
    "Perf counters are not supported in multi-threaded cases.\n";
```

src/perf_counters.cc
```diff
@@ -0,0 +1,108 @@
+#include "perf_counters.h"
+
+#if defined HAS_LIBPFM
+#include <cstring>
+
```
dominichamon, line 5: nit: missing include for <vector>

src/perf_counters.cc
```diff
@@ -0,0 +1,108 @@
...
+  const int mode = PFM_PLM3;  // user mode only
+  for (size_t i = 0; i < counter_names.size(); ++i) {
+    const bool is_first = i == 0;
+    struct perf_event_attr attr;
+    std::memset(&attr, 0, sizeof(perf_event_attr));
+    attr.size = sizeof(attr);
+    const int group_id = !is_first ? counter_ids[0] : -1;
+    const auto& name = counter_names[i];
+    if (name.empty()) {
```
dominichamon, line 39: you might want to use CHECK here too if it's not too catastrophic

src/benchmark_runner.cc
mtrofin: there are a few places we need to pass the counters pointer, `perf_counters_measurement_ptr` simplifies needing to check in every place whether we can do that.

mtrofin: oh, CHECK here logs, doesn't crash. OK.

src/perf_counters.cc
mtrofin: for what

src/perf_counters.cc
mtrofin: I'll leave it to the user of the object to decide what to do with the outcome, so I CHECK in benchmark-runner.cc

CMakeLists.txt
```diff
@@ -270,6 +270,30 @@ cxx_feature_check(STEADY_CLOCK)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+# If successful, the following variables will be defined:
```
dominichamon, line 273: it might be cleaner to put this in a separate cmake file and call it as a function like `cxx_feature_check`. optional though.

src/CMakeLists.txt
```diff
@@ -28,6 +28,12 @@ target_include_directories(benchmark PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
     )
 
+# libpfm, if available
+if (LIBPFM)
```
dominichamon, line 32: should this be `if (HAVE_LIBPFM)`?

src/perf_counters.cc
```diff
@@ -0,0 +1,114 @@
...
+    }
+    pfm_perf_encode_arg_t arg;
+    std::memset(&arg, 0, sizeof(pfm_perf_encode_arg_t));
+    arg.attr = &attr;
+
+    const int pfm_get =
+        pfm_get_os_event_encoding(name.c_str(), mode, PFM_OS_PERF_EVENT, &arg);
+    if (pfm_get != PFM_SUCCESS) {
+      std::cerr << "Unknown counter name: " << name << "\n";
+      return NoCounters();
```
dominichamon, line 54: do you want to skip adding all the others here? we could add the ones we recognise at least and continue (having logged the error). wdyt?

src/perf_counters.cc
```diff
@@ -0,0 +1,114 @@
...
+}
+#else   // defined HAVE_LIBPFM
+const bool PerfCounters::kSupported = false;
+
+bool PerfCounters::Initialize() { return false; }
+
+PerfCounters PerfCounters::Create(
+    const std::vector<std::string>& counter_names) {
+  if (!counter_names.empty()) {
+    std::cerr << "Performance counters not supported.";
```
dominichamon, line 106: you might want to `GetErrorLogInstance() <<` instead of assuming `std::cerr`. optional.

CMakeLists.txt
mtrofin: I can do that in a follow-up cl?

src/CMakeLists.txt
mtrofin: done.

src/perf_counters.cc
mtrofin: it may be lost to the user - failing fast lets them correct and restart.

src/perf_counters.cc
dominichamon: \<vector\>

CMakeLists.txt
dominichamon: sg

mtrofin:
> the bazel build-and-test is failing due to `GTEST_SKIP` use which isn't available in the version we use (https://github.com/google/googletest/archive/3f0cf6b62ad1eb50d8736538363d3580dd640c3e.zip)

Ah, so should we document that instead of '$ git clone https://github.com/google/googletest.git benchmark/googletest' in the  [installation section](https://github.com/google/benchmark#installation)? In that case, what's the recommended alternative to GTEST_SKIP (just silently skipping?)

Additionally, should we move the buildbots to a more recent version (probably `git clone` isn't ideal for dependency management, anyway)?

dominichamon:
> > the bazel build-and-test is failing due to `GTEST_SKIP` use which isn't available in the version we use (https://github.com/google/googletest/archive/3f0cf6b62ad1eb50d8736538363d3580dd640c3e.zip)
> 
> Ah, so should we document that instead of '$ git clone https://github.com/google/googletest.git benchmark/googletest' in the [installation section](https://github.com/google/benchmark#installation)? In that case, what's the recommended alternative to GTEST_SKIP (just silently skipping?)

i think documenting it is fine as any recent version is compatible with this version. it's only when trying to use new features we run into problems :)

I haven't used GTEST_SKIP but yes, silently skipping is fine by me.

> 
> Additionally, should we move the buildbots to a more recent version (probably `git clone` isn't ideal for dependency management, anyway)?

we can update bazel to a newer version but i'd want it to be a version that is readily available and i'm not sure 1.10 (which include GTEST_SKIP) is available everywhere trivially.

mtrofin:
> > > the bazel build-and-test is failing due to `GTEST_SKIP` use which isn't available in the version we use (https://github.com/google/googletest/archive/3f0cf6b62ad1eb50d8736538363d3580dd640c3e.zip)
> > 
> > 
> > Ah, so should we document that instead of '$ git clone https://github.com/google/googletest.git benchmark/googletest' in the [installation section](https://github.com/google/benchmark#installation)? In that case, what's the recommended alternative to GTEST_SKIP (just silently skipping?)
> 
> i think documenting it is fine as any recent version is compatible with this version. it's only when trying to use new features we run into problems :)
> 
> I haven't used GTEST_SKIP but yes, silently skipping is fine by me.
> 
> > Additionally, should we move the buildbots to a more recent version (probably `git clone` isn't ideal for dependency management, anyway)?
> 
> we can update bazel to a newer version but i'd want it to be a version that is readily available and i'm not sure 1.10 (which include GTEST_SKIP) is available everywhere trivially.

Ah, I see. OK. I'll prep a separate PR for the doc.

mtrofin:
> > > > the bazel build-and-test is failing due to `GTEST_SKIP` use which isn't available in the version we use (https://github.com/google/googletest/archive/3f0cf6b62ad1eb50d8736538363d3580dd640c3e.zip)
> > > 
> > > 
> > > Ah, so should we document that instead of '$ git clone https://github.com/google/googletest.git benchmark/googletest' in the [installation section](https://github.com/google/benchmark#installation)? In that case, what's the recommended alternative to GTEST_SKIP (just silently skipping?)
> > 
> > 
> > i think documenting it is fine as any recent version is compatible with this version. it's only when trying to use new features we run into problems :)
> > I haven't used GTEST_SKIP but yes, silently skipping is fine by me.
> > > Additionally, should we move the buildbots to a more recent version (probably `git clone` isn't ideal for dependency management, anyway)?
> > 
> > 
> > we can update bazel to a newer version but i'd want it to be a version that is readily available and i'm not sure 1.10 (which include GTEST_SKIP) is available everywhere trivially.
> 
> Ah, I see. OK. I'll prep a separate PR for the doc.

Tried an alternative where we leave the instructions to do git clone for googletest, but I check if the feature I want - GTEST_SKIP - is defined. Some bots seem to use the latest googletest, and presumably we'd move there eventually? 
If we like this alternative, then I can just document the method in the installation instructions (as a 'warning - if you author tests using features such as...")
wdyt?

src/perf_counters.cc
dominichamon: ^^

dominichamon (approved):
LGTM

@LebedevRI do you want to take a look at this too?

CMakeLists.txt
```diff
@@ -270,6 +270,29 @@ cxx_feature_check(STEADY_CLOCK)
...
+# HAVE_LIBPFM.
+# Set BENCHMARK_ENABLE_LIBPFM to 0 to disable, regardless of libpfm presence.
+include(CheckIncludeFile)
+include(CheckLibraryExists)
+ENABLE_LANGUAGE(C)
```
chfast, line 278: This should be lowercase.

CMakeLists.txt
```diff
@@ -270,6 +270,29 @@ cxx_feature_check(STEADY_CLOCK)
...
+
+if (BENCHMARK_ENABLE_LIBPFM)
+  check_library_exists(libpfm.a pfm_initialize "" HAVE_LIBPFM_INITIALIZE)
```
chfast, line 281: This could be encapsulated as `FindPFM.cmake` module and used as `find_package(PFM)`. I can do this conversion if desired.

src/perf_counters.cc
```diff
@@ -0,0 +1,116 @@
...
+        << PerfCounterValues::kMaxCounters << "\n";
+    return NoCounters();
+  }
+  std::vector<int> counter_ids(counter_names.size());
+
+  const int mode = PFM_PLM3;  // user mode only
+  for (size_t i = 0; i < counter_names.size(); ++i) {
+    const bool is_first = i == 0;
+    struct perf_event_attr attr;
+    std::memset(&attr, 0, sizeof(perf_event_attr));
```
chfast, line 39: You can just do `perf_event_attr attr{};` without `memset()`.

CMakeLists.txt
mtrofin: done

src/perf_counters.cc
mtrofin: oh - yes. thanks!

CMakeLists.txt
mtrofin: thanks! I can take a stab at it once build bots green up, as a last step.

.github/workflows/build-and-test-perfcounters.yml
```diff
@@ -0,0 +1,44 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, ubuntu-16.04, ubuntu-20.04]
+        build_type: ['Release', 'Debug']
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: install libpfm
+      run: sudo apt-get install libpfm4-dev
```
dominichamon, line 23: i think this is better future proofed as `sudo apt install libpfm4-dev` as ubuntu is moving from `apt-get` to `apt`

src/perf_counters.h
```diff
@@ -0,0 +1,155 @@
...
+  // specified. The values are user-mode only. The counter name format is
+  // implementation and OS specific.
+  // TODO: once we move to C++-17, this should be a std::optional, and then the
+  // IsValid() boolean can be dropped.
+  static PerfCounters Create(const std::vector<std::string>& counter_names);
+
+  // Take a snapshot of the current value of the counters into the provided
+  // valid PerfCounterValues storage. The values are populated such that:
+  // names()[i]'s value is (*values)[i]
+  BENCHMARK_ALWAYS_INLINE inline bool Snapshot(PerfCounterValues* values) {
```
dominichamon, line 83: you don't need both `BENCHMARK_ALWAYS_INLINE` and `inline` (in fact it's a build warning on Windows)

mtrofin: done

CMakeLists.txt
mtrofin: done

dominichamon:
It looks like you're merging all these changes into the single commit. You know you can make multiple commits against the same PR right? You just push them into the branch. That would also make it a bit easier to review just the deltas for each commit.

src/perf_counters.cc
mtrofin: Done.

.github/workflows/build-and-test-perfcounters.yml
mtrofin: done

mtrofin:
> It looks like you're merging all these changes into the single commit. You know you can make multiple commits against the same PR right? You just push them into the branch. That would also make it a bit easier to review just the deltas for each commit.

Now I remember - sorry. And they can be squashed then at the end, when pushing to main, correct?

dominichamon:
> > It looks like you're merging all these changes into the single commit. You know you can make multiple commits against the same PR right? You just push them into the branch. That would also make it a bit easier to review just the deltas for each commit.
> 
> Now I remember - sorry. And they can be squashed then at the end, when pushing to main, correct?

exactly :)

README.md
```diff
@@ -1068,6 +1070,42 @@ order to manually set the time unit, you can specify it manually:
 BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 ```
```
dominichamon, line 1072: maybe this should be a separate doc under `docs` that is linked to from here. a README over 1k lines is a bit heavy.

src/perf_counters.cc
```diff
@@ -0,0 +1,114 @@
+#include "perf_counters.h"
+
+#if defined HAVE_LIBPFM
+#include <cstring>
```
dominichamon, line 4: sorry, i just spotted these includes (string, vector) would be needed if `HAVE_LIBPFM` is not defined too.

mtrofin: done.

README.md
mtrofin: done.

docs/perf_counters.md
```diff
@@ -0,0 +1,35 @@
...
+The feature does not require modifying benchmark code. Counter collection is
+handled at the boundaries where timer collection is also handled. 
+
+To opt-in:
+
+*  Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
+*  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
+
+To use, pass a comma-separated list of counter names through the
+`-benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
```
LebedevRI, line 28: `--benchmark_perf_counters` i guess

src/benchmark.cc
```diff
@@ -163,11 +170,23 @@ void State::PauseTiming() {
   // Add in time accumulated so far
   CHECK(started_ && !finished_ && !error_occurred_);
   timer_->StopTimer();
+  if (perf_counters_measurement_) {
+    auto measurements = perf_counters_measurement_->StopAndGetMeasurements();
+    for (const auto& name_and_measurement : measurements) {
+      auto name = name_and_measurement.first;
+      auto measurement = name_and_measurement.second;
+      CHECK_EQ(counters[name], 0.0);
+      counters[name] = Counter(measurement, Counter::kAvgIterations);
```
LebedevRI, line 179: Will this play well with threads?

src/perf_counters.cc
```diff
@@ -0,0 +1,114 @@
+#include "perf_counters.h"
```
LebedevRI, line 1: Common license blurb missing

src/perf_counters.h
```diff
@@ -0,0 +1,158 @@
+#ifndef BENCHMARK_PERF_COUNTERS_H
```
LebedevRI, line 1: License blurb missing

src/perf_counters.h
```diff
@@ -0,0 +1,158 @@
...
+  uint64_t operator[](size_t pos) const { return values_[kPadding + pos]; }
+
+  static constexpr size_t kMaxCounters = 3;
+
+ private:
+  friend class PerfCounters;
+  // Get the byte buffer in which perf counters can be captured.
+  // This is used by PerfCounters::Read
+  std::pair<char*, size_t> get_data_buffer() {
+    return {reinterpret_cast<char*>(values_.data()),
```
LebedevRI, line 43: This looks like type punning. I'd prefer to avoid that if possible :)

docs/perf_counters.md
mtrofin: done

src/perf_counters.cc
mtrofin: done

dominichamon:
thanks!

PR #1115, opened 2021-04-21
Generate HTML charts script

Auburn:
Hi, I've created a python script for generating html charts from multiple JSON benchmark outputs. I read that you were looking for a chart generator for results so hopefully this can be of use.

The python script generates an HTML file that uses [Chart.js](https://www.chartjs.org/) and [Materialize](https://materializecss.com/) to create a good looking and responsive experience

Here is an example output: https://auburn.github.io/fastnoise2benchmarking/

The script supports configuration of how results are grouped into charts and also how the charts are grouped using regex. Naming of charts and data points is also done through regex. The HTML header file is used for configuring Chart.js formatting which can be used to colour code results and provide clickable links on data points as in my example.

I have not used Python/HTML/JS much before so let me know if I'm doing anything non standard, or if it can be improved in any way.


Auburn:
@googlebot I signed it!

dominichamon:
The results look great!

It'll take a while to review because that's quite a bit of Python so please bear with us. A few things that might be useful:

* putting it in a subfolder under tools so the html and py files are kept together but apart from other tools
* Maybe modularising the python a bit
* Tests would be ideal so we can add them to the CI
* Documentation so people know how to use it would be amazing.

dominichamon:
@Auburn do you think you'll be able to continue to work on this? I'm really excited to see it land, but it could be cleaned up a bit first.

Auburn:
I can write some documentation if you want, the command line parameters are already documented and descriptions of each can be seen by using -h / --help

Can you elaborate on what you mean by these things:
- Maybe modularising the python a bit
- Tests would be ideal so we can add them to the CI

What do you mean by "modularising the python" I'm not that familiar with python

What kind of tests would you want for this?

dominichamon:
> I can write some documentation if you want, the command line parameters are already documented and descriptions of each can be seen by using -h / --help
> 
> Can you elaborate on what you mean by these things:
> 
> * Maybe modularising the python a bit
> * Tests would be ideal so we can add them to the CI
> 
> What do you mean by "modularising the python" I'm not that familiar with python


It's one giant python script right now. I wonder if it could be split into files, each of which does _one part_ of the whole process.

eg https://stackoverflow.com/questions/7216993/how-to-split-python-script-in-parts-and-to-import-the-parts-in-a-loop

> 
> What kind of tests would you want for this?

https://realpython.com/python-testing/

unit tests covering the functional blocks.

dominichamon:
(if you feel this is too much because you're new to Python, say so. we can always get this in and then have an open issue asking for help cleaning it up :) )

Auburn:
I can work on modularising a bit more, I would like some help setting up tests though. It's not something I've touched before even outside of python. What kind of things would the tests look at, would it just check that the json output is properly parsed?

dominichamon:
i think tests for each function can be a good start:

* `try_regex_find` is simple, but we should make sure tests that cover edge cases work as expected
* `parse_benchmark_file` is the big one.. that should check that some sample json parses as expected. you might find it useful to extract the file open from the parsing so the test can focus on parsing some json passed in as a string.
* extract out the html printing code into a function and write tests that confirm the HTML looks good for a given benchmark passed in

Auburn:
I did make the code more modular and added some more comments, but I don't have the time or knowledge to write tests. If you wanted to take it anyway and have an open issue to clean it up as you suggested, I think that is the best solution. If not you can just leave this closed

PR #1118, opened 2021-04-24
fix cmake issue with referencing a non-existing function argument

germasch:
according to https://cmake.org/cmake/help/latest/command/function.html,
"Referencing to ARGV# arguments beyond ARGC have undefined behavior.",
which I hit with cmake 3.19.7.

This uses ARGC to check whether ARGV1 has been passed before referencing
it.


dominichamon:
travis issue seems unrelated (i think i might replace it entirely with github actions).

thank you!

Issue #525, opened 2018-02-07
Custom context fields


mmcloughlin:
I am interested in adding custom context fields to benchmarks. As far as I can tell this isn't possible right now? If not would it be possible to add?

The use case I have in mind is adding our software build version to the context. This way output JSON/CSV data files can be passed around or stored and contain the information required to reproduce results.

Not critical since this information can be stored elsewhere. But it might be nice. Thoughts?

dominichamon:
It sounds like a nice addition. maybe a method that can be called from the
main, or a key-value set passed in to the main macro maybe?

Dominic Hamon | Google
*There are no bad ideas; only good ideas that go horribly wrong.*

On Wed, Feb 7, 2018 at 2:40 PM, Michael McLoughlin <notifications@github.com

mmcloughlin:
Thanks for the quick reply.

I would have thought it made sense to leave the `BENCHMARK_MAIN` macro as it is, since adding context fields doesn't seem like the most common use case. But a method that could be called from a custom `main` function would be ideal.

EricWF:
I think this is a bug we should try to fix on the `v2` branch using JSON.

I think the best way would be to make the `BenchmarkFamilies` class part of the public interface. I'll take a stab at something.

mmcloughlin:
Just thought I'd bump this, since I find myself wishing I had this feature a second time :) Right now I have a benchmark that uses a GPU and I would like to be able to output GPU information to the context.

LebedevRI:
You currently could use `SetLabel()`, though that is not really a proper solution.

marehr:
> You currently could use `SetLabel()`, though that is not really a proper solution.

But, `SetLabel()` is for each individual benchmark, right?

I also wish I could set some global meta-data like the build commit or the used library versions.

dominichamon:
A couple of options then:

1. a flag that takes a key-value pair and can be repeated: `--benchmark_context="myfoo=bar"`
2. a function that can be called from a custom main: `benchmark::AddContext(std::string key, std::string value)`
3. (longer term): Something with the `BenchmarkFamilies` interface.

If anyone wants to take a crack at #2 i think it's the quickest option. #1 is maybe more flexible (or a combination of 1 and 2) but would be a little more work.

mmcloughlin:
Since posting this issue two years ago I now have additional use cases for it. Adding software version information is still useful, but I'd also like to be able to query more information about system and process configuration, for example:

* What is CPU frequency set to? 
[...]

PR #1127, opened 2021-04-28
Add `benchmark_context` flag that allows per-run custom context.

dominichamon:
Add support for key-value flags in general.
Added test for key-value flags.
Added `benchmark_context` flag.
Output content of `benchmark_context` to base reporter.

Solves the first part of #525.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "benchmark/benchmark.h"
+
 #include "benchmark_api_internal.h"
 #include "benchmark_runner.h"
 #include "internal_macros.h"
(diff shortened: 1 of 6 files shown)


jkeiser:
Is there a document this ought to be added to, or could you put a quickie example use in this PR so people can reference it when it lands? Thanks :)

src/benchmark.cc
```diff
@@ -446,6 +451,7 @@ void PrintUsageAndExit() {
           "          [--benchmark_out_format=<json|console|csv>]\n"
           "          [--benchmark_color={auto|true|false}]\n"
           "          [--benchmark_counters_tabular={true|false}]\n"
+          "          [--benchmark_context=<extra_context>]\n"
```
jkeiser, line 454: Maybe `[--benchmark_context=<key>=<value>,...]` ?

dominichamon:
> Is there a document this ought to be added to, or could you put a quickie example use in this PR so people can reference it when it lands? Thanks :)

Done

LebedevRI (approved):
Perhaps we might want to have an API away to add those later?

dominichamon:
> Perhaps we might want to have an API away to add those later?

yep, agreed. that's the next PR.

PR #1129, opened 2021-04-29
RFC: Remove unnecessary reporter subclasses from public API

dominichamon:
The `BenchmarkReporter` class is meant to be part of the public API so
custom reporters can be built. However, there isn't a good reason to
expose the `ConsoleReporter`, `CSVReporter`, or `JSONReporter` classes.

[Searches on github](https://github.com/search?l=C%2B%2B&q=%22benchmark%3A%3AJSONReporter%22&type=Code) suggest that they are referenced but it seems like users are using their own command line parsing to select between reporters, which is unnecessary as `benchmark::Initialize` takes `argc` and `argv` and will do the right thing.


dominichamon:
given the change to the API i think this would require a bump to 1.6

dominichamon:
Does anyone think this is a terrible idea and if so why? :)

src/console_reporter.h
```diff
@@ -0,0 +1,37 @@
+#include "benchmark/benchmark.h"
```
LebedevRI, line 1: license blurb missing

PR #1135, opened 2021-05-04
Perf counters logging

debashish-ghosh:
Fixes google/benchmark/#1134

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -434,7 +434,7 @@ namespace internal {
 
 void PrintUsageAndExit() {
   fprintf(stdout,
-          "benchmark"
+          "%s"
           " [--benchmark_list_tests={true|false}]\n"
           "          [--benchmark_filter=<regex>]\n"
           "          [--benchmark_min_time=<min_time>]\n"
@@ -446,7 +446,11 @@ void PrintUsageAndExit() {
           "          [--benchmark_out_format=<json|console|csv>]\n"
           "          [--benchmark_color={auto|true|false}]\n"
           "          [--benchmark_counters_tabular={true|false}]\n"
-          "          [--v=<verbosity>]\n");
+#if defined HAVE_LIBPFM
+          "          [--benchmark_perf_counters=<counter1[,counter2...]>]\n"
+#endif
+          "          [--v=<verbosity>]\n",
+          BenchmarkReporter::Context::executable_name);
   exit(0);
 }
 
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -38,7 +38,7 @@ PerfCounters PerfCounters::Create(
     return NoCounters();
   }
   if (counter_names.size() > PerfCounterValues::kMaxCounters) {
-    GetErrorLogInstance()
+    VLOG(2)
         << counter_names.size()
         << " counters were requested. The minimum is 1, the maximum is "
         << PerfCounterValues::kMaxCounters << "\n";
@@ -54,7 +54,7 @@ PerfCounters PerfCounters::Create(
     const int group_id = !is_first ? counter_ids[0] : -1;
     const auto& name = counter_names[i];
     if (name.empty()) {
-      GetErrorLogInstance() << "A counter name was the empty string\n";
+      VLOG(2) << "A counter name was the empty string\n";
       return NoCounters();
     }
     pfm_perf_encode_arg_t arg{};
@@ -63,7 +63,7 @@ PerfCounters PerfCounters::Create(
     const int pfm_get =
         pfm_get_os_event_encoding(name.c_str(), mode, PFM_OS_PERF_EVENT, &arg);
     if (pfm_get != PFM_SUCCESS) {
-      GetErrorLogInstance() << "Unknown counter name: " << name << "\n";
+      VLOG(2) << "Unknown counter name: " << name << "\n";
       return NoCounters();
     }
     attr.disabled = is_first;
@@ -85,7 +85,7 @@ PerfCounters PerfCounters::Create(
       }
     }
     if (id < 0) {
-      GetErrorLogInstance()
+      VLOG(2)
           << "Failed to get a file descriptor for " << name << "\n";
       return NoCounters();
     }
@@ -93,7 +93,7 @@ PerfCounters PerfCounters::Create(
     counter_ids[i] = id;
   }
   if (ioctl(counter_ids[0], PERF_EVENT_IOC_ENABLE) != 0) {
-    GetErrorLogInstance() << "Failed to start counters\n";
+    VLOG(2) << "Failed to start counters\n";
     return NoCounters();
   }
 
@@ -117,7 +117,7 @@ bool PerfCounters::Initialize() { return false; }
 PerfCounters PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
   if (!counter_names.empty()) {
-    GetErrorLogInstance() << "Performance counters not supported.";
+    VLOG(3) << "Performance counters not supported.\n";
   }
   return NoCounters();
 }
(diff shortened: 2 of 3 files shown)


src/perf_counters.cc
```diff
@@ -38,7 +38,7 @@ PerfCounters PerfCounters::Create(
     return NoCounters();
   }
   if (counter_names.size() > PerfCounterValues::kMaxCounters) {
-    GetErrorLogInstance()
+    VLOG(2)
```
dominichamon, line 41: i think this should still be an error as the user has attempted to set counters.

src/perf_counters.cc
```diff
@@ -54,7 +54,7 @@ PerfCounters PerfCounters::Create(
     const int group_id = !is_first ? counter_ids[0] : -1;
     const auto& name = counter_names[i];
     if (name.empty()) {
-      GetErrorLogInstance() << "A counter name was the empty string\n";
+      VLOG(2) << "A counter name was the empty string\n";
```
dominichamon, line 57: same here, as it results in the user not getting what they want.

src/perf_counters.cc
```diff
@@ -63,7 +63,7 @@ PerfCounters PerfCounters::Create(
     const int pfm_get =
         pfm_get_os_event_encoding(name.c_str(), mode, PFM_OS_PERF_EVENT, &arg);
     if (pfm_get != PFM_SUCCESS) {
-      GetErrorLogInstance() << "Unknown counter name: " << name << "\n";
+      VLOG(2) << "Unknown counter name: " << name << "\n";
```
dominichamon, line 66: same.

src/perf_counters.cc
```diff
@@ -85,15 +85,15 @@ PerfCounters PerfCounters::Create(
       }
     }
     if (id < 0) {
-      GetErrorLogInstance()
+      VLOG(2)
```
dominichamon, line 88: same

src/perf_counters.cc
```diff
@@ -85,15 +85,15 @@ PerfCounters PerfCounters::Create(
...
           << "Failed to get a file descriptor for " << name << "\n";
       return NoCounters();
     }
 
     counter_ids[i] = id;
   }
   if (ioctl(counter_ids[0], PERF_EVENT_IOC_ENABLE) != 0) {
-    GetErrorLogInstance() << "Failed to start counters\n";
+    VLOG(2) << "Failed to start counters\n";
```
dominichamon, line 96: same

src/perf_counters.cc
```diff
@@ -117,7 +117,7 @@ bool PerfCounters::Initialize() { return false; }
 PerfCounters PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
   if (!counter_names.empty()) {
-    GetErrorLogInstance() << "Performance counters not supported.";
+    VLOG(3) << "Performance counters not supported.\n";
```
dominichamon, line 120: VLOG(1) or VLOG(2) for this?

src/perf_counters.h
```diff
@@ -94,15 +94,15 @@ class PerfCounters final {
   // names()[i]'s value is (*values)[i]
   BENCHMARK_ALWAYS_INLINE bool Snapshot(PerfCounterValues* values) {
 #ifndef BENCHMARK_OS_WINDOWS
-    assert(values != nullptr);
-    assert(IsValid());
-    auto buffer = values->get_data_buffer();
-    auto read_bytes = ::read(counter_ids_[0], buffer.first, buffer.second);
-    return static_cast<size_t>(read_bytes) == buffer.second;
+    if (values != nullptr && IsValid()) {
```
dominichamon, line 97: why is this better than asserting?

debashish-ghosh: If BENCHMARK_OS_WINDOWS is defined, then it simply returns false without any asserts. This seems to me as non-uniform behavior.

But this change is not necessary and can be reverted.

src/perf_counters.cc
debashish-ghosh: The behavior is same even if the user doesn't give any benchmark_perf_counters command-line option.

dominichamon: if the user doesn't set the flag the counter_names size is > maxCounters?

that seems like a bug further upstream no?

src/perf_counters.cc
debashish-ghosh: Okay. I thought 1 was default (it is 0). I can change all the above levels to 1.
The long term solution is to validate the command-line args, and print the diagnostics before even starting the first benchmark.
But this will take a lot of code changes. The current counter-error-logging code is tightly coupled with counter-reading code.

src/perf_counters.cc
debashish-ghosh: Oops, that was for the comment below. I'll update this part.

src/perf_counters.cc
debashish-ghosh: The behavior is same even if the user doesn't give any ```--benchmark_perf_counters``` command-line option. We need to differentiate between user-gave-empty-flag vs user-gave-no-flag.

src/perf_counters.cc
mtrofin: The current error logging code assumes we're in PerfCounters::Create because we actually want to create counters. I think the fix is to catch outside calling this that either they aren't supported, or the vector is empty, and not call it.

src/benchmark.cc
```diff
@@ -446,7 +446,11 @@ void PrintUsageAndExit() {
           "          [--benchmark_out_format=<json|console|csv>]\n"
           "          [--benchmark_color={auto|true|false}]\n"
           "          [--benchmark_counters_tabular={true|false}]\n"
-          "          [--v=<verbosity>]\n");
+#if defined HAVE_LIBPFM
```
mtrofin, line 449: I'd propose separating this out of this PR.

That aside, I wouldn't condition this on HAVE_LIBPFM, for 2 reasons:

* while currently the case, !HAVE_LIBPFM isn't equivalent to perf counters not supported; for instance, if someone implements it on windows (libpfm is a linux thing).
* the flag is always available, just doesn't function if perflab support isn't there.

My preference would be to add an extra note after the format of the option if !PerfCounters::kSupported, like for example before `exit(0)`

```
if (!PerfCounters::kSupported) fprintf(stdout, "NOTE: --benchmark_perf_counters not supported on this platform/build. See documentation for details");
```

src/perf_counters.h
mtrofin: This shouldn't even be called on windows or on cases where the counters aren't supported; the windows test is to circumvent dealing with ::read's definition being different on windows.

That aside, I agree with you, why not have the behavior consistent - I'd propose having this in a different patch, though, it's not logging - related.

src/benchmark.cc
debashish-ghosh: Yes. Now I'm thinking about both build time support and runtime support. This is definitely going in it's own separate direction. I'd better not include it in this PR.

debashish-ghosh:
Understanding the libpfm APIs. I'll update this in sometime.

mtrofin:
> Understanding the libpfm APIs. I'll update this in sometime.

Sounds good. I can fix the spurious logging in the meantime, if you are OK with that?=.

debashish-ghosh:
Yeah. Sure

debashish-ghosh:
I tested this change and it looks better now.

However, there is still an issue of error logs at each benchmark if you give a wrong counter name.
Furthermore, if you provide a comma separated list of counters, some of which are invalid (e.g. typo), then it ignores the entire list.

What I was expecting is that the error should be printed before the first benchmark is run. And whatever counters were valid, should be included in the benchmark results

PR #1137, opened 2021-05-04
Add API to benchmark allowing for custom context to be added

### LINKED ISSUE #525: Custom context fields

dominichamon:
Fixes #525

### DIFF
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -24,10 +24,7 @@
 #include <vector>
 
 #include "check.h"
-#include "commandlineflags.h"
 #include "string_util.h"
-
-DECLARE_kvpairs(benchmark_context);
 
 namespace benchmark {
 namespace internal {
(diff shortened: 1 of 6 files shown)


src/reporter.cc
```diff
@@ -71,6 +76,12 @@ void BenchmarkReporter::PrintBasicContext(std::ostream *out,
     Out << kv.first << ": " << kv.second << "\n";
   }
 
+  if (internal::global_context != nullptr) {
```
LebedevRI, line 79: I was actually expecting that `FLAGS_benchmark_context` would be parsed into `internal::global_context`.

dominichamon: oh that would be smart.

LebedevRI (commented):
Looks about right.
Tests missing.
Also, i think you may want to explicitly spell out in the docs, and test,
what happens when there's key collision.

dominichamon:
> Looks about right.
> Tests missing.
> Also, i think you may want to explicitly spell out in the docs, and test,
> what happens when there's key collision.

done, done, done :)

LebedevRI (approved):

dominichamon:
waiting for CI. Thanks, @LebedevRI.

PR #1141, opened 2021-05-06
cmake: Add explicit BENCHMARK_ENABLE_LIBPFM option

chfast:


CMakeLists.txt
```diff
@@ -270,6 +270,7 @@ cxx_feature_check(STEADY_CLOCK)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+option(BENCHMARK_ENABLE_LIBPFM "Enable performance counters provided by libpfm" OFF)
```
dominichamon, line 273: most of the options are up at the top of the file. maybe add it there instead?

chfast: ok

dominichamon:
thanks!

PR #1143, opened 2021-05-06
Cleanup Clang's -Wreserved-identifier instances

LebedevRI:
Identifiers in global namespace can not start with `_`, or end with `__`.
Clang now warns about these.
As far as i can tell these are the only places where it fires in benchmark,
other than the `googletest` code.

### DIFF
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -24,6 +24,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cinttypes>
 #include <condition_variable>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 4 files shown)


src/benchmark_register.cc
```diff
@@ -35,11 +36,6 @@
 #include <sstream>
 #include <thread>
 
-#ifndef __STDC_FORMAT_MACROS
```
dominichamon, line 38: was this added to protect on certain platforms?

LebedevRI: https://en.cppreference.com/w/cpp/types/integer

> The C99 standard suggests that C++ implementations should not define the above limit, constant, or format macros unless the macros __STDC_LIMIT_MACROS, __STDC_CONSTANT_MACROS or __STDC_FORMAT_MACROS (respectively) are defined before including the relevant C header (stdint.h or inttypes.h). This recommendation was not adopted by any C++ standard and was removed in C11. However, some implementations (such as glibc 2.17) try to apply this rule, and it may be necessary to define the __STDC macros; C++ compilers may try to work around this by automatically defining them in some circumstances.

So i think we should be fine with just using `<cinttypes>`.

dominichamon: sgtm. we'll hear back quickly if not :D

LebedevRI:

PR #1105, opened 2021-03-28
Implementation of random interleaving. 

haih-g:
See http://github.com/google/benchmark/issues/1051 for the feature requests.

Committer: Hai Huang (http://github.com/haih-g)

On branch fr-1051
Changes to be committed:
modified:   include/benchmark/benchmark.h
modified:   src/benchmark.cc
new file:   src/benchmark_adjust_repetitions.cc
new file:   src/benchmark_adjust_repetitions.h
modified:   src/benchmark_api_internal.cc
modified:   src/benchmark_api_internal.h
modified:   src/benchmark_register.cc
modified:   src/benchmark_runner.cc
modified:   src/benchmark_runner.h
modified:   test/CMakeLists.txt
new file:   test/benchmark_random_interleaving_gtest.cc

PR #1148, opened 2021-05-10
Refactor `BenchmarkInstance`

dominichamon:
Pulling out the refactoring of `BenchmarkInstance` from the rest of #1105


LebedevRI (approved):
Thank you!
I assume (and this looks like it), this is a pure NFC cleanup.
If then, LG.

PR #1149, opened 2021-05-10
Implement random interleaving (merge result)

dominichamon:
This follows on from #1105 as a merge. I can't push to the original branch to update the original PR.


dominichamon:
@haih-g i can't push back onto your branch to update the original PR. maybe you can either take from here, or we can use this as the PR to land the feature?

LebedevRI:
This can be split up further - there's essentially two features here: random interleaving of repetitions, and fragmentation of the repetitions into smaller ones.

dominichamon:
@haih-g i'm also seeing a delay before the results are printed. is that expected? i thought we still printed results as they came in. maybe i missed something in the merge?

PR #1160, opened 2021-05-31
Fix perf counter argument parsing

0xfeedface:
This fixes a bug where providing multiple perf counters via `--benchmark_perf_counters` does not parse correctly:

    % ./benchmark_bin --benchmark_perf_counters=instructions,cache-misses,branch-misses
    ...
    Unknown counter name: cache-misses,branch-misses


dominichamon:
Well spotted, and thank you for also updating the tests.

PR #1162, opened 2021-06-01
32 bit action

dmah42:


LebedevRI:
@dmah42 we now apparently have some 32-bit CI coverage via #1704,
but that is windows / MSYS2 only. Should this PR be closed?

dmah42:
> @dmah42 we now apparently have some 32-bit CI coverage via #1704, but that is windows / MSYS2 only. Should this PR be closed?

probably.  i'd need to start over anyway now if i wanted to get it working.

PR #1163, opened 2021-06-01
Random interleaving of benchmark repetitions - the sequel (fixes #1051)

### LINKED ISSUE #1051: [FR] Using Random Interleaving to reduce run-to-run variance.
**tl;dr** Introduce random interleaving that is able to reduce run-to-run variance by 40% and make it the default behavior. No change on the command line. The output is equivalent to setting `--benchmark_repetitions=12 --benchmark_min_time=0.042`. This won’t increase test execution time for most cases as `12 x 0.042 = 0.5` which is the default min time. To disable, set `--benchmark_enable_random_interleaving=false --benchmark_repetitions=1 --benchmark_min_time=0.5`. One can also adjust `--benchmark_repetitions` and `--benchmark_min_time` to balance between variance reduction and execution cost.

Being a dynamic measurement, microbenchmarking is vulnerable to noise, or run-to-run variance. Even with identical software versions and testing environments, there may be differences on the performance measurement if a microbenchmark is executed twice. Heightened run-to-run variance is a key challenge against reliable performance measurement and reduces the effectiveness of microbenchmarking as a way to track performance regressions.

One cause of the run-to-run variance is the dynamic state of the process, including but not limited to cache warmup, memory allocation, and thread scheduling. When running the benchmarking process twice, even with identical software versions and testing environments, the dynamic state of the process may change, and benchmarks that are sensitive to the states may see different performance numbers. 

We use a microbenchmark to depict the impact of the dynamic state of the process. The microbenchmark measures the latency in nanoseconds for a random API, which generates statistics on a large dataset by evaluating a model on each slice of the data. We set the `--benchmark_repetitions` to `12` and run two runs. We also set `--benchmark_min_time` to `0.5 / 12` so the total execution time does not change.

image2
*Figure 1: Run 1 - most sample points above 43M.*

image3
*Figure 2: Run 2 - most sample points below 43M.

As shown in Figure 1 and 2, in one run most of the sample points are above 43M while in the other run most sample points are below 43M. Given we are running the two runs with the identical software version and testing environment, the only difference would be the dynamic state of the benchmark process.

We propose a novel approach, Random Interleaving, to execute microbenchmarks to lower the run-to-run variance without increasing the execution cost.

In our approach, all the iterations for a single API are not executed in one-shot. Instead, the iterations are divided into multiple chunks. Each chunk will be executed in one-shot. Because we still run the same number of iterations, we don’t increase the execution cost (modulo benchmark startup/shutdown costs).

Usually a single microbenchmark executable contains multiple benchmarks. In our approach, we interleave the execution of the chunks from one benchmark with that of the chunks from others. Thus in between each chunk, we introduce a chance to change the initial state of the process for the next chunk, and the state may affect the performance of the next chunk. To better understand this, consider cache warm-up, new objects created in memory, new threads started by the scheduler, etc, each of them may change the state of the process and affect the performance. If instead we run all the chunks from the same benchmark continuously, only the first or the first couple of chunks may exercise the state changes. With interleaving, each chunk now has an opportunity to exercise the state changes. Worth noting is that in real production, no API being benchmarked is expected to be executed repeatedly in many iterations. Instead, multiple related APIs are executed in coordination to carry out a use case or use scenario. Interleaving is able to make the benchmark execution mimicking the real production environment.

In addition, the interleaving is done in a random fashion. This helps mimic that in a real production environment, different use cases or use scenarios could happen at the same time, and related APIs may be executed in various orders. 

A user is able to control the number of chunks to balance between, within a single chunk, how much time should be spent to exercise state changes, and how much time should be spent to run iterations and measure performance.

We also capture a performance data point per each chunk. Previously the microbenchmark framework only produced a single data point per repetition of a benchmark because all iterations were executed in one-shot. Now we have multiple data points per benchmark. This enables users to run A/B t-Test or u-Test (both require multiple sample points from both A and B tests) to gauge the statistical significance of performance differences.

This would be different from the [existing tooling](https://github.com/google/benchmark/blob/master/docs/tools.md#u-test) that runs a u-Test between two sets of repetitions of benchmarks. That exists to suggest any statistically significant delta between benchmarks, while this would enable statistical analysis of a single benchmark across chunks of iterations.

Furthermore, the multiple data points expose the intra-run variance which were not available before. The intra-run variance is useful information for users to understand the nature of the API, whether they are sensitive to state change in the process, and may reveal optimization opportunities.

image1
*Figure 3: Run with Random Interleaving - See sample points both above and below 43M.*

With random interleaving, we are able to exercise different dynamic states of the process, and see sample points both above and below 43M. This enables us to expose the variance within a single run. By calculating the mean of the 12 sample points per run, we are able to average out the variance within each run. We run the benchmark 50 runs, and for each run we capture the mean latency out of the 12 sample points. We then calculate the STDEV / MEAN over the 50 means. The run-to-run variance is reduced by 40%.

Config | # Runs | STDEV% | MAD%
-------|--------|----------|-------
`--benchmark_repetitions=12 --benchmark_min_time=0.042` with random interleaving | 50 | 1.01% | 0.67%
`--benchmark_repetitions=12 --benchmark_min_time=0.042` without random interleaving | 50 | 1.37% | 1.25%

Worth noting that `--benchmark_repetitions=12 --benchmark_min_time=0.042` is not equivalent to `--benchmark_repetitions=1 --benchmark_min_time=0.5`. In the former, the setup / teardown are executed 12 times, while only once in the latter. 

```c++
static void BM_foo(benchmark::State& state) {
  // Setup code
  for (auto _ : state) {
    // Benchmark iteration code
  }
  // Teardown code
}
```

Thus the table above is not necessarily a comparison between random interleaving and the default behavior. We run the comparison against the default behavior for ~800 microbenchmarks, and the average improvement on run-to-run variance is ~50%. 

Also worth noting is that random interleaving is reducing run-to-run variance, but at a potential cost of increased intra-run variance. Actually it reduces the run-to-run variance by exposing the variance within each run where the variance can be averaged out and thus not exposed as run-to-run variance. If the purpose is to track regression, reducing run-to-run variance is more important. We are not aware of any but there might be cases where the intra-run variance is interested, and random interleaving may not be the right solution for it.

Yet another note is there is no one-size fit all parameter for `--benchmark_repetitions` in random interleaving. We tested 10, 12, and 20, and 12 maximizes the variance reduction for the ~800 microbenchmarks we have tested. 

Random interleaving may increase the execution time if:

1.   The microbenchmark runs < 12 iterations by default
1.   The setup / teardown is costly

In which cases, a user wants to pick a different `--benchmark_repetitions` to balance the benefit and additional cost, if any, of random interleaving.

In general, a higher `--benchmark_repetitions` means more chances to exercise different states of the process, and results in less run-to-run variance. But because we want to keep the total execution time constant, a higher `--benchmark_repetitions` means a lower `--benchmark_min_time`. If the benchmark is expected to warm up the cache, less `--benchmark_min_time` per repetition may not be enough to warm up the cache warm-up, and the user may not measure the right thing. Furthermore, the user may end up with less number of iterations (cold cache means long latency, which in turn means less iteration in a fixed amount of time), and less number of iterations may lead to increased variance.

If possible, it is recommended to increase `--benchmark_repetitions` without decreasing the `--benchmark_min_time`. A user will spend more time to run the benchmark, but with much stabler results.

If the setup / teardown cost is high, then one may want to reduce `--benchmark_repetitions` to bring down the execution time. However, the user may end up with heightened run-to-run variance which renders the result inconclusive.

LebedevRI:
Inspired by the original implementation by Hai Huang @haih-g
from https://github.com/google/benchmark/pull/1105.

The original implementation had design deficiencies that
weren't really addressable without redesign, so it was reverted.

In essence, the original implementation consisted of two separateable parts:
* reducing the amount time each repetition is run for, and symmetrically increasing repetition count
* running the repetitions in random order

While it worked fine for the usual case, it broke down when user would specify repetitions
(it would completely ignore that request), or specified per-repetition min time (while it would
still adjust the repetition count, it would not adjust the per-repetition time,
leading to much greater run times)

Here, like i was originally suggesting in the original review, i'm separating the features,
and only dealing with a single one - running repetitions in random order.

Now that the runs/repetitions are no longer in-order, the tooling may wish to sort the output,
and indeed `compare.py` has been updated to do that: #1168.

### DIFF
--- a/docs/random_interleaving.md
+++ b/docs/random_interleaving.md
@@ -1,0 +1,13 @@
+<a name="interleaving" />
+
+# Random Interleaving
+
+[Random Interleaving](https://github.com/google/benchmark/issues/1051) is a
+technique to lower run-to-run variance. It randomly interleaves repetitions of a
+microbenchmark with repetitions from other microbenchmarks in the same benchmark
+test. Data shows it is able to lower run-to-run variance by
+[40%](https://github.com/google/benchmark/issues/1051) on average.
+
+To use, you mainly need to set `--benchmark_enable_random_interleaving=true`,
+and optionally specify non-zero repetition count `--benchmark_repetitions=9`
+and optionally decrease the per-repetition time `--benchmark_min_time=0.1`.
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -33,8 +33,10 @@
 #include <cstdlib>
 #include <fstream>
 #include <iostream>
+#include <limits>
 #include <map>
 #include <memory>
+#include <random>
 #include <string>
 #include <thread>
 #include <utility>
@@ -72,6 +74,10 @@ DEFINE_double(benchmark_min_time, 0.5);
 // The number of runs of each benchmark. If greater than 1, the mean and
 // standard deviation of the runs will be reported.
 DEFINE_int32(benchmark_repetitions, 1);
+
+// If set, enable random interleaving of repetitions of all benchmarks.
+// See http://github.com/google/benchmark/issues/1051 for details.
+DEFINE_bool(benchmark_enable_random_interleaving, false);
 
 // Report the result of each benchmark repetitions. When 'true' is specified
 // only the mean, standard deviation, and other statistics are reported for
@@ -297,23 +303,68 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
   context.name_field_width = name_field_width;
 
   // Keep track of running times of all instances of each benchmark family.
-  std::map<int /*family_index*/, std::vector<BenchmarkReporter::Run>>
-      complexity_reports;
+  std::map<int /*family_index*/, BenchmarkReporter::PerFamilyRunReports>
+      per_family_reports;
 
   if (display_reporter->ReportContext(context) &&
       (!file_reporter || file_reporter->ReportContext(context))) {
     FlushStreams(display_reporter);
     FlushStreams(file_reporter);
 
-    for (const BenchmarkInstance& benchmark : benchmarks) {
-      std::vector<BenchmarkReporter::Run>* complexity_reports_for_family =
-          nullptr;
-      if (benchmark.complexity() != oNone)
-        complexity_reports_for_family =
-            &complexity_reports[benchmark.family_index()];
+    size_t num_repetitions_total = 0;
 
-      RunResults run_results =
-          RunBenchmark(benchmark, complexity_reports_for_family);
+    std::vector<internal::BenchmarkRunner> runners;
+    runners.reserve(benchmarks.size());
+    for (const BenchmarkInstance& benchmark : benchmarks) {
+      BenchmarkReporter::PerFamilyRunReports* reports_for_family = nullptr;
+      if (benchmark.complexity() != oNone)
+        reports_for_family = &per_family_reports[benchmark.family_index()];
+
+      runners.emplace_back(benchmark, reports_for_family);
+      int num_repeats_of_this_instance = runners.back().GetNumRepeats();
+      num_repetitions_total += num_repeats_of_this_instance;
+      if (reports_for_family)
+        reports_for_family->num_runs_total += num_repeats_of_this_instance;
+    }
+    assert(runners.size() == benchmarks.size() && "Unexpected runner count.");
+
+    std::vector<int> repetition_indices;
+    repetition_indices.reserve(num_repetitions_total);
+    for (size_t runner_index = 0, num_runners = runners.size();
+         runner_index != num_runners; ++runner_index) {
+      const internal::BenchmarkRunner& runner = runners[runner_index];
+      std::fill_n(std::back_inserter(repetition_indices),
+                  runner.GetNumRepeats(), runner_index);
+    }
+    assert(repetition_indices.size() == num_repetitions_total &&
+           "Unexpected number of repetition indexes.");
+
+    if (FLAGS_benchmark_enable_random_interleaving) {
+      std::random_device rd;
+      std::mt19937 g(rd());
+      std::shuffle(repetition_indices.begin(), repetition_indices.end(), g);
+    }
+
+    for (size_t repetition_index : repetition_indices) {
+      internal::BenchmarkRunner& runner = runners[repetition_index];
+      runner.DoOneRepetition();
+      if (runner.HasRepeatsRemaining()) continue;
+      // FIXME: report each repetition separately, not all of them in bulk.
+
+      RunResults run_results = runner.GetResults();
+
+      // Maybe calculate complexity report
+      if (const auto* reports_for_family = runner.GetReportsForFamily()) {
+        if (reports_for_family->num_runs_done ==
+            reports_for_family->num_runs_total) {
+          auto additional_run_stats = ComputeBigO(reports_for_family->Runs);
+          run_results.aggregates_only.insert(run_results.aggregates_only.end(),
+                                             additional_run_stats.begin(),
+                                             additional_run_stats.end());
+          per_family_reports.erase(
+              (int)reports_for_family->Runs.front().family_index);
+        }
+      }
 
       Report(display_reporter, file_reporter, run_results);
     }
@@ -471,6 +522,7 @@ void PrintUsageAndExit() {
           "          [--benchmark_filter=<regex>]\n"
           "          [--benchmark_min_time=<min_time>]\n"
           "          [--benchmark_repetitions=<num_repetitions>]\n"
+          "          [--benchmark_enable_random_interleaving={true|false}]\n"
           "          [--benchmark_report_aggregates_only={true|false}]\n"
           "          [--benchmark_display_aggregates_only={true|false}]\n"
           "          [--benchmark_format=<console|json|csv>]\n"
@@ -495,6 +547,8 @@ void ParseCommandLineFlags(int* argc, char** argv) {
                         &FLAGS_benchmark_min_time) ||
         ParseInt32Flag(argv[i], "benchmark_repetitions",
                        &FLAGS_benchmark_repetitions) ||
+        ParseBoolFlag(argv[i], "benchmark_enable_random_interleaving",
+                      &FLAGS_benchmark_enable_random_interleaving) ||
         ParseBoolFlag(argv[i], "benchmark_report_aggregates_only",
                       &FLAGS_benchmark_report_aggregates_only) ||
         ParseBoolFlag(argv[i], "benchmark_display_aggregates_only",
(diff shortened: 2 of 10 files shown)


docs/random_interleaving.md
```diff
@@ -0,0 +1,13 @@
...
+# Random Interleaving
+
+[Random Interleaving](https://github.com/google/benchmark/issues/1051) is a
+technique to lower run-to-run variance. It randomly interleaves repetitions of a
+microbenchmark with repetitions from other microbenchmarks in the same benchmark
+test. Data shows it is able to lower run-to-run variance by
+[40%](https://github.com/google/benchmark/issues/1051) on average.
+
+To use, you mainly need to set `--benchmark_enable_random_interleaving=true`,
+and optionally specify non-zero repetition count ``--benchmark_repetitions=9``
```
haih-g, line 12: ``--benchmark_repetitions=9`` --> `--benchmark_repetitions=9`

``--benchmark_min_time=0.1`` --> `--benchmark_min_time=0.1`

haih-g:
@LebedevRI Thanks for doing this. You may also want to include the follow test cases from test/benchmark_random_interleaving_gtest.cc

```
TEST_F(BenchmarkTest, Match1) {...}
TEST_F(BenchmarkTest, Match1WithRepetition) {...}
TEST_F(BenchmarkTest, Match1WithRandomInterleaving) {...}
```

src/benchmark.cc
```diff
@@ -268,45 +307,53 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
...
-      flushStreams(file_reporter);
+    FlushStreams(display_reporter);
+    FlushStreams(file_reporter);
+
+    size_t num_repetitions_total = 0;
+
+    std::vector<internal::BenchmarkRunner> runners;
+    runners.reserve(benchmarks.size());
+    for (const BenchmarkInstance& benchmark : benchmarks) {
+      assert(!FLAGS_benchmark_enable_random_interleaving ||
```
dominichamon, line 320: this assert won't fire in release builds. it would be better to `if` and log/return.

src/benchmark.cc
```diff
@@ -268,45 +307,53 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
...
+
+    std::vector<int> repetition_indices;
+    repetition_indices.reserve(num_repetitions_total);
+    for (size_t runner_index = 0, num_runners = runners.size();
+         runner_index != num_runners; ++runner_index) {
+      const internal::BenchmarkRunner& runner = runners[runner_index];
+      std::fill_n(std::back_inserter(repetition_indices),
+                  runner.get_num_repeats(), runner_index);
+    }
+    assert(repetition_indices.size() == num_repetitions_total &&
```
dominichamon, line 337: might be better to use CHECK here.

src/benchmark_runner.cc
```diff
@@ -135,237 +137,217 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+         ((i.results.real_time_used >= 5 * min_time) && !b.use_manual_time());
+}
 
-    IterationResults i;
-    // Acquire the measurements/counters from the manager, UNDER THE LOCK!
-    {
-      MutexLock l(manager->GetBenchmarkMutex());
-      i.results = manager->results;
-    }
+void BenchmarkRunner::do_one_repetition() {
```
dominichamon, line 268: this should be `DoOneRepetition` (google style naming)

src/benchmark_runner.cc
```diff
@@ -135,237 +137,217 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+  assert(has_repeats_remaining() && "Already done all repetitions?");
```
dominichamon, line 269: CHECK?

src/benchmark_runner.cc
```diff
@@ -135,237 +137,217 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
-      memory_manager->Stop(&memory_result);
-    }
+  ++num_repetitions_done;
+}
 
-    // Ok, now actualy report.
-    BenchmarkReporter::Run report =
-        CreateRunReport(b, i.results, memory_iterations, memory_result,
-                        i.seconds, repetition_index);
+RunResults&& BenchmarkRunner::get_results() {
```
dominichamon, line 334: `GetResults`

src/benchmark_runner.h
```diff
@@ -42,6 +47,54 @@ struct RunResults {
   bool file_report_aggregates_only = false;
 };
 
+class BenchmarkRunner {
+ public:
+  BenchmarkRunner(const benchmark::internal::BenchmarkInstance& b_,
+                  std::vector<BenchmarkReporter::Run>* complexity_reports_);
```
dominichamon, line 53: style guide has changed. `complexity_reports` is non-optional so should be a non-const reference (https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs).. you can change here or we can do it in another PR.

src/benchmark.cc
LebedevRI: What are the guidelines for when to choose which? This one isn't going to be hit, it's mainly only a dev sanity check.

src/benchmark.cc
LebedevRI: Or better yet, fix this for real :)

src/benchmark_runner.h
LebedevRI: This can not be a reference.

src/benchmark.cc
dominichamon: CHECK just gives a bit nicer log on failure, but there's no specific guideline. I try to use CHECK when i think a user _might_ hit an issue.

src/benchmark.cc
```diff
@@ -297,23 +303,69 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
...
+    for (size_t repetition_index : repetition_indices) {
+      internal::BenchmarkRunner& runner = runners[repetition_index];
+      runner.DoOneRepetition();
+      if (runner.HasRepeatsRemaining()) continue;
+      // FIXME: report each repetition separately, not all of them in bulk.
+
+      RunResults run_results = runner.GetResults();
+
+      // Maybe calculate complexity report
+      if (BenchmarkReporter::PerFamilyRunReports* reports_for_family =
```
dominichamon, line 357: can this be `if (const auto* reports_for_family = runner.GetReportsForFamily())`?

i think we don't mutate, right? (and we can use auto to remove some repetition).

src/benchmark.cc
LebedevRI: This won't be hit unless the code was changed, or the code is miscompiled.

dominichamon (approved):

LebedevRI:
Thank you for the review!

PR #1165, opened 2021-06-02
Introduce per-family instance index

LebedevRI:
Much like it makes sense to enumerate all the families,
it makes sense to enumerate stuff within families.
Alternatively, we could have a global instance index,
but i'm not sure why that would be better.

This will be useful when the benchmarks are run not in order,
for the tools to sort the results properly.


LebedevRI:
I'm just going to merge this, it's rather identical to the #1164

PR #1166, opened 2021-06-03
Make 'complexity reports' cache per-family, not global

LebedevRI:
While the current variant works, it assumes that all the instances of
a single family will be run together, with nothing inbetween them.
Naturally, that won't work once the runs may be interleaved.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -265,8 +265,9 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
   BenchmarkReporter::Context context;
   context.name_field_width = name_field_width;
 
-  // Keep track of running times of all instances of current benchmark
-  std::vector<BenchmarkReporter::Run> complexity_reports;
+  // Keep track of running times of all instances of each benchmark family.
+  std::map<int /*family_index*/, std::vector<BenchmarkReporter::Run>>
+      complexity_reports;
 
   // We flush streams after invoking reporter methods that write to them. This
   // ensures users get timely updates even when streams are not line-buffered.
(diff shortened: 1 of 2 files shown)


src/benchmark_runner.cc
```diff
@@ -179,11 +180,11 @@ class BenchmarkRunner {
 
     // Maybe calculate complexity report
     if ((b.complexity() != oNone) && b.last_benchmark_instance) {
```
LebedevRI, line 182: Yes, `b.last_benchmark_instance` will need fixing too, not here.

src/benchmark.cc
```diff
@@ -265,8 +265,9 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
   BenchmarkReporter::Context context;
   context.name_field_width = name_field_width;
 
-  // Keep track of running times of all instances of current benchmark
-  std::vector<BenchmarkReporter::Run> complexity_reports;
+  // Keep track of running times of all instances of each benchmark family.
+  std::map<int /*family_index*/, std::vector<BenchmarkReporter::Run>>
```
dominichamon, line 269: if we're worried about performance, this could be a `std::vector<std::vector<BenchmarkReporter::Run>>` where the family index is the index into the outer vector.

src/benchmark_runner.cc
dominichamon: you can also, i think, replace the check against `b.complexity()` with a check against `complexity_reports != nullptr` which might be a bit simpler to reason about.

src/benchmark.cc
LebedevRI: Better yet, i don't create entries into this map unless we will need them, see code below.

src/benchmark_runner.cc
```diff
@@ -361,7 +362,7 @@ class BenchmarkRunner {
                         i.seconds, repetition_index, repeats);
 
     if (!report.error_occurred && b.complexity() != oNone)
```
dominichamon, line 364: same here, regarding checking against existence of complexity reports.

LebedevRI:
@dominichamon does this look good to you now?

src/benchmark.cc
dominichamon: yeah, map access is still _much_ slower than vector index lookup. not going to block this CL on it though as there's no proof it's slowing anything down.

dominichamon (approved):

PR #1167, opened 2021-06-03
Enable various sanitizer builds in github actions

dominichamon:


dominichamon:
failing due to existing sanitizer badness (yay?). making a new PR to try to solve that first.

PR #1170, opened 2021-06-03
Enable msan

dmah42:


LebedevRI:
Pretty sure this got merged along the way, we clearly have msan CI step now.

PR #1171, opened 2021-06-03
Use modern clang/libc++ for sanitizers

dominichamon:
* Instead of pulling from the old repo, use the new repo
* Don't use clang-3.8, use modern clang
* Update to ubuntu-latest


dominichamon:
ubuntu-latest + latest libc++ + latest libc++abi + latest clang == failing regex checks on build configuration. ugh.

LebedevRI:
Seems reasonable

PR #1179, opened 2021-06-16
Provide helpers to create integer lists for the given ranges.

PCMan:
This can be used together with ArgsProduct() to allow multiple ranges
with different multipliers and mixing dense and sparse ranges.

Example:
```
BENCHMARK(MyTest)->ArgsProduct({
  CreateRange(0, 1024, /*multi=*/32),
  CreateRange(0, 100, /*multi=*/4),
  CreateDenseRange(0, 4, /*step=*/1)
});
```

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1649,6 +1649,21 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
   BENCHMARK_UNREACHABLE();
 }
 
+// Creates a list of integer values for the given range and multiplier.
+// This can be used together with ArgsProduct() to allow multiple ranges
+// with different multiplers.
+// Example:
+// ArgsProduct({
+//   CreateRange(0, 1024, /*multi=*/32),
+//   CreateRange(0, 100, /*multi=*/4),
+//   CreateDenseRange(0, 4, /*step=*/1),
+// });
+std::vector<int64_t> CreateRange(int64_t lo, int64_t hi, int multi);
+
+// Creates a list of integer values for the given range and step.
+std::vector<int64_t> CreateDenseRange(int64_t start, int64_t limit,
+                                      int step);
+
 }  // namespace benchmark
 
 #endif  // BENCHMARK_BENCHMARK_H_
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -458,4 +458,20 @@ void ClearRegisteredBenchmarks() {
   internal::BenchmarkFamilies::GetInstance()->ClearBenchmarks();
 }
 
+std::vector<int64_t> CreateRange(int64_t lo, int64_t hi, int multi) {
+  std::vector<int64_t> args;
+  internal::AddRange(&args, lo, hi, multi);
+  return args;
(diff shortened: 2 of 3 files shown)


src/benchmark_register.cc
```diff
@@ -458,4 +458,20 @@ void ClearRegisteredBenchmarks() {
...
+
+std::vector<int64_t> CreateDenseRange(int64_t start, int64_t limit,
+                                      int64_t step) {
+  CHECK_LE(start, limit);
+  std::vector<int64_t> args;
+  for (int64_t arg = start; arg <= limit; arg += step) {
+    args.push_back({arg});
+  }
+  return args;
+}
```
dominichamon, line 475: can `DenseRange` be rewritten to call this directly?

PCMan: Yes it's possible, but this might not make the code more DRY.
This `CreateRange` returns a vector<int64> while `DenseRange() generates vector<vector<int>>`.
To do this a loop of vector.push_back is still need to convert things like {0, 1, 2, 3} to {{0}, {1}, {2}, {3}}, which is nearly the same as the original code.

PCMan: This will become something like:
```
Benchmark* Benchmark::DenseRange(int64_t start, int64_t limit, int step) {
  CHECK(ArgsCnt() == -1 || ArgsCnt() == 1);
  for (int64_t arg : CreateDenseRange(start, limit, step)) {
    args_.push_back({arg});
  }
  return this;
}
```
Nearly the same as the original, but a temp vector is added.

include/benchmark/benchmark.h
```diff
@@ -1649,6 +1649,21 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
...
+// Creates a list of integer values for the given range and multiplier.
+// This can be used together with ArgsProduct() to allow multiple ranges
+// with different multiplers.
+// Example:
+// ArgsProduct({
+//   CreateRange(0, 1024, /*multi=*/32),
+//   CreateRange(0, 100, /*multi=*/4),
+//   CreateDenseRange(0, 4, /*step=*/1),
+// });
+std::vector<int64_t> CreateRange(int64_t lo, int64_t hi, int64_t multi);
```
dominichamon, line 1661: `multi` should be `int`

include/benchmark/benchmark.h
```diff
@@ -1649,6 +1649,21 @@ inline double GetTimeUnitMultiplier(TimeUnit unit) {
...
+
+// Creates a list of integer values for the given range and step.
+std::vector<int64_t> CreateDenseRange(int64_t start, int64_t limit,
+                                      int64_t step);
```
dominichamon, line 1665: `step` should be `int`

include/benchmark/benchmark.h
PCMan: Fixed. Thanks!

include/benchmark/benchmark.h
PCMan: Fixed. Thanks!

dominichamon:
thanks!

PR #1182, opened 2021-06-18
Add missing trailing commas

### LINKED ISSUE #1181: [BUG] Custom context with JSON file reporter produces invalid JSON
**Describe the bug**
Adding a custom context (`--benchmark_context`) and using the JSON reporter (`--benchmark_out_format`), the line preceding the custom context entry does not have a comma at the end resulting in an invalid JSON file.

**System**
  - OS: macOS
  - Compiler and version: Apple clang version 12.0.5

**To reproduce**
1. sync to commit 80a6261
2. cmake -DCMAKE_BUILD_TYPE=Release ...
3. make

**Expected behavior**
The line before the custom context entry should have a comman (`,`) at the end if there is a context entry.

**Screenshots**
Screenshot 2021-06-18 at 15 50 55

**Additional context**
N/A

dominichamon:
Fixes #1181

### DIFF
--- a/src/json_reporter.cc
+++ b/src/json_reporter.cc
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "benchmark/benchmark.h"
-#include "complexity.h"
-
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
@@ -25,6 +22,8 @@
 #include <tuple>
 #include <vector>
 
+#include "benchmark/benchmark.h"
+#include "complexity.h"
 #include "string_util.h"
 #include "timers.h"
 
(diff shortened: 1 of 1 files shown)


src/json_reporter.cc
```diff
@@ -162,11 +162,11 @@ bool JSONReporter::ReportContext(const Context& context) {
 #else
   const char build_type[] = "debug";
 #endif
-  out << indent << FormatKV("library_build_type", build_type) << "\n";
+  out << indent << FormatKV("library_build_type", build_type) << ",\n";
```
miki-shapr, line 165: Unfortunately JSON does not allow trailing commas, so this would still result in invalid JSON files.

src/json_reporter.cc
```diff
@@ -162,11 +162,11 @@ bool JSONReporter::ReportContext(const Context& context) {
...
 
   if (internal::global_context != nullptr) {
     for (const auto& kv: *internal::global_context) {
-      out << indent << FormatKV(kv.first, kv.second) << "\n";
+      out << indent << FormatKV(kv.first, kv.second) << ",\n";
```
miki-shapr, line 169: Same as above.

miki-shapr (changes requested):
Only append a comma at the end of a line if it is followed by another entry.

dominichamon:
> Only append a comma at the end of a line if it is followed by another entry.

i think i got it. can you approve if you think so @miki-shapr ?

miki-shapr (approved):
LGTM 👍

Issue #910, opened 2019-11-26
Will google benchmark be adding an Abseil dependency?


alexhadd:
I'd like to use this library for micro-benchmarks but I am concerned about the possibility of this library taking on a dependency on Abseil, like Google Test is doing soon. On the Abseil website, i[t is mentioned](https://abseil.io/about/philosophy#compatibility-with-google-oss-projects) that many Google OSS projects will be incorporating Abseil. Is Google Benchmark planning on adding Abseil soon?

LebedevRI:
Almost 2 years ago I wanted to use abseil in my library (https://github.com/darktable-org/rawspeed/issues/122),
but a lot of water has flown since then, many things have happened;
also it's still not packaged in at least debian - to me personally that speaks volumes.
So by now i **_personally_** i hope nothing will use it..
Not that there is any chance of that happening, or much/any choices for avoiding it after it happens :)

> library taking on a dependency on Abseil, like Google Test is doing soon

Uh oh.

dominichamon:
It's been on my backlog/wishlist for a while because it has some nice features that would allow us to remove some bits we've implemented here (flags, string formatting). However, I believe that would first require us to move wholesale to bazel (i haven't found a nice cmake abseil hook) which isn't likely to happen soon.

dominichamon:
https://abseil.io/docs/cpp/quickstart-cmake

some of the code we have is starting to deviate from the absl versions (which we use internally) and it would be really good to converge.

now there's a cmake option may be the time to adopt it for command line  flags and logging, at least.

PR #1183, opened 2021-06-22
Introduce a dependency on Abseil

mtrofin:
The main motivation is enabling the benchmark library to play well with
hosts that use Abseil, especially for command line flag functionality.
This also opens up the possibility of reusing functionality such as
string utils, concurrency, or logging.

This was previously discussed in issue #910. This patch:

- adds the dependency to Abseil. The library will either be downloaded
(at a fixed, specific commit) or the user may specify their own location
via the ABSEIL_PATH cmake flag

- replaces flag functionality with Abseil's.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,7 @@
 workspace(name = "com_github_google_benchmark")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 http_archive(
     name = "rules_cc",
@@ -9,11 +10,10 @@ http_archive(
     sha256 = "d7dc12c1d5bc1a87474de8e3d17b7731a4dcebcfb8aa3990fe8ac7734ef12f2f",
 )
 
-http_archive(
+git_repository(
     name = "com_google_absl",
-    sha256 = "f41868f7a938605c92936230081175d1eae87f6ea2c248f41077c8f88316f111",
-    strip_prefix = "abseil-cpp-20200225.2",
-    urls = ["https://github.com/abseil/abseil-cpp/archive/20200225.2.tar.gz"],
+    commit = "4a23151e7ee089f54f0575f0a6d499e3a3fb6728",
+    remote = "https://github.com/abseil/abseil-cpp.git",
 )
 
 http_archive(
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -1,0 +1,22 @@
+include(FetchContent)
+
+set(ABSL_ENABLE_INSTALL ON)
+if (NOT (DEFINED ABSEIL_PATH))
+  message("ABSEIL_PATH not provided. Downloading.")
+  FetchContent_Declare(
+    abseil
+    GIT_REPOSITORY    https://github.com/abseil/abseil-cpp.git
+    GIT_TAG           4a23151e7ee089f54f0575f0a6d499e3a3fb6728
+  )
+  FetchContent_GetProperties(abseil)
+  if(NOT abseil_POPULATED)
+      FetchContent_Populate(abseil)
+      add_subdirectory(${abseil_SOURCE_DIR} ${abseil_BINARY_DIR})
+      set(ABSEIL_PATH ${abseil_SOURCE_DIR})
+  endif()
+else()
+    add_subdirectory(${ABSEIL_PATH} ${CMAKE_CURRENT_BINARY_DIR}/abseil-build)
+endif()
+
+include_directories(${ABSEIL_PATH})
+list(APPEND BENCHMARK_CXX_LIBRARIES absl::flags absl::flags_parse)
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -7,10 +7,14 @@ if (NOT (DEFINED ABSEIL_PATH))
   execute_process(COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" .
     RESULT_VARIABLE result 
     WORKING_DIRECTORY ${ABSEIL_PREFIX})
-  execute_process(COMMAND ${CMAKE_COMMAND} --build .
-    WORKING_DIRECTORY ${ABSEIL_PREFIX})
   if(result)
     message(FATAL_ERROR "CMake step for abseil failed: ${result}")
+  endif()
+
+  execute_process(COMMAND ${CMAKE_COMMAND} --build .
+    WORKING_DIRECTORY ${ABSEIL_PREFIX} RESULT_VARIABLE result)
+  if(result)
+    message(FATAL_ERROR "Download step for abseil failed: ${result}")
   endif()
 
   set(ABSEIL_PATH ${ABSEIL_PREFIX}/abseil-src)
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -18,6 +18,14 @@ if (NOT (DEFINED ABSEIL_PATH))
   endif()
 
   set(ABSEIL_PATH ${ABSEIL_PREFIX}/abseil-src)
+
+  execute_process(COMMAND ${PYTHON_EXECUTABLE} 
+    ${ABSEIL_PATH}/absl/copts/generate_copts.py 
+    RESULT_VARIABLE result)
+  if(result)
+    message(FATAL_ERROR "Config step for abseil failed: ${result}")
+  endif()
+
   add_subdirectory(${ABSEIL_PATH} ${ABSEIL_PREFIX}/abseil-build)
 else()
     add_subdirectory(${ABSEIL_PATH} ${CMAKE_CURRENT_BINARY_DIR}/abseil-build)
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -1,19 +1,20 @@
-include(FetchContent)
 
 set(ABSL_ENABLE_INSTALL ON)
 if (NOT (DEFINED ABSEIL_PATH))
   message("ABSEIL_PATH not provided. Downloading.")
-  FetchContent_Declare(
-    abseil
-    GIT_REPOSITORY    https://github.com/abseil/abseil-cpp.git
-    GIT_TAG           4a23151e7ee089f54f0575f0a6d499e3a3fb6728
-  )
-  FetchContent_GetProperties(abseil)
-  if(NOT abseil_POPULATED)
-      FetchContent_Populate(abseil)
-      add_subdirectory(${abseil_SOURCE_DIR} ${abseil_BINARY_DIR})
-      set(ABSEIL_PATH ${abseil_SOURCE_DIR})
+  set(ABSEIL_PREFIX "${benchmark_BINARY_DIR}/third_party/abseil")
+  configure_file(${benchmark_SOURCE_DIR}/cmake/Abseil.cmake.in ${ABSEIL_PREFIX}/CMakeLists.txt @ONLY)
+  execute_process(COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" . 
+    RESULT_VARIABLE result 
+    WORKING_DIRECTORY ${ABSEIL_PREFIX})
+  execute_process(COMMAND ${CMAKE_COMMAND} --build .
+    WORKING_DIRECTORY ${ABSEIL_PREFIX})
+  if(result)
+    message(FATAL_ERROR "CMake step for abseil failed: ${result}")
   endif()
+
+  set(ABSEIL_PATH ${ABSEIL_PREFIX}/abseil-src)
+  add_subdirectory(${ABSEIL_PATH} ${ABSEIL_PREFIX}/abseil-build)
 else()
     add_subdirectory(${ABSEIL_PATH} ${CMAKE_CURRENT_BINARY_DIR}/abseil-build)
 endif()
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -19,13 +19,6 @@ if (NOT (DEFINED ABSEIL_PATH))
 
   set(ABSEIL_PATH ${ABSEIL_PREFIX}/abseil-src)
 
-  execute_process(COMMAND ${PYTHON_EXECUTABLE} 
-    ${ABSEIL_PATH}/absl/copts/generate_copts.py 
-    RESULT_VARIABLE result)
-  if(result)
-    message(FATAL_ERROR "Config step for abseil failed: ${result}")
-  endif()
-
   add_subdirectory(${ABSEIL_PATH} ${ABSEIL_PREFIX}/abseil-build)
 else()
     add_subdirectory(${ABSEIL_PATH} ${CMAKE_CURRENT_BINARY_DIR}/abseil-build)
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -4,7 +4,7 @@ if (NOT (DEFINED ABSEIL_PATH))
   message("ABSEIL_PATH not provided. Downloading.")
   set(ABSEIL_PREFIX "${benchmark_BINARY_DIR}/third_party/abseil")
   configure_file(${benchmark_SOURCE_DIR}/cmake/Abseil.cmake.in ${ABSEIL_PREFIX}/CMakeLists.txt @ONLY)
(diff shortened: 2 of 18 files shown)


dominichamon:
I think you need to add a bazel dependency on @com_google_absl in the BUILD file.

cmake/Abseil.cmake
```diff
@@ -0,0 +1,22 @@
+include(FetchContent)
+
+set(ABSL_ENABLE_INSTALL ON)
```
dominichamon, line 3: should we use `BENCHMARK_DOWNLOAD_DEPENDENCIES` too from GoogleTest.cmake?

cmake/Abseil.cmake
```diff
@@ -0,0 +1,22 @@
...
+if (NOT (DEFINED ABSEIL_PATH))
+  message("ABSEIL_PATH not provided. Downloading.")
+  FetchContent_Declare(
```
dominichamon, line 6: added in 3.11 and we don't require that (to support older linux distributions).

i think you want to follow what the GoogleTest.cmake script does.

src/commandlineflags.h
```diff
@@ -5,30 +5,6 @@
...
-  int32_t FLAG(name) = benchmark::Int32FromEnv(#name, default_val)
-#define DEFINE_double(name, default_val) \
-  double FLAG(name) = benchmark::DoubleFromEnv(#name, default_val)
-#define DEFINE_string(name, default_val) \
-  std::string FLAG(name) = benchmark::StringFromEnv(#name, default_val)
-#define DEFINE_kvpairs(name, default_val)         \
-  std::map<std::string, std::string> FLAG(name) = \
-      benchmark::KvPairsFromEnv(#name, default_val)
-
 namespace benchmark {
```
dominichamon, line 8: do we need the rest of this file?

we used to set the default values from the environment... does abseil do that?

cmake/Abseil.cmake
mtrofin: I considered this. The way I see it is that GoogleTest is not a required dependency - it's only so if you want tests enabled. Abseil is a required dependency.

cmake/Abseil.cmake
dominichamon: https://github.com/google/benchmark/issues/603 for the last discussion about bumping cmake versions.

mtrofin: Ugh. Lots of cmake wizardry. Got it.

src/commandlineflags.h
mtrofin: I can look into it, my plan was to tackle any more involved logic as a subsequent step; e.g. flag validation (as well as kvpairs)

src/perf_counters.cc
```diff
@@ -33,7 +33,7 @@ const bool PerfCounters::kSupported = true;
 bool PerfCounters::Initialize() { return pfm_initialize() == PFM_SUCCESS; }
 
 PerfCounters PerfCounters::Create(
-    const std::vector<std::string>& counter_names) {
+    const std::vector<std::string>&& counter_names) {
```
pleroy, line 36: At the risk of showing my incompetence, could you explain these changes from `const&` to `const&&`?  Const rvalue references are pretty uncommon in my experience, and I'd be surprised if Abseil was inviting their usage.  At a minimum, a comment would be welcome.

mtrofin: Good catch, it's not needed.

mtrofin:
Punting this to when we move past Ubuntu 14.04 LTS - absl requires a compiler version higher than the version of either gnu/clang on that Linux distro.

oontvoo:
Hi, any idea on exactly when it would be that we move past Ubuntu 14.04 LTS?
It's a pretty old version of ubuntu by now 

This patch would benefit a lot of things internally.
Thanks

dominichamon:
https://google.github.io/benchmark/dependencies.html

i believe we can move past 14.04 at this point by quite some way. perhaps we can even drop support for older compilers...

dominichamon:
we now have updated the dependencies, if someone wants to have another go at this. it will up the compiler version, which may mean the removal of some bots that test older versions.

mtrofin:
I can resurrect the patch.

mtrofin:
reopening

WORKSPACE
```diff
@@ -9,11 +10,10 @@ http_archive(
     sha256 = "d7dc12c1d5bc1a87474de8e3d17b7731a4dcebcfb8aa3990fe8ac7734ef12f2f",
 )
 
-http_archive(
+git_repository(
```
dominichamon, line 13: https://docs.bazel.build/versions/main/external.html#repository-rules suggests a preference for `http_archive`

src/benchmark.cc
```diff
@@ -599,21 +569,20 @@ int InitializeStreams() {
...
-  internal::ParseCommandLineFlags(argc, argv);
-  internal::LogLevel() = FLAGS_v;
+std::vector<char*> Initialize(int* argc, char** argv) {
+  using namespace benchmark;
+  BenchmarkReporter::Context::executable_name =
+      (argc && *argc > 0) ? argv[0] : "unknown";
+
+  auto unparsed = absl::ParseCommandLine(*argc, argv);
+  *argc = static_cast<int>(unparsed.size());
+  benchmark::internal::ValidateCommandLineFlags();
```
dominichamon, line 579: nit: don't need the `benchmark::` prefix here

src/mutex.h
```diff
@@ -2,32 +2,16 @@
...
 #include "check.h"
 
-// Enable thread safety attributes only with clang.
-// The attributes can be safely erased when compiling with other compilers.
-#if defined(HAVE_THREAD_SAFETY_ATTRIBUTES)
-#define THREAD_ANNOTATION_ATTRIBUTE_(x) __attribute__((x))
-#else
-#define THREAD_ANNOTATION_ATTRIBUTE_(x)  // no-op
-#endif
+#define THREAD_ANNOTATION_ATTRIBUTE_ THREAD_ANNOTATION_ATTRIBUTE__
```
oontvoo, line 9: Not needed?

src/benchmark.cc
```diff
@@ -56,74 +59,80 @@
...
-// pairs. Kept internal as it's only used for parsing from env/command line.
-BM_DEFINE_kvpairs(benchmark_context, {});
-
-// The level of verbose logging to output
-BM_DEFINE_int32(v, 0);
+ABSL_FLAG(
+    bool, benchmark_list_tests, false,
+    "Print a list of benchmarks. This option overrides all other options.");
+
+ABSL_FLAG(std::string, benchmark_filter, ".",
```
oontvoo, line 66: Please change this back to "" (empty).

(We intentionally changed this from "." (ie all) to empty (no benchmarks )to support various internal users)

oontvoo (changes requested):

LebedevRI:
@mtrofin since you are the one who last updated `benchmark` in llvm, i feel like this question is uniquely suited for you:
how do you envision the future of the library in llvm?

mtrofin:
> @mtrofin since you are the one who last updated `benchmark` in llvm, i feel like this question is uniquely suited for you: how do you envision the future of the library in llvm?

Not sure that qualifies me in any particular way, so apologies if the following will disappoint :) Also, if it's unrelated to this patch, we can probably start a thread on llvm-discuss or wherever it'd be more appropriate for llvm-test-suite. I remember someone (maybe you) pointing out there are other copies in llvm elsewhere, so maybe we should talk within the community about consolidating them, if there's value in that.

To try and answer your question; for context, I pulled it in llvm-test-suite for 2 reasons: the main one was that I wanted to leverage perf-counters for experiments. The second was the observation that folks were making changes here and then cherry-picking them over to the llvm-test-suite copy, but other features / fixes weren't pulled over to llvm-test-suite (so more code divergence, which seemed like 'a bad thing'). This gives me the limited perspective through which I'd say we'd keep the llvm-test-suite copy sync-ed whenever folks in the second group wanted new features enabled.

This doesn't address the multiple copies of benchmark in llvm, though, nor does it have a reasoned position as to why do it this way (i.e. pull when needed), other than that 'it's similar in spirit to how it's been done so far' (with the benefit of not adding to code divergence). So maybe a more appropriate way would be to sync periodically? WDYT?

src/mutex.h
mtrofin: keeps the changes below simpler. If you prefer, I can remove it and replace THREAD_ANNOTATION_ATTRIBUTE_ with THREAD_ANNOTATION_ATTRIBUTE__

src/benchmark.cc
mtrofin: sorry, must have missed it in the rebase (was going to do a read-over, but wanted to make sure I resurrect the pull first). Thanks for catching this!

LebedevRI:
I'm asking because my opinion on this change keeps fluctuating between:
* meh, i'll manage
* this greatly simplifies upstreaming for changes! (i.e. either soft fork or hard fork)

src/mutex.h
oontvoo: if it's not too much trouble, perhaps it'd be great to do that (since you're already overhauling things here) :)
(Optional request, though)

oontvoo:
> > I'm asking because my opinion on this change keeps fluctuating between:
> > 
> > @mtrofin since you are the one who last updated benchmark in llvm, i feel like this question is uniquely suited for you:
>> how do you envision the future of the library in llvm?

On related topic, with this patch going in, doesn't it mean you'd have to get absl as an LLVM's dependency too? (it isn't nowadays, for various reasons). ... may've opened up a can of worms 🤔

mtrofin:
> > > I'm asking because my opinion on this change keeps fluctuating between:
> > > @mtrofin since you are the one who last updated benchmark in llvm, i feel like this question is uniquely suited for you:
> > > how do you envision the future of the library in llvm?
> 
> On related topic, with this patch going in, doesn't it mean you'd have to get absl as an LLVM's dependency too? (it isn't nowadays, for various reasons). ... may've opened up a can of worms 🤔

The dependency would kick in if LLVM_BUILD_BENCHMARKS were enabled; by default, it'd be downloaded as a build dependency.

So I think there would be at least 2 problems:
- for folks that want LLVM_BUILD_BENCHMARKS on, but don't want auto-downloads; 
- if there are build issues or other issues you mentioned
not sure what the precedent is for the former; on the latter, do you know what the issues may be?

oontvoo:
> > > > I'm asking because my opinion on this change keeps fluctuating between:
> > > > @mtrofin since you are the one who last updated benchmark in llvm, i feel like this question is uniquely suited for you:
> > > > how do you envision the future of the library in llvm?
> > 
> > 
> > On related topic, with this patch going in, doesn't it mean you'd have to get absl as an LLVM's dependency too? (it isn't nowadays, for various reasons). ... may've opened up a can of worms 🤔
> 
> The dependency would kick in if LLVM_BUILD_BENCHMARKS were enabled; by default, it'd be downloaded as a build dependency.
> 
> So I think there would be at least 2 problems:
> 
> * for folks that want LLVM_BUILD_BENCHMARKS on, but don't want auto-downloads;
> * if there are build issues or other issues you mentioned
>   not sure what the precedent is for the former; on the latter, do you know what the issues may be?
Nothing concrete ( was thinking of the supported gcc versions ... abseil seems to stop support older gcc sooner but looking at the llvm's doc now, it's basically the same).
So nothing that I can see yet :)

dominichamon:
note: you'll need to remove those failing actions if we can't support 14.04/16.04 any more.

mtrofin:
> I'm asking because my opinion on this change keeps fluctuating between:
> 
> * meh, i'll manage
> * this greatly simplifies upstreaming for changes! (i.e. either soft fork or hard fork)

I see - IIUC, you're not pushing back with the abseil adoption, but (combining with what @oontvoo was saying) we should consider what we do with the other copy in llvm.

@dominichamon, I think we should circle up to llvm-dev the abseil dep, because of the potential impact to the llvm-proper package. I'll start a thread.

dominichamon:
> > I'm asking because my opinion on this change keeps fluctuating between:
> > 
> > * meh, i'll manage
> > * this greatly simplifies upstreaming for changes! (i.e. either soft fork or hard fork)
> 
> I see - IIUC, you're not pushing back with the abseil adoption, but (combining with what @oontvoo was saying) we should consider what we do with the other copy in llvm.
> 
> @dominichamon, I think we should circle up to llvm-dev the abseil dep, because of the potential impact to the llvm-proper package. I'll start a thread.

sgtm

oontvoo:
> > > I'm asking because my opinion on this change keeps fluctuating between:
> > > 
> > > * meh, i'll manage
> > > * this greatly simplifies upstreaming for changes! (i.e. either soft fork or hard fork)
> > 
> > 
> > I see - IIUC, you're not pushing back with the abseil adoption, but (combining with what @oontvoo was saying) we should consider what we do with the other copy in llvm.
> > @dominichamon, I think we should circle up to llvm-dev the abseil dep, because of the potential impact to the llvm-proper package. I'll start a thread.
> 
> sgtm

+1

cmake/Abseil.cmake
```diff
@@ -0,0 +1,29 @@
+set(ABSL_ENABLE_INSTALL ON CACHE BOOL "" FORCE)
+if (NOT (DEFINED ABSEIL_PATH))
+  message("ABSEIL_PATH not provided. Downloading.")
+  set(ABSEIL_PREFIX "${benchmark_BINARY_DIR}/third_party/abseil")
+  configure_file(${benchmark_SOURCE_DIR}/cmake/Abseil.cmake.in ${ABSEIL_PREFIX}/CMakeLists.txt @ONLY)
+  execute_process(COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" .
```
petrhosek, line 6: This doesn't pass through CMake options to the Abseil build and instead relies on defaults. For example, CMake will use `/usr/bin/cc` and `/usr/bin/c++` as the default compiler on UNIX-like systems but that may be undesirable. In our case, it's going to fail because those binaries don't exists on our bots at all. Rather, we always specify the compiler manually via `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER`. The same applies to all other tools. Ideally we would pass all of these variables from the benchmark build to the abseil build.

cmake/Abseil.cmake
```diff
@@ -0,0 +1,29 @@
+set(ABSL_ENABLE_INSTALL ON CACHE BOOL "" FORCE)
+if (NOT (DEFINED ABSEIL_PATH))
+  message("ABSEIL_PATH not provided. Downloading.")
+  set(ABSEIL_PREFIX "${benchmark_BINARY_DIR}/third_party/abseil")
+  configure_file(${benchmark_SOURCE_DIR}/cmake/Abseil.cmake.in ${ABSEIL_PREFIX}/CMakeLists.txt @ONLY)
```
petrhosek, line 5: What's the motivation for using the template here rather than including `ExternalProject_Add` here directly?

gjasny:
Hello,

some time ago I noticed that an LTS abseil used within my project and the webrtc abseil version generate the same symbols despite the goal of abseil to make LTS versions co-linkable with other abseil versions. The filed issue https://github.com/abseil/abseil-cpp/issues/851 does not see much traction.

I could imagine that as long as the issue is not resolved mixing webrtc, grpc, and google-benchmark might lead into an interesting dependency chaos.

Thanks,
Gregor

LebedevRI:
> @LebedevRI , you also mentioned a potential upside

My upside was that i would not have to upstream my changes the library once i fork it over this dependency.
Somehow i suspect i'm not the only one, uhm, unhappy about the proposed dependency,
so llvm and whoever else are welcomed to switch upstreams.

For the life of me, i just don't understand why google thinks that forcing changes down people's throats
is a good strategy, but then this is by far not the first time it happens (see numerous google chrome fiaskos).

oontvoo:
> @oontvoo, what was the main motivation you'd benefit from the abseil dep? [...] @dominichamon , IIRC the main problem we hit when pulling in our repo was around flags, is that correct?

The #1 motivation was flags, yes. (Internally we have an ABSL_FLAG --benchmark_filter, whereas the OSS benchmark here uses its own facility)

Additionally, there are other features from absl that we could benefit, (for eg., most recently:  https://github.com/google/benchmark/pull/1230 )

oontvoo:
> @oontvoo, what was the main motivation you'd benefit from the abseil dep? [...] @dominichamon , IIRC the main problem we hit when pulling in our repo was around flags, is that correct?

The #1 motivation was flags, yes. (Internally we have an ABSL_FLAG --benchmark_filter, whereas the OSS benchmark here uses its own facility)

Additionally, there are other features from absl that we could benefit, (for eg., most recently:  https://github.com/google/benchmark/pull/1230 )

PR #1184, opened 2021-06-23
Fix potential memory leaks reported by valgrind

amadio:
This is what valgrind shows:
```
MPK ==2243078== 24 bytes in 1 blocks are still reachable in loss record 1 of 4
==2243078==    at 0x484005F: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2243078==    by 0x166473: allocate (new_allocator.h:121)
==2243078==    by 0x166473: allocate (allocator.h:173)
==2243078==    by 0x166473: allocate (alloc_traits.h:460)
==2243078==    by 0x166473: _M_allocate (stl_vector.h:346)
==2243078==    by 0x166473: _M_allocate (stl_vector.h:343)
==2243078==    by 0x166473: _M_create_storage (stl_vector.h:361)
==2243078==    by 0x166473: _Vector_base (stl_vector.h:305)
==2243078==    by 0x166473: vector (stl_vector.h:524)
==2243078==    by 0x166473: GetLoadAvg (sysinfo.cc:682)
==2243078==    by 0x166473: benchmark::CPUInfo::CPUInfo() (sysinfo.cc:707)
==2243078==    by 0x1665A4: benchmark::CPUInfo::Get() (sysinfo.cc:698)
==2243078==    by 0x15D628: benchmark::BenchmarkReporter::Context::Context() (reporter.cc:83)
==2243078==    by 0x130B13: RunBenchmarks (benchmark.cc:239)
==2243078==    by 0x130B13: benchmark::RunSpecifiedBenchmarks(benchmark::BenchmarkReporter*, benchmark::BenchmarkReporter*) (benchmark.cc:404)
==2243078==    by 0x111AB3: main (math_benchmarks.cc:217)
==2243078== 
MPK ==2243078== 32 bytes in 1 blocks are still reachable in loss record 2 of 4
==2243078==    at 0x484005F: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2243078==    by 0x16235C: benchmark::SystemInfo::Get() (sysinfo.cc:711)
==2243078==    by 0x15D630: benchmark::BenchmarkReporter::Context::Context() (reporter.cc:83)
==2243078==    by 0x130B13: RunBenchmarks (benchmark.cc:239)
==2243078==    by 0x130B13: benchmark::RunSpecifiedBenchmarks(benchmark::BenchmarkReporter*, benchmark::BenchmarkReporter*) (benchmark.cc:404)
==2243078==    by 0x111AB3: main (math_benchmarks.cc:217)
==2243078== 
MPK ==2243078== 72 bytes in 1 blocks are still reachable in loss record 3 of 4
==2243078==    at 0x484005F: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2243078==    by 0x166599: benchmark::CPUInfo::Get() (sysinfo.cc:698)
==2243078==    by 0x15D628: benchmark::BenchmarkReporter::Context::Context() (reporter.cc:83)
==2243078==    by 0x130B13: RunBenchmarks (benchmark.cc:239)
==2243078==    by 0x130B13: benchmark::RunSpecifiedBenchmarks(benchmark::BenchmarkReporter*, benchmark::BenchmarkReporter*) (benchmark.cc:404)
==2243078==    by 0x111AB3: main (math_benchmarks.cc:217)
==2243078== 
MPK ==2243078== 192 bytes in 1 blocks are still reachable in loss record 4 of 4
==2243078==    at 0x484005F: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2243078==    by 0x166AA3: allocate (new_allocator.h:121)
==2243078==    by 0x166AA3: allocate (allocator.h:173)
==2243078==    by 0x166AA3: allocate (alloc_traits.h:460)
==2243078==    by 0x166AA3: _M_allocate (stl_vector.h:346)
==2243078==    by 0x166AA3: _M_allocate (stl_vector.h:343)
==2243078==    by 0x166AA3: void std::vector >::_M_realloc_insert(__gnu_cxx::__normal_iterator > >, benchmark::CPUInfo::CacheInfo const&) (vector.tcc:440)
==2243078==    by 0x165E01: push_back (stl_vector.h:1198)
==2243078==    by 0x165E01: benchmark::(anonymous namespace)::GetCacheSizesFromKVFS() (sysinfo.cc:286)
==2243078==    by 0x166447: GetCacheSizes (sysinfo.cc:414)
==2243078==    by 0x166447: benchmark::CPUInfo::CPUInfo() (sysinfo.cc:705)
==2243078==    by 0x1665A4: benchmark::CPUInfo::Get() (sysinfo.cc:698)
==2243078==    by 0x15D628: benchmark::BenchmarkReporter::Context::Context() (reporter.cc:83)
==2243078==    by 0x130B13: RunBenchmarks (benchmark.cc:239)
==2243078==    by 0x130B13: benchmark::RunSpecifiedBenchmarks(benchmark::BenchmarkReporter*, benchmark::BenchmarkReporter*) (benchmark.cc:404)
==2243078==    by 0x111AB3: main (math_benchmarks.cc:217)
==2243078== 
```

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -706,8 +706,8 @@ std::vector<double> GetLoadAvg() {
 }  // end namespace
 
 const CPUInfo& CPUInfo::Get() {
-  static const CPUInfo* info = new CPUInfo();
-  return *info;
+  static const CPUInfo info;
+  return info;
 }
 
 CPUInfo::CPUInfo()
@@ -718,8 +718,8 @@ CPUInfo::CPUInfo()
       load_avg(GetLoadAvg()) {}
 
 const SystemInfo& SystemInfo::Get() {
-  static const SystemInfo* info = new SystemInfo();
-  return *info;
+  static const SystemInfo info;
+  return info;
 }
 
 SystemInfo::SystemInfo() : name(GetSystemName()) {}


src/sysinfo.cc
```diff
@@ -706,8 +706,8 @@ std::vector<double> GetLoadAvg() {
 }  // end namespace
 
 const CPUInfo& CPUInfo::Get() {
-  static const CPUInfo* info = new CPUInfo();
-  return *info;
+  static const CPUInfo info;
```
dominichamon, line 709: this is not safe as it has a non-trivial destructor. The destructor will be executed by the thread which starts the termination of the program, but other threads may not have terminated yet and _might_ still be referencing it.

The only safe way to manage this is having safe initialization with no destruction. Either declaring things as `constexpr`, or heap allocating the static object.

dominichamon (changes requested):

src/sysinfo.cc
amadio: Yes, I agree that that destruction can become a problem with this version of the changes. However, if we can find a solution that does not trip valgrind into thinking there is a memory leak, that would be better. I don't think using `constexpr` is possible, because `CPUInfo` queries information about the hardware at runtime. Maybe a call to delete the allocated object can be registered with `atexit()`?

dominichamon: that's pretty much how the global destructors run now (hand waving around a bit) and it suffers from the same issue.

we just don't know what people are doing with the library and threads that could cause issues.

having a way to tell valgrind "this is not a leak" would be great. with LSAN we can pass in a [leak suppression file](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer) but i don't think valgrind has something similar, does it?

amadio: It does, you can pass a leak suppression file. I guess that if there's no easy way to silence valgrind without causing other problems, then I will just apply this change locally within my project, as I always run the benchmarks with a single thread there and this change is not an issue.

dominichamon:
i'm going to close this out. i appreciate you trying to help, but the change just isn't the right approach to sorting out the leak.

 i think maybe we could have a leak suppression file checked in but it would also need some documentation to go with it.

amadio:
Ok, no problem, I've applied the changed locally, as I don't need multi-threading with benchmark. One possible solution, however, could be to just replace `new CPUInfo()` with a `std::shared_pointer` or `std::unique_pointer` holding the `CPUInfo`. If you think that would work, I can check if valgrind still would show warnings and if not, open a new pull request with the change.

dominichamon:
sadly neither would work. both have non-trivial destructors.

PR #1190, opened 2021-06-26
Use C++11 atomic_signal_fence for ClobberMemory

mtrofin:


dominichamon:
we avoid C++11 in public headers until we can guarantee support in the older compilers we support.

dominichamon:
> we avoid C++11 in public headers until we can guarantee support in the older compilers we support.

apparently this ship sailed a while ago :D

PR #1204, opened 2021-07-20
Add a temporary `--benchmarks` flag for internal compatibility

oontvoo:
Details:

Internally, we use `--benchmarks` flag to specify which benchmarks to run. In the OSS version here, however, there is only `--benchmark_filter` (which does the same thing).
This patch added the alias so that we could gradually migrate users over to using `--benchmark_filter`

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -64,6 +64,11 @@ DEFINE_bool(benchmark_list_tests, false);
 // linked into the binary are run.
 DEFINE_string(benchmark_filter, ".");
 
+// Alias of benchmark_filter.
+// FIXME(vyng) Remove this once we've migrated everyone to using
+// benchmark_filter
+DEFINE_string(benchmarks, ".");
+
 // Minimum number of seconds we should run benchmark before results are
 // considered significant.  For cpu-time based tests, this is the lower bound
 // on the total cpu time used by all threads that make up the test.  For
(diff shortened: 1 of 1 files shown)


oontvoo:
@dominichamon

LebedevRI:
This is missing some words.

oontvoo:
> This is missing some words.

Thanks - it got truncated because the sentence was too long. I've updated it to be shorter

LebedevRI:
> > This is missing some words.
> 
> Thanks - it got truncated because the sentence was too long. I've updated it to be shorter

I was mainly talking about the PR description

oontvoo:
> > > This is missing some words.
> > 
> > 
> > Thanks - it got truncated because the sentence was too long. I've updated it to be shorter
> 
> I was mainly talking about the PR description

Updated the first comment to have more details.

pleroy:
This PR looks like a terrible idea to me.  There is no way that the flag can ever be removed once it's added, because any user of the library is free to use it instead of `--benchmark_filter`.  The `FIXME` is not going to help (and it won't be fixed for a long time anyway).

I surely hope that there is a way to deal with internal compatibility without polluting the public API.

dominichamon:
i suggested this because we only ever update the internal version from the external version, and any internal-only changes would be overwritten every time.

however, @pleroy, you're absolutely right and maybe we need to just handle this internally as we migrate, and trust our tooling to identify this issue when imports happen.

pleroy:
Maybe you could have a small file with a module initializer that copies `--benchmarks` to `--benchmark_filter` if it is set, and include that at link time in the `BUILD` file.

oontvoo:
Thanks - closing this PR now. No longer needed as we'll go with a different approach

PR #1206, opened 2021-07-29
Remove dead code from PredictNumItersNeeded

### LINKED ISSUE #1205: PredictNumItersNeeded has unreachable code
1. It is, as far as I can tell, mathematically impossible for [this `min`](https://github.com/google/benchmark/blob/main/src/benchmark_runner.cc#L242) to return anything other than 10.
2. Unless there is a call to `PredictNumItersNeeded` beyond the one in `DoOneRepetition`, `multiplier <= 1.0` is impossible and thus [this statement](https://github.com/google/benchmark/blob/ab74ae5e104f72fa957c1712707a06a781a974a6/src/benchmark_runner.cc#L243) will never be executed.
   1. `multiplier <= 1.0` implies that `isSignificant` is true, so it remains unchanged by the time it arrives at the if.
   1. `multiplier <= 1.0` implies `i.seconds >= min_time * 1.4`
   1. `ShouldReportResults` is always called before `PredictNumItersNeeded`, if `i.seconds >= min_time` then the loop is broken and `PredictNumItersNeeded` is never called.
   1. ii and iii together imply that `multiplier <= 1.0` is never true.
Beyond this, it seems weird to have clearly exceeded the minimum time and then decide that doubling the number of iterations is necessary.

kuzi117:
Each of the commit messages contain the rationale behind modifying or deleting each of these lines.

Closes #1205.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -240,7 +240,6 @@ IterationCount BenchmarkRunner::PredictNumItersNeeded(
   // expansion should be 14x.
   bool is_significant = (i.seconds / min_time) > 0.1;
   multiplier = is_significant ? multiplier : 10.0;
-  if (multiplier <= 1.0) multiplier = 2.0;
 
   // So what seems to be the sufficiently-large iteration count? Round up.
   const IterationCount max_next_iters = static_cast<IterationCount>(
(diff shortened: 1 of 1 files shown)


kuzi117:
Force push was to modify the numbering in the commit message. I assumed github markup would be applied but it wasn't.

dominichamon:
maths checks out. tests _should_ pass or the universe is broken.

dominichamon:
phew! universe isn't broken. thank you!

kuzi117:
Glad I wasn't crazy, thanks!

PR #1208, opened 2021-08-11
Introduce accessors for currently public data members (threads and thread_index)

oontvoo:
Also deprecate the direct access to these fields.

Motivations:

Our internal library provides accessors for those fields because the styleguide disalows accessing classes' data members directly (even if they're const).
There has been a discussion to simply move internal library to make its fields public similarly to the OSS version here, however, the concern is that these kinds of direct access would prevent many types of future design changes (eg how/whether the values would be stored in the data member)

I think the concensus in the end is that we'd change the external library for this case.
AFAIK, there are three important third_party users that we'd need to migrate: tcmalloc, abseil and tensorflow.
Please let me know if I'm missing anyone else.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -667,11 +667,13 @@ class State {
   BENCHMARK_DEPRECATED_MSG("use 'range(1)' instead")
   int64_t range_y() const { return range(1); }
 
+  // Number of threads concurrently executing the benchmark.
   BENCHMARK_ALWAYS_INLINE
   int threads() { return threads_; }
 
+  // Index of the executing thread. Values from [0, threads).
   BENCHMARK_ALWAYS_INLINE
-  int get_thread_index() { return thread_index; }
+  int get_thread_index() { return thread_index_; }
 
   BENCHMARK_ALWAYS_INLINE
   IterationCount iterations() const {
@@ -709,10 +711,6 @@ class State {
   // Container for user-defined counters.
   UserCounters counters;
 
-  // Index of the executing thread. Values from [0, threads).
-  const int thread_index
-      BENCHMARK_DEPRECATED_MSG("Use get_thread_index() instead.");
-
  private:
   State(IterationCount max_iters, const std::vector<int64_t>& ranges,
         int thread_i, int n_threads, internal::ThreadTimer* timer,
@@ -725,7 +723,7 @@ class State {
   bool KeepRunningInternal(IterationCount n, bool is_batch);
   void FinishKeepRunning();
 
-  // Number of threads concurrently executing the benchmark.
+  const int thread_index_;
   const int threads_;
 
   internal::ThreadTimer* const timer_;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -140,13 +140,13 @@ thread exits the loop body. As such, any global setup or teardown you want to
 do can be wrapped in a check against the thread index:
 
 static void BM_MultiThreaded(benchmark::State& state) {
-  if (state.thread_index() == 0) {
+  if (state.get_thread_index() == 0) {
     // Setup code here.
   }
   for (auto _ : state) {
     // Run the test as normal.
   }
-  if (state.thread_index() == 0) {
+  if (state.get_thread_index() == 0) {
     // Teardown code here.
   }
 }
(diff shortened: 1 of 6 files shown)


oontvoo:
@dominichamon

include/benchmark/benchmark.h
```diff
@@ -667,6 +667,12 @@ class State {
   BENCHMARK_DEPRECATED_MSG("use 'range(1)' instead")
   int64_t range_y() const { return range(1); }
 
+  BENCHMARK_ALWAYS_INLINE
+  int num_threads() { return threads; }
```
dominichamon, line 671: hm. i'd really like to name this `threads()` right away to follow https://google.github.io/styleguide/cppguide.html#Function_Names 

given i just released v1.5.6 i think we can make this API change and the next release will be v1.6.0.

anyone building against HEAD should be flexible about API changes i think.

oontvoo: My concern is it'd break tensorflow_serving  if we delete `int threads` and replace it with `int threads()` It'd be better (less headache for me, mostly) if there were intermediate states where I could migrate people gradually without forcing the updates on them ("take this or you'd break" kind of change is really stressful).

LebedevRI:
> Our internal library provides accessors for those fields because the styleguide disalows accessing classes' data members directly (even if they're const).

Could you please link the specific point?
https://google.github.io/styleguide/cppguide.html#Access_Control indeed says
> Make classes' data members private, unless they are [constants](https://google.github.io/styleguide/cppguide.html#Constant_Names)

but https://google.github.io/styleguide/cppguide.html#Constant_Names doesn't make it obvious that it only applies to immediate constants, and i'm not quite seeing the wording that makes it clear that immutable invariant data members
are still data members.

LebedevRI:
Sry, mouse was at the wrong position.

oontvoo:
> Could you please link the specific point?
> https://google.github.io/styleguide/cppguide.html#Access_Control indeed says
> 
> > Make classes' data members private, unless they are [constants](https://google.github.io/styleguide/cppguide.html#Constant_Names)
> 
> but https://google.github.io/styleguide/cppguide.html#Constant_Names doesn't make it obvious that it only applies to immediate constants, and i'm not quite seeing the wording that makes it clear that immutable invariant data members
> are still data members.

While I agree with you that the definition of "constants", as written there in the styleguide, seems like it could be open to slightly different interpretations, in practice, I don't think that's the intent. 

Furthermore, the value of `threads` and `thread_index` here aren't technically "fixed duration of the program" (if you consider un-init state being different from assigned value),  which means they aren't qualified to be considered constants.

include/benchmark/benchmark.h
dominichamon: does tensorflow_serving link to HEAD? i would hope most clients are pinned to a version or aware that they need to fix stuff against HEAD...

oontvoo: tensorflow_serving depends on tensorflow (against a specific version/hash commit), which then links to a specific benchmark version (`64cb55e91067860548cb95e012a38f2e5b71e026` right now).

But actually, I've done a quick search, only two tests are using this - I think it's small enough that it's not a problem.

Renamed the function to `threads()` and make the field private.

oontvoo:
@dominichamon , @LebedevRI  :  I've resolved all the failures and I think the diff is now ready to review. PTAL. thanks!

include/benchmark/benchmark.h
```diff
@@ -667,6 +667,14 @@ class State {
   BENCHMARK_DEPRECATED_MSG("use 'range(1)' instead")
   int64_t range_y() const { return range(1); }
 
+  // Number of threads concurrently executing the benchmark.
+  BENCHMARK_ALWAYS_INLINE
+  int threads() const { return threads_; }
+
+  // Index of the executing thread. Values from [0, threads).
+  BENCHMARK_ALWAYS_INLINE
+  int get_thread_index() const { return thread_index_; }
```
dominichamon, line 676: this can also be just `thread_index()` right?

dominichamon:
@LebedevRI any concerns with me merging this and releasing a 1.6.0 to cover the change in API?

oontvoo:
Hi, friendly ping? Thanks

LebedevRI:
> @LebedevRI any concerns with me merging this and releasing a 1.6.0 to cover the change in API?

Sorry, i have not see this comment.
I don't have any further thoughts here, that is why i'm not commenting further.

dominichamon:
@dominichamon note to myself to create a v1.6 release next time.

PR #1215, opened 2021-08-23
Fix links to further doc in user_guide.md

mjacobse:
Refactoring in 201b981a moved most of the documentation from `README.md` to `docs/user_guide.md`. Some links from `README.md` to other `docs/*.md` files ended up unchanged in `docs/user_guide.md`. Those links were now broken as they did not link from outside the `docs` directory anymore, but from inside it. Removing the leading `docs/` for these links fixes this.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -38,9 +38,9 @@
 
 [Setting the Time Unit](#setting-the-time-unit)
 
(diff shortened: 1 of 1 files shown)


mjacobse:
@googlebot I signed it!

dominichamon:
thank you!

bots should be unaffected so i'll merge when they go green.

Issue #1146, opened 2021-05-07
[FR] Relative Standard Deviation


LebedevRI:
While we have  Standard Deviation aggregate metric, it is measured in time units,
which means it might be too descriptive.

It would be kinda good to have same-ish data, that does not require also looking at median.
I guess [Relative Standard Deviation](https://en.wikipedia.org/wiki/Coefficient_of_variation), aka `\sigma / |\mu|`, fits the bill,
but naturally it is in percent, while all other current stats are in time units.
So at a minimum we'll need to be able to specify the units of a statistic.

PR #1219, opened 2021-09-03
Statistics: add support for percentage unit in addition to time

LebedevRI:
I think, `stddev` statistic is useful, but confusing.

What does it mean if `stddev` of `1ms` is reported?
Is that good or bad? If the `median` is `1s`,
then that means that the measurements are pretty noise-less.

And what about `stddev` of `100ms` is reported?
If the `median` is `1s` - awful, if the `median` is `10s` - good.

And hurray, there is just the statistic that we need:
https://en.wikipedia.org/wiki/Coefficient_of_variation

But, naturally, that produces a value in percents,
but the statistics are currently hardcoded to produce time.

So this refactors thinkgs a bit, and allows a percentage unit for statistics.

I'm not sure whether or not `benchmark` would be okay
with adding this `RSD` statistic by default,
but regales, that is a separate patch.

Refs. https://github.com/google/benchmark/issues/1146


include/benchmark/benchmark.h
```diff
@@ -450,6 +450,8 @@ enum BigO { oNone, o1, oN, oNSquared, oNCubed, oLogN, oNLogN, oAuto, oLambda };
 
 typedef uint64_t IterationCount;
 
+enum StatisticUnit { Time, Percentage };
```
dominichamon, line 453: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

`kTime` and `kPercentage` please

src/console_reporter.cc
```diff
@@ -142,10 +142,16 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
-  } else {
+  } else if (result.run_type != Run::RT_Aggregate ||
+             result.aggregate_unit == StatisticUnit::Time) {
     const char* timeLabel = GetTimeUnitString(result.time_unit);
     printer(Out, COLOR_YELLOW, "%s %-4s %s %-4s ", real_time_str.c_str(), timeLabel,
             cpu_time_str.c_str(), timeLabel);
+  } else {
+    assert(result.aggregate_unit == StatisticUnit::Percentage);
+    printer(Out, COLOR_YELLOW, "%10.2f %-4s %10.2f %-4s ",
+            (100. * result.real_accumulated_time), "%",
```
dominichamon, line 153: bit awkward this is called `_time` and now isn't always one. maybe a rename? or is that too intrusive?

src/statistics.cc
```diff
@@ -167,13 +168,15 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
...
-    // We will divide these times by data.iterations when reporting, but the
-    // data.iterations is not nessesairly the scale of these measurements,
-    // because in each repetition, these timers are sum over all the iterations.
-    // And if we want to say that the stats are over N repetitions and not
-    // M iterations, we need to multiply these by (N/M).
-    data.real_accumulated_time *= iteration_rescale_factor;
-    data.cpu_accumulated_time *= iteration_rescale_factor;
+    if (data.aggregate_unit == StatisticUnit::Time) {
+      // We will divide these times by data.iterations when reporting, but the
+      // data.iterations is not nessesairly the scale of these measurements,
```
dominichamon, line 173: while you're here: "necessarily"

dominichamon (commented):
i always thought that alongside median we should have other percentiles, instead of mean and stddev, because this gives you a better idea of the distribution of results. maybe the CV is a good middle ground instead of arguing about which percentiles to show.

i'd be supportive of replacing stddev with stddev/mean (and we calculate both anyway...)

src/console_reporter.cc
LebedevRI: I keep delaying dealing with this, so i still don't want to change this now,
but this should be a vector of measurements, just two hardcoded "timers" isn't right.

src/statistics.cc
LebedevRI: If only Qt/Qtcreator would finally add spell checking support (:

LebedevRI:
@dominichamon thank you for taking a look!

> i always thought that alongside median we should have other percentiles, instead of mean and stddev, because this gives you a better idea of the distribution of results. maybe the CV is a good middle ground instead of arguing about which percentiles to show.

Aha, sounds good to me.

> i'd be supportive of replacing stddev with stddev/mean (and we calculate both anyway...)
I'm not really sure about completely replacing it, but then we also don't want to bloat the output

with too many enabled-by-default statistics.

PR #1221, opened 2021-09-03
Console reporter: if statistic produces percents, format it as such

LebedevRI:


LebedevRI:
Now with tests!

src/console_reporter.cc
```diff
@@ -161,10 +161,17 @@ void ConsoleReporter::PrintRunData(const Run& result) {
   for (auto& c : result.counters) {
     const std::size_t cNameLen = std::max(std::string::size_type(10),
                                           c.first.length());
-    auto const& s = HumanReadableNumber(c.second.value, c.second.oneK);
+    std::string s;
     const char* unit = "";
-    if (c.second.flags & Counter::kIsRate)
-      unit = (c.second.flags & Counter::kInvert) ? "s" : "/s";
+    if (result.run_type == Run::RT_Aggregate &&
```
dominichamon, line 166: i think we repeat these two lines in a few places now... worth having a utility method on `result`?

dominichamon (approved):

src/console_reporter.cc
LebedevRI: The formatting here is different, there is no space separator before the unit,
and the time unit (seconds) isn't printed in the unit specified for the benchmark,
but in an automatically picked one. So i'm not sure yet how to refactor this.

LebedevRI:

src/console_reporter.cc
dominichamon: i meant the conditional (`RT_Aggregate` && `kPercentage`).

PR #1225, opened 2021-09-08
Optimized docs installation

xvitaly:
- Always use GNUInstallDirs in docs installation target.
- Added an option to disable docs installation.


CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ else()
   set(BENCHMARK_BUILD_32_BITS OFF CACHE BOOL "Build a 32 bit version of the library - unsupported when using MSVC)" FORCE)
 endif()
 option(BENCHMARK_ENABLE_INSTALL "Enable installation of benchmark. (Projects embedding benchmark may want to turn this OFF.)" ON)
+option(BENCHMARK_ENABLE_DOCS_INSTALL "Enable installation of documentation." ON)
```
LebedevRI, line 28: I would've suggested `BENCHMARK_INSTALL_DOCS`, but then there's already `BENCHMARK_ENABLE_INSTALL`, so i guess this is fine.

LebedevRI (approved):

CMakeLists.txt
xvitaly: Sure. Fixed.

LebedevRI: (Note how i worded that, i wasn't asking to change the diff because it was consistent)

@dominichamon do we want to be consistent with the existing options, or have maybe better(worse?) new options?

dominichamon:
i like this version where you enable the install, and then choose whether to install the docs.

LebedevRI:
> i like this version where you ENABLE the INSTALL, and then choose whether to INSTALL the DOCS.
> 
> :)

If i'm reading this right, you are okay with the current diff, in which case please feel free to merge.

dominichamon:
> > i like this version where you ENABLE the INSTALL, and then choose whether to INSTALL the DOCS.
> > :)
> 
> If i'm reading this right, you are okay with the current diff, in which case please feel free to merge.

will do.. awaiting final bot checks.

PR #850, opened 2019-08-06
Custom user counters: add invert modifier.

LebedevRI:
While current counters can e.g. answer the question
"how many items is processed per second", it is impossible to get
it to tell "how many seconds it takes to process a single item".

The solution is to add a yet another modifier `kInvert`,
that is *always* considered last, which simply inverts the answer.

Fixes #781, #830, #848.

PR #1227, opened 2021-09-12
Support for building with LLVM clang-10/clang-11 on Windows.

alisenai:
# Issue:
Benchmark does not compile with LLVM clang-10/clang-11 on Windows.

## Relevant Compiler Output:
```
In file included from benchmark/src/benchmark_name.cc:15:
benchmark/include/benchmark/benchmark.h:408:5: error: enumerator value is not representable in the underlying type 'int' [-Werror,-Wmicrosoft-enum-value]
    kInvert = 1U << 31U
    ^
1 error generated.
```

## Reason For Issue:
  The enum `Flags` on line 384 of `benchmark/include/benchmark/benchmark.h`.
  The issue arises due to the pull request https://github.com/google/benchmark/pull/850 being merged in. \
  As I currently understand, the underlying type of an `enum` is somewhat implementation-defined, which is assumed by the previously mentioned line / pull request. This is described [here](https://timsong-cpp.github.io/cppwp/n4659/dcl.enum#7).

## Fix:
  Let the compiler decide the underlying type for the `Flags` enum by removing the unsigned specification of `U`.

## More Information:
- Platform:
  - Windows-10.0.19042
- Architecture:
  - amd64
- (Known) Affected Compilers: 
  - [Windows LLVM Clang-10.0.0](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.1.0/LLVM-11.1.0-win64.exe)
  - [Windows LLVM Clang-11.0.0](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.1.0/LLVM-10.0.0-win64.exe)
- (Known) Working Compilers Post Patch: 
  - [Windows LLVM Clang-10.0.0](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.1.0/LLVM-11.1.0-win64.exe)
  - [Windows LLVM Clang-11.0.0](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.1.0/LLVM-10.0.0-win64.exe)
  - Linux Clang-10

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -381,31 +381,31 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 // This class is used for user-defined counters.
 class Counter {
  public:
-  enum Flags : unsigned {
+  enum Flags {
     kDefaults = 0,
     // Mark the counter as a rate. It will be presented divided
     // by the duration of the benchmark.
-    kIsRate = 1U << 0U,
+    kIsRate = 1 << 0,
     // Mark the counter as a thread-average quantity. It will be
     // presented divided by the number of threads.
-    kAvgThreads = 1U << 1U,
+    kAvgThreads = 1 << 1,
     // Mark the counter as a thread-average rate. See above.
     kAvgThreadsRate = kIsRate | kAvgThreads,
     // Mark the counter as a constant value, valid/same for *every* iteration.
     // When reporting, it will be *multiplied* by the iteration count.
-    kIsIterationInvariant = 1U << 2U,
+    kIsIterationInvariant = 1 << 2,
     // Mark the counter as a constant rate.
     // When reporting, it will be *multiplied* by the iteration count
     // and then divided by the duration of the benchmark.
     kIsIterationInvariantRate = kIsRate | kIsIterationInvariant,
     // Mark the counter as a iteration-average quantity.
     // It will be presented divided by the number of iterations.
-    kAvgIterations = 1U << 3U,
+    kAvgIterations = 1 << 3,
     // Mark the counter as a iteration-average rate. See above.
     kAvgIterationsRate = kIsRate | kAvgIterations,
 
     // In the end, invert the result. This is always done last!
-    kInvert = 1U << 31U
+    kInvert = 1 << 31
   };
 
   enum OneK {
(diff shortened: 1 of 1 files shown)


include/benchmark/benchmark.h
```diff
@@ -385,27 +385,27 @@ class Counter {
     kDefaults = 0,
```
LebedevRI, line 385: Does the issue also go away if you simply add `U` suffix here instead?

alisenai: Just tried it but, unfortunately, it does not. Nor does casting everything to an `unsigned`. Seems like the compiler is not attempting to change the underlying type - Or doesn't support it?.

include/benchmark/benchmark.h
```diff
@@ -385,27 +385,27 @@ class Counter {
...
     // Mark the counter as a iteration-average quantity.
     // It will be presented divided by the number of iterations.
-    kAvgIterations = 1U << 3U,
+    kAvgIterations = 1 << 3,
     // Mark the counter as a iteration-average rate. See above.
     kAvgIterationsRate = kIsRate | kAvgIterations,
 
     // In the end, invert the result. This is always done last!
-    kInvert = 1U << 31U
+    kInvert = 1 << 31
```
chadaustin, line 408: Isn't `1 << 31` UB on platforms with 32-bit signed int? Instead, would it make sense to write `enum Flags : uint32_t` and bring back `1U << 31U` here?

alisenai: I believe you are correct in stating that `1 << 31` is UB. I must have mistaken an enum's ability to account for this when it [fits all underlying values](https://timsong-cpp.github.io/cppwp/n4659/dcl.enum#7).

I also tried a similar solution as you are mentioning when I originally made a patch (see https://github.com/alisenai/benchmark/commit/e75182f7edfd702020f668f72ee142d29c6efad1), though I do not recall why I decided against it (?).

alisenai: Oh, right, as the `README.md` states:
> The library can be used with C++03 (...)

Since enum type specifiers were only added in c++11 (correct me if I'm wrong), `enum Flags : uint32_t` would not compile when including this header in a C++03 file/project.

Though, looking through the rest of the file, maybe a similar solution can be implemented?https://github.com/google/benchmark/blob/60b16f11a30146ac825b7d99be0b9887c24b254a/include/benchmark/benchmark.h#L577-L581

PR #1228, opened 2021-09-13
Added Doxygen support

### LINKED ISSUE #1125: [FR] Doxygen

No API docs are published for the benchmark library.

Publish basic Doxygen HTML docs online. This is fairly straightforward with Travis + GitHub pages, if you don't want to use a separate doc hosting site.

xvitaly:
Added Doxygen support. Can be enabled with `-DBENCHMARK_ENABLE_DOXYGEN:BOOL=ON`.

Closes #1125.


xvitaly:
Also we can drop a separate `docs` installation, because Doxygen will automatically render all Markdown files to HTML including working hyperlinks.

xvitaly:
Any reviewers?

dominichamon:
hi!

i'm not very familiar with how doxygen and cmake interact and where the doxygen output ends up. 

would this allow us to create a github action that would build the documentation and check it in to `docs/` ?

dominichamon:
> > hi!
> > i'm not very familiar with how doxygen and cmake interact and where the doxygen output ends up.
> > would this allow us to create a github action that would build the documentation and check it in to `docs/` ?
> 
> Oh please no.
> At most it should be auto-published as part of the pages (the site)

`docs/` is pages (the site).

LebedevRI:
> > > hi!
> > > i'm not very familiar with how doxygen and cmake interact and where the doxygen output ends up.
> > > would this allow us to create a github action that would build the documentation and check it in to `docs/` ?
> > 
> > 
> > Oh please no.
> > At most it should be auto-published as part of the pages (the site)
> 
> `docs/` is pages (the site).

In this repo in this branch? My point still stands then.

xvitaly:
> would this allow us to create a github action that would build the documentation and check it in to docs/ ?

Sure. But I suggest `gh-pages` branch.

xvitaly:
Added GitHub action with Doxygen.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,30 @@
...
+      run: |
+        mkdir build
+    - name: Building HTML using Doxygen
+      run: |
+        cmake -S . -B build -DBENCHMARK_ENABLE_TESTING:BOOL=OFF -DBENCHMARK_ENABLE_DOXYGEN:BOOL=ON
+        cmake --build build --target benchmark_doxygen
+    - name: Deploying built HTML to GitHub pages
+      uses: JamesIves/github-pages-deploy-action@4.0.0
+      with:
+        branch: gh-pages
```
dominichamon, line 29: our pages are linked to main, iiuc, and there is no gh-pages branch.

also this would override the docs that are already in `docs/` i think.. wouldn't it?

xvitaly: I think that pushing a lot of dynamically generated static assets to `docs` directory is not a good idea.

dominichamon: but are the results of this push accessible anywhere if our project is set up with `main` as the pages branch?

xvitaly: I have a better idea. I will push changes to my fork shortly and we can see the final results.

xvitaly: Site index: https://xvitaly.github.io/benchmark/

Static page example: https://xvitaly.github.io/benchmark/md_docs_user_guide.html

dominichamon: some nitpicks: the main page doesn't have working links, and the user guide code samples aren't syntax formatted. i don't know if these are solvable.

i think what i would expect as an end state is the existing docs working as they are (https://google.github.io/benchmark) with a separate location (https://google.github.io/benchmark/api or /dev ?) that has just the generated code documentation.

or, we can just have the simple version where there's an option to generate doxygen locally for people that want it and leave it there.

xvitaly: > some nitpicks: the main page doesn't have working links, and the user guide code samples aren't syntax formatted. i don't know if these are solvable.

Very old doxygen version, I think. Doxygen 1.9.1 from Fedora 34 generate working links.

> i think what i would expect as an end state is the existing docs working as they are

All these docs already built with Doxygen and pushed to the website.

> or, we can just have the simple version where there's an option to generate doxygen locally for people that want it and leave it there.

I like this option.

LebedevRI: > > or, we can just have the simple version where there's an option to generate doxygen locally for people that want it and leave it there.
> 
> I like this option.

Sure, that sounds good.

xvitaly: > Sure, that sounds good.

Removed GitHub actions file and disabled the Doxygen target by default.

xvitaly: Also we can install Doxygen files to `${CMAKE_CURRENT_BINARY_DIR}` and disable installation of files from `docs/` directory, because generated by Doxygen HTMLs already includes all of them.

xvitaly:
Updated my PR. Now it will generate and install generated HTML pages if `BENCHMARK_ENABLE_DOXYGEN` is set. If not, it will install documentation files from `docs` directory.

dominichamon:
it might be a good idea to have a single build bot (action) to protect the doxygen creation step against regressions. ie, an ubuntu vanilla build that just sets the doxygen flag and builds.

xvitaly:
> it might be a good idea to have a single build bot (action) to protect the doxygen creation step against regressions. ie, an ubuntu vanilla build that just sets the doxygen flag and builds.

Done.

dominichamon (approved):

PR #1230, opened 2021-09-17
Align benchmark::State to a cacheline.

ckennelly:
This can avoid interference with neighboring objects and stabilize
benchmark results.


oontvoo:
@dominichamon  Another reason to have absl dependency :) Would make this PR so much simpler.

dominichamon:
do you have the results from a run with and without this change?

ckennelly:
@dominichamon I only have an internal microbenchmark, which demonstrated higher variability across runs (especially when combined with a change to TCMalloc) without the alignment attribute.

LebedevRI:
This seems plausible (although hard to tell without a test), but is this only about the `State`?
What about `ThreadTimer`, `ThreadManager`, `PerfCountersMeasurement`?

chfast (commented):
Can you use `std::hardware_destructive_interference_size` from C++17 if available?

include/benchmark/benchmark.h
```diff
@@ -290,11 +290,49 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
...
+// space, while underestimates tend to waste more time.
+#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
+#endif
+
+// Indicates that the declared object be cache aligned using
+// `BENCHMARK_INTERNAL_CACHELINE_SIZE` (see above).
+#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED \
+  __attribute__((aligned(BENCHMARK_INTERNAL_CACHELINE_SIZE)))
+#elif defined(_MSC_VER)
+#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
```
LebedevRI, line 323: Already caught by `#ifndef BENCHMARK_INTERNAL_CACHELINE_SIZE` block.

dmah42:
> This seems plausible (although hard to tell without a test), but is this only about the `State`? What about `ThreadTimer`, `ThreadManager`, `PerfCountersMeasurement`?

we could try to run some tests, but i don't think it should block this PR (which i totally missed).

or just align all the things.

dmah42:
> Can you use `std::hardware_destructive_interference_size` from C++17 if available?

not in the public header.  _soon_.

include/benchmark/benchmark.h
```diff
@@ -290,11 +290,49 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
...
+#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED \
+  __declspec(align(BENCHMARK_INTERNAL_CACHELINE_SIZE))
+#else
+#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
+#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED
+#endif  // __GNUC__
+
```
LebedevRI: I think it might be cleaner to close the `__GNUC__` block immediately after `cpu arch`,
then define a fallback cacheline size, and then define the `BENCHMARK_INTERNAL_CACHELINE_ALIGNED`?

dmah42: i think you're right. i'm slowly going cross-eyed looking at this.  thanks for making sense of it.

include/benchmark/benchmark.h
```diff
@@ -290,11 +290,50 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
...
+#elif defined(_MSC_VER)
+// A reasonable default guess.  Note that overestimates tend to waste more
+// space, while underestimates tend to waste more time.
+#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
+#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED \
+  __declspec(align(BENCHMARK_INTERNAL_CACHELINE_SIZE))
+#else
+#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED
+#endif
```
LebedevRI, line 330:
```suggestion
#if defined(__GNUC__)
// Determine the cacheline size based on architecture
#if defined(__i386__) || defined(__x86_64__)
#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
#elif defined(__powerpc64__)
#define BENCHMARK_INTERNAL_CACHELINE_SIZE 128
#elif defined(__aarch64__)
#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
#elif defined(__arm__)
// Cache line sizes for ARM: These values are not strictly correct since
// cache line sizes depend on implementations, not architectures.  There
// are even implementations with cache line sizes configurable at boot
// time.
#if defined(__ARM_ARCH_5T__)
#define BENCHMARK_INTERNAL_CACHELINE_SIZE 32
#elif defined(__ARM_ARCH_7A__)
#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
#endif  // ARM_ARCH
#endif  // arches
#endif  // __GNUC__

#ifndef BENCHMARK_INTERNAL_CACHELINE_SIZE
// A reasonable default guess.  Note that overestimates tend to waste more
// space, while underestimates tend to waste more time.
#define BENCHMARK_INTERNAL_CACHELINE_SIZE 64
#endif

#if defined(__GNUC__)
// Indicates that the declared object be cache aligned using
// `BENCHMARK_INTERNAL_CACHELINE_SIZE` (see above).
#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED \
  __attribute__((aligned(BENCHMARK_INTERNAL_CACHELINE_SIZE)))
#elif defined(_MSC_VER)
#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED \
  __declspec(align(BENCHMARK_INTERNAL_CACHELINE_SIZE))
#else
#define BENCHMARK_INTERNAL_CACHELINE_ALIGNED
#endif
```

dmah42 (dismissed):

dmah42:
yeah your way is better because it separates the definition of the cacheline size from the macro that uses it.

include/benchmark/benchmark.h
```diff
@@ -290,11 +290,50 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #define BENCHMARK_OVERRIDE
 #endif
 
+#if defined(__GNUC__)
```
LebedevRI, line 293: Not really sure if this outermost guard is needed, but then i don't know what set of per-cpu defines MSVC provides.

LebedevRI (approved):
Yay, thank you!

PR #1235, opened 2021-10-05
Allow UseRealTime() to take an optional parameter

oontvoo:
Internally, we have a few users who would like to call this function conditionally (ie, based one some flag)
And because this function is usually called on the `BENCHMARK()` instantiation macro it's quite awkward to conditionalised the `UseRealTime` right now.

Making it taking a parameter will make it easier.
Eg.,
`BENCHMARK(BM_MyTest)->UseRealTime(GetFlags(FLAGS_someflags));`

(Similar changes will be proposed to the internal API)


oontvoo:
Withdrawing this since the internal change wasn't blessed - and `Apply()`can be used.
Sorry for the noise!

PR #1238, opened 2021-10-11
Introduce additional memory metrics

oontvoo:
- added total_allocated_bytes and net_heap_growth
  - updated reporter code to report these, if available.
  - move MemoryManager class closer to the top of the header (needed to reference the MemoryManager::Result class in benchmark.h:line 1444, which is before it's currently declared)
 - add an overload `MemoryManager::Stop(Result&)` to replace `MemoryManager::Stop(Result*)` because the styleguide advises using a reference type for output param that is *not* expected to be null.
 - 
@dominichamon

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -180,6 +180,7 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 #include <cassert>
 #include <cstddef>
 #include <iosfwd>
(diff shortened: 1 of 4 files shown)


oontvoo:
I must be missing something here but I don't understand why a few workflows are complaining about this patch failing to build with cxx03. (It *should* be failing with cxx03 as the minimum C++ version required by this repo is C++11, right?)

PR #1240, opened 2021-10-14
cmake: allow to use package config from build directory

### LINKED ISSUE #1239: [BUG] Cannot use CMake package config from build directory
**Describe the bug**
Currently, it is not possible to use the CMake package config directly from the build directory. Instead, an install step is required to generate all the necessary CMake files. This makes the benchmark CMake package config less useful.

There are several problems with the current implementation:
1. Package config files are written to the `<build-dir>/src/generated` directory which is less intuitive than the commonly used build directory root, i.e., `<build-dir>`.
2. `benchmarkTargets.cmake` is not generated during CMake configuration time. Consequently, trying to include `benchmarkConfig.cmake` using the `find_package` mechanism will fail. 

**System**
  - OS: Ubuntu 20.04
  - Compiler and version: GCC 9 

**To reproduce**
1. sync to commit 0baacde3618ca617da95375e0af13ce1baadea47 (main branch)
2. Configure using CMake
3. Compile
4. Try to use the generated CMake package config by configuring a project depending on Google benchmark

**Expected behavior**

CMake package config files are written to Google benchmark build directory and can be used without an install step.

sergiud:
Fixes #1239.

### DIFF
--- a/cmake/Config.cmake.in
+++ b/cmake/Config.cmake.in
@@ -1,3 +1,5 @@
+@PACKAGE_INIT@
+
 include (CMakeFindDependencyMacro)
 
 find_dependency (Threads)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -64,7 +64,7 @@ set_target_properties(benchmark_main PROPERTIES
 target_link_libraries(benchmark_main PUBLIC benchmark::benchmark)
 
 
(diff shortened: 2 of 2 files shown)


dominichamon:
i'm not a cmake expert by any stretch but this looks reasonable to me. i'll leave this open for now until someone else can comment though.

LebedevRI:
Rebase please?
Will try to look at this next.

PR #1244, opened 2021-10-14
cmake: make package config relocatable

### LINKED ISSUE #1243: [BUG] CMake package config is not relocatable
**Describe the bug**

The CMake package config is currently not relocatable since the installed version contains absolute paths such as these (taken from `benchmarkTargets.cmake`):
```cmake
set_target_properties(benchmark::benchmark PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "-pthread;/usr/lib/x86_64-linux-gnu/librt.so"
)
```
which may not exist on the system employing the package config, e.g., when cross-compiling. Also, hardcoding `-pthread` is problematic  since the flag may not be available during package config use.

**System**
  - OS: Ubuntu 20.04
  - Compiler and version: GCC 9

**To reproduce**
1. sync to commit 0baacde3618ca617da95375e0af13ce1baadea47
2. Run CMake with a toolchain file
3. Cross-compile and install
4. Move toolchain sysroot to a different location
5. Linker error

**Expected behavior**

CMake package config must not contain any absolute paths.

sergiud:
Fixes #1243.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,9 @@ string(SUBSTRING ${VERSION} 0 1 GENERIC_LIB_SOVERSION)
 include(CheckCXXCompilerFlag)
 include(AddCXXCompilerFlag)
 include(CXXFeatureCheck)
+include(CheckLibraryExists)
+
+check_library_exists(rt shm_open "" HAVE_LIB_RT)
 
 if (BENCHMARK_BUILD_32_BITS)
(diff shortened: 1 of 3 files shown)


LebedevRI:
If this works and doesn't break anything i think this is good.

LebedevRI:
Rebase please.

PR #1245, opened 2021-10-15
Fix -Wdeprecated-declarations warning triggered by clang-cl.

bc-lee:
WebRTC uses Google Benchmarks as a dependency and uses Chromium's build
infrastructure. Chromium is compiled using clang-cl on Windows, and the
-Wdeprecated-declarations warning is triggered. Because clang-cl accepts
gcc's diagnostic prama and defines the `__clang__` macro,
using it can solve this issue.

Bug: webrtc:13280

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -378,7 +378,7 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
 
 // Disable deprecated warnings temporarily because we need to reference
 // CSVReporter but don't want to trigger -Werror=-Wdeprecated-declarations
-#ifdef __GNUC__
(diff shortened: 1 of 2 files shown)


bc-lee:
The `BENCHMARK_DEPRECATED_MSG` macro is currently only defined in gcc/clang, so I intentionally used the `__clang__` preprocessor definition.
If the deprecation warning for MSVC is also enabled later, it must be handled separately. Fortunately, once #1256 is merged, it would be less error-prone. Because if someone wants to rewrite `BENCHMARK_WARNING_MSG` for MSVC, they would of course think they should also modify `BENCHMARK_DISABLE_DEPRECATED_WARNING` / `BENCHMARK_RESTORE_DEPRECATED_WARNING`.

PR #1254, opened 2021-10-18
[RFC] Adding API for setting/getting benchmark_filter flag?

oontvoo:
This PR is more of a request-for-comment - open to other ideas/suggestions as well.

Details:
Internally, this flag has different implementations(absl vs benchmark) and since the proposal to add absl as a dependency was rejected, it would be nice to have a reliable (and less hacky) way to access this flag internally.
(Actually, reading it isn't much a problem but setting it is).

Internally, we have a sizeable number users to use absl::SetFlags to set this flag. This will not work with benchmark-flags.

Another motivation is that not all users use the command line flag. Some prefer to programmatically set this value.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -21,6 +21,7 @@ TEST_ARGS = ["--benchmark_min_time=0.01"]
 PER_SRC_TEST_ARGS = ({
     "user_counters_tabular_test.cc": ["--benchmark_counters_tabular=true"],
     "repetitions_test.cc": [" --benchmark_repetitions=3"],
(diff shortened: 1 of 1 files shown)


oontvoo:
> This is missing the second part of the reasoning -

I've updated the first comment. Please let me know if that's still not clear.
To answer the questions here:

>  what's it used for? What is the workflow?
This is meant to allow users (binaries/tests) to specify what benchmarks they want to run without having to use the command line flag.
Why? One, because not all binaries use command line flag - some prefer to set it programmatically. And two, more importantly, because internally we implement this flag with absl::Flags and users use absl::SetFlag to set it. This will not work with benchmark's command line flag.

PR #1255, opened 2021-10-18
Update README.md to stop recommending cmake command that doesn't work

oontvoo:
`cmake -DCMAKE_BUILD_TYPE=Release -S . -B "build"` produces CMake error (for googletest)
Tested with on Linux (`cmake version 3.18.4`) and OSX (`cmake version 3.20.0-rc5`)


Celthi:
I met the same issue with running the command `cmake -DCMAKE_BUILD_TYPE=Release -S . -B "build"  `
My cmake is 
cmake -version                                                                                                                                                                                                                                                           (git)-[main] 
cmake version 3.18.4

CMake suite maintained and supported by Kitware (kitware.com/cmake).

Issue #1257, opened 2021-10-20
[BUG] ArgNames results in running a family of benchmarks with the same name


LeonScroggins:
**Describe the bug**
The first `std::string` in the vector passed to `ArgNames` is used for all benchmarks in the same family.

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Android
  - Compiler and version: Not sure - whatever my Android checkout is using

**To reproduce**
1. sync to commit `ea5a5bbff491fd625c6e3458f6edd680b8bd5452`
2. Write a benchmark that takes two arguments and two names. e.g.

```
  void BM_phony(benchmark::State& benchState) {
      int64_t x = 0;
      const int64_t inc = benchState.range();
      for (auto _ : benchState) {
          x += inc;
      }
  }
  BENCHMARK(BM_phony)->Arg(1)->Arg(2)->ArgNames({"one", "two"});
```
3. make && run

**Expected behavior**
I would expect to see the benchmark run twice, once with the name "one" and once with "two".

**Actual behavior**
Although the second run uses the proper number at the end (`2`), it uses the first string (`"one"`).
```
BM_phony/one:1        0.000 ns        0.000 ns   1000000000
BM_phony/one:2        0.000 ns        0.000 ns   1000000000
```

LebedevRI:
Please refer to https://github.com/google/benchmark/blob/main/docs/user_guide.md#passing-arguments,
you want `->Args({1, 2})->ArgNames({"one", "two"})` i think.
Also, looks like `ArgNames` isn't documented there, might be good to document it (aka, patches welcomed).

LeonScroggins:
It turns out there's already a way to achieve what I wanted to do:

```
BENCHMARK(BM_phony)->Arg(1)->ArgName("one")
                   ->Arg(2)->ArgName("two");
```

LebedevRI:
> It turns out there's already a way to achieve what I wanted to do:
> 
> ```
> BENCHMARK(BM_phony)->Arg(1)->ArgName("one")
>                    ->Arg(2)->ArgName("two");
> ```

Would it be useful to document that?

raccoonmonk:
> It turns out there's already a way to achieve what I wanted to do:
> 
> ```
> BENCHMARK(BM_phony)->Arg(1)->ArgName("one")
>                    ->Arg(2)->ArgName("two");
> ```

Since when? it doesn't work for me
```
BENCHMARK(BM_TabSelect)
[...]

PR #1258, opened 2021-10-21
Treat Benchmark::ArgName like Benchmark::Arg

LeonScroggins:
Bug: #1257
Test: TODO

Benchmark::Arg creates an instance of the Benchmark which accepts a
single argument.

Similarly, Benchmark::Args creates an instance of the Benchmark which
accepts multiple arguments.

Benchmark::ArgNames works like ::Args - it must match the number of
arguments passed to Args, and it matches each Name to an Arg.

Previously, within a family of Benchmarks there was no way to name the
Arg(s) of the second (and so on) instances differently from the first
instance. Add this feature by treating ArgName like Arg - convert
Benchmark::arg_names_ to a vector of vectors (like args_). Each inner
vector corresponds to an instance.

Possible regressions:
1) If a client was depending on ArgNames to name the arguments for every
instance in a family, it will now only name the first instance. This
will still be supported by calling ArgNames multiple times with the same
values.
    - Alternatively, the library could support this use by using the
    last set of ArgNames for any further instances.
2) Previously, a client could call ArgNames or ArgName multiple times on
the same Benchmark, and only the last one would be respected. I don't
expect this to happen in practice.

Change-Id: Ide3f8cb433e219042a3506c192e8c4287ae55780

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1069,8 +1069,8 @@ class Benchmark {
 
   std::string name_;
   AggregationReportMode aggregation_report_mode_;
-  std::vector<std::string> arg_names_;       // Args for all benchmark runs
-  std::vector<std::vector<int64_t> > args_;  // Args for all benchmark runs
+  std::vector<std::vector<std::string>> arg_names_;  // Args for all benchmark runs
+  std::vector<std::vector<int64_t> > args_;          // Args for all benchmark runs
   TimeUnit time_unit_;
   int range_multiplier_;
   double min_time_;
--- a/src/benchmark_api_internal.cc
+++ b/src/benchmark_api_internal.cc
@@ -29,14 +29,16 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
       threads_(thread_count) {
   name_.function_name = benchmark_.name_;
 
+  const auto* arg_names = per_family_instance_idx < benchmark->arg_names_.size() ?
+          &(benchmark->arg_names_[per_family_instance_idx]) : nullptr;
   size_t arg_i = 0;
   for (const auto& arg : args) {
     if (!name_.args.empty()) {
       name_.args += '/';
     }
 
-    if (arg_i < benchmark->arg_names_.size()) {
-      const auto& arg_name = benchmark_.arg_names_[arg_i];
+    if (arg_names && arg_i < arg_names->size()) {
+      const auto& arg_name = arg_names->operator[](arg_i);
       if (!arg_name.empty()) {
         name_.args += StrFormat("%s:", arg_name.c_str());
       }
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -291,13 +291,13 @@ Benchmark* Benchmark::ArgsProduct(
 
 Benchmark* Benchmark::ArgName(const std::string& name) {
   BM_CHECK(ArgsCnt() == -1 || ArgsCnt() == 1);
-  arg_names_ = {name};
+  arg_names_.push_back({name});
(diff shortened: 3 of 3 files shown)


dominichamon:
> I haven't added tests or documentation, but I wanted to get some feedback on the direction. I think this makes sense given how Arg works, and it allows me to do what I originally wanted, using something like
> 
> `BENCHMARK(BM_phony)->Arg(1)->Arg(2)->ArgName("one")->ArgName("two");`

with this change, does this become something like: `BENCHMARK(BM_phony)->Args({1,2})->ArgNames({"one", "two"});` ?

LeonScroggins:
> with this change, does this become something like: `BENCHMARK(BM_phony)->Args({1,2})->ArgNames({"one", "two"});` ?

The behavior of the code you wrote is unchanged. It will create a single instance of `BM_phony`, which runs using two named arguments. I think the output would look something like

```
BM_phony/one:1/two:2        0.000 ns        0.000 ns   1000000000
```

But that's not what I want. I want to run the benchmark twice (or two instances in the same family), each time with a single argument. And I want to add names to specify what each instance's value means. (It's an enum, so seeing that it's `1` in the output isn't very helpful.)

Today,

> `BENCHMARK(BM_phony)->Arg(1)->Arg(2)->ArgName("one")->ArgName("two");`

gives the output:

```
BM_phony/two:1        0.000 ns        0.000 ns   1000000000
BM_phony/two:2        0.000 ns        0.000 ns   1000000000
```

Note how "two" replaced "one"; it is now used for all instances of the benchmark. Maybe this is useful behavior, but I suspect it's just incidental. My proposal is to change how that same line works. With this commit,

> `BENCHMARK(BM_phony)->Arg(1)->Arg(2)->ArgName("one")->ArgName("two");`

gives the output:

```
BM_phony/one:1        0.000 ns        0.000 ns   1000000000
BM_phony/two:2        0.000 ns        0.000 ns   1000000000
```

I personally think this is very useful. That said, I've figured out that I can achieve the same result with the current code, by using

```
BENCHMARK(BM_phony)->Arg(1)->ArgName("one");
BENCHMARK(BM_phony)->Arg(2)->ArgName("two");
```

It's slightly more wordy. It is perhaps a little easier to follow, although the less wordy version could (with my commit) be written as

```
BENCHMARK(BM_phony)->Arg(1)->ArgName("one")
                   ->Arg(2)->ArgName("two");
```

LebedevRI:
This seems to lack test coverage.
What happens if you have `BENCHMARK(BM_phony)->Arg(1)->Arg(2)->ArgName("one")->ArgName("two")->Threads(1)->Threads(2);` ? I'm not sure `per_family_instance_idx` gives you what you expect it does.

`

LeonScroggins:
> This seems to lack test coverage.

Indeed! [See my original comment](https://github.com/google/benchmark/pull/1258#pullrequestreview-785994959):

> I haven't added tests or documentation, but I wanted to get some feedback on the direction.

Regarding:
> What happens if you have `BENCHMARK(BM_phony)->Arg(1)->Arg(2)->ArgName("one")->ArgName("two")->Threads(1)->Threads(2);` ? I'm not sure `per_family_instance_idx` gives you what you expect it does.

That may be true; I've only recently started using google-benchmark, and I'm not very familiar with the internals. I don't know what to expect when you make multiple calls to `Threads()` regardless of my change.

But I think there's two different concerns here:
1. Is the behavior I'm trying to support desirable? Given that I ultimately found a way to do it without any changes to google-benchmark, maybe there's no need to change anything!
2. Is this the right way to do it? Again, maybe not. I'm no longer blocked by this, so I likely won't have a lot of time to work out the right way to do it.

So I'll go ahead and close this for now.

PR #1180, opened 2021-06-16
Add wheel and sdist building action to GH Workflow

njunge94:
Working to fix #1177. I added a workflow for building wheels and an sdist and uploading it to GitHub. I set the proposed wheel building workflow to manual dispatch as of now, so it can be checked whether or not it works. 

For later, it could be worth adding a publishing job running after release/tagging/manually, but I don't know how you guys handle this for Python (it would require a PyPI token secret added to this repo for instance), so I left it out for now. Would be excited about some feedback.

PR #1263, opened 2021-10-26
Fix dependency typo and unpin cibuildwheel version in wheel building …

nicholasjng:
…action.

Apologies for abandoning the Python wheels, I got sidetracked by some things over the last few months. In its current form, the wheel building action did not work due to a careless typo by me, that is fixed with this commit.

I started a first hail-mary workflow run in my fork just a few minutes ago. The results can be found [here](
https://github.com/nicholasjng/benchmark/actions/runs/1385963105). In short, as a summary:
* macOS builds succeed (Python 3.6-3.9, macOS 11.0, x86 64-bit)
* Linux builds fail, presumably due to an environment misconfiguration (it fails to locate Bazel in the first place).
* Windows builds fail due to a linker error coming from pybind11 (?).

So that is currently 1/3. If someone has experience with setting up Bazel inside GitHub Actions (specifically Linux and Windows), I would love to get some direction on this.

cc @yashk2810 (who caught the typo [here](https://github.com/google/benchmark/pull/1180#discussion_r727229147))

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -48,7 +48,7 @@ jobs:
         env:
           CIBW_BUILD: 'cp36-* cp37-* cp38-* cp39-*'
         run: |
-          pip install cibuildwheelcibuildwheel==2.0.0a2
+          pip install cibuildwheel
           python -m cibuildwheel --output-dir wheelhouse
 
       - name: Upload wheels
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -19,21 +19,15 @@ jobs:
       - name: Build and check sdist
         run: |
           python setup.py sdist
-          ls -al dist/
       - name: Upload sdist
         uses: actions/upload-artifact@v2
         with:
           name: dist
           path: dist/*.tar.gz
 
-  build_wheels:
-    name: Build google-benchmark wheels on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      # let runner finish even if there is a failure
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+  build_linux:
+    name: Build google-benchmark manylinux wheels
+    runs-on: ubuntu-latest
 
     steps:
       - name: Check out Google Benchmark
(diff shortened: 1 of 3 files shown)


nicholasjng:
Good news, I have been able to get all wheels to build on all platforms after some tweaks. 

Completed run: https://github.com/nicholasjng/benchmark/actions/runs/1434685479

Left TODO are:
* Testing the created wheels by running the Python bindings test.
* Emulating bazel inside the Docker container to allow for aarch64/ppc64le/i686 manylinux builds.
* win32 builds (these fail due to an MSVC linking error, no idea how to fix this yet).

dominichamon:
nice!  is it ready for me to review?

nicholasjng:
Yes, please!

.github/workflows/wheels.yml
```diff
@@ -19,21 +19,88 @@ jobs:
...
+          pip install cibuildwheel
+          python -m cibuildwheel --output-dir wheelhouse
+
+      - name: Upload macOS wheels
+        uses: actions/upload-artifact@v2
+        with:
+          name: dist
+          path: wheelhouse/*.whl
+
   build_wheels:
```
dominichamon, line 101: maybe "build_windows" to match the others?

dominichamon (commented):
did you want to have this run on push instead of manually?

nicholasjng:
TBH, I am not sure: The wheels should be built for every new release, since PyPI also groups them by release (and the release numbers are part of the wheel names). 

I have seen repos with the trigger set to published release, but given that the Python bindings version and the Google Benchmark version are not the same, this would fail if a new GBM version gets published without a new Python bindings version. One could prevent this, though, by checking the release tag against the current. What do you think?

dominichamon:
> TBH, I am not sure: The wheels should be built for every new release, since PyPI also groups them by release (and the release numbers are part of the wheel names).

gotcha.

> I have seen repos with the trigger set to published release, but given that the Python bindings version and the Google Benchmark version are not the same,

i'm missing something, why are they not the same?

> this would fail if a new GBM version gets published without a new Python bindings version. One could prevent this, though, by checking the release tag against the current. What do you think?

i don't know exactly how this would work but i would like to see this automated if at all possible.  if not possible, then could you add instructions to the releasing doc? https://google.github.io/benchmark/releasing.html

nicholasjng:
> i'm missing something, why are they not the same?

That I am not sure about, either.

> i don't know exactly how this would work but i would like to see this automated if at all possible. if not possible, then could you add instructions to the releasing doc? https://google.github.io/benchmark/releasing.html

This is definitely possible. I did this on a hobby project previously, I think I will use the logic from there. 

Regarding the versioning, I see two possibilities (largely):
1) Sync the Python bindings version to match GBM, which would produce a gap (0.2.0 -> 1.6.x), but alas. Then, trigger wheel building on a new release, and add a reminder to bump the Python bindings version as well to the releasing doc.
2) Keep the version as is. Then, depending on the latest PyPI version, wheels may or may not have to be built upon release, which can probably be determined by a clever `if`-clause in the GitHub Action (TBA).

I personally favor option 1), but ultimately it is of course your call. Also, I should add a wheel pushing step to upload the built wheels to PyPI (using `twine`). Does this repo have a PyPI API Token as a GitHub secret? (That is usually how these uploading jobs work).

dominichamon:
option 1 sounds great and would make a lot of sense to users.

no, we've had no need for a pypi api token.  let's tackle that as a separate PR maybe :)

nicholasjng:
I added wheel test commands and a published release trigger in b1e8848. Every built wheel (except macOS ARM, which cannot yet be tested on GHA runners) passes the tests (one run through the Python bindings).

Also, I edited the `setup.py` to build wheels for minimum `macosx_10_9` on x86, which supports a broader range of OSX releases - with the previous setup, everyone on anything older than macOS 11 would have been left out.

dominichamon:
happy for me to merge?

nicholasjng:
Yeah, I guess we can do the PyPI publishing job separately (I would need your input on that again). 

Would you like me to add a note about bumping the Python bindings version before tagging/release?

dominichamon:
oh, yes if the version needs to be bumped somewhere new then that should be in the docs/releasing.md doc.

nicholasjng:
> oh, yes if the version needs to be bumped somewhere new then that should be in the docs/releasing.md doc.

Done in 62e5038.

dominichamon:
thanks!

Issue #743, opened 2018-12-11
There is no per-benchmark setup when arguments are used


travisdowns:
One of the main reasons to set up to a test fixture is to factor out any expensive setup code for your test, so that it runs the minimum number of times.

However, I can't find the right place to put such setup code, assuming the setup depends on the parameters exposed by `State::range()`. Ideally, you'd call the setup once for each parameter combination. If you put it in the fixture constructor however, it will only be called once benchmark registered using that fixture so it's too coarse (and of course as a consequence `State::range()` isn't available there anyways).

If you put it in `Fixture::Setup` it will be called on _every_ benchmark method invocation, including the repeated invocations as the framework searches for the right number of iterations. This means the same data may be repeatedly generated, which also adds noise to the benchmark (e.g., because memory is repeatedly being allocated and de-allocated).

It would be ideal to have a setup method that is called "per-benchmark" (i.e., per line in the tabular output).

LebedevRI:
https://github.com/google/benchmark#basic-usage
> The benchmark library will measure and report the timing for code within the for(...) loop.

So just put it outside (before) of the `for (auto _ : state)` loop.

Does this answer your question?

travisdowns:
@LebedevRI - no, that is essentially the same as putting it in the `Setup` method (plus some extra duplication) - it will be called not once per benchmark, but once every time the benchmark method is invoked (which is 10+ times as the benchmark hunts for right number of iterations).

LebedevRI:
Ah i see.

> If you put it in `Fixture::Setup` it will be called on _every_ benchmark method invocation, including the repeated invocations as the framework searches for the right number of iterations. This means the same data may be repeatedly generated, which also adds noise to the benchmark (e.g., because memory is repeatedly being allocated and de-allocated).

To be noted, the time outside of the loop isn't counted, thus, there will be no direct noise.

But yes, that is not possible. I'm not sure how of a disruptive change it would be to support that.

travisdowns:
> To be noted, the time outside of the loop isn't counted, thus, there will be no direct noise.

Yes, I didn't mean "direct" noise. What I mean is that constantly allocating and deallocating the per-benchmark data adds an unnecessary source of variance. For example, the new allocations may have a different alignment, or no longer be backed by huge pages, happen to alias differently in the L2 cache, end up coming from a different NUMA node, whatever.

[...]

PR #1269, opened 2021-10-30
Add Setup/Teardown option on Benchmark.

oontvoo:
Motivations:
- feature parity with our internal library. (which has ~718 callers)
- more flexible than cordinating setup/teardown code inside the benchmark routine.

### DIFF
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -113,11 +113,11 @@ PYBIND11_MODULE(_benchmark, m) {
           "complexity",
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
-          py::arg("complexity") = benchmark::oAuto)
-      .def("setup", &Benchmark::Setup,
-           py::return_value_policy::reference_internal)
-      .def("teardown", &Benchmark::Teardown,
-           py::return_value_policy::reference_internal);
+          py::arg("complexity") = benchmark::oAuto);
+  //      .def("setup", &Benchmark::Setup,
+  //      py::return_value_policy::reference_internal)
+  // .def("teardown", &Benchmark::Teardown,
+  //       py::return_value_policy::reference_internal);
 
   using benchmark::Counter;
   py::class_<Counter> py_counter(m, "Counter");
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -114,9 +114,13 @@ PYBIND11_MODULE(_benchmark, m) {
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
           py::arg("complexity") = benchmark::oAuto)
-      .def("setup", &Benchmark::Setup, py::return_value_policy::reference)
-      .def("teardown", &Benchmark::Teardown,
-           py::return_value_policy::reference);
+      .def("setup",
+           (Benchmark * (Benchmark::*)(benchmark::State&)) & Benchmark::Setup,
+           py::return_value_policy::reference)
+      .def(
+          "teardown",
+          (Benchmark * (Benchmark::*)(benchmark::State&)) & Benchmark::Teardown,
+          py::return_value_policy::reference);
 
   using benchmark::Counter;
   py::class_<Counter> py_counter(m, "Counter");
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -114,9 +114,10 @@ PYBIND11_MODULE(_benchmark, m) {
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
           py::arg("complexity") = benchmark::oAuto)
-      .def("setup", &Benchmark::Setup, py::return_value_policy::reference)
+      .def("setup", &Benchmark::Setup,
+           py::return_value_policy::reference_internal)
       .def("teardown", &Benchmark::Teardown,
-           py::return_value_policy::reference);
+           py::return_value_policy::reference_internal);
 
   using benchmark::Counter;
   py::class_<Counter> py_counter(m, "Counter");
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -113,7 +113,10 @@ PYBIND11_MODULE(_benchmark, m) {
           "complexity",
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
-          py::arg("complexity") = benchmark::oAuto);
+          py::arg("complexity") = benchmark::oAuto)
+      .def("setup", &Benchmark::Setup, py::return_value_policy::reference)
+      .def("teardown", &Benchmark::Teardown,
+           py::return_value_policy::reference);
 
   using benchmark::Counter;
   py::class_<Counter> py_counter(m, "Counter");
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -114,13 +114,9 @@ PYBIND11_MODULE(_benchmark, m) {
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
           py::arg("complexity") = benchmark::oAuto)
-      .def("setup",
-           (Benchmark * (Benchmark::*)(benchmark::State&)) & Benchmark::Setup,
-           py::return_value_policy::reference)
-      .def(
-          "teardown",
-          (Benchmark * (Benchmark::*)(benchmark::State&)) & Benchmark::Teardown,
-          py::return_value_policy::reference);
+      .def("setup", &Benchmark::Setup, py::return_value_policy::reference)
+      .def("teardown", &Benchmark::Teardown,
+           py::return_value_policy::reference);
 
   using benchmark::Counter;
   py::class_<Counter> py_counter(m, "Counter");
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -114,10 +114,6 @@ PYBIND11_MODULE(_benchmark, m) {
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
           py::arg("complexity") = benchmark::oAuto);
-  //      .def("setup", &Benchmark::Setup,
-  //      py::return_value_policy::reference_internal)
-  // .def("teardown", &Benchmark::Teardown,
-  //       py::return_value_policy::reference_internal);
 
   using benchmark::Counter;
   py::class_<Counter> py_counter(m, "Counter");
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -244,10 +244,10 @@ information about the machine on which the benchmarks are run.
 
 ## Setup/Teardown
 
-Global setup/teardown specific to each benchmark group can be done by
+Global setup/teardown specific to each benchmark can be done by
 passing a callback to Setup/Teardown:
 
-The setup/teardown callbacks will be invoked once for every group.
+The setup/teardown callbacks will be invoked once for each benchmark.
 If the benchmark is multi-threaded (will run in k threads), they will be invoked exactly once before
 each run with k threads.
 If the benchmark uses different size groups of threads, the above will be true for each size group.
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -18,6 +18,8 @@
 
 [Runtime and Reporting Considerations](#runtime-and-reporting-considerations)
 
+[Setup/Teardown] (#setup-teardown)
+
 [Passing Arguments](#passing-arguments)
 
 [Custom Benchmark Name](#custom-benchmark-name)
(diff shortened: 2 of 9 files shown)


LebedevRI:
There's fixtures already. Why do we need two ways to slice bread?

oontvoo:
> There's fixtures already. Why do we need two ways to slice bread?

The Fixture::Setup/Teardown don't get called at the same time as these (see requirement described in the code comment).  The fixture  also requires boilerplate code for a new class  whereas this new Setup/Teardown could be used with a simple routine.

LebedevRI:
So now we will have two concepts under the same name, and they will do diametrically opposite things?
I agree that one-time global state init/deinit is sometime useful,
but what exactly there can't be achieved via fixture's constructor/destructor?

oontvoo:
> So now we will have two concepts under the same name, and they will do diametrically opposite things? 

We could rename them and/or provide further docs to avoid confusions.  I don't think it's a new "confusion" per se, In gtest, for eg, we already have fixture::setup vs  static setup (ie., once-per-instance vs once-per-class).

> I agree that one-time global state init/deinit is sometime useful, but what exactly there can't be achieved via fixture's > constructor/destructor?

The most common use case we see in google3:  different routines that want to use the similar code for init *but* want to ensure the init is done once for each group of benchmarks (and not all the time).
Eg

```
void Setup(int id, benchmark::State& s) {
    // Populate a cache or setup some test devices
}

// Benchmark reading from map using method a
void BM_read_a(benchmark::State&) {}

// Note: you want the state to be initted for each group of threads, not every time the benchmark routine is run
// Specifically, here you want the `Setup()` to be called 4 TIMES:
//  read_a_arg_128_16_threads , read_a_arg_128_32_threads
//  read_a_arg_256_16_threads , read_a_arg_256_32_threads
BENCHMARK(BM_read_a)->Setup(bind_front(Setup, 2)->Args(128)->Arg(256)->Threads(16)->Threads(32);

// Benchmark reading from map using method b
void BM_read_b(benchmark::State&) {}

BENCHMARK(BM_read_b)->Setup(bind_front(Setup, 2)->Arg(128)->Arg(256)->Threads(16)->Threads(32);

< more routines using similar setup ...> 
```

This would be quite hair to do with the Fixture. Of course, not impossible - but not very user friendly.

Btw, looking at the notes we have internally about the Fixture's implementation here, it seemed like the intention was to have the Fixture::Setup behave the same way as what the Setup() proposed here would, but for some reason, it doesn't? Do you know/remember why? Was there any other discussions regarding this?

Last but not least,  this is part of our effort to unified google internal benchmark API with the oss one. For each disparity, we  try to evaluate carefully whether changes should be made internally or externally (sometimes, it's both). Obviously this doesn't mean it should give the authority to change whatever I want in the OSS Benchmark - I'm just providing the background so hopefully you'd keep an open mind for feature-proposals ... They aren't proposed for no reason. You may not see some features being used externally, it doesn't mean they aren't used all.

dominichamon:
basically, it was considered redundant to have this type of fixture.  if you want behaviour for every benchmark run, you can put it outside the timing loop (before or after).

oontvoo:
[more changes planned  - including updating docs] Please hold off reviewing ....

include/benchmark/benchmark.h
```diff
@@ -968,6 +969,25 @@ class Benchmark {
...
+  // "setup"/"teardown".
+  //
+  // If the benchmark uses different size groups of threads (e.g. via
+  // ThreadRange), the above will be true for each size group.
+  //
+  // The callback will be passed the number of threads for this benchmark run.
+  //
+  // The callback must not be self-deleting.  The Benchmark
+  // object takes ownership of the callback std::function object.
+  Benchmark* SetUp(std::function<void(benchmark::State&)>);
```
LebedevRI, line 987: Can you add motivational test why this can't just be a function pointer like see e.g. `BigOFunc`?

src/benchmark_runner.cc
```diff
@@ -290,6 +291,7 @@ void BenchmarkRunner::DoOneRepetition() {
                                          has_explicit_iteration_count ||
                                          ShouldReportIterationResults(i);
 
+    b.Tearndown();
```
LebedevRI, line 294: This is surely `Teardown`

src/benchmark_api_internal.cc
```diff
@@ -78,6 +78,11 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
   if (!benchmark_.thread_counts_.empty()) {
     name_.threads = StrFormat("threads:%d", threads_);
   }
+
+#if defined(BENCHMARK_HAS_CXX11)
```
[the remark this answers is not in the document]
oontvoo: How about if it's *used* in cxx03, then it won't be able to use these new `SetUp()/Teardown()`?
Or alternatively, #ifdef here so that it takes a pointer in cxx03 and std::function in cxx11? (not sure this would be better - only more confusing)

Our styleguide (at least internally)  prefers `std::function` over raw function pointer (unless there is legit performance concern, which doesn't seem apply here since ).
Furthermore, how long are we planning on supporting 03? I understand we don't want to break existing users, but limiting our API simply due to compatibility with a (very) old C++ version doesn't seem ideal.

src/benchmark_runner.cc
oontvoo: What do you mean?  The `Teardown` callback is passed to the benchmark instance (`b`) so I wrap the call to it over this method . 
Alternatively, we could do `b.teardown_(....)` ?

oontvoo: OOOH, I see now.  you were pointing out the typo!

src/benchmark_api_internal.cc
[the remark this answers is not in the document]
LebedevRI: What i'm saying is that `Benchmark` class **can not** have different layout (different member variables)
depending on the C++ standard with which the TU into which it's included is compiled,
because the library will itself be compiled with C++11+, and it will expect those fields to be there.
This is a fundamental correctness problem.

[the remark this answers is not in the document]
dominichamon: as an OSS project, there are some parts of the style guide we can't/don't necessarily support.  another example is in non-const reference arguments, which we pioneered ;)

C++03 compatibility is still required, even if support is best effort, due to use on platforms where the standard, although old, is the only one supported by compilers for that platform.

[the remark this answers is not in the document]
oontvoo: Ok, that's reasonable. Let's make these take an function pointers.

include/benchmark/benchmark.h
```diff
@@ -982,11 +980,9 @@ class Benchmark {
   //
   // The callback will be passed the number of threads for this benchmark run.
   //
-  // The callback must not be self-deleting.  The Benchmark
-  // object takes ownership of the callback std::function object.
-  Benchmark* SetUp(std::function<void(benchmark::State&)>);
-  Benchmark* TearDown(std::function<void(benchmark::State&)>);
-#endif
+  // The callback must not be NULL or self-deleting.
+  Benchmark* Setup(void (*setup)(benchmark::State&));
```
dominichamon, line 984: should the parameter to the setup/teardown methods be `const`?  i can't imagine we want to change the state during setup or teardown.  it should be safe to, but there's a possibility of breaking some implicit invariant if `State` becomes more complicated.

src/benchmark_runner.cc
```diff
@@ -291,7 +291,7 @@ void BenchmarkRunner::DoOneRepetition() {
                                          has_explicit_iteration_count ||
                                          ShouldReportIterationResults(i);
 
-    b.Tearndown();
+    b.Teardown();
```
dominichamon, line 294: nit: this could come right after the call to `DoNIterations` so it's bracketed by `Setup` and `Teardown`.

include/benchmark/benchmark.h
oontvoo: Good question - hadn't thought of it. I just made it non-const to be consistent (`state` has been tossed around as non-const ref everywhere) 
Practically speaking, the state object passed to these callbacks is thrown away after they're invoked so changes on it has no effect.

How strongly do you feel about changing this to const?

src/benchmark_runner.cc
oontvoo: done.

include/benchmark/benchmark.h
dominichamon: if it's thrown away, even more strongly. not because it could cause a bug but because it indicates the user _can't_ change it prior to the benchmark being called.

oontvoo: Good point! Done! :)

include/benchmark/benchmark.h
```diff
@@ -968,6 +968,22 @@ class Benchmark {
...
+  // If the benchmark is multi-threaded (will run in k threads concurrently),
+  // the setup callback will be be invoked exactly once (not k times) before
+  // each run with k threads. Time allowing (e.g. for a short benchmark), there
+  // may be multiple such runs per benchmark, each run with its own
+  // "setup"/"teardown".
+  //
+  // If the benchmark uses different size groups of threads (e.g. via
+  // ThreadRange), the above will be true for each size group.
+  //
+  // The callback will be passed the number of threads for this benchmark run.
```
dominichamon, line 981: specifically it's passed the state which includes the number of threads...

dominichamon:
can you also please update the docs (user guide) to include this feature?

docs/user_guide.md
```diff
@@ -238,6 +240,38 @@ times and statistical results across these repetitions will also be reported.
...
+```c++
+static void DoSetup(const benchmark::State& state) {
+}
+
+static void DoTeardown(const benchmark::State& state) {
+}
+
+static void BM_func(benchmark::State& state) {...}
+
+BENCHMARK(BM_func)->Arg(1)->Arg(3)->Threads(16)->Threads(32)->Setup(DoSetup)->Teardonw(DoTeardown);
```
dominichamon, line 266: typo: Teardonw -> Teardown

oontvoo: sorry!

docs/user_guide.md
```diff
@@ -238,6 +240,38 @@ times and statistical results across these repetitions will also be reported.
 As well as the per-benchmark entries, a preamble in the report will include
 information about the machine on which the benchmarks are run.
 
+<a name="setup-teardown" />
+
+## Setup/Teardown
+
+Global setup/teardown specific to each benchmark group can be done by
```
dominichamon, line 247: i think it's just "each benchmark".  ie, a benchmark function defines a benchmark family and the args/threads when enumerated define "benchmarks".

dominichamon:
just to check, i think this does _not_ resolve #743, right?

dominichamon:
also please check the CI failures, especially the rather nasty looking segfault in the python bindings run.

test/benchmark_setup_teardown_test.cc
```diff
@@ -0,0 +1,100 @@
...
+#include <cassert>
+#include <cstdlib>
+#include <cstring>
+#include <iostream>
+#include <limits>
+#include <string>
+
+#include "benchmark/benchmark.h"
+
+#ifdef BENCHMARK_HAS_CXX11
```
LebedevRI, line 11: benchmark is always compiled with c++11-capable compiler.

oontvoo:
> just to check, i think this does _not_ resolve #743, right?

From my understanding, yes it does, actually.  They were asking for one setup/teardown "per line in the tabular output", which is what this does.

bindings/python/google_benchmark/benchmark.cc
```diff
@@ -113,7 +113,14 @@ PYBIND11_MODULE(_benchmark, m) {
           "complexity",
           (Benchmark * (Benchmark::*)(benchmark::BigO)) & Benchmark::Complexity,
           py::return_value_policy::reference,
-          py::arg("complexity") = benchmark::oAuto);
+          py::arg("complexity") = benchmark::oAuto)
+      .def("setup",
+           (Benchmark * (Benchmark::*)(benchmark::State&)) & Benchmark::Setup,
```
oontvoo, line 118: This still seems     ... wrong? 
@dominichamon  @LebedevRI

oontvoo: P.S: actually i've figured it out - the explicit cast was wrong. (and it's not needed here anyway)

src/benchmark_runner.cc
```diff
@@ -279,7 +279,9 @@ void BenchmarkRunner::DoOneRepetition() {
   // is *only* calculated for the *first* repetition, and other repetitions
   // simply use that precomputed iteration count.
   for (;;) {
+    b.Setup();
     i = DoNIterations();
+    b.Teardown();
 
     // Do we consider the results to be significant?
     // If we are doing repetitions, and the first repetition was already done,
```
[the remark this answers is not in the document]
oontvoo: Done - added setup/teardown there too.

test/benchmark_setup_teardown_test.cc
```diff
@@ -0,0 +1,100 @@
...
+  assert(concurrent::setup_call.load(std::memory_order_relaxed) == 3);
+  assert(concurrent::teardown_call.load(std::memory_order_relaxed) == 3);
+
+  // 3 group of threads calling this function.
+  assert((5 + 10 + 15) ==
+         concurrent::func_call.load(std::memory_order_relaxed));
+
+  return 0;
+}
+
```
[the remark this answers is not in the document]
oontvoo: Yes - it should be run for each repetition.

oontvoo:
The current ubuntu failures seem really unrelated (eg not finding "master" tag ....).
I've run the tests on my machine and they all passed! (I know this generally isn't good enough but this change isn't platform dependent so if it works on one, it should work on most :) )

dominichamon:
i think not finding master is due to #1285. fix landing soon.

dominichamon:
thank you!

we got there :)

Issue #1172, opened 2021-06-05
[BUG] The library does not build with clang


kpyrkosz:
**Describe the bug**
[  3%] Building CXX object src/CMakeFiles/benchmark.dir/console_reporter.cc.o
[  9%] Building CXX object src/CMakeFiles/benchmark.dir/colorprint.cc.o
[  9%] Building CXX object src/CMakeFiles/benchmark.dir/benchmark.cc.o
[ 12%] Building CXX object src/CMakeFiles/benchmark.dir/commandlineflags.cc.o
[ 15%] Building CXX object src/CMakeFiles/benchmark.dir/csv_reporter.cc.o
[ 18%] Building CXX object src/CMakeFiles/benchmark.dir/json_reporter.cc.o
[ 21%] Building CXX object src/CMakeFiles/benchmark.dir/log.cc.o
[ 24%] Building CXX object src/CMakeFiles/benchmark.dir/reporter.cc.o
[ 27%] Building CXX object src/CMakeFiles/benchmark.dir/sleep.cc.o
[ 30%] Building CXX object src/CMakeFiles/benchmark.dir/string_util.cc.o
[ 33%] Building CXX object src/CMakeFiles/benchmark.dir/sysinfo.cc.o
[ 36%] Building CXX object src/CMakeFiles/benchmark.dir/walltime.cc.o
[ 39%] Building CXX object src/CMakeFiles/benchmark.dir/complexity.cc.o
[ 42%] Building CXX object src/CMakeFiles/benchmark.dir/re_std.cc.o
/tmp/benchmark/src/complexity.cc:89:10: error: variable 'sigma_gn' set but not used [-Werror,-Wunused-but-set-variable]

**System**
`uname -a`
Linux playground 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux
`clang --version`
clang version 13.0.0 (https://github.com/llvm/llvm-project.git db45746821ab01a54f8df033991c3280c4284e3b)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/bin

**To reproduce**
Standard build procedure, cmake && make

**Expected behavior**
The library builds

**Additional context**
https://github.com/google/benchmark/blob/main/src/complexity.cc#L85
[...]

PR #1270, opened 2021-10-30
Work around spurious WASM compilation error with Apple clang

rickwebiii:
When compiling with Clang on MacOS targeting Emscripten, I get the following spurious warning:
```
/Users/rickweber/Projects/SEAL/thirdparty/benchmark-src/src/complexity.cc:85:10: error: variable 'sigma_gn' set but not used [-Werror,-Wunused-but-set-variable]
  double sigma_gn = 0.0;

```

The warning is definitely spurious because `sigma_gn` gets incremented in the for loop in `MinimalLeastSq`.

Proposing a workaround for a spurious compilation warning (and hence error due to -Wall) when compiling with Clang on MacOS targeting emscripten.

I don't like it, so if anyone has a better approach, I'm open to other ideas.

### DIFF
--- a/src/complexity.cc
+++ b/src/complexity.cc
@@ -82,7 +82,16 @@ std::string GetBigOString(BigO complexity) {
 LeastSq MinimalLeastSq(const std::vector<int64_t>& n,
                        const std::vector<double>& time,
                        BigOFunc* fitting_curve) {
+// When compiling benchmarks in WASM on MacOS with Clang, Clang emits an erroneous warning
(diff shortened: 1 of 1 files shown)


rickwebiii:
Doh, let me make sure this isn't fixed in mainline first...

LebedevRI:
Not sure how github allows PR's with so mismatched branches.
This was fixed in e991355c02b93fe17713efe04cbc2e278e00fdbd / https://github.com/google/benchmark/issues/1172.

rickwebiii:
I see that now. Sorry for the noise.

PR #1272, opened 2021-11-04
Fix error with Fix Werror=old-style-cast 

bensuperpc:
Fix error with Fix Werror=old-style-cast


bensuperpc:
@googlebot I signed it!

dominichamon:
thanks!

PR #1276, opened 2021-11-07
Fix warning with MacOS

bensuperpc:
Use reinterpret_cast instead old style cast on MacOS, to fix warning with -Wold-style-cast, more C++ compliant.
It triggers an error when we compiled with gtest.


dominichamon:
nit: is this line > 80 characters?  can you run `clang-format` with the `Google` style on it please?

LebedevRI:
Please don't just reformat the whole file, only the new change.

dominichamon:
thanks!

PR #1283, opened 2021-11-14
Remove bazelbuild/rules_cc dependency

rHermes:
It seems according to [1] that bazelbuild/rules_cc has been put on hold
and that the recommended way for now, is to use the native cc rules.

[1]: https://github.com/bazelbuild/rules_go/pull/2950


dominichamon:
thanks!

PR #1284, opened 2021-11-16
Googletest renamed master to main today.

IbrH:
"master" does not exist anymore, it was changed to "main". As a result cmake throws an error when trying to checkout master.


dominichamon:
closing in favour of #1285 (only because the CLA has been signed on that one).

thank you!

PR #1285, opened 2021-11-16
Googletest renamed master to main today.

dondenton:
It looks like google test migrated from master to main, so I made the corresponding change in the cmake.in file.


yzhang93:
Waiting for the merge of this PR to be able to build other projects

Issue #1147, opened 2021-05-07
[FR] compare.py: compute overall stats - e.g. geomean


LebedevRI:
While just comparing the runs is good, it would also be nice
to be able to judge the changes overall. Are they a win or not? 

At a minimum, i'd like to see geomean.

PR #1289, opened 2021-11-23
compare.py: compute and print 'OVERALL GEOMEAN' aggregate

LebedevRI:
Despite the wide variety of the features we provide,
some people still have the audacity to complain and demand more.

Concretely, i *very* often would like to see the overall result
of the benchmark. Is the 'new' better or worse, overall,
over all the non-aggregate time/cpu measurements.

This comes up for me most often when i want to quickly see
what effect some LLVM optimization change has on the benchmark.

The idea is straight-forward, just produce four lists:
wall times for LHS benchmark, CPU times for LHS benchmark,
wall times for RHS benchmark, CPU times for RHS benchmark;
then compute geomean for each one of those four lists,
and compute the two percentage change between
* geomean wall time for LHS benchmark and geomean wall time for RHS benchmark
* geomean CPU time for LHS benchmark and geomean CPU time for RHS benchmark
and voila!

It is complicated by the fact that it needs to graciously handle
different time units, so pandas.Timedelta dependency is introduced.
That is the only library that does not barf upon floating times,
i have tried numpy.timedelta64 (only takes integers)
and python's datetime.timedelta (does not take nanosecons),
and they won't do.

Fixes https://github.com/google/benchmark/issues/1147

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -7,7 +7,9 @@ import re
 import copy
 import random
 
-from scipy.stats import mannwhitneyu
+from scipy.stats import mannwhitneyu, gmean
+from numpy import array
+from pandas import Timedelta
 
 
 class BenchmarkColor(object):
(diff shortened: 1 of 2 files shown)


tools/gbench/report.py
```diff
@@ -7,7 +7,9 @@
 import copy
 import random
 
-from scipy.stats import mannwhitneyu
+from scipy.stats import mannwhitneyu, gmean
+from numpy import array
+from pandas import Timedelta
```
dominichamon, line 12: pandas should be added to `requirements.txt` i think.

LebedevRI: Thinking about it, should i just hardcode the multipliers needed?

dominichamon: i have no issue with using pandas. i think it's cleaner to use something that is known and does the job with a well-known API.

chfast:
How to use this? I'm doing regular comparison and don't see any "OVERALL GEOMEAN".

LebedevRI:
> How to use this? I'm doing regular comparison and don't see any "OVERALL GEOMEAN".

It's not under any of the options, it's just there. Perhaps the `benchmark/tools` you use doesn't have it?

chfast:
Aaa, I was on `master` branch...

PR #1290, opened 2021-11-24
Add clang-tidy check

### LINKED ISSUE #1126: [FR] clang-tidy

Please adjust the API and inner code to pass clang-tidy. Right now, attempts to use Google Benchmark trigger hundreds of clang-tidy warnings.

For example, the `state` variable is observed by clang-tidy to violate Google's C++ guidelines, specifically the `google-runtime-references` rule.

* Configure a `.clang-tidy` file for Google Benchmark's inner code.
* Design the API so that user code will not trigger clang-tidy, even if all the rules are turned on.

dominc8:
Closes #1126


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: install clang-tidy
+      run: sudo apt update && sudo apt -y install clang-tidy
+
+    - name: create build environment
+      run: cmake -E make_directory ${{ runner.workspace }}/_build
+
+    - name: configure cmake
```
dominichamon, line 21: do you need cmake to run clang-tidy?

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,36 @@
+name: clang-tidy
+
+on:
+  push: {}
```
dominichamon, line 4: on pull request would be good too to catch regressions

.github/workflows/clang-tidy.yml
dominc8: Indirectly because clang-tidy uses compile_commands.json and I generate it using this cmake command.

dominichamon: ah makes sense. thanks.

.github/workflows/clang-tidy.yml
dominichamon: could you add this to the change? then i can merge it.

dominc8: Done, and what do you think about enabling other clang-tidy checks? Because there is like 455 of them in total, and here I enabled only one class of them (clang static analyzer) that has 113 checks. Some other checks/groups might be nice to have

dominichamon:
can we exclude running on third party (i think just excluding `_build` would be sufficient):

`clang-tidy-10 -p=/home/runner/work/benchmark/_build /home/runner/work/benchmark/_build/third_party/googletest/src/googlemock/src/gmock-all.cc`

.github/workflows/clang-tidy.yml
dominichamon: starting small but passing is fine.

we can always enable more as we go :)

dominc8:
clang-tidy runs through all files in compile_commands.json, I added variable which disables googletest sources from this, other option is to run sed and remove all sources that are from third_party.

dominichamon:
think this is ready. thanks so much!

test/skip_with_error_test.cc
```diff
@@ -119,7 +119,7 @@ ADD_CASES("BM_error_during_running", {{"/1/threads:1", true, "error message"},
 
 void BM_error_during_running_ranged_for(benchmark::State& state) {
   assert(state.max_iterations > 3 && "test requires at least a few iterations");
-  int first_iter = true;
+  [[maybe_unused]] bool first_iter = true;
```
dominichamon, line 122: this is a c++17 extension, no?

our bazel build should be setting c++11 but it passed CI. i'm a bit surprised.

dominc8: I think that compilers tend to ignore attributes which they do not recognize, at least this is what it seems to be happening. Locally I added some random attribute like [[asdfasdf]] and it compiled just fine, although clang-tidy threw an error. So it seems that compilers ignore those attributes but clang-tidy recognizes them. This [[maybe_unused]] is to prevent clang-tidy from throwing *clang-analyzer-deadcode.DeadStores*, but it can be also done by adding `(void)first_iter` after `first_iter=false`.

dominichamon: i spotted this because my local build (osx clang bazel (which sets c++11)) failed on this line.  my linux bazel build doesn't fail (which is why the CI bots aren't failing).

this may be opening up the question: can we support c++17 yet? :D  @LebedevRI ?

cmake/GoogleTest.cmake
```diff
@@ -38,7 +38,11 @@ add_subdirectory(${GOOGLETEST_SOURCE_DIR}
                  ${GOOGLETEST_BINARY_DIR}
                  EXCLUDE_FROM_ALL)
 
-set_target_properties(gtest PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:gtest,INTERFACE_INCLUDE_DIRECTORIES>)
-set_target_properties(gtest_main PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:gtest_main,INTERFACE_INCLUDE_DIRECTORIES>)
-set_target_properties(gmock PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:gmock,INTERFACE_INCLUDE_DIRECTORIES>)
-set_target_properties(gmock_main PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:gmock_main,INTERFACE_INCLUDE_DIRECTORIES>)
+if(NOT DEFINED GTEST_COMPILE_COMMANDS)
+    set(GTEST_COMPILE_COMMANDS ON)
+endif()
```
LebedevRI, line 43: Necro-review:
I do not understand what the point of `EXPORT_COMPILE_COMMANDS` is,
if we default to `ON` and then explicitly enable it in that job.
Should it be always-on?

PR #1292, opened 2021-11-29
Add pandas to tools/ requirements.txt

jackgerrits:
#1289 added a usage of `pandas` to the `gbench` module in the `tools/` directory. That PR added `pandas` to the root `requirements.txt` but not the `requirements.txt` in the `tools/` directory.


LebedevRI:
Why do we have two requirements.txt in the first place?

jackgerrits:
I'm not sure, perhaps the tools one is just stale?

PR #1293, opened 2021-11-29
Set C++ 17

gerzin:
On MacOs I got this error:

/Users/gerardozinno/repos/parallel-cellular-automata/tests/ext/benchmark/test/skip_with_error_test.cc:122:5: error: use of the 'maybe_unused' attribute is a
      C++17 extension [-Werror,-Wc++17-extensions]
  [[maybe_unused]] bool first_iter = true;

So I simply and only added `set(CMAKE_CXX_STANDARD 17)` on line 17 of the CMakeLists.txt, everything worked fine after that and the framework compiled without any problem

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,8 @@ endforeach()
 
 project (benchmark VERSION 1.6.0 LANGUAGES CXX)
 
+set(CMAKE_CXX_STANDARD 17)
+
 option(BENCHMARK_ENABLE_TESTING "Enable testing of the benchmark library." ON)
 option(BENCHMARK_ENABLE_EXCEPTIONS "Enable the use of exceptions in the benchmark library." ON)
 option(BENCHMARK_ENABLE_LTO "Enable link time optimisation of the benchmark library." OFF)


LebedevRI:
The right fix would be to drop the attribute, and instead use cast-to-void silencing.

gerzin:
@LebedevRI yes, your solution seems much better. I just wanted to highlight the problem to someone who had familiarity with this code base. I'm just an end user who had the problem and saw a quick fix to it.

LebedevRI:
It is somewhat weird that the CI is green though, why is it not complaining?

dominichamon:
i had this conversation in the PR that introduced the issue.  It seems that many compilers ignore the attribute but OSX Clang doesn't. and bazel explicitly sets C++11 but only runs on ubuntu on CI while cmake (i think) finds the highest available C++ std.

i agree that the `(void)` cast is the best cross-platform support unless we can start considering C++17... :D

dominichamon:
running CI to see if any of our older bots complain.  though this doesn't mean we do want to go to c++17 yet given the plethora of non-CI-covered platforms we try to support.

LebedevRI (changes requested):

PR #1295, opened 2021-12-01
Add pandas to benchmark/tools/requirements.txt

nimrod-gileadi:
Commit d6ba952fc15dbdc54fda46b99babdb9db5fbe58b added pandas as a dependency to the compare.py script, and updated benchmark/requirements.txt, but this file wasn't updated.


LebedevRI:
Why do we have two requirements.txt?
Can we remove one of them?

dominichamon:
we have multiple because we have two python installations: bindings and tools.

`WORKSPACE` refers to the root requirements.txt
`setup.py` refers to bindings/python/requirements.txt
`docs/tools.md` refers to tools/requirements.txt

i think setup.py is correct as it's installing the requirements for the python bindings for the library.

i think docs/tools.md is correct as it's setting up the tooling requirements.

i think the root requirements.txt is redundant and WORKSPACE should instead refer to bindings/python/requirements.txt

nimrod-gileadi:
Are you expecting an action on my part? I'm happy to leave this to library maintainers to sort out.

LebedevRI:
> Are you expecting an action on my part? I'm happy to leave this to library maintainers to sort out.

As per previous comment, let's remove root  requirements.txt, and change WORKSPACE file to point to this file instead.

PR #1298, opened 2021-12-02
clang-tidy: readability-redundant and performance

dominc8:
What do you think about enabling these checks? performance-* in this MR is mainly about changing references to const references and adding std::move if possible, readability-redundant-* mostly removes calling default constructors in initializer lists and doubled "private:" lines. Also there is changing this c++17 [[maybe_unused]] to void cast.


dominichamon:
i'm happy with it.  quite happy to take any robot-driven recommendations.

@LebedevRI may have stronger opinions

include/benchmark/benchmark.h
```diff
@@ -1676,10 +1675,7 @@ class ConsoleReporter : public BenchmarkReporter {
     OO_Defaults = OO_ColorTabular
   };
   explicit ConsoleReporter(OutputOptions opts_ = OO_Defaults)
-      : output_options_(opts_),
```
oontvoo, line 1679: I'm slightly suspicious of this change ... it's making these fields un-initted, no?
(I've just had to debug a problem very recently when I "forgot" to init the field and assume it'd be null/zero-initialised)

dominichamon: it depends on the type being constructed: https://abseil.io/tips/146

the field that's being removed here is `prev_counters_` which is a `UserCounters`, or `std::map<>`.  this has a user-provided (non-default) constructor and so will be value-initialized and safe to use.

however, you're right to be suspicious in general and this is extremely subtle.

test/skip_with_error_test.cc
```diff
@@ -119,12 +119,13 @@ ADD_CASES("BM_error_during_running", {{"/1/threads:1", true, "error message"},
...
 void BM_error_during_running_ranged_for(benchmark::State& state) {
   assert(state.max_iterations > 3 && "test requires at least a few iterations");
-  [[maybe_unused]] bool first_iter = true;
+  bool first_iter = true;
   // NOTE: Users should not write the for loop explicitly.
   for (auto It = state.begin(), End = state.end(); It != End; ++It) {
     if (state.range(0) == 1) {
       assert(first_iter);
       first_iter = false;
+      (void)first_iter;
```
oontvoo, line 128: I dont think this is a good change - using an explicit unused attribute is almost always better than the (void) cast. (because it's clearer)

dominichamon: unfortunately the attribute is not ignored on OSX Clang at c++11 and is causing a build error.

include/benchmark/benchmark.h
oontvoo: Ah, I see. That's really subtle! But then, it's C++ - i should expect nothing less! 🙃

dominichamon:
thanks

PR #1302, opened 2021-12-09
Avoid errors due to "default label in switch which covers all enumeration values" in Windows codepath

mstorsjo:
This applies a fix that used to exist in LLVM's downstream copy of
this library, from
https://github.com/llvm/llvm-project/commit/948ce4e6edec6ad3cdf1911fc3e8e9569140d4ff.

I presume this warning isn't present if built with MSVC or Clang-cl,
but it's printed in MinGW mode. As the benchmark library adds
-Werror, this is a fatal error when builtin MinGW mode.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -344,6 +344,7 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesWindows() {
     C.num_sharing = static_cast<int>(B.count());
     C.level = Cache->Level;
     C.size = Cache->Size;
+    C.type = "Unknown";
     switch (Cache->Type) {
       case CacheUnified:
         C.type = "Unified";
@@ -356,9 +357,6 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesWindows() {
         break;
       case CacheTrace:
         C.type = "Trace";
-        break;
-      default:
-        C.type = "Unknown";
         break;
     }
     res.push_back(C);


dominichamon:
thanks!

mtrofin:
(Based on mstorsjo's comment in https://reviews.llvm.org/D112012) Would we want to soften -Werror in `benchmark`?

dominichamon:
i prefer not to soften it. i rather like that we have a strict build requirement.

mstorsjo:
> i prefer not to soften it. i rather like that we have a strict build requirement.

Would you consider softening it in the case when benchmark is a subcomponent as part of a larger build? (I.e. cmake root for isn’t equal to the current directory.) Alternatively a way to opt out of the behavior? I appreciate the desire to keep it strict when focusing on benchmark itself, but when pulled in as a dependency of something else, `-Werror` often has a tendency to cause a bit of frustration. In the case of LLVM also, LLVM does have an option for enabling `-Werror`, but one can’t use that for disabling the flag for this subproject.

dominichamon:
there's a weird virality here, i think.  if benchmark is included as a subproject, we have to adopt the least strict options of all projects that include it.  but the opposite is true, that the parent project has to adopt the most strict of its subprojects.

ideally, imo, a subproject should be built as a separate unit and linked in to the parent project build. building everything all-in-one just causes these sorts of issues.

LebedevRI:
We already have `BENCHMARK_ENABLE_WERROR`/`BENCHMARK_FORCE_WERROR` though.

dominichamon:
😂 .. the project is moving too fast for me to keep track of.

@mstorsjo does this help? 😁

mstorsjo:
> We already have `BENCHMARK_ENABLE_WERROR`/`BENCHMARK_FORCE_WERROR` though.

Oh, indeed, that does help. That's not available in the old copy of benchmark present in LLVM, but I guess one could suggest backporting 285e5e97715b887d53c99ddf0a736a2a6baad257 to LLVM's copy - or just update the whole library to the latest version.

dominichamon:
updating llvm to the latest sounds like a good idea _anyway_ given the constant stream of changes/improvements.

mstorsjo:
> updating llvm to the latest sounds like a good idea _anyway_ given the constant stream of changes/improvements.

Yeah, that's of course the ideal. But doing that might end up exposing other issues (which of course would be great to fix and sort out) which would be better handled by someone with more time to look into it...

mtrofin:
I'll do an update to latest, and also then pre-set BENCHMARK_ENABLE_WERROR off.

mstorsjo:
> I'll do an update to latest, and also then pre-set BENCHMARK_ENABLE_WERROR off.

Thanks, that'd be appreciated!

mtrofin:
Waiting for #1305, which should unblock refreshing the llvm copy.

PR #1308, opened 2021-12-17
Cache PerfCounters instance in PerfCountersMeasurement

### LINKED ISSUE #1306: [BUG] User counters only showing results for first five of total set of permutations
**Describe the bug**
I'm trying to use libpfm and the performance counters offered by this library, but it only appears to give me some of the results

**System**
  - OS: Debian 11
  - Compiler and version:  gcc 10.2

**To reproduce**
1. sync to commit master
2. compile example with "-DHAVE_LIBPFM" / link libpfm
3. run with "--benchmark_perf_counters=CYCLES"
4. See output

**Expected behavior**
Metrics are given for all permutations

**Screenshots**
image
image

**Additional context**
sample reproducible 

```
#include <cstdlib>

static std::vector<std::int64_t> A{1, 10};
static std::vector<std::int64_t> B{100, 500, 1000};
static std::vector<std::int64_t> C{50, 100, 150, 200};

template<std::size_t X, std::size_t Y, bool B>
struct Fixture : public benchmark::Fixture
{
public:
    void SetUp(::benchmark::State const& state) override
    {
        std::uint32_t a = state.range(0);
        std::uint32_t b = state.range(1);

        doubles.clear();
        for (std::size_t i = 0; i < a; ++i)
            doubles.emplace_back(b);
    }

    void run(::benchmark::State& state)
    {
        for (auto _ : state)
            for (auto& vec : doubles)
                for (auto& d : vec)
                    d += 1;
    }

    std::vector<std::vector<double>> doubles;
};

BENCHMARK_TEMPLATE_DEFINE_F(Fixture, _3_1_1, 3, 1, 1)(benchmark::State& state)
{
    run(state);
}
BENCHMARK_REGISTER_F(Fixture, _3_1_1)->Unit(benchmark::kNanosecond)->ArgsProduct({A, B, C});

BENCHMARK_TEMPLATE_DEFINE_F(Fixture, _3_1_0, 3, 1, 0)(benchmark::State& state)
{
    run(state);
}
BENCHMARK_REGISTER_F(Fixture, _3_1_0)->Unit(benchmark::kNanosecond)->ArgsProduct({A, B, C});

int main(int argc, char** argv)
{
    ::benchmark::Initialize(&argc, argv);
    ::benchmark::RunSpecifiedBenchmarks();
}
```

taoliq:
This patch fixes #1306, by reducing the pinned instances of
PerfCounters.

The issue is caused by creating multiple pinned events in the
same thread, doing so results in the Snapshot(PerfCounterValues* values)
failing, and that's now discoverable.
Creating multile pinned events is an unsupported behavior currently.
The error would be detected at read() time, not
perf_event_open() / iotcl() time.

The unsupported benavior above is confirmed by Stephane Eranian @seranian,
and he also pointed the dectection method.

Finished this patch under the guidance of Mircea Trofin @mtrofin.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -184,7 +184,10 @@ void State::PauseTiming() {
   BM_CHECK(started_ && !finished_ && !error_occurred_);
   timer_->StopTimer();
   if (perf_counters_measurement_) {
-    auto measurements = perf_counters_measurement_->StopAndGetMeasurements();
+    std::vector<std::pair<std::string, double>> measurements;
+    if (!perf_counters_measurement_->Stop(measurements)) {
+      BM_CHECK(false) << "Perf counters read the value failed.";
+    }
     for (const auto& name_and_measurement : measurements) {
       auto name = name_and_measurement.first;
       auto measurement = name_and_measurement.second;
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -15,6 +15,7 @@
 #include "perf_counters.h"
 
 #include <cstring>
+#include <memory>
 #include <vector>
 
 #if defined HAVE_LIBPFM
@@ -104,7 +105,7 @@ PerfCounters PerfCounters::Create(
   return PerfCounters(counter_names, std::move(counter_ids));
 }
 
-PerfCounters::~PerfCounters() {
+void PerfCounters::CloseCounters() const {
   if (counter_ids_.empty()) {
     return;
   }
@@ -126,7 +127,44 @@ PerfCounters PerfCounters::Create(
   return NoCounters();
 }
 
-PerfCounters::~PerfCounters() = default;
+void PerfCounters::CloseCounters() const {}
 #endif  // defined HAVE_LIBPFM
+
+Mutex PerfCountersMeasurement::mutex_;
+int PerfCountersMeasurement::ref_count_ = 0;
+PerfCounters PerfCountersMeasurement::counters_ = PerfCounters::NoCounters();
+
+PerfCountersMeasurement::PerfCountersMeasurement(
+    const std::vector<std::string>& counter_names)
+    : start_values_(counter_names.size()), end_values_(counter_names.size()) {
+  MutexLock l(mutex_);
+  if (ref_count_ == 0) {
+    counters_ = PerfCounters::Create(counter_names);
+  }
+  // We chose to increment it even if `counters_` ends up invalid,
+  // so that we don't keep trying to create, and also since the dtor
+  // will decrement regardless of `counters_`'s validity
+  ++ref_count_;
+
+  BM_CHECK(!counters_.IsValid() || counters_.names() == counter_names);
+}
+
+PerfCountersMeasurement::~PerfCountersMeasurement() {
+  MutexLock l(mutex_);
+  --ref_count_;
+  if (ref_count_ == 0) {
+    counters_ = PerfCounters::NoCounters();
+  }
+}
+
+PerfCounters& PerfCounters::operator=(PerfCounters&& other) noexcept {
+  if (this != &other) {
+    CloseCounters();
+
+    counter_ids_ = std::move(other.counter_ids_);
+    counter_names_ = std::move(other.counter_names_);
+  }
+  return *this;
+}
 }  // namespace internal
 }  // namespace benchmark
--- a/src/perf_counters.h
+++ b/src/perf_counters.h
@@ -17,11 +17,13 @@
 
 #include <array>
 #include <cstdint>
+#include <memory>
 #include <vector>
 
 #include "benchmark/benchmark.h"
 #include "check.h"
 #include "log.h"
+#include "mutex.h"
 
 #ifndef BENCHMARK_OS_WINDOWS
 #include <unistd.h>
@@ -71,12 +73,14 @@ class PerfCounters final {
   // True iff this platform supports performance counters.
   static const bool kSupported;
 
-  bool IsValid() const { return is_valid_; }
+  bool IsValid() const { return !counter_names_.empty(); }
   static PerfCounters NoCounters() { return PerfCounters(); }
 
-  ~PerfCounters();
+  ~PerfCounters() { CloseCounters(); }
   PerfCounters(PerfCounters&&) = default;
   PerfCounters(const PerfCounters&) = delete;
+  PerfCounters& operator=(PerfCounters&&) noexcept;
+  PerfCounters& operator=(const PerfCounters&) = delete;
 
   // Platform-specific implementations may choose to do some library
   // initialization here.
@@ -111,24 +115,27 @@ class PerfCounters final {
  private:
   PerfCounters(const std::vector<std::string>& counter_names,
                std::vector<int>&& counter_ids)
-      : counter_ids_(std::move(counter_ids)),
-        counter_names_(counter_names),
-        is_valid_(true) {}
-  PerfCounters() : is_valid_(false) {}
+      : counter_ids_(std::move(counter_ids)), counter_names_(counter_names) {}
+  PerfCounters() = default;
+
+  void CloseCounters() const;
 
   std::vector<int> counter_ids_;
-  const std::vector<std::string> counter_names_;
-  const bool is_valid_;
+  std::vector<std::string> counter_names_;
 };
 
 // Typical usage of the above primitives.
 class PerfCountersMeasurement final {
  public:
-  PerfCountersMeasurement(PerfCounters&& c)
-      : counters_(std::move(c)),
-        start_values_(counters_.IsValid() ? counters_.names().size() : 0),
-        end_values_(counters_.IsValid() ? counters_.names().size() : 0) {}
+  PerfCountersMeasurement(const std::vector<std::string>& counter_names);
+  ~PerfCountersMeasurement();
 
+  // The only way to get to `counters_` is after ctor-ing a
+  // `PerfCountersMeasurement`, which means that `counters_`'s state is, here,
+  // decided (either invalid or valid) and won't change again even if a ctor is
+  // concurrently running with this. This is preferring efficiency to
+  // maintainability, because the address of the static can be known at compile
+  // time.
   bool IsValid() const { return counters_.IsValid(); }
 
   BENCHMARK_ALWAYS_INLINE void Start() {
@@ -136,30 +143,33 @@ class PerfCountersMeasurement final {
     // Tell the compiler to not move instructions above/below where we take
     // the snapshot.
     ClobberMemory();
-    counters_.Snapshot(&start_values_);
+    valid_read_ &= counters_.Snapshot(&start_values_);
     ClobberMemory();
   }
 
-  BENCHMARK_ALWAYS_INLINE std::vector<std::pair<std::string, double>>
-  StopAndGetMeasurements() {
+  BENCHMARK_ALWAYS_INLINE bool Stop(
+      std::vector<std::pair<std::string, double>>& measurements) {
     assert(IsValid());
     // Tell the compiler to not move instructions above/below where we take
     // the snapshot.
     ClobberMemory();
-    counters_.Snapshot(&end_values_);
+    valid_read_ &= counters_.Snapshot(&end_values_);
     ClobberMemory();
 
-    std::vector<std::pair<std::string, double>> ret;
     for (size_t i = 0; i < counters_.names().size(); ++i) {
       double measurement = static_cast<double>(end_values_[i]) -
                            static_cast<double>(start_values_[i]);
-      ret.push_back({counters_.names()[i], measurement});
+      measurements.push_back({counters_.names()[i], measurement});
     }
-    return ret;
+
+    return valid_read_;
   }
 
  private:
-  PerfCounters counters_;
+  static Mutex mutex_;
+  GUARDED_BY(mutex_) static int ref_count_;
+  GUARDED_BY(mutex_) static PerfCounters counters_;
+  bool valid_read_ = true;
   PerfCounterValues start_values_;
   PerfCounterValues end_values_;
 };
--- a/test/perf_counters_gtest.cc
+++ b/test/perf_counters_gtest.cc
@@ -11,6 +11,7 @@ struct MsgHandler {
 #endif
 
 using benchmark::internal::PerfCounters;
+using benchmark::internal::PerfCountersMeasurement;
 using benchmark::internal::PerfCounterValues;
 
 namespace {
@@ -93,6 +94,53 @@ TEST(PerfCountersTest, Read2Counters) {
   EXPECT_TRUE(counters.Snapshot(&values2));
   EXPECT_GT(values2[0], 0);
   EXPECT_GT(values2[1], 0);
+}
+
+TEST(PerfCountersTest, ReopenExistingCounters) {
+  // The test works (i.e. causes read to fail) for the assumptions
+  // about hardware capabilities (i.e. small number (3-4) hardware
+  // counters) at this date.
+  if (!PerfCounters::kSupported) {
+    GTEST_SKIP() << "Test skipped because libpfm is not supported.\n";
+  }
+  EXPECT_TRUE(PerfCounters::Initialize());
+  std::vector<PerfCounters> counters;
+  counters.reserve(6);
+  for (int i = 0; i < 6; i++)
+    counters.push_back(PerfCounters::Create({kGenericPerfEvent1}));
+  PerfCounterValues values(1);
+  EXPECT_TRUE(counters[0].Snapshot(&values));
+  EXPECT_FALSE(counters[4].Snapshot(&values));
+  EXPECT_FALSE(counters[5].Snapshot(&values));
+}
+
+TEST(PerfCountersTest, CreateExistingMeasurements) {
+  // The test works (i.e. causes read to fail) for the assumptions
+  // about hardware capabilities (i.e. small number (3-4) hardware
+  // counters) at this date,
+  // the same as previous test ReopenExistingCounters.
+  if (!PerfCounters::kSupported) {
+    GTEST_SKIP() << "Test skipped because libpfm is not supported.\n";
+  }
+  EXPECT_TRUE(PerfCounters::Initialize());
+  std::vector<PerfCountersMeasurement> perf_counter_measurements;
+  std::vector<std::pair<std::string, double>> measurements;
+
+  perf_counter_measurements.reserve(10);
+  for (int i = 0; i < 10; i++)
+    perf_counter_measurements.emplace_back(
+        std::vector<std::string>{kGenericPerfEvent1});
+
+  perf_counter_measurements[0].Start();
+  EXPECT_TRUE(perf_counter_measurements[0].Stop(measurements));
+
+  measurements.clear();
+  perf_counter_measurements[8].Start();
+  EXPECT_FALSE(perf_counter_measurements[8].Stop(measurements));
+
+  measurements.clear();
+  perf_counter_measurements[9].Start();
+  EXPECT_FALSE(perf_counter_measurements[9].Stop(measurements));
 }
 
 size_t do_work() {
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -152,8 +152,7 @@ BenchmarkRunner::BenchmarkRunner(
       has_explicit_iteration_count(b.iterations() != 0),
       pool(b.threads() - 1),
       iters(has_explicit_iteration_count ? b.iterations() : 1),
-      perf_counters_measurement(
-          PerfCounters::Create(StrSplit(FLAGS_benchmark_perf_counters, ','))),
+      perf_counters_measurement(StrSplit(FLAGS_benchmark_perf_counters, ',')),
       perf_counters_measurement_ptr(perf_counters_measurement.IsValid()
                                         ? &perf_counters_measurement
                                         : nullptr) {


taoliq:
I think there is a better way to use `valid_read_`, which does not change the return type of StopAndGetMeasurements().
Looking forward to your advice.

mtrofin:
> …nces (fixes #1306)
> 
> Finished this patch under the guidance of Mircea Trofin @mtrofin. This unsupported behavior is confirmed by Stephane Eranian @seranian.
You mean "the behavior of not supporting `perf_event_open`-ing more than once per counter per thread" - because "unsupported behavior" would indicate we're somehow depending on unsupported behavior, which would be bad.
> 
> The issue is, it is not supported to creat multiple pinned events in the same threads, doing so results in the Snapshot(PerfCounterValues* values) failing, and that's now discoverable.
typo: creat->create; threads->thread

Also, I recommend starting with the fix and leaving credits at the end.
> 
> The error would be detected at read() time, not perf_event_open() / iotcl() time, pointed by Stephane Eranian @seranian.

mtrofin:
> I think there is a better way to use `valid_read_`, which does not change the return type of `StopAndGetMeasurements()`. Looking forward to your advice.

We use BM_CHECK in `benchmark_runner.cc` to message if we cannot set up counters; what would the alternative be?

LebedevRI:
Is this fixing two bugs at once?
* pause/resume
* ???

I do not understand why `PerfCounters::Create` is hoisted out of `BenchmarkRunner::BenchmarkRunner()`,
and i do not believe this change is correct, at least semantically.

src/perf_counters.h
```diff
@@ -136,17 +136,18 @@ class PerfCountersMeasurement final {
...
     // the snapshot.
     ClobberMemory();
-    counters_.Snapshot(&start_values_);
+    valid_read_ &= counters_.Snapshot(&start_values_);
     ClobberMemory();
   }
 
-  BENCHMARK_ALWAYS_INLINE std::vector<std::pair<std::string, double>>
+  BENCHMARK_ALWAYS_INLINE
+  std::pair<std::vector<std::pair<std::string, double>>, bool>
```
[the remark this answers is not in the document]
taoliq: - test/perf_counter_gtest.cc has been updated. To make sure trigger the unsupported behavior, 4 counters (each hold 2 hardware events) are created.
- test/perf_counter_test.cc will be updated soon.

test/perf_counters_gtest.cc
```diff
@@ -95,6 +95,22 @@ TEST(PerfCountersTest, Read2Counters) {
   EXPECT_GT(values2[1], 0);
 }
 
+TEST(PerfCountersTest, UnsupportedMutipleHardwareCounters) {
```
mtrofin, line 98: `UnsupportedMultipleHardwareCounters` is misleading, because it seems to indicate we can't collect more than one counter (i.e. more than, for example, `CYCLES`). How about `ReopenExistingCounters` (no need to say it's unsupported).

Could you also add a test for `PerfCountersMeasurement::StopAndGetMeasurement` usage, that's how the benchmark uses the counters, it is representative of the way the problem surfaced, easier to understand, too (you just need 2 calls to Create and 2 StopAndGetMeasurement calls), and also tests that API's return of the bool.

Also - I think it's be easier in both tests if you just used one of the events, e.g. `kGenericPerfEvent1`. Otherwise it's confusing what the problem is: it's not that you track 2 different events; it's that you track the same one twice.

taoliq:
> Is this fixing two bugs at once?
> 
>     * pause/resume
> 
>     * ???

It fix only one bug, the output would show some wired values(0 or negative) for performance counter.
I didn't catch the word `pause/resume`.  May I have some more explanation?

> I do not understand why `PerfCounters::Create` is hoisted out of `BenchmarkRunner::BenchmarkRunner()`, and i do not believe this change is correct, at least semantically.

You are right. It is not obvious at semantic. We should go deeper to see the difference.
Let's focus on one difference after https://github.com/google/benchmark/commit/fbc31405b2a2838560e155d7aec937e135ae0d81
The difference is the order of creating and running among all BenchmarkRunner.

Let's say, 10 BenchmarkRunners, runner1, runner2, ... runner10.
and 1 counter (cycle)

before the commit
instantiate runner1, run runner1, instantiate runner2, run runner2, ... instantiate runner10, run runner10.
at the commit
instantiate runner1, runner2 ... runner10, and then run them in order or ramdomly.

Notice, a BenchmarkRunner will hold a PerfCounters, which will hold the file descriptor of performance counter in hardware.
If we print the file descriptor in PerfCounter::Create()
In my machine, 
before the commit
3, 3, 3, ... 3
at the commit
3, 4, 5, ... 12

We could see, before the commit, the same file descriptor is used because the previous instance dtor() is invoked and the fd is released.
at the commit, too many pinned events are created, which is the unsupported behavior.

test/perf_counters_gtest.cc
taoliq: unit tests are updated

Call Creat() twice is not enought to trigger the error. 
Both in #1306 or locally, for single events, it is able to create 5 PerfCounter withough wrong return value.
So I create 10 instances in the unit tests.

LebedevRI:
> at the commit, too many pinned events are created, which is the unsupported behavior.

Ok, so the TLDR is: there is a limit on the number of readers for each event.

@mtrofin it sounds like what this really needs is a caching layer/multiplexing,
so that opening the same perf counter again and again doesn't actually open it?

mtrofin:
> > at the commit, too many pinned events are created, which is the unsupported behavior.
> 
> Ok, so the TLDR is: there is a limit on the number of readers for each event.
> 
> @mtrofin it sounds like what this really needs is a caching layer/multiplexing, so that opening the same perf counter again and again doesn't actually open it?

Not sure it needs to be that complicated - @taoliq's change does that in essence: we set up one file descriptor. Then we just read that fd. Since we run one benchmark iteration at a time, sequentially, it doesn't matter that the fd is reused (well... there's a larger chance of overflow, but that's a separate thing).

Or is this about a design consideration? In that case, all we'd need to do is have PerfCounters::Create maintain some global state.

LebedevRI:
> > > at the commit, too many pinned events are created, which is the unsupported behavior.
> > 
> > 
> > Ok, so the TLDR is: there is a limit on the number of readers for each event.
> > @mtrofin it sounds like what this really needs is a caching layer/multiplexing, so that opening the same perf counter again and again doesn't actually open it?
> 
> Not sure it needs to be that complicated - @taoliq's change does that in essence: we set up one file descriptor. Then we just read that fd. Since we run one benchmark iteration at a time, sequentially, it doesn't matter that the fd is reused (well... there's a larger chance of overflow, but that's a separate thing).
> 
> Or is this about a design consideration? In that case, all we'd need to do is have PerfCounters::Create maintain some global state.

It is a design consideration, yes.

mtrofin:
> > > > at the commit, too many pinned events are created, which is the unsupported behavior.
> > > 
> > > 
> > > Ok, so the TLDR is: there is a limit on the number of readers for each event.
> > > @mtrofin it sounds like what this really needs is a caching layer/multiplexing, so that opening the same perf counter again and again doesn't actually open it?
> > 
> > 
> > Not sure it needs to be that complicated - @taoliq's change does that in essence: we set up one file descriptor. Then we just read that fd. Since we run one benchmark iteration at a time, sequentially, it doesn't matter that the fd is reused (well... there's a larger chance of overflow, but that's a separate thing).
> > Or is this about a design consideration? In that case, all we'd need to do is have PerfCounters::Create maintain some global state.
> 
> It is a design consideration, yes.

Got it, sgtm.

LebedevRI:
Also, which libpfm version is that? Is this a known (and documented!) limitation, or should this be reported upstream?

mtrofin:
> Also, which libpfm version is that? Is this a known (and documented!) limitation, or should this be reported upstream?

It's a `perf_event_open` issue. Libpfm just translates counter names to IDs.

mtrofin:
@LebedevRI This is the significant blurb from the `perf_event_open` [man page](https://man7.org/linux/man-pages/man2/perf_event_open.2.html):

> _pinned_  the pinned bit specifies that the counter should always be on the CPU if at all possible.  It applies only to hardware counters and only to group leaders. **If a pinned counter cannot be put onto the CPU (e.g., because there              are not enough hardware counters or because of a conflict with some other event), then the counter goes into an 'error' state, where reads return end-of-file (i.e., read(2) returns 0) until the counter is subsequently enabled or disabled.**

(and we want the counters pinned for better accuracy)

BTW, @taoliq, the tests should be tied to a specific architecture (intel), because the number of counters it takes to request until we get the error behavior is related to the limitations on that architecture (so max 3 on Intel, IIUC). Maybe also add a comment on the test spelling out that the test works (i.e. causes `read` to fail) for the assumptions about hardware capabilities at this date - so a future maintainer would know to bump the number of counter requests in the test over a future higher limit, should that happen (i.e. if the hardware were to support, say, 5 pinned counters at a latter date). You could alternatively just request an aberrantly large number of counters, like 20, and not make the test architecture-dependent (but I'd still put a comment there to explain all this, i.e. that we go with `20` because we expect it to be a 'too large' number of counters, irrespective of the architecture, now or in the future)

src/benchmark_runner.h
```diff
@@ -82,7 +83,6 @@ class BenchmarkRunner {
   // So only the first repetition has to find/calculate it,
   // the other repetitions will just use that precomputed iteration count.
 
-  PerfCountersMeasurement perf_counters_measurement;
   PerfCountersMeasurement* const perf_counters_measurement_ptr;
```
dominichamon, line 86: nit: this no longer needs the `_ptr` suffix as it's the only way we store a reference to it.

src/perf_counters.h
dominichamon, line 144: i think this return type is getting to the point where we should consider a separate data structure.  i'd also be ok with a boolean return for validity and an out param for the vector of measurements.

dominichamon (changes requested):
i think @LebedevRI is right about hoisting the perf counters out of the runner (which may negate some of my comments) and a layer to avoid recreation of existing counters is probably a better design.

this is starting to feel "hacked in".

src/perf_counters.cc
```diff
@@ -32,17 +32,22 @@ const bool PerfCounters::kSupported = true;
 
 bool PerfCounters::Initialize() { return pfm_initialize() == PFM_SUCCESS; }
 
-PerfCounters PerfCounters::Create(
+PerfCounters* PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
+  static std::unique_ptr<PerfCounters> perf_counters_ptr;
```
mtrofin, line 37: How about this instead, so we also exit cleanly (i.e. cleaning up file descriptors):
- don't have the statics here (lines 37-38)
- add a `int ref_count_ =1` field to `PerfCounters`, and a `void refcount() { ++ref_count; }`
- `PerfCountersMeasurement`:
   - has a static `std::unique_ptr<PerfCounters> counters_` field.
   - its ctor takes a `const std::vector<std::string>& counter_names`. It checks if `counters_` is nullptr, if so, creates it with the given `counter_names` and returns its contained ptr. If it's not nullptr, then sanity check the `counter_names` match what we have in `counters_`, and `counters_->refcount(); return counters_.get()`.
   -  the dtor of `PerfCounters` decrements `ref_count_`, and when that hits 0, closes the file descriptors as it does today

Add a comment on `PerfCountersMeasurement` that the ctor is not threadsafe (which should be fine, the scenario is that these are create upfront, at set up); and that the string vector must match what's cached (which is both fine and unavoidable, the whole point is that we can't have too many simultaneously `perf_event_open`-ed counters.)

src/perf_counters.h
taoliq: Add PerfCountersMeasurement::Result as the returned data structure.

src/benchmark.cc
```diff
@@ -184,8 +184,10 @@ void State::PauseTiming() {
   BM_CHECK(started_ && !finished_ && !error_occurred_);
   timer_->StopTimer();
   if (perf_counters_measurement_) {
-    auto measurements = perf_counters_measurement_->StopAndGetMeasurements();
-    for (const auto& name_and_measurement : measurements) {
+    internal::PerfCountersMeasurement::Result measurement_result;
```
dominichamon, line 187: this would be slightly simpler if `Stop` returns the validity of the read:

```c++
internal::PerfCountersMeasurement::Results = measurements;
bool is_valid = perf_counters_measurement_->Stop(&measurements);
BM_CHECK(is_valid) << "Perf counters value read failed";
for (const auto& name_and_measurement : measurements) {
...
}
```

src/perf_counters.cc
```diff
@@ -128,5 +129,22 @@ PerfCounters PerfCounters::Create(
 
 PerfCounters::~PerfCounters() = default;
 #endif  // defined HAVE_LIBPFM
+
+PerfCountersMeasurement* PerfCountersMeasurement::Create(
+    const std::vector<std::string>& counter_names) {
+  static std::unique_ptr<PerfCountersMeasurement> perf_counters_measurement_ptr;
+  static std::vector<std::string> cached_counter_names;
+
+  if (perf_counters_measurement_ptr) {
```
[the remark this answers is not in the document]
taoliq: Put those initialization code a into lambda, and then use `std::call_once` to invoke that.
Have I got that right?

src/benchmark.cc
taoliq: If `Stop` returns the validity, do we need to remove `PerfCountersMeasurement::Results` 
and continue to use `std::vector<std::pair<std::string, double>> measurements`?

mtrofin: No, just return a bool from `Stop`. It already takes an object to pour in the results. So the code becomes simpler, like @dominichamon  pointed out:

```
if(!perf_counters_measurement_->Stop(&measurement_result))
  BM_CHECK(false) << "Perf counters value read failed"
```
@dominichamon  - is there a more appropriate way to report the error (`std:err << ` and exit?)

test/perf_counters_gtest.cc
```diff
@@ -95,6 +96,54 @@ TEST(PerfCountersTest, Read2Counters) {
   EXPECT_GT(values2[1], 0);
 }
 
+TEST(PerfCountersTest, ReopenExistingCounters) {
+  // The test works (i.e. causes read to fail) for the assumptions
```
mtrofin, line 100: specify what those hardware capabilities are: small number (3-4) hardware counters.

src/perf_counters.cc
mtrofin: OK, but this still relies on process shutdown to clean up everything, which is not quite best-practice.

I fixed my earlier comment above, it wasn't quite right; also, if thread safety is a concern, it can easily be mitigated there by having a mutex for operations on the shared `counters` and the `ref_count_` field.

taoliq: Implemented this in be1b7.
Mutex will be added soon (some bugs appeared).

src/perf_counters.h
```diff
@@ -124,44 +126,44 @@ class PerfCounters final {
...
     // the snapshot.
     ClobberMemory();
-    counters_.Snapshot(&start_values_);
+    valid_read_ &= counters_->Snapshot(&start_values_);
     ClobberMemory();
   }
 
-  BENCHMARK_ALWAYS_INLINE std::vector<std::pair<std::string, double>>
-  StopAndGetMeasurements() {
+  BENCHMARK_ALWAYS_INLINE bool Stop(std::vector<std::pair<std::string, double>> *measurements) {
```
dominichamon, line 143: you assume `measurements` is non-null, so why not receive it as a reference instead of a pointer?

dominichamon (changes requested):
i appreciate all the work here.. it's really close i think to being done, and it's a big improvement.

src/perf_counters.h
mtrofin: @dominichamon do we follow the Google coding style here? (asking to learn, I kind of assumed it)

mtrofin, line 143: Why did we rename this to `Stop`?

src/perf_counters.cc
```diff
@@ -128,5 +129,33 @@ PerfCounters PerfCounters::Create(
...
+    const std::vector<std::string>& counter_names)
+    : start_values_(counter_names.size()),
+      end_values_(counter_names.size()) {
+  {
+    // TODO: Add MutexLock
+    //MutexLock l(mutex_);
+    if (!counters_) {
+      PerfCounters perf_counters = PerfCounters::Create(counter_names);
+      counters_ = std::unique_ptr<PerfCounters> (
+          new PerfCounters(std::move(perf_counters)));
```
mtrofin, line 146: This is can be simplified if PerfCounters::Create returns a std::unique_ptr; OR come to think of it, I think we wanted PerfCounters to be allocatable inlined (i.e. on the stack, or as a field of a class/struct), so we'd avoid a dereference and added noise to measurements.

So let's have `counters_` be a PerfCounters object, invalid initially (add a default ctor for `PerfCounters` to do that) and a `reset()` function on it that does what the dtor does today (the dtor would call it, too, for completeness). The design is logically the same.

src/perf_counters.h
dominichamon: yes we do.

mtrofin: Ack  - @taoliq, that's why we pass by mutable pointer, because we'd only pass by reference if it were const, and that wouldn't be applicable here.

dominichamon: the google style guide allows non-const references on parameters (since last year)

mtrofin: Oh! Great! (also, finally!). Thanks!

taoliq: I did it to keep consistance, 
because when I moved the result measurements from return value to the parameter, 
I noticed there are other `void ::Stop(Result* result)`.

src/perf_counters.cc
taoliq: When I tried to make `counters_` be a PerfCounters object, here is a problem.
The `counters_` is constructed firstly, and the `counter_names` is passed in later.
How about add move assignment to `PerfCounters`, so we could use `PerfCounters::Create()` and `PerfCounters::NoCounters()` without any modification.

mtrofin: By all means - add what you need.

mtrofin:
> > you also want to clear counter_ids_ after all is cleaned up.
> 
> ```
> ~PerfCounters() { CloseCounters(); }
> ```
> 
> Just remove the code from d-tor outside, logically the same as before. Is this enough to clear counter_ids_?

The difference is that before that was happening when the object was being destroyed, so there was no subsequent state to observe. Now there is: I can hang on to the object after calling `CloseCounters()`. So clearing, combined with computing IsValid() based on counter_ids_, avoids that sort of problem: the object is either "invalid" - meaning no counter_ids_ - or valid, case in which it has counter_ids_ which need eventual closing.

taoliq:
> looks fine to me, one thing, there's a message at the top of the review `This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.` - some branching snafu?

I think it's caused by my force-pushed commits. After the last force-pushed commit, here is no message like that.

taoliq:
> You can probably add a test that repeatedly trying to create PerfCountesMeasurement with garbage input only outpus error message once.

I didn't add test like that. I don't know how to check the error message. 
Is there any test which is similar to that?

taoliq:
I saw the CI failed in this PR, [here](https://github.com/google/benchmark/runs/4839509854?check_suite_focus=true)
The failed case also appeared in my forked branch at a low probability.
I think the flake is not due to my change, and I reproduce that in the main branch without my change.
see here
https://github.com/taoliq/benchmark/actions/runs/1738261739
https://github.com/taoliq/benchmark/actions/runs/1738261739/attempts/3
https://github.com/taoliq/benchmark/actions/runs/1738261739/attempts/2
https://github.com/taoliq/benchmark/actions/runs/1718360927

dominichamon (approved):

dominichamon:
LGTM. @mtrofin can you weigh in (and anyone else that wants to)?

src/perf_counters.cc
```diff
@@ -126,7 +127,45 @@ PerfCounters PerfCounters::Create(
   return NoCounters();
 }
 
-PerfCounters::~PerfCounters() = default;
+void PerfCounters::CloseCounters() const {}
 #endif  // defined HAVE_LIBPFM
+
+Mutex PerfCountersMeasurement::mutex_;
+int PerfCountersMeasurement::ref_count_ = 0;
+PerfCounters PerfCountersMeasurement::counters_ = PerfCounters::NoCounters();
```
LebedevRI, line 135: GUARDED_BY(PerfCountersMeasurement::mutex) int PerfCountersMeasurement::ref_count_ = 0;
GUARDED_BY(PerfCountersMeasurement::mutex) PerfCounters PerfCountersMeasurement::counters_ = PerfCounters::NoCounters();

src/perf_counters.cc
```diff
@@ -126,7 +127,45 @@ PerfCounters PerfCounters::Create(
...
+    : start_values_(counter_names.size()), end_values_(counter_names.size()) {
+  {
+    MutexLock l(mutex_);
+    if (ref_count_ == 0) {
+      counters_ = PerfCounters::Create(counter_names);
+    }
+    ++ref_count_;
+  }
+
+  BM_CHECK(!counters_.IsValid() || counters_.names() == counter_names);
```
LebedevRI, line 148: You are accessing `counters_` without the lock here

LebedevRI (changes requested):
Could you please add thread safety annotations first?

taoliq:
Clear nits mentioned by @mtrofin .

mtrofin (approved):

src/perf_counters.h
```diff
@@ -111,55 +115,61 @@ class PerfCounters final {
...
-    return ret;
+
+    return valid_read_;
   }
 
  private:
-  PerfCounters counters_;
+  static Mutex mutex_;
+  static int ref_count_;
+  static PerfCounters counters_;
```
LebedevRI, line 171: Oh, these are declared here, i was afraid that they have completely global scope.
But i think that means the GUARDED_BY should also be here, not at the definition.

LebedevRI (approved):
LG, seems fine now.
Thanks.

dominichamon:
thank you!

taoliq:
Thanks for your patience with the work.

PR #1311, opened 2021-12-28
letting benchmark to download google test

Huzaifa696:


nirvdrum:
FYI, the README shows you how to set this value when invoking `cmake` such that you don't need to make code modifications.

dmah42:
this is deliberately disabled on HEAD as it can be overridden per the docs.

Issue #826, opened 2019-06-04
Warnings C4244, C4305, C4722, ... when compiling with MS Visual Studio 2017 15.9.12


Blonder:
googlebenchmark\src\benchmark_runner.cc(267): warning C4244: 'initializing': conversion from 'double' to 'benchmark::IterationCount', possible loss of data
googlebenchmark\src\benchmark_runner.cc(267): warning C4244: 'initializing': conversion from 'double' to 'const benchmark::IterationCount', possible loss of data
d:\thirdparty\trunk\googlebenchmark\test\statistics_gtest.cc(25): warning C4305: 'argument': truncation from 'double' to 'RawType'
googlebenchmark\test\statistics_gtest.cc(24): warning C4244: 'argument': conversion from 'double' to 'RawType', possible loss of data
googlebenchmark\src\benchmark_register.h(25): warning C4244: '*=': conversion from 'int' to 'T', possible loss of data
googlebenchmark\third_party\googletest\src\googlemock\include\gmock\gmock-matchers.h(156): warning C4244: 'argument': conversion from 'const M' to 'const Element', possible loss of data
googlebenchmark\src\check.h(41): warning C4722: 'benchmark::internal::CheckHandler::~CheckHandler': destructor never returns, potential memory leak
googlebenchmark\src\benchmark_runner.cc(267): warning C4244: 'initializing': conversion from 'double' to 'benchmark::IterationCount', possible loss of data
googlebenchmark\src\benchmark_runner.cc(267): warning C4244: 'initializing': conversion from 'double' to 'const benchmark::IterationCount', possible loss of data
googlebenchmark\src\check.h(41): warning C4722: 'benchmark::internal::CheckHandler::~CheckHandler': destructor never returns, potential memory leak
googlebenchmark\test\statistics_gtest.cc(25): warning C4305: 'argument': truncation from 'double' to 'RawType'
googlebenchmark\test\statistics_gtest.cc(24): warning C4244: 'argument': conversion from 'double' to 'RawType', possible loss of data
googlebenchmark\src\check.h(41): warning C4722: 'benchmark::internal::CheckHandler::~CheckHandler': destructor never returns, potential memory leak
googlebenchmark\src\check.h(41): warning C4722: 'benchmark::internal::CheckHandler::~CheckHandler': destructor never returns, potential memory leak
googlebenchmark\src\check.h(41): warning C4722: 'benchmark::internal::CheckHandler::~CheckHandler': destructor never returns, potential memory leak
[...]

PR #1316, opened 2022-01-07
Destructor not returning is expected in some cases

staffantj:
Address what appears to be the final part of #826 

Per https://github.com/google/benchmark/issues/826#issuecomment-851995549 we are safe to ignore the warning
in this case.

It is unfortunate that the declspec used for the BENCHMARK_NORETURN macro does not cause the compiler to elide
this warning, as one might imagine it should.


dominichamon:
thanks!  can you fix the linter warning from the bots please? :)

dominichamon:
thanks!

PR #1331, opened 2022-01-19
Refine the User Guide CPU Frequency Scaling section

### LINKED ISSUE #1327: [BUG] user guide has a suboptimal script for managing CPU scaling
**Describe the bug**

The script for managing CPU scaling in the user guide doesn't work. If it did I don't think it would not be a recommendable idiom.

**System**

Linux

**To reproduce**

See https://github.com/google/benchmark/commit/9e859f5bf55073a2b4c785b97b9cb78cbdaf127e

The current sample script:

1. Won't disable CPU scaling on all CPUs because it uses a single ? rather than a * in the glob.
2. It'll run the benchmark up to 10 times, since it runs the benchmark within the for loop and that is how many CPUs it might match.
3. ...so it'll run the benchmark N times with only one CPU set to "performance" each time.
4. It may encourage people to run benchmarks as root, since that is usually what it takes to modify CPU scaling.
5. Omits mention of the `cpupower` command. I learned of that command from this user guide many years ago, so I think it is good to mention it.  In fact, this is how I noticed the problem.  I forgot what the command was called so I came back here looking for it.  :-)

**Expected behavior**

The script:

1. Remembers the CPU scaling for the first CPU
2. Sets CPU scaling to performance on all CPUs
3. Runs the benchmark
4. Sets CPU scaling to the original value on all CPUs

However, the cpupower scripts make that hard because they don't provide a nice scriptable way to get the current governor.  Maybe suggest a script like this, named something like `with_performance`:

```
#!/bin/sh
read prior_governor < /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
sudo cpupower frequency-set --governor performance
"$@"
sudo cpupower frequency-set --governor "$prior_governor"
```

And then people can run benchmarks like this in their shell:

```
$ with_performance ./my_benchmark
```

**Additional context**

This changed very recently.  See #1325 and #1048.

matta:
Here is how I'd fix #1327

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1227,16 +1227,10 @@ If you see this error:
 ***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
 ```
 
-you might want to disable the CPU frequency scaling while running the benchmark using a script like:
+you might want to disable the CPU frequency scaling while running the benchmark:
 
 ```bash
-for f in /sys/devices/system/cpu/cpu?/cpufreq/scaling_governor
-  do
-  before=`cat $f`
-  echo "performance" > $f
-  ./mybench  # run your benchmark here
-  echo $before > $f
-done
+sudo cpupower frequency-set --governor performance
+./mybench
+sudo cpupower frequency-set --governor powersave
 ```
-
-note, you will have to run this as root (under sudo).
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1227,10 +1227,36 @@ If you see this error:
 ***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
 ```
 
-you might want to disable the CPU frequency scaling while running the benchmark:
+you might want to disable the CPU frequency scaling while running the
+benchmark.  Exactly how to do this depends on the Linux distribution,
+desktop environment, and installed programs.  Specific details are a moving
+target, so we will not attempt to exhaustively document them here.
+
+One simple option is to use the `cpupower` program to change the
+performance governor to "performance".  This tool is maintained along with
+the Linux kernel and provided by your distribution.
+
+It must be run as root, like this:
 
 ```bash
 sudo cpupower frequency-set --governor performance
-./mybench
-sudo cpupower frequency-set --governor powersave
 ```
+
+After this you can verify that all CPUs are using the performance governor
+by running this command:
+
+```bash
+cpupower frequency-info -o proc
+```
+
+The benchmarks you subsequently run will have less variance.
+
+Note that changing the governor in this way will not persist across
+reboots.  To set the governor back, run the first command again with the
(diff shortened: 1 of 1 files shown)


docs/user_guide.md
```diff
@@ -1227,16 +1227,36 @@ If you see this error:
...
-  before=`cat $f`
-  echo "performance" > $f
-  ./mybench  # run your benchmark here
-  echo $before > $f
-done
+cpupower frequency-info -o proc
 ```
 
-note, you will have to run this as root (under sudo).
+The benchmarks you subsequently run will have less variance.
```
dominichamon, line 1252: we used to also suggest putting the governor back to how it was before, which was the intent of the script. so i think a fuller example (for completeness) that saves the current governor settings and resets them would be ideal.

though i understand that's not straightforward using cpupower's output. the best i have is

```bash
#!/bin/bash                                                                                                                                             
BENCHMARK=$1

CPUS=$(sudo cpupower frequency-info -o proc | tail -n +2 | tr -s ' ' | cut -d\  -f2,14)

declare -A prev_cpus=()

while IFS= read -r c
do
  IFS=' ' read CPU GOVERNOR <<< $c
  prev_cpus[$CPU]=$GOVERNOR
done <<< "$CPUS"

echo "setting all CPUs to 'performance'"
sudo cpupower frequency-set -g performance

echo "running $BENCHMARK"
$BENCHMARK

echo "setting all CPUs back to previous states"
for i in "${!prev_cpus[@]}"
do
  echo "setting cpu $i back to ${prev_cpus[$i]}"
  sudo cpupower --cpu $i frequency-set -g ${prev_cpus[$i]}
done
```

maybe it's worth having this checked in to `tools` and pointing folks at it.

matta: I did write a script like yours, but I found the workflow annoying. What I've always done in the past is set the machine to performance mode for the entire duration of an interactive benchmarking session. What I'd advocate for in the user guide is a description of how to achieve this workflow.

This isn't to say that a script supporting the "performance while running one command" must be excluded (see idea below).

In practice there is a delay setting the governors with `cpupower`, which makes sense because there is a short delay before they take full effect.  I've also got 24 cores, so cpupower spat out 48 lines of output each time.  The delay and the output made it annoying to use for every single benchmark invocation.  Using a script wrapper is also yet another thing to remember, integrate into a workflow (which may be script driven, etc.), forget to run, etc.

One idea: if the script had optional `--enable` and `--disable` options that saved and restored settings, respectively, both workflows could be supported.  Maybe it would save them somewhere under [`$XDG_STATE_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).  It could be used like this:

```
with_performance ./mybenchmark
```

or like this:

```
with_performance --enable
./mybenchmark
make
./mybenchmark
./make
./mybenchmark
etc.
with_performance --disable
```

so both workflows would be supported.

dominichamon: makes total sense. perhaps having this conversation here will help people too :)

Issue #661, opened 2018-08-24
Build Error w/ msvc-141 while targeting Universal Windows Platform


natxete88:
### Summary
While attempting to build google/benchmark with cmake for a Windows UWP (Universal Windows Platform) target the following errors can be observed.

![error](https://user-images.githubusercontent.com/42657589/44603055-1ba56000-a7a7-11e8-942d-d8af25c96b90.PNG)

These symbols are disabled/undefined by `VersionHelper.h:18 -- #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)`
Yet required in the `sysinfo.cc:526 -- #elif defined BENCHMARK_OS_WINDOWS` block.

### Steps to reproduce

The above can be reproduced (so long as you have VS2017 or can compile with the MSVC v141 toolchain) with the following steps:

```bash
$ git clone https://github.com/google/benchmark.git
$ cd benchmark
$ mkdir -p build && cd build
$ cmake -G "Visual Studio 15" -DBENCHMARK_ENABLE_GTEST_TESTS=OFF -DCMAKE_BUILD_TYPE=RELEASE  -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0 ..

```
*Note: the CMAKE_SYSTEM_NAME/VERSION settings drive the cmake config step for UWP apps*

Then open the .sln in build/ and run a build.

### Fix

* The issue:
![bug](https://user-images.githubusercontent.com/42657589/44603059-24963180-a7a7-11e8-852b-4d4c2df84a75.PNG)

* The fix:
![fix](https://user-images.githubusercontent.com/42657589/44603078-3546a780-a7a7-11e8-8205-976d7109197a.PNG)

### Alternative

The frequency estimation in the block in question is sampling/inferring the MHz of the system (presumably for performance measures). As this particular block is platform(Windows Specific) wouldn't usage of the [`QueryPerformanceFreq()` API (windows.h)](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644905(v=vs.85).aspx) be more appropriate/accurate ?

Happy to try and submit a PR for this if our thinking on this is correct and acceptable.

dominichamon:
I don't know much about windows development, but from reading the API it seems reasonable that this would be a better source of data than the registry.

[...]

PR #1333, opened 2022-01-25
Use Win32 API only for Win32 apps

batortaller:
This PR introduces two new platform defines `BENCHMARK_OS_WINDOWS_WIN32` and `BENCHMARK_OS_WINDOWS_RT`. These macros determine the Windows API types while `BENCHMARK_OS_WINDOWS` remains a define for the Window platform.

Using these new defines the code that uses the Win32 API in `sysinfo.cc` is guarded by `BENCHMARK_OS_WINDOWS_WIN32` so that the code can compile for non Win32 apps (for example UWP apps).

With this modification the code will compile for UWP apps but the CPU frequency information will be less reliable than with Win32. Unfortunately I could not find a method to query the CPU frequency on Windows that is CPU architecture independent. IMHO the crude estimate for the CPU clock is still better than not being able to use the library since this information is not essential.

Related issue: https://github.com/google/benchmark/issues/661

Note: the `QueryPerformanceFreq()` function mentioned in the issue is not reliable and gave incorrect information in my case.


LebedevRI:
Please fix the CLA issue.

batortaller:
> Please fix the CLA issue.

@LebedevRI the CLA issue is fixed

LebedevRI:
Rather not familiar with the windows side of things, but seems reasonable, and CI is happy.

Issue #1328, opened 2022-01-19
[FR] Setting the time unit globally


zchrissirhcz:
**Is your feature request related to a problem? Please describe.**
The benchmark output time unit is `ns` on default. I would like do make all my benchmarks time cost reported in `ms` time unit.
I can modify each benchmark:
```
BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
```
But modifying to each benchmark makes me frustrated, so much duplicated work.

**Describe the solution you'd like**
`./my_benchmark --benchmark_time_unit=ms`

**Describe alternatives you've considered**
Support environment variable `BENCHMARK_TIME_UNIT`.
Also support globally setup this time unit in C++ code.

**Additional context**

dominichamon:
Fixed by #1337

PR #1337, opened 2022-02-02
Allow setting the default time unit globally

batortaller:
This PR introduces the concept of default time unit to the benchmark framework. If the `--benchmark_time_unit` command line argument is specified then `SetDefaultTimeUnit` is called which stores the value in a global variable. The initial value of this global is `kNanosecond` so the library works same if the flag is not specified. The `Benchmark` class is extended with a `use_default_time_unit_` bool member. The value of this bool is set to false if the `Unit` method is called so the default time unit does not affect tests where the unit is explicitly specified.

The default time unit can also be set from C++ by calling the `SetDefaultTimeUnit` function at the beginning of the program. 

Related issue: https://github.com/google/benchmark/issues/1328


src/benchmark_register.cc
```diff
@@ -203,6 +203,7 @@ Benchmark::Benchmark(const char* name)
     : name_(name),
       aggregation_report_mode_(ARM_Unspecified),
       time_unit_(kNanosecond),
```
LebedevRI, line 205: Can't you just `time_unit_(GetDefaultTimeUnit())`?

batortaller: Yes, that is perfectly reasonable.
1667d8e353d7c0abb3298883c53d9ec6470bf9cf

src/benchmark_register.cc
```diff
@@ -202,7 +202,8 @@ bool FindBenchmarksInternal(const std::string& re,
 Benchmark::Benchmark(const char* name)
     : name_(name),
       aggregation_report_mode_(ARM_Unspecified),
-      time_unit_(kNanosecond),
+      time_unit_(GetDefaultTimeUnit()),
+      use_default_time_unit_(true),
```
dominichamon, line 206: i don't understand why you need this boolean.

if you initialise to the default, and then the user overrides it by calling `Unit`, then you can just return `time_unit_` knowing it's either the default or the override.  what is this boolean protecting against?

batortaller: If you use the predefined macros `BENCHMARK`, `BENCHMARK_F` then the `Benchmark` objects are created by global initialization so argument parsing and calling `SetDefaultTimeUnit` (or even `main`) will be done later.

src/benchmark.cc
```diff
@@ -529,6 +542,23 @@ void AddCustomContext(const std::string& key, const std::string& value) {
   }
 }
 
+namespace {
+
+inline TimeUnit GetTimeUnitFromString(const std::string& unit_string) {
```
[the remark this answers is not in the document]
batortaller: 9ebe93e0798ac5a1aee66ebd6a0953dea817485f

src/benchmark.cc
```diff
@@ -529,6 +542,23 @@ void AddCustomContext(const std::string& key, const std::string& value) {
...
+  if (unit_string == "s") {
+    return kSecond;
+  } else if (unit_string == "ms") {
+    return kMillisecond;
+  } else if (unit_string == "us") {
+    return kMicrosecond;
+  } else if (unit_string == "ns") {
+    return kNanosecond;
+  }
+  BENCHMARK_UNREACHABLE();
```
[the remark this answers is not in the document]
dominichamon: we have a `BM_CHECK` macro (https://github.com/google/benchmark/blob/main/src/check.h) that allows for a check to be coupled with some logging if it fails.

src/benchmark.cc
```diff
@@ -513,6 +517,15 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* display_reporter,
...
 }
 
+namespace {
+// stores the time unit benchmarks use by default
+TimeUnit default_time_unit = kNanosecond;
+}  // namespace
+
+TimeUnit GetDefaultTimeUnit() { return default_time_unit; }
+
+void SetDefaultTimeUnit(TimeUnit unit) { default_time_unit = unit; }
```
dominichamon, line 527: do you need this setter method?  seems like just inlining the assignment below in `SetDefaultTimeUnitFromFlag` is just as readable

src/benchmark_register.cc
dominichamon: ah very well spotted, thanks :)

dominichamon (approved):

dominichamon:
LGTM.  and thanks for adding the tests :)

i'll wait for @LebedevRI to review.

src/benchmark.cc
batortaller: This function is exposed to the header so you can set the default from your app if you'd like that. I use this function below instead of inlining to have a single point where this global is set.

include/benchmark/benchmark.h
```diff
@@ -1122,7 +1129,10 @@ class BENCHMARK_EXPORT Benchmark {
   AggregationReportMode aggregation_report_mode_;
   std::vector<std::string> arg_names_;       // Args for all benchmark runs
   std::vector<std::vector<int64_t> > args_;  // Args for all benchmark runs
+
   TimeUnit time_unit_;
+  bool use_default_time_unit_;
```
LebedevRI, line 1134: I thought we agreed that this was not needed now that `time_unit_` is init'd to the `GetDefaultTimeUnit()`?

test/time_unit_gtest.cc
```diff
@@ -0,0 +1,37 @@
...
+TEST(DefaultTimeUnitTest, TimeUnitIsNotSet) {
+  DummyBenchmark benchmark;
+  EXPECT_EQ(benchmark.GetTimeUnit(), kNanosecond);
+}
+
+TEST(DefaultTimeUnitTest, DefaultIsSet) {
+  DummyBenchmark benchmark;
+  EXPECT_EQ(benchmark.GetTimeUnit(), kNanosecond);
+  SetDefaultTimeUnit(kMillisecond);
+  EXPECT_EQ(benchmark.GetTimeUnit(), kMillisecond);
```
LebedevRI, line 24: The pr's description says:
```
The default time unit can also be set from C++ by calling the SetDefaultTimeUnit function
at the beginning of the program.
```
but the current design allows for the default to be changed retroactively?

include/benchmark/benchmark.h
batortaller: If you create global test objects (like how it is shown in the examples with the helper macros) then `time_unit_` is initialized before the arguments are parsed.

test/time_unit_gtest.cc
batortaller: The default has to be set before `::benchmark::RunSpecifiedBenchmarks()`. Calling `SetDefaultTimeUnit` does not change the default after that.

LebedevRI:
No opinion here.

batortaller:
@dominichamon do you need anything from my end?

include/benchmark/benchmark.h
```diff
@@ -337,6 +337,15 @@ BENCHMARK_EXPORT size_t
 RunSpecifiedBenchmarks(BenchmarkReporter* display_reporter,
                        BenchmarkReporter* file_reporter, std::string spec);
 
+// TimeUnit is passed to a benchmark in order to specify the order of magnitude
+// for the measured time.
+enum TimeUnit { kNanosecond, kMicrosecond, kMillisecond, kSecond };
+
+BENCHMARK_EXPORT TimeUnit GetDefaultTimeUnit();
+
+// Sets the default time unit the benchmarks use
```
dominichamon, line 346: add a comment documenting that it only works if called prior to the main benchmark loop please.

src/benchmark.cc
```diff
@@ -520,6 +524,15 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* display_reporter,
...
```
dominichamon, line 534: it's a shame we  can't catch this being called at the wrong time.  in `State` we can check `!started_ && !finished_` but i don't think we expose that anywhere, and you don't have `State` here anyway.

i don't think there's anything to be done, but i wanted to flag it in case you had an idea.

batortaller: Can't think of a good solution either. Maybe putting it in the `internal` namespace thus not recommending the usage of this (or just removing it from the header). But I'd rather leave the option to set this programatically.

dominichamon (approved):

PR #1339, opened 2022-02-03
Add pandas and numpy to tools/requirements.txt

batortaller:
Add the appropriate requirements that `tools/gbench/report.py` uses to `tools/requirements.txt`. Right now `tools/compare.py` can fail if someone relies on `tools/requirements.txt`.

### DIFF
--- a/tools/requirements.txt
+++ b/tools/requirements.txt
@@ -1,1 +1,3 @@
 scipy>=1.5.0
+pandas>=1.1.5
+numpy>=1.21


LebedevRI:
See #1295

LebedevRI:
#1295 is stuck, i don't have a preference which one fixes the problem.
I was mainly highlighting https://github.com/google/benchmark/pull/1295#issuecomment-993621380
> As per previous comment, let's remove root requirements.txt, and change WORKSPACE file to point to this file instead.

PR #1347, opened 2022-02-14
Check for macro existence before using

oontvoo:
@dominichamon


dominichamon (approved):
this is safe.. i'm not sure it does what is expected in every case (is WINAPI_FAMILY_PARTITION defined whenever WIN32 should be?)

but it's clearly safer than what we have.

oontvoo:
> i'm not sure it does what is expected in every case (is WINAPI_FAMILY_PARTITION defined whenever WIN32 should be?)

From my understanding, _WIN32 is defined by the compiler, whereas WINAPI_FAMILY_PARTITION is defined in `winapifamily.h`.
So it's not guaranteed that we'll always see WINAPI_FAMILY_PARTITION. (unless we want to conditional include that header, too) WDYT?

PR #1356, opened 2022-02-24
@platforms is magical; remove it from WORKSPACE.

junyer:


junyer:
Sorry, @dominichamon, I only just twigged to this after I realised that it isn't in RE2's `WORKSPACE` file and yet building benchmarks still succeeded.

PR #1355, opened 2022-02-24
Make generate_export_header.bzl work for Windows.

junyer:
While I'm here, bring the generated code slightly closer to what CMake
would generate nowadays.

Fixes #1351.

PR #1359, opened 2022-03-03
Add PyPI upload job to wheel building workflow

nicholasjng:
This commit adds a job running after the wheel building job responsible for uploading the built wheels to PyPI.
The job only runs on successful completion of all build jobs, and uploads to PyPI using a secret added to the Google Benchmark repo (TBD).
Also, the `setup-python` action has been bumped to the latest version v3.

TODO: 
* Verify that the credential setup is correct (by trying a wheel building CI job)
* Make sure the token is added to this repo's Github secrets under the name `pypi_password` (or I can change the name to whatever it is now)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.9
 
@@ -37,7 +37,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Set up Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.9
 
@@ -80,7 +80,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Set up Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.9
 
@@ -111,7 +111,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Set up Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.9
 
@@ -131,3 +131,19 @@ jobs:
         with:
           name: dist
           path: wheelhouse/*.whl
+
+  pypi_upload:
+    name: Publish google-benchmark wheels to PyPI
+    needs: [build_sdist, build_linux, build_macos, build_windows]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+    steps:
+    - uses: actions/download-artifact@v2
+      with:
+        name: artifact
+        path: dist
+
+    - uses: pypa/gh-action-pypi-publish@v1.5.0
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_PASSWORD }}


dominichamon:
this looks fine to me, but i don't have access to the google-benchmark repository on pypl to get the secrets.

nicholasjng:
Okay, do you happen to know who does? I could ping them about the plan here and issuing an upload token.

dominichamon:
https://pypi.org/user/benchmark-py/ is the maintainer. i have no idea who that is.

nicholasjng:
Ah, I thought that was you. The account there points to a google email address (benchmark-py[AT]google[.]com), maybe they will answer. (Though I do not assume they will give me the token, but maybe they have some info)

dominichamon:
i've also contacted them :)

dominichamon:
i now have access to the internal group, and am working on getting access to the password for PyPI.

dominichamon:
i have now updated `PYPI_PASSWORD` to the API project.

nicholasjng:
Ok, so the repo has the API token secret as `PYPI_PASSWORD`? It's all case sensitive, so I need to upgrade the secret name from the lower case, then.

dominichamon:
yes.. i added it as lower case but it ended up showing up as upper case.

nicholasjng:
I changed the spelling to uppercase and assigned a proper job name, PTAL.

dominichamon:
looks reasonable to me. can we test it before it's merged?

dominichamon (approved):

nicholasjng:
You could dispatch the workflow manually from the GitHub Actions page.

dominichamon:
> You could dispatch the workflow manually from the GitHub Actions page.

i can't figure out how to do it premerge. "To trigger the workflow_dispatch event, your workflow must be in the default branch. " https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

dominichamon:
looks like the windows build failed https://github.com/google/benchmark/runs/5499504439?check_suite_focus=true

nicholasjng:
That's a large number of linker errors. Some recent change broke the linkage inside the Windows runner, I guess?

nicholasjng:
I happen to have a Windows 10 x86_64 machine on hand, so I can try to repro this locally (and hopefully come up with a fix). I expect this will involve a Windows-specific change to the `setup.py`, since apparently other actions involving Windows builds succeed (?). I'm on it.

dominichamon:
maybe related to https://github.com/google/benchmark/pull/1355 ?

nicholasjng:
> maybe related to #1355 ?

Definitely looks like it. It seems to find pybind11 though, so that's a plus. 

Unfortunately `cibuildwheel` does not show build logs on successful builds, so I cannot easily diff between this attempt and the last successful build from the `v1.6.1` release. I will keep you in the loop about my progress debugging this.

nicholasjng:
https://github.com/google/benchmark/pull/1321#issuecomment-1038935168

This comment of yours mentions the library will now be built as shared by default - does this have an effect on the bindings? (Then the fix might be as easy as toggling that option to the previous value)

nicholasjng:
I set up Bazel and a proper toolchain on my Windows, and I am able to reproduce this locally, so that's a first step.

Perhaps we need to fiddle with the `BazelExtension`'s build command. Could you point me to the shared build flags on Windows? Thanks!

dominichamon:
https://github.com/google/benchmark/runs/5552873494?check_suite_focus=true

looks like there's an issue in the definition of the upload step @nicholasjng

nicholasjng:
> https://github.com/google/benchmark/runs/5552873494?check_suite_focus=true
> 
> looks like there's an issue in the definition of the upload step @nicholasjng

Yup, saw it too. I know what this is related to, this is on me due to some misconfigured names. Will send a fix shortly

dominichamon:
https://pypi.org/project/google-benchmark/ it worked!

now i think i need to add running this to the release documentation.

PR #1363, opened 2022-03-09
Appended additional BSD 3-Clause to LICENSE

oontvoo:
Needed for export_header module

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -200,3 +200,33 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+   
+   
+                 BSD 3-Clause License
+
+Copyright (c) [year], [fullname]
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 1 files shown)


dominichamon:
thanks

LebedevRI:
This change seems to be missing about 500 words of motivation.

LebedevRI:
(I would recommend reverting this immediately)

oontvoo:
This was needed because PR/1321 included code that had this license (the generate_export_header).
This change should have been done there but I guess we didn't have a git license compliant-lint check.

LebedevRI:
This change should be a lot more explicit about that, it should call out that said new license is only for that single specific file,
otherwise this can be read as if the whole project is now double-licensed. This is a problem.

dominichamon:
why would it be a problem that the project as a whole is double-licensed under these licenses, which are completely compatible?

LebedevRI:
> why would it be a problem that the project as a whole is double-licensed under these licenses, which are completely compatible?

Please consult a lawyer.
Until then, please revert this.

LebedevRI:
> This change should be a lot more explicit about that, it should call out that said new license is only for that single specific file, otherwise this can be read as if the whole project is now double-licensed. This is a problem.

oontvoo:
@LebedevRI  I've sent PR/1366 . PTAL. Thanks!

oontvoo:
from my interpretation, LebedevRI's concern is that it's not OK to retrospectively re-license code (which is what this PR sort of accidentally did). I think that's a reasonable point.

PR #1366, opened 2022-03-10
Update LICENSE file to clearly state which file needs BSD 3

oontvoo:
Follow up to PR/1363

@LebedevRI  @dominichamon

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -202,6 +202,8 @@
    limitations under the License.
    
    
+Only benchmark/config/generate_export_header.bzl depends on the following licence:
+
                  BSD 3-Clause License
 
 Copyright (c) [year], [fullname]


jwnimmer-tri:
Regarding #1321 up through this PR.

The bzl file in question here came from Drake, a project that I help [maintain](https://github.com/RobotLocomotion/drake/pulse/monthly) on behalf of Toyota Research.

If you like, I have authority to CLA the bzl contents over to the standard googlebenchmark license so that you don't need to carry around this license addendum forevermore.  Let me know if you'd like me to help with that?

For a mechanism, I suppose I could open a PR that removes the BSD-3-Clause contents from LICENSE as well as the bzl file itself, and my signing of that CLA on _that_ PR would would suffice?

dominichamon:
i appreciate the offer, but it would need to be changed in the original source for us to inherit the new license.  this doesn't cause any licensing issues as far as i'm aware as BSD-3 and Apache-2 are compatible, so i don't believe any further action is required.

PR #1369, opened 2022-03-14
Add BENCHMARK_STATIC_DEFINE to the Python bindings' `cc_binary` local…

### LINKED ISSUE #1367: [BUG] Linker errors during wheel builds on Windows
**Describe the bug**
Building a wheel for the Python bindings on Windows with an MSVC toolchain fails with a large number of linker errors.

**System**
  - OS: Win10, x64
  - Compiler and version: MSVC bundled with Visual Studio 22

**To reproduce**
1. sync to main@HEAD (`git checkout main`)
2. Using a system Python (I tried with Python 3.9, assumed available on the command line as "python"):
3. `python -m venv venv --system-site-packages --upgrade-deps`
4. .\venv\Scripts\Activate.ps1
5. `python -m pip install wheel`
6. `python -m pip wheel .`

**Expected behavior**
The wheel is built and deposited in the `build` folder.

**Additional context**
This issue is Windows-specific, Linux and macOS wheels build properly.

This was discovered yesterday by @dominichamon and I when trying to build wheels for PyPI releases, check the pull request discussion [here](https://github.com/google/benchmark/pull/1359). I have since been able to repro this on my local machine as well.

This regression likely comes from the recent work on creating export headers in #1321 and #1355. Pinging the authors of the PRs @sergiud and @junyer for visibility / ideas. (Since the c++ library can be built under Windows, I assume that something in the Bazel config of the Python bindings broke.)

Logs / traceback of the GitHub job: https://github.com/google/benchmark/runs/5499504439?check_suite_focus=true

nicholasjng:
… defines.

Fixes #1367.

This commit fixes the previous breakage in Python wheel builds for Windows by adding a `local_defines` field to the `cc_binary` generated in the process of the Python bindings builds. This define is being
picked up by the auto-generated export header `benchmark_export.h`, unsetting the benchmark export macro.
Furthermore, the `linkshared` and `linkstatic` attributes are passed booleans now instead of ints, making the command more directly interpretable to the human reader.

The fix was suggested by @junyer in the corresponding GitHub issue thread https://github.com/google/benchmark/issues/1367 - thank you for the suggestion!

Wheel CI is green with this change @dominichamon - see https://github.com/nicholasjng/benchmark/actions/runs/1980232291.

### DIFF
--- a/bindings/python/build_defs.bzl
+++ b/bindings/python/build_defs.bzl
@@ -8,12 +8,13 @@ def py_extension(name, srcs, hdrs = [], copts = [], features = [], deps = []):
         shared_lib_name = name + shared_lib_suffix
         native.cc_binary(
             name = shared_lib_name,
-            linkshared = 1,
-            linkstatic = 1,
+            linkshared = True,
+            linkstatic = True,
             srcs = srcs + hdrs,
             copts = copts,
             features = features,
             deps = deps,
+            local_defines = ["BENCHMARK_STATIC_DEFINE"],
         )
 
     return native.py_library(


dominichamon:
fantastic. thank you!

PR #1371, opened 2022-03-15
Change artifact download name to dist to match upload name

nicholasjng:
The uploaded artifact name is `dist` in every wheel build job, as well as on the completed jobs page.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -139,7 +139,7 @@ jobs:
     steps:
     - uses: actions/download-artifact@v2
       with:
-        name: artifact
+        name: dist
         path: dist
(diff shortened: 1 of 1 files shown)


nicholasjng:
Fyi: The previously configured name `artifact`, which resulted in errors, is the default name given by GitHub Actions if no `name` is given to the uploaded artifact (in our case, the created wheels) in the upload step.

So an equivalent solution would be to delete the `name: dist` lines for every one of the four preceding upload steps, and then download the (default-named) `artifact`. I like the naming here as it is (`artifact` is not really an expressive name imho), but YMMV.

dominichamon:
i agree

PR #1373, opened 2022-03-16
Build `//:benchmark` as a static library only.

junyer:
If someone or something ever needs the dynamic library as a Bazel build
artifact, we can figure that out for them then, but right now, there is
no strong reason to be wrangling various `export.h`-controlling macros.

Fixes #1372.

PR #1374, opened 2022-03-17
Fix Bazel build breakage caused by commit 6a894bd.

junyer:


junyer:
Ping, @dominichamon. :)

dominichamon (approved):

dominichamon:
still confused about why the CI didn't catch this until after i merged. every action is running on push and pull_request.

junyer:
https://github.com/google/benchmark/pull/1373/checks shows that only the CLA check ran that time. O_o

dominichamon:
that's on me then. i should have seen it.

thanks for the quick fix!

PR #1380, opened 2022-04-02
Dynamically detect PMU capabilities through libpfm

damageboy:
- Instead of allowing for up to 3 counters, libpfm's internal
  capabilities of reporting PMU info are used to manage a per-PMU
  "registry" adn dynamically allocate slots according to the specific
  counter requested
- In this PR/commit, it is still impossible to get more detailed (x86)
  information about each specific counter properties in terms of being
  a fixed/non-fixed counter, due to what seems to be a lack of API
  surface on libpfm itself:
  https://sourceforge.net/p/perfmon2/mailman/message/37631173/
- The maximal number of counters was bumped from 3 to 63, which together
  with the current padding "scheme" means we pre-allocate
  64 in-place counter slots (64-bits each) per measurement instance


src/perf_counters.cc
```diff
@@ -26,9 +26,121 @@
...
+    return true;
+  }
+
+ private:
+  pfm_pmu_t pmu_id_;
+  const char* name_;
+  const char* desc_;
+  std::unordered_map<int, uint64_t> counter_ids_;
+  std::unordered_map<int, uint64_t> fixed_counter_ids_;
+  uint64_t available_counters_;
```
dominichamon, line 96: https://google.github.io/styleguide/cppguide.html#Integer_Types

for pretty good reason, we try to stick to `int` rather than `uint64_t`.  obviously if 64-bit is required that is fine, but please don't use unsigned for these.  instead used signed and check for over/underflow instead.

src/perf_counters.cc
```diff
@@ -26,9 +26,121 @@
...
+class PMURegistry {
+ public:
+  ~PMURegistry() = default;
+  PMURegistry(PMURegistry&&) = default;
+  PMURegistry(const PMURegistry&) = delete;
+  PMURegistry& operator=(PMURegistry&&) noexcept;
+  PMURegistry& operator=(const PMURegistry&) = delete;
+  PMURegistry() {}
+
+  bool EnlistCounter(const std::basic_string<char>& name, struct perf_event_attr &attr_base) {
```
dominichamon, line 109: why not just `const std::string& name`?

src/perf_counters.cc
```diff
@@ -26,9 +26,121 @@
...
+    attr_base.size = sizeof(attr_base);
+    pfm_perf_encode_arg_t encoding{};
+    encoding.attr = &attr_base;
+
+    const auto pfm_get = pfm_get_os_event_encoding(
+        name.c_str(), PFM_PLM3, PFM_OS_PERF_EVENT, &encoding);
+    if (pfm_get != PFM_SUCCESS) {
+      GetErrorLogInstance() << "Unknown counter name: " << name << "\n";
+      return false;
```
dominichamon, line 118: just checking design decision: given the perf counters are requested by users would it not be better to hard-fail here with an assert?

src/perf_counters.h
```diff
@@ -49,13 +49,15 @@ namespace internal {
 // operator[]) of this object.
 class PerfCounterValues {
  public:
-  explicit PerfCounterValues(size_t nr_counters) : nr_counters_(nr_counters) {
-    BM_CHECK_LE(nr_counters_, kMaxCounters);
+  explicit PerfCounterValues(size_t nr_counters)
```
dominichamon, line 52: `int` please instead of `size_t`.

src/perf_counters.cc
```diff
@@ -26,9 +26,121 @@
...
+               << ") in pmu " << name_ << " (" << counter_ids_.size() << "/" << available_counters_ << "\n";
+
+    return true;
+  }
+
+ private:
+  pfm_pmu_t pmu_id_;
+  const char* name_;
+  const char* desc_;
+  std::unordered_map<int, uint64_t> counter_ids_;
```
dominichamon, line 94: i think you're missing a header for this?

dominichamon (changes requested):
also check the format issues please. use clang-format, ideally.

src/perf_counters.cc
damageboy: Well, in this case, I'm just trying to stick to whatever libpfm does:

```c
typedef struct {
	...
	uint64_t		code;	/* event raw code (not encoding) */
    ...
} pfm_event_info_t;
```

Do you prefer something else given their choice of types?

src/perf_counters.cc
damageboy: Agree 100% percent, this is why I'm always returning `false` from the internal api and later failing with:

```cpp
    if (!ok) {
      GetErrorLogInstance() << "Failed to register counter: " << name << "\n";
      return NoCounters();
    }
```

src/perf_counters.h
damageboy: Just to be clear, the original code in the `main` branch opted for this:

https://github.com/google/benchmark/blob/60b16f11a30146ac825b7d99be0b9887c24b254a/src/perf_counters.h#L52

Do you want me to change here as a cleanup side-effect in this PR?

src/perf_counters.cc
damageboy: Yeah, right, this only fails on mac/windows for some reason, but 100% correct.

damageboy:
> also check the format issues please. use clang-format, ideally.

Yeah, sorry about that, I was editing the file through a different project with its own .clang-format and it got mixed up.

src/perf_counters.cc
dominichamon: i think the map should stay as `uint64_t` as we have no way to know what they're using for ids (they could be 64-bit UUID or something).  the _counters_ should change to int though.

src/perf_counters.cc
dominichamon: ok. i'd also be ok with us just asserting/CHECKing here, given it's all one library. up to you.

src/perf_counters.h
dominichamon: if you don't mind, that would be great.  if it causes problems we can do it (i can) as a followup.

src/perf_counters.cc
damageboy: So, this introduces a warning about comparting `.size()` which returns a `size_t` to `int` and the possible loss of integer precision. Do you want me to cast the `available_counters_` to `size_t` for the comparison, or just leave as is?

dominichamon: ideally, we'd check the returned `.size()` is within bounds for int(64_t) and then cast to int(64_t).

i'm much more concerned about the unsigned bit than the 64-bit bit (though if we have more counters than fit in 32-bit... :O

damageboy: OK, This feels over the top to me considering the amount of `.size()` and the use of `size_t` everywhere, due to STL, but there you go, updated the code.

dominichamon:
can you check those bot failures please?

damageboy:
> can you check those bot failures please?

Sure on it!

damageboy:
> can you check those bot failures please?

Well, apart from the formatting error, the build errors are from that (apparently) not so innocent that was part of the pre-existing code that you requested: https://github.com/google/benchmark/pull/1380#discussion_r841551848

Reverting that unrelated change seems to have solved the issue locally for me when building with:

```shell
cmake ../ -G Ninja -DCMAKE_BUILD_TYPE=Release -DBENCHMARK_ENABLE_WERROR=ON -DBENCHMARK_ENABLE_TESTING=OFF
```

The formatting issues have been fixed too.

dominichamon:
> > can you check those bot failures please?
> 
> Well, apart from the formatting error, the build errors are from that (apparently) not so innocent that was part of the pre-existing code that you requested: [#1380 (comment)](https://github.com/google/benchmark/pull/1380#discussion_r841551848)
> 
> Reverting that unrelated change seems to have solved the issue locally for me when building with:
> 
> ```shell
> cmake ../ -G Ninja -DCMAKE_BUILD_TYPE=Release -DBENCHMARK_ENABLE_WERROR=ON -DBENCHMARK_ENABLE_TESTING=OFF
> ```
> 
> The formatting issues have been fixed too.

ah crap, sorry for the extra churn then.

src/perf_counters.cc
```diff
@@ -19,16 +19,133 @@
...
+ public:
+  ~SinglePMURegistry() = default;
+  SinglePMURegistry(SinglePMURegistry&&) = default;
+  SinglePMURegistry(const SinglePMURegistry&) = delete;
+  SinglePMURegistry& operator=(SinglePMURegistry&&) noexcept;
+  SinglePMURegistry& operator=(const SinglePMURegistry&) = delete;
+
+  SinglePMURegistry(pfm_pmu_t pmu_id)
+      : pmu_id_(pmu_id), available_counters_(0), available_fixed_counters_(0) {
+    {
```
dominichamon, line 42: why the extra braces here?

src/perf_counters.cc
```diff
@@ -19,16 +19,133 @@
...
+class PMURegistry {
+ public:
+  ~PMURegistry() = default;
+  PMURegistry(PMURegistry&&) = default;
+  PMURegistry(const PMURegistry&) = delete;
+  PMURegistry& operator=(PMURegistry&&) noexcept;
+  PMURegistry& operator=(const PMURegistry&) = delete;
+  PMURegistry() {}
+
+  bool EnlistCounter(const std::string& name,
```
dominichamon, line 113: nit: can you be consistent about `type&` vs `type &`? clang-format will generally pick one so i'm not sure why it didn't here.

damageboy: Done. Again, mishap with using two clang-formats, will be more careful in the future.

src/perf_counters.cc
damageboy: Removed, mistake...

dominichamon:
> @dominichamon I'm having a hard time reproing the build issues on a pristine ubuntu 16.04 with gcc-5 using a local docker image. Do you have any ideas on how I can repro the build breakage? Referring to this:
> 
> https://github.com/google/benchmark/runs/5850874474?check_suite_focus=true:
> 
> ```
> /__w/benchmark/benchmark/src/perf_counters.cc:145:52:   required from here
> /usr/include/c++/5/bits/hashtable_policy.h:85:34: error: no match for call to '(const std::hash<pfm_pmu_t>) (const pfm_pmu_t&)'
>   noexcept(declval<const _Hash&>()(declval<const _Key&>()))>
> ```

i haven't seen this error before. it smells a bit like an ABI mismatch, which suggests the bots are misconfigured, but i don't think this has been true until now.

it suggests we don't have a `std::hash` for `pfm_pmu_t`...

damageboy:
> en this error before. it smells a bit like an ABI mismatch, which suggests the bots are misconfigured, but i don't think this has been true until now.

Yes, I realize this, to make things worse, this doesn't happen on a brand new printine ubuntu 16.04 docker image I ran locally.
It just compiles cleanly... :(

src/perf_counters.cc
```diff
@@ -19,16 +19,132 @@
...
+
+    BM_VLOG(2) << "Registered counter: " << event_id << " (" << info.name
+               << " - " << info.desc << ") in pmu " << name_ << " ("
+               << counter_ids_.size() << "/" << available_counters_ << "\n";
+
+    return true;
+  }
+
+ private:
+  pfm_pmu_t pmu_id_;
```
mtrofin, line 94: Nit: can you set `pmu_id_` to an invalid value here? Easier to avoid uninitialized values (even if you set it later in the ctor, setting it at decl is easier to maintain)

or could it be const?

src/perf_counters.cc
```diff
@@ -19,16 +19,132 @@
...
 #endif
 
 namespace benchmark {
 namespace internal {
 
-constexpr size_t PerfCounterValues::kMaxCounters;
-
 #if defined HAVE_LIBPFM
+
+class SinglePMURegistry {
```
mtrofin, line 33: can you add some comments - purpose of the class, etc?

src/perf_counters.cc
```diff
@@ -19,16 +19,132 @@
...
+    available_fixed_counters_ = pmu_info.num_fixed_cntrs;
+
+    BM_VLOG(1) << "PMU: " << pmu_id << " " << name_ << " " << desc_ << "\n";
+    BM_VLOG(1) << "     counters: " << available_counters_
+               << " fixed: " << available_fixed_counters_ << "\n";
+  }
+
+  const char* name() const { return name_; }
+
+  bool AddCounter(int event_id) {
```
mtrofin, line 63: please add tests.

dominichamon:
ping .. are you planning on coming back to finish this off? it would be really good to be able to merge it.

dmah42:
one more check to see if there's any plans to come back to this...

PR #1383, opened 2022-04-08
Shut down Bazel gracefully and revert wheel build strategy to job matrix

nicholasjng:
This commit adds a `bazel shutdown` command to the setuptools BazelExtension. This has the effect that wheel builds shut down the Bazel server and terminate gracefully after the build, something that was previously an issue on Windows builds.

Since the windows-specific `--no-clean` flag option to `pip wheel` becomes unnecessary due to this change, this change has the side-effect that GitHub Actions wheel builds via `cibuildwheel` can now be written as a compact job matrix again, which leads to a lot of deduplicated code in the corresponding workflow file.

Lastly, some GitHub-provided actions (checkout, setup-python, upload/download-artifact) were bumped to the latest v3 version.

As always, a successful workflow run with this change can be found here: https://github.com/nicholasjng/benchmark/actions/runs/2115208956

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install Python 3.9
         uses: actions/setup-python@v3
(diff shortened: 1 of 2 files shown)


nicholasjng:
I got this idea while discussing https://github.com/google/jax/issues/10132. They explicitly include a `bazel shutdown` command in their Python build script (they build a wheel for a different package in the repository, which is a slightly different workflow than building with `setup.py` like we are here).
I gave it a shot for solving the `pip wheel` errors seen previously on Windows due to conflicts with Bazel, and it worked! Now we can canonicalize the build job into a job matrix again, which cuts a lot of duplicate lines from the workflow file.

PR #1392, opened 2022-05-10
Add installation and build instructions for Python bindings

nicholasjng:
This commit adds a small section on how to install and build Python
bindings wheels to the readme.

Notes were added that clearly state availability of Python wheels based
on Python version and OS/architecture combinations.

For the guide to build a wheel from source, the best practice of
creating a virtual environment and activating it before build was
detailed. Also, a note on the required installation of Bazel was added,
with a link to the official docs on installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ IRC channels:
 
 [Assembly Testing Documentation](docs/AssemblyTests.md)
 
+[Building and installing Python bindings](docs/python_bindings.md)
+
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -47,6 +47,41 @@ IRC channels:
 
 [Assembly Testing Documentation](docs/AssemblyTests.md)
 
+## Python bindings
```
dominichamon, line 50: can you add this as a doc in docs/ and link to it instead? that way it gets installed with the package.

dominichamon:
thanks!

Issue #1330, opened 2022-01-19
[FR] Expose the verbosity command line option value


yurikhan:
**Is your feature request related to a problem? Please describe.**
The benchmark library recoginzes a command line option `--v` to set verbosity.

A benchmark program might want to set its own verbosity based on that flag.

**Describe the solution you'd like**
Have a declaration in the public interface that lets the calling program know the verbosity value.

**Describe alternatives you've considered**
* The calling program could pre-parse the command line for the `--v` option before handing it off to `benchmark::Initialize`. It would then have to duplicate the splitting, integer conversion, and range checking code for that option’s value.
* The calling program could declare `extern int32_t FLAGS_v;`. This basically breaks encapsulation and will break if the library’s representation of the option changes.
* The calling program could have a separate verbosity option. This is awkward for the user who will have to understand why a program has two verbosity options.

Matthdonau:
Can be closed i guess?

dominichamon:
fixed by #1397

PR #1397, opened 2022-05-17
Add option to get the verbosity provided by commandline flag -v (#1330)

Matthdonau:


Matthdonau:
Last push should fix the clang format issue.

dominichamon:
just made one tweak to make sure the test is configured correctly in all cases (replace assert with explicit test failure)

Matthdonau:
@dominichamon this pattern you replaced is also present in spec_arg_test.cc. Maybe you want or want not to change it there as well if asserting is a problem. Just wanted to let you know.

dominichamon:
thanks. i'll take a look at that. i must have missed it.

Issue #1130, opened 2021-04-30
[FR] Allow for warm up before running each benchmark


Adrodoc:
**Is your feature request related to a problem? Please describe.**
Some code is slow at first but becomes faster after some time. In my example I am benchmarking an MPI application and the first few benchmark iterations run a lot slower (I suspect because the memory of the MPI-Window is not fully initialized and put into cache).
For example here are the timings of the first 10 iterations of some benchmark I am working with:
```
seconds=0.0433436
seconds=0.0381923
seconds=0.0215415
seconds=0.0117373
seconds=0.0098229
seconds=0.0090367
seconds=0.0100212
seconds=0.0068646
seconds=0.0071745
seconds=0.0073567
```
As you can see it takes a few iterations to warm up and then it stays at about 0.007 seconds which is almost 6 times faster than the initial iteration.

**Describe the solution you'd like**
I would like to specify a `MinWarmUpTime` that acts like `MinTime` to dynamically determine a number of iterations to run a warm up. All timings of the warm up should be discarded (to not impact the reported average) and they should not be used to determine the number of iterations for the actual benchmark.
After the warm up the iteration count of the benchmark itself would be determined using the next few iterations as normal using `MinTime`.

**Describe alternatives you've considered**
You could also specify a fixed iteration count for warm up, but then users would have to first find a good count manually (this is what I am doing now).
You could also specify a fixed `WarmUpTime` and always run the warm up for that time. This might be easier than dynamically determine a number of iterations and should work fine too.

Matthdonau:
Can be closed i guess?

dominichamon:
Fixed by #1399

PR #1399, opened 2022-05-21
Introduce warmup phase to BenchmarkRunner (#1130)

Matthdonau:
In order to account for caching effects in user
benchmarks introduce a new command line option
"--benchmark_min_warmup_time"
which allows to specify an amount of time for
which the benchmark should be run before results
are meaningful.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1036,7 +1036,7 @@ class BENCHMARK_EXPORT Benchmark {
   // Set the minimum amount of time to run the benchmark before taking runtimes
   // of this benchmark into account. This
   // option overrides the `benchmark_min_warmup_time` flag.
-  // REQUIRES: `t > 0` and `Iterations` has not been called on this benchmark.
+  // REQUIRES: `t >= 0` and `Iterations` has not been called on this benchmark.
   Benchmark* MinWarmUpTime(double t);
 
   // Specify the amount of iterations that should be run by this benchmark.
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1033,6 +1033,12 @@ class BENCHMARK_EXPORT Benchmark {
   // REQUIRES: `t > 0` and `Iterations` has not been called on this benchmark.
   Benchmark* MinTime(double t);
 
+  // Set the minimum amount of time to run the benchmark before taking runtimes
+  // of this benchmark into account. This
+  // option overrides the `benchmark_min_warmup_time` flag.
+  // REQUIRES: `t > 0` and `Iterations` has not been called on this benchmark.
+  Benchmark* MinWarmUpTime(double t);
+
   // Specify the amount of iterations that should be run by this benchmark.
   // REQUIRES: 'n > 0' and `MinTime` has not been called on this benchmark.
   //
@@ -1145,6 +1151,7 @@ class BENCHMARK_EXPORT Benchmark {
 
   int range_multiplier_;
   double min_time_;
+  double min_warmup_time_;
   IterationCount iterations_;
   int repetitions_;
   bool measure_process_cpu_time_;
@@ -1539,6 +1546,7 @@ struct BENCHMARK_EXPORT BenchmarkName {
   std::string function_name;
   std::string args;
   std::string min_time;
+  std::string min_warmup_time;
   std::string iterations;
   std::string repetitions;
   std::string time_type;
--- a/src/benchmark_api_internal.cc
+++ b/src/benchmark_api_internal.cc
@@ -25,6 +25,7 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
       statistics_(benchmark_.statistics_),
       repetitions_(benchmark_.repetitions_),
       min_time_(benchmark_.min_time_),
+      min_warmup_time_(benchmark_.min_warmup_time_),
       iterations_(benchmark_.iterations_),
       threads_(thread_count) {
   name_.function_name = benchmark_.name_;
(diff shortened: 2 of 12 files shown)


src/benchmark_runner.cc
```diff
@@ -258,19 +263,81 @@ bool BenchmarkRunner::ShouldReportIterationResults(
...
+  // wrong guess since the warmup loops might be considerably slower (e.g
+  // because of caching effects).
+  const IterationCount i_backup = iters;
+
+  for (;;) {
+    b.Setup();
+    i_warmup = DoNIterations();
+    b.Teardown();
+
+    const bool finsish = ShouldReportIterationResults(i_warmup);
```
Matthdonau, line 309: typo: finish

src/benchmark_api_internal.cc
```diff
@@ -50,6 +51,11 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
     name_.min_time = StrFormat("min_time:%0.3f", benchmark_.min_time_);
   }
 
+  if (!IsZero(benchmark->min_warmup_time_)) {
+    name_.min_warmup_time =
+        StrFormat("min_warmup_time:%0.3f", benchmark_.min_warmup_time_);
```
dominichamon, line 56: why is it important to have this in the name?

src/benchmark_register.cc
```diff
@@ -355,9 +356,17 @@ Benchmark* Benchmark::MinTime(double t) {
   return this;
 }
 
+Benchmark* Benchmark::MinWarmUpTime(double t) {
+  BM_CHECK(t > 0.0);
```
dominichamon, line 360: >=, i think. given the default is 0.0, i can imagine a user might want to ensure one benchmark _isn't_ run with warmup.

src/benchmark_runner.cc
```diff
@@ -258,19 +263,81 @@ bool BenchmarkRunner::ShouldReportIterationResults(
...
+          !b.use_manual_time());
+}
+
+double BenchmarkRunner::GetMinTimeToApply() const {
+  // In order to re-use functionality to run and measure benchmarks for running
+  // a warmup phase of the benchmark, we need a way of telling whether to apply
+  // min_time or min_warmup_time. This function will figure out if we are in the
+  // warmup phase and therefore need to apply min_warmup_time or if we already
+  // in the benchmarking phase and min_time needs to be applied.
+  if (!warmup_done)
```
dominichamon, line 281: i would flip this around and simplify: `return warmup_done ? min_time : min_warmup_done;`

src/benchmark_runner.cc
```diff
@@ -258,19 +263,81 @@ bool BenchmarkRunner::ShouldReportIterationResults(
...
+    return min_warmup_time;
+  else
+    return min_time;
+}
+
+void BenchmarkRunner::FinishWarmUp(const IterationCount& i) {
+  warmup_done = true;
+  iters = i;
```
dominichamon, line 289: i'm not sure why you're setting `iters` here if you then ignore it later.

Matthdonau: If I don't reset it here to the value it was before warmup started, the first run of the benchmarking phase will start with the last value of iters of the warmup phase, right?
The description of the issue explicitly stated, that the benchmarkings phase determination of a proper iters count should be unaffected by the warmup phase.

src/benchmark_runner.cc
Matthdonau: good call. done

src/benchmark_runner.cc
Matthdonau: doen

src/benchmark_api_internal.cc
Matthdonau: I don't really think it is important. However I would assume its the same reasoning as for having min_time in the name of the benchmark which I assume is:
Something might change on the results depending on this value. In this case this might be figuring out how long to run this benchmark on a certain platform until I am sure the caches are warmed up which might be easier if I can register the benchmark a few times with different warm up phases to see where the results saturate. If you think it shouldn't be in the name I am happy to remove it :)

src/benchmark_register.cc
Matthdonau: So you are suggesting >=.
If yes, good point done.

Matthdonau: So you proposing to check it for == 0?
Because this is probably is more interesting to the user than ensuring its positive and therefore set. Agree done

Matthdonau: figured it out. You meant >=0.0. My bad wasn't paying enough attention. Will fix it immediately

Matthdonau:
Failing test also sometimes fails in main, therefore I assume it is a known issue and not related to code changes of this PR?

src/benchmark_api_internal.cc
dominichamon: it's fine :)

src/benchmark_runner.cc
dominichamon: ah yes, sorry. i missed that we were _resetting_ it and thought maybe we were setting it to what it was at the end of the warmup (the opposite of what your comment claims).

include/benchmark/benchmark.h
```diff
@@ -1033,6 +1033,12 @@ class BENCHMARK_EXPORT Benchmark {
   // REQUIRES: `t > 0` and `Iterations` has not been called on this benchmark.
   Benchmark* MinTime(double t);
 
+  // Set the minimum amount of time to run the benchmark before taking runtimes
+  // of this benchmark into account. This
+  // option overrides the `benchmark_min_warmup_time` flag.
+  // REQUIRES: `t > 0` and `Iterations` has not been called on this benchmark.
```
dominichamon, line 1039: t >= 0

Matthdonau: good call. done

dominichamon:
can you add this to the user guid in `docs/` please?

Matthdonau:
Tried my best adding a documentation to the user guid

dominichamon:
looks great, thank you!

PR #1400, opened 2022-05-23
Add support to get clock for new architecture CSKY

zixuan-wu:
it's like what loongarch does to get cycle clock for CSKY by gettimeofday function.


dominichamon:
i have no way to test this or to confirm it works, but it doesn't break anything so.. :)

thanks!

Issue #1303, opened 2021-12-10
[BUG] Graphic Bug: When the length of the name of the function to benchmark is too long the output is not aligned in the table (see screenshot)


gerzin:
**Describe the bug**
The tabular output is not aligned when the length of the name of the function plus the parameter passed to it gets too long.
The screenshot attached should be self explanatory.

**System**
Which OS, compiler, and compiler version are you using:
  - OS: MacOs 10.15.3
  - Compiler and version: Homebrew clang version 12.0.1

**To reproduce**
benchmark a function named BM_GridCreationAndDeletion and pass to it the following arguments 1e1, 1e2, ..., 1e5.
The last benchmark should overflow the table boundaries.

**Expected behavior**
All the text should fit in the table

**Screenshots**
<img width="678" alt="Screenshot 2021-12-10 at 19 14 38" src="https://user-images.githubusercontent.com/15021033/145622052-88aeb750-f49e-4abf-b619-1b6a3a84f056.png">

dominichamon:
it seems it's maxing out at 32, but i'm not sure why... we start with a width of 10 and iterate through the benchmarks tracking the max so far against the benchmark name's `str()` method. this returns the joined string of name, args, min time, iterations, repetitions, time type, and threads without a maximum.

the only thing i can think of is that "%-*s" has a maximum width of 32, but a quick test shows that's not the case.

hmm...

dominichamon:
i've tried to reproduce this and i can't but i'm wondering if it's a combination of the long name and the long times.

dominichamon:
i'm convinced this is due to the time being longer than expected (13 characters).  we can't calculate this width in advance (naturally) so the best we can do is extend it (to match CPU at 15), but that will also be fragile in the future.

gerzin:
> i've tried to reproduce this and i can't but i'm wondering if it's a combination of the long name and the long times.

I got it when running  [this](https://github.com/gerzin/parallel-cellular-automata/blob/main/tests/benchmarks/bench_sequential_automaton.cpp) benchmark on an old macbook pro (if you'r pc isn't too powerful it may happen on yours too) .

The only dependencies needed to run it are a couple of headers in the repo plus the framework.

Matthdonau:
Would switching to a scientific notation in such cases be a solution (e.g when exceeding a billion switching to 1e9 <timeunint>)?

Matthdonau:
Added PR as a suggestion. Not sure if scientific notation is desired :)

PR #1402, opened 2022-05-26
Report large numbers in scientific notation in console reporter (#1303)

Matthdonau:
Report all time numbers > 10 digits in scientific notation with
4 decimal places. This is necessary since only 10 digits
are currently reserved for the time columns (Time and CPU).
If exceeding 10 digits the output isnt properly aligned anymore.

### DIFF
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -99,6 +99,9 @@ static void IgnoreColorPrint(std::ostream& out, LogColor, const char* fmt,
 }
 
 static std::string FormatTime(double time) {
+  // For the time columns of the console printer 13 digits are reserved. One of
+  // them is a space and max two of them are the time unit (e.g ns). That puts
+  // us at 10 digits usable for the number.
   // Align decimal places...
   if (time < 1.0) {
     return FormatString("%10.3f", time);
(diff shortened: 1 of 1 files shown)


dominichamon:
JSON output isn't affected, so any serious tooling should be fine.  it _might_ cause some readability issues if the numbers are mixed but i'm not sure how much that matters.  i'm _tempted_ to suggest switching to scientific notation more generally for the console reporter...

Matthdonau:
IMHO setting the time unit to ns in a benchmark which exceeds 10 digits ns is simply a false usage of the library and the proper solution is to switch to a different time unit (which the library already supports). So in my point of view this is a fix for a very exotic use case in order to not mess up the formatting.  Therefore readability isn't really one of my concerns here since even if 11 digits would be supported, 11 digit numbers are also not very readable.
I don't have a strong opinion on switching to scientific notation in general (could also be a setting?) I think this would be better suited for an extra issue (which I would be happy to implement). The "bug" fixed with this PR is as said imho a misuse and shouldn't be a strong argument for switching to an other notation in general.

dominichamon:
ah, so the user could also workaround by setting a better time unit (micro or milli).

this change LGTM. i'm going to give it a bit to see if anyone else has a stronger opinion.

Matthdonau:
Sounds good!

LebedevRI:
This is fine i guess.

dominichamon:
thanks!

PR #1404, opened 2022-06-02
added INSTRUCTIONS and CYCLES hardware perf counters on MacOS.

robwyatt:
MacOS has a semi undocumented API in libpthread that returns the INSTRUCTION and CYCLE counts of the calling thread. The API is thread_selfcounts and is available on Intel and Apple Silicon hardware.  Using the API, this PR cleanly integrates with the existing performance counter code and allows those two counters.

For example you can use `--benchmark_perf_counters=INSTRUCTIONS,CYCLES` just like you can on Linux but no external dependencies are required as everything needed is in libpthread.

The performance counter documentation has been updated to reflect this change.

### DIFF
--- a/docs/perf_counters.md
+++ b/docs/perf_counters.md
@@ -1,5 +1,3 @@
-<a name="perf-counters" />
-
 # User-Requested Performance Counters
 
 When running benchmarks, the user may choose to request collection of
@@ -9,14 +7,30 @@ performance improvement matches expectations.
 
 This feature is available if:
 
-* The benchmark is run on an architecture featuring a Performance Monitoring
-  Unit (PMU),
-* The benchmark is compiled with support for collecting counters. Currently,
-  this requires [libpfm](http://perfmon2.sourceforge.net/) be available at build
-  time
+* The benchmark is run on an architecture featuring a Performance Monitoring Unit (PMU),
+* The benchmark is compiled with support for collecting counters. 
+
 
 The feature does not require modifying benchmark code. Counter collection is
 handled at the boundaries where timer collection is also handled. 
+
+The counter values are reported back through the [User Counters](../README.md#custom-counters)
+mechanism, meaning, they are available in all the formats (e.g. JSON) supported
+by User Counters.
+
+## MacOS
+MacOS, on Apple Silicon and Intel, has built in support for per thread instruction 
+and cycle counters. These counters can be queried by the by semi-undocumented API 
+in libpthread `thread_selfcounts`. Benchmark support for these counters is always 
+enabled as it requires no additional dependencies.
+
+To use, pass a comma-separated list of counter names through the 
+`--benchmark_perf_counters` flag. The only available counter names 
+are `CYCLES` and `INSTRUCTIONS`.
+
+## Linux
+Currently, this requires [libpfm](http://perfmon2.sourceforge.net/) be available 
+at build time.
 
 To opt-in:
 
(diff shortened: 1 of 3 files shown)


LebedevRI:
Shouldn't this go into perfmon itself?

robwyatt:
I looked at that but it's a hell of an effort to get perfmon to build on MacOS and support the whole API but it's difficult because it uses read() with a driver provided fd to get data - this is not how MacOS works and from user mode isn't really possible. 

It also seems like software engineering overkill to make a a whole library to wrap a single API that is built in to the OS (the OS already does a lot of work to make those two counters behave the same on both platforms, the counters are thread specific and context switching and processor migration is handled so the counters are rock solid). This PR just calls that API where the perform version calls read(), to me it seems like a fair change and it has zero overhead and zero additional dependencies, ultimately its a 50 line addition, no existing code was changed. I understand if you don't agree but I'd like to hear your thoughts on how to get around the file descriptor/read() problem.

Now, if we had easy access to the whole set of counters then the full library would make sense but the read() problem is not trivial even from kernel mode. Getting access to the the hardware counters on MacOS is a challenge and completely different on Intel and Arm so it would be two completely independent implementations and on top of that kernel drivers are tricky especially on Apple Silicon. On both platforms Apple assume they have sole access to the counters and writing a kernel driver would break things like instruments and the existing profile tools.

Sorry for the long reply, but I did look at all the options and I'm pretty sure this is the best solution.

docs/perf_counters.md
```diff
@@ -1,5 +1,3 @@
-<a name="perf-counters" />
```
dominichamon, line 1: can you leave this in please? (or replace it with the equivalent github markdown anchor method)

src/perf_counters.cc
```diff
@@ -114,6 +114,50 @@ void PerfCounters::CloseCounters() const {
     close(fd);
   }
 }
+#elif defined(BENCHMARK_OS_MACOSX)
```
dominichamon, line 117: i wonder if it's worth splitting this file into perf_counters_linux.cc and perf_counters_osx.cc

(then we can selectively compile on top of using these guards)

src/perf_counters.h
```diff
@@ -35,6 +35,13 @@
 #pragma warning(disable : 4251)
 #endif
 
+#ifdef BENCHMARK_OS_MACOSX
+  //this is the OSX syscall wrapper function, implemented in libpthread, that returns
+  //per thread instruction and cycle counters. The counts remain correct over context
+  //switches and core migrations.
+  extern "C"  int thread_selfcounts(int type, void *buf, size_t nbytes);
```
dominichamon, line 42: is this api stable over osx/libpthread version updates?

dominichamon:
please check the clang format (and any other build failures) :)

src/perf_counters.h
robwyatt: Its been stable for years, it was added in 2017 and is in the source code as syscall 186, its on Intel and Arm versions of MacOS, its technically on the XNU OS's but you'll have a hard time calling it on anything else because of sandboxing.

src/perf_counters.cc
robwyatt: not a bad idea

src/perf_counters.h
dominichamon: SGTM .. thanks for the complete answer.

bafanahub (approved):

dmah42:
@robwyatt did you intend to get back to this?

PR #1408, opened 2022-06-08
[bazel] Build libpfm as a dependency to allow collection of perf counters

rajachan:
This commit builds libpfm using rules_foreign_cc and lets the default
build of the benchmark library support perf counter collection without
needing additional work from users.
    
Tested with a custom target:
    
    bazel run \
            --override_repository=com_github_google_benchmark=/home/raghu/benchmark \
            -c opt :test-bench -- "--benchmark_perf_counters=INSTRUCTIONS,CYCLES"
    Using profile: local
    
    <snip>
    
    ----------------------------------------------------------------------
    Benchmark            Time             CPU   Iterations UserCounters...
    ----------------------------------------------------------------------
    BM_Test      0.279 ns        0.279 ns   1000000000 CYCLES=1.00888 INSTRUCTIONS=2

### DIFF
--- a/docs/perf_counters.md
+++ b/docs/perf_counters.md
@@ -12,16 +12,17 @@ This feature is available if:
 * The benchmark is run on an architecture featuring a Performance Monitoring
   Unit (PMU),
 * The benchmark is compiled with support for collecting counters. Currently,
-  this requires [libpfm](http://perfmon2.sourceforge.net/) be available at build
-  time
+  this requires [libpfm](http://perfmon2.sourceforge.net/), which is built as a
+  dependency via Bazel.
 
 The feature does not require modifying benchmark code. Counter collection is
 handled at the boundaries where timer collection is also handled. 
 
 To opt-in:
-
-*  Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
-*  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
+* If using a Bazel build, add `--define pfm=1` to your buid flags
+* If using CMake:
+  * Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
+  * Enable the CMake flag `BENCHMARK_ENABLE_LIBPFM` in `CMakeLists.txt`.
 
 To use, pass a comma-separated list of counter names through the
 `--benchmark_perf_counters` flag. The names are decoded through libpfm - meaning,
--- a/AUTHORS
+++ b/AUTHORS
@@ -50,6 +50,7 @@ Norman Heino <norman.heino@gmail.com>
 Oleksandr Sochka <sasha.sochka@gmail.com>
 Ori Livneh <ori.livneh@gmail.com>
 Paul Redmond <paul.redmond@gmail.com>
+Raghu Raja <raghu@enfabrica.net>
 Radoslav Yovchev <radoslav.tm@gmail.com>
 Rainer Orth <ro@cebitec.uni-bielefeld.de>
 Roman Lebedev <lebedev.ri@gmail.com>
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -18,6 +18,14 @@ config_setting(
     visibility = [":__subpackages__"],
 )
 
+config_setting(
+    name = "perfcounters",
+    define_values = {
+        "pfm": "1",
+    },
+    visibility = [":__subpackages__"],
+)
+
 cc_library(
     name = "benchmark",
     srcs = glob(
@@ -41,7 +49,16 @@ cc_library(
     # Using `defines` (i.e. not `local_defines`) means that no
     # dependent rules need to bother about defining the macro.
     linkstatic = True,
-    defines = ["BENCHMARK_STATIC_DEFINE"],
+    defines = [
+        "BENCHMARK_STATIC_DEFINE",
+    ] + select({
+        ":perfcounters": ["HAVE_LIBPFM"],
+        "//conditions:default": [],
+    }),
+    deps = select({
+        ":perfcounters": ["@libpfm//:libpfm"],
+        "//conditions:default": [],
+    }),
 )
 
 cc_library(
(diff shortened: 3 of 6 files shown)


dominichamon:
it'll need to be added to cmake too. and do we actually want to push the cost of the extra library on to all end users, even if they don't care about perf counters?

raghupazl:
>  it'll need to be added to cmake too. and do we actually want to push the cost of the extra library on to all end users, even if they don't care about perf counters?

Now that bazelisk builds it on behalf of users, it should be a self-contained target with no meaningful impact for those who do not care about the feature... unless you think there's a runtime cost associated with this path? As an alternative, I can add a separate target with libpfm support, giving folks an option to use either one. Thoughts?

raghupazl:
Hmm, the sourceforge links seem to have expired, hence the build failures. Let me find a canonical source for the package.

rajachan:
> it'll need to be added to cmake too. and do we actually want to push the cost of the extra library on to all end users, even if they don't care about perf counters?

Now that bazelisk builds it on behalf of users, it should be a self-contained target with no meaningful impact for those who do not care about the feature... unless you think there's a runtime cost associated with this path? As an alternative, I can add a separate target with libpfm support, giving folks an option to use either one. Thoughts?

rajachan:
Fixed build failures.

rajachan:
`//test:perf_counters_gtest` and `//test:perf_counters_test` are failing. Are these tests that never ran before in the CI environment, given the perf counter path was not built?

```
[  FAILED  ] PerfCountersTest.ReopenExistingCounters (0 ms)
[ RUN      ] PerfCountersTest.CreateExistingMeasurements
test/perf_counters_gtest.cc:139: Failure
Value of: perf_counter_measurements[8].Stop(measurements)
  Actual: true
Expected: false
test/perf_counters_gtest.cc:143: Failure
Value of: perf_counter_measurements[9].Stop(measurements)
  Actual: true
Expected: false
[  FAILED  ] PerfCountersTest.CreateExistingMeasurements (0 ms)
```

dominichamon:
yes, it's possible. but it suggests that the bazel config is not quite the same as the cmake one.

rajachan:
Perhaps I am not following your ask. CMake uses `check_library_exists` to build with the perf counter support when the library is available on the system benchmark is getting built on. Unlike Bazel, CMake doesn't let us vendor a library in a hermetic way. I did notice `BENCHMARK_ENABLE_LIBPFM` was set to OFF, forcefully disabling the counters. Turning that back on still doesn't let the test pass as I expected. It also doesn't look like the tests are using the CMake build?

dominichamon:
> Perhaps I am not following your ask. CMake uses `check_library_exists` to build with the perf counter support when the library is available on the system benchmark is getting built on. Unlike Bazel, CMake doesn't let us vendor a library in a hermetic way. I did notice `BENCHMARK_ENABLE_LIBPFM` was set to OFF, forcefully disabling the counters. Turning that back on still doesn't let the test pass as I expected. It also doesn't look like the tests are using the CMake build?

yes, there are perf counter tests that use cmake (build-and-test-perfcounters).  with your change, we are now also running these same tests on the bazel build for the first time, but they're failing. which suggests _something_ about how the bazel build is configured is different from the [cmake build-and-test-perfcounters build](https://github.com/google/benchmark/blob/main/.github/workflows/build-and-test-perfcounters.yml).

my initial guess would be: the cmake version doesn't run the tests as they fail due to permissions issues.

so to be equivalent, you may need to just filter out the pfm tests from the bazel workflow on the CI bots.

> To clarify, I am using the same define, HAVE_LIBPFM, that CMake uses here:
> 
> ```
> # libpfm, if available                                                          
> if (HAVE_LIBPFM)                                                                
>   target_link_libraries(benchmark PRIVATE pfm)                                  
>   add_definitions(-DHAVE_LIBPFM)                                                
> endif()  
> ```

rajachan:
Apologies for letting this sit, I've been taking some time off. I hope to get back to this next week.

LebedevRI:
If this is bazel-specific, it should not touch non-bazel stuff.

rajachan:
Addressed the CMake situation, removed the explicit CMakeLists.txt change to always turn on PFM... Added a config_setting to make this optional for Bazel builds too. Please take a look.

LebedevRI:
As long as it's bazel-only i have no comments :)

docs/perf_counters.md
```diff
@@ -12,16 +12,17 @@ This feature is available if:
...
+  dependency via Bazel.
 
 The feature does not require modifying benchmark code. Counter collection is
 handled at the boundaries where timer collection is also handled. 
 
 To opt-in:
-
-*  Install `libpfm4-dev`, e.g. `apt-get install libpfm4-dev`.
-*  Enable the cmake flag BENCHMARK_ENABLE_LIBPFM.
+* If using a Bazel build, add `--define pfm=1` to your buid flags
```
dominichamon, line 22: you may want to change the default to 0 if we're going to document the "opt-in" procedure. wdyt?

dominichamon:
thanks for coming back to it. i just had one comment :)

docs/perf_counters.md
rajachan: If pfm=0 is specified or if no define for pfmm is provided, it will be disabled, which is the default behavior in the config_setting.

dominichamon:
```
config_setting(
    name = "perfcounters",
    define_values = {
        "pfm": "1",
    },
    visibility = [":__subpackages__"],
)
```

doesn't this mean it's defaulted to "1"?

rajachan: That config_setting's value to match is 1, but if you see where I use it:

```
    defines = [
        "BENCHMARK_STATIC_DEFINE",
    ] + select({
        ":perfcounters": ["HAVE_LIBPFM"],
        "//conditions:default": [],
    }),
    deps = select({
        ":perfcounters": ["@libpfm//:libpfm"],
        "//conditions:default": [],
    }),
```

We don't enable PFM or take a dependency on it if unless config_setting matches to true.

dominichamon: ok! thanks

dominichamon (approved):

PR #1409, opened 2022-06-10
Fix warnings -Wshift-sign-overflow, -Wzero-as-null-pointer

OfekShilon:


OfekShilon:
Apparently backwards support prevents changing NULL to nullptr - only our own build demanded it. I'll close the PR.

Issue #1340, opened 2022-02-03
[BUG] DoNotOptimize() adds overhead with extra copy of argument(gcc)


alexgpg:
**Describe the bug**

On GCC DoNotOptimize() does full copy of an argument if it's not a pointer and slows down benchmark.
If an argument is big enough there is a memcpy() call for copy the agrument.
An argument object can be a big object so benchmark::DoNotOptimize() could add sufficient overhead and affect to benchmark results.

**System**
  - OS: Linux 5.16.2
  - Compiler and version:
    - GCC 11.0 (but looks like versions 4.1.2 - 11.12(all available on godbolt) is also affected)
    - clang 13.0.0
  - Google Benchmark: 1.6.1

**To reproduce**

Code to reproduce

1. Benchmark code

```c++
#include <benchmark/benchmark.h>
#include <array>

// Easy way to get a big size object.
static std::array<int, 1000000> arr{};

static void BM_Array(benchmark::State& state) {
  for (auto _ : state) {
    benchmark::DoNotOptimize(arr);
  }
}

static void BM_PtrToArray(benchmark::State& state) {
  for (auto _ : state) {
    benchmark::DoNotOptimize(&arr);
  }
}

BENCHMARK(BM_Array);
BENCHMARK(BM_PtrToArray);

BENCHMARK_MAIN();
```
2. Run for g++ and clang++

Run with g++

```
g++ -g -Ofast -march=native -lbenchmark -lpthread -o bench.bin main.cc
./bench.bin
--------------------------------------------------------
Benchmark              Time             CPU   Iterations
--------------------------------------------------------
BM_Array           81655 ns        81588 ns         8603
BM_PtrToArray      0.242 ns        0.241 ns   1000000000

```

**BM_PtrToArray is 338539 times(!) slower.**

Run with clang++

```
clang++ -g -Ofast -march=native -lbenchmark -lpthread -o bench.bin main.cc
./bench.bin
--------------------------------------------------------
Benchmark              Time             CPU   Iterations
--------------------------------------------------------
BM_Array           0.245 ns        0.244 ns   1000000000
BM_PtrToArray      0.240 ns        0.240 ns   1000000000
```

Results are approximately equal.

**Expected behavior**

Zero or a litle overhead fot DoNotOptimize(). No full copy of an argument.
Equal or approximately equal overhead as for clang.

**Additional context**

It's gcc assembly code for DoNotOptimize(). There is memcpy() call in code.

```
[...]

PR #1410, opened 2022-06-11
Fix DoNotOptimize() GCC copy overhead (#1340)

alexgpg:
- Fixed issue with GCC copy overhead
- Assembly tests for issued cases were added
- Specific compiler versions required for reliable tests passed assembly tests
  was added

## Fix/Workaround for the bug

The solution is the split DoNotOptimize() in two cases - value fits in register and value doesn't fit in a register. And use case-specific asm constraint.

The solution works with GCC starting 5 version because the fix requires std::is_trivially_copyable feature which is available since GCC 5 and newer.

## Add specific compiler versions for assembly tests

- Assembly tests are inherently non-portable. So explicitly add GCC and Clang versions required for reliable tests passed
- Write a warning message if the current compiler version isn't supported

## How to check

Use this **[Dockerfile](https://gist.github.com/alexgpg/959f7ded943e7d42100a2013c8f0f697#file-dockerfile)**.
Copy it and run

```
docker build --tag googlebench:1340fix .
```

If docker build successfully that means the branch builds and tests passed.

## Notes

- Fallback for GCC version < 5 still exists but it uses the "m" constraint which means a little bit more overhead in some cases

- Why versions are 5.5.0 for GCC and 5.0.0 for Clang?
  I took Clang version from [CI configuration](https://www.travis-ci.org/github/google/benchmark/jobs/741836197#L51)
  I use GCC 5.5.0 cause it's a version available for Ubuntu 14.04 and supports std::is_trivially_copyable

- Is it right that run_state_assembly_test_CHECK tests failed?
  Yes, that's right. But it's not related with this PR/commits.
  Test run_state_assembly_test_CHECK doesn't work since [commit: Enable -Wconversion (#1390) ](https://github.com/google/benchmark/commit/8d86026c67e41b1f74e67c1b20cc8f73871bc76e)

## Topics to discuss

- Is a fallback for GCC < 5 is needed or it's ok to drop support for GCC < 5 right now in this PR?
- Is more assembly tests for is needed? **I'm very happy to hear ideas for test cases**

## Links

- [Issue: [BUG] DoNotOptimize() adds overhead with extra copy of argument(gcc)](https://github.com/google/benchmark/issues/1340)
- [Dockerfile](https://gist.github.com/alexgpg/959f7ded943e7d42100a2013c8f0f697#file-dockerfile) for check build and tests
- [GCC: Bug 105519 - Unnecessary memcpy() copy for empty asm volatile](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105519)

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -446,6 +446,7 @@ inline BENCHMARK_ALWAYS_INLINE void ClobberMemory() {
 // intended to add little to no overhead.
 // See: https://youtu.be/nXaxk27zwlk?t=2441
 #ifndef BENCHMARK_HAS_NO_INLINE_ASSEMBLY
+#if !defined(__GNUC__) || defined(__llvm__) || defined(__INTEL_COMPILER)
 template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
   asm volatile("" : : "r,m"(value) : "memory");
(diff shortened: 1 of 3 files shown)


dominichamon:
i don't see any issue with retaining the GCC < 5 fallback for now.  we run (and are built for) a large number of platforms with varying compiler support.  this PR is already a huge step forward for the core platforms :D

i'm sad that the assembly tests have bit-rotted. the goal of them was to protect the core loop assembly to be as minimal as possible (as subtle changes in class layout can break this).  i think more would be good if someone wanted to take on the challenge of "ideal assembly" across multiple compilers and platforms, but it is always going to be brittle and difficult to maintain.

alexgpg:
Code formating fix added.

alexgpg:
>i think more would be good if someone wanted to take on the challenge of "ideal assembly" across multiple compilers and platforms, but it is always going to be brittle and difficult to maintain.

I think it's an interesting(and quite hard) task. I can try to fix some tests for "supported" versions. That will be the first step for re-enabling assembly tests on CI. The next step is running tests across compiler versions and start thinking about portable/non-portable parts.

But I can't promise anything. I'm not sure how much time I have for that.

Anyway, I can remove the "supported compiler versions" part from this PR if it doesn't fit here.

alexgpg:
Update:

- Formating fixed again. I had some issues with formatter but now it looks ok
- Rebased on top of the main branch

Note:

All clang sanitizers build fail with the same error message:

```
CMake Error at CMakeLists.txt:309 (message):
```

include/benchmark/benchmark.h
```diff
@@ -459,6 +460,57 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
   asm volatile("" : "+m,r"(value) : : "memory");
 #endif
 }
+#else
+#if defined(BENCHMARK_HAS_CXX11) && (__GNUC__ >= 5)
```
dominichamon, line 464: i think #elif simplifies this slightly and removes one of the nested #ifs

dominichamon:
> Update:
> 
> * Formating fixed again. I had some issues with formatter but now it looks ok
> * Rebased on top of the main branch
> 
> Note:
> 
> All clang sanitizers build fail with the same error message:
> 
> ```
> CMake Error at CMakeLists.txt:309 (message):
>   Failed to determine the source files for the regular expression backend
> ```

sanitizer failure is happening at HEAD too. i haven't had a chance to figure out what changed.

include/benchmark/benchmark.h
alexgpg: Good point. I've added a fix.

dclowes:
I believe that this commit causes a problem that does not exist at the previous commit.

The error occurs on both  debian:latest (with gcc 10.2.1) and centos:7 (with gcc 11.2.1) containers hosted on Debian 11 (Linux 5.10.0-14-amd64) on an AMD 5950X.

The error message, together with a cut down program that produces it, follows:

`root@e4e9d70cd077:/tmp/benchmark# g++ junk.cpp 
In file included from junk.cpp:1:
/usr/local/include/benchmark/benchmark.h: In function 'void Ctor_1c(benchmark::State&)':
/usr/local/include/benchmark/benchmark.h:487:3: error: impossible constraint in 'asm'
  487 |   asm volatile("" : "+r"(value) : : "memory");
      |   ^~~
root@e4e9d70cd077:/tmp/benchmark# cat junk.cpp 
#include <cstring>
uint8_t pad_1[6] { 0x00 + 35, 0x40 + 20, 2, 10, 0x20, 0};
uint8_t pad_6[6] { 0x00 + 35, 0x40 + 20, 2, 10, 0x20, 0};
class Event {
    public:
    Event(const uint8_t *evt) {
        memmove(&evt_, evt, 6);
    }
    private:
    uint8_t evt_[6];
};
static void Ctor_1c(benchmark::State& state)
{
  Event e(pad_1);
  for (auto _ : state) {
    benchmark::DoNotOptimize(e = Event(&pad_6[0]));
  }
}
BENCHMARK(Ctor_1c);
BENCHMARK_MAIN();
`

dominichamon:
https://gcc.godbolt.org/z/a6GxxYPGo at least works

equivalent code is https://gcc.godbolt.org/z/rE1G6fYd1

alexgpg:
I'll take a look. I think it's because of GCC can't store in register(use +r constraint) small array(with size smaller or equal than register). Probably it can be fixed with adding one more condition to template and use +m for the case. 

@dclowes could you post a Dockerfile? It will be helpful.

alexgpg:
It's definitely a case that breaks a new code.

Here is an even smaller code to reproduce.

```C++
char buffer3[3] = "";
benchmark::DoNotOptimize(buffer3);

```
Looks like it can be fixed with a mix of an old and new approach with saving all desirable requirements.
I'll post my thoughts(and solution, I hope) soon(in a couple of days).

Issue #838, opened 2019-07-17
Store args in output files


tesch1:
It would be useful to (optionally?) store the arg(s) in the output file.

Let's say you have a benchmark that has different performance for different sizes of N.  And you want to view a plot the performance of that function across a relevant range of N.  Currently, for both JSON and csv, the N value is stored embedded in the benchmark name.  So in order to use N as an axis, you need to parse it out.

For JSON it would be nice to be able to import a dataset and use the Ns without parsing them out of the name field.  This could be addressed at the same time as #814 - there could be a flag to switch between the current mode and a new mode of allowing multiple entries with the same "name" but different "args" entries.  (I assume the reason that "name" and "run_name" are the same is in order to always have a unique "name" (?))

For plotting in, ie gnuplot, having a column(s) in the CSV with the args would also be useful.  And maybe extra newlines between name blocks.

LebedevRI:
> This could be addressed at the same time as #814 - there could be a flag to switch between the current mode and a new mode of allowing multiple entries with the same "name" but different "args" entries.

It shouldn't be a flag; more of that info should be serialized in JSON indeed.

tesch1:
Here's an implementation proposal, the goal is to make it easy to import a json file into existing libraries (ie pandas) and not have to muck around with the field names.  Please let me know what you think:

- output a key "function_name" to each entry in the JSON's "benchmark" which contains the function_name (+ optionally aggregate_name)
- add keys "arg1","arg2",... for each of the args

For csv...  I'm less sure of this being comprehensive... basically I want to be able to run gnuplot on a csv output file without mucking about too much.  (It seems csv is not actually being deprecated?)
- add a column that contains the same "function_name"
- add a column for each arg
- add extra breaks to the output file to match gnuplot's "dataset" structuring: https://stackoverflow.com/questions/35785030/plot-multiple-datasets-from-one-file

Probably the csv columns should be added after the existing ones, so as not to break existing column index dependencies, even though it's aesthetically 'alternative'.

LebedevRI:
> For csv...

I'm pretty sure it is being deprecated, so don't worry about CSV reporter, only JSON should be extended.

tesch1:
(ps. this is also related somehow to #190 )

Matthdonau:
Hi @LebedevRI as this issue is pretty old I wanted to ask if this is still desired.
If yes, I would suggest:
```
    {
      "name": "BM_ComplexityCaptureArgs/capture_test/2/4",
[...]

PR #1415, opened 2022-06-20
Report arguments and benchmark function name in json report #838

Matthdonau:


src/json_reporter.cc
```diff
@@ -232,6 +232,27 @@ void JSONReporter::PrintRunData(Run const& run) {
...
+  std::string func_name = run.run_name.str();
+  auto pos = run.run_name.str().find('/');
+  // Only if "/" is contained in run_name we need to do something.
+  if (pos != std::string::npos) func_name = func_name.substr(0, pos);
+  out << indent << FormatKV("function_name", func_name) << ",\n";
+
+  // Print all arguments
+  out << indent << "\"arguments\": "
+      << "[";
+  if (run.arguments.size() > 0) {
```
Matthdonau, line 246: Currently if no arguments are present the output is:
arguments: []
Not sure if the arguments entry shouldn't even be there if there are no arguments.

dominichamon: imo, having an empty array for arguments instead of skipping the field is appropriate.  this allows parsers to assume it exists rather than having to check.

src/json_reporter.cc
```diff
@@ -232,6 +232,27 @@ void JSONReporter::PrintRunData(Run const& run) {
...
+    for (auto& arg : run.arguments) {
+      out << "\"" << arg << "\"";
```
dominichamon, line 248: why strings? json can have number types as values.

src/json_reporter.cc
```diff
@@ -232,6 +232,27 @@ void JSONReporter::PrintRunData(Run const& run) {
       << FormatKV("per_family_instance_index", run.per_family_instance_index)
       << ",\n";
   out << indent << FormatKV("run_name", run.run_name.str()) << ",\n";
+
+  // Get the name of the benchmark function
+  std::string func_name = run.run_name.str();
+  auto pos = run.run_name.str().find('/');
```
dominichamon, line 238: you don't need to do any of this.  `std::string func_name = run.run_name.function_name;`

Matthdonau: I saw that field but realised when using BENCHMARK_CAPTURE .function_name also holds the test_case_name as well. The question is should the test_case_name be included in the reported function name or should it not.

dominichamon: yes for two(ish) reasons:

1. the json "function name" should be the benchmark instance "function name" no matter how that ends up being defined by the code
2. if the function name shouldn't include the test_case_name, that's a broader discussion than the JSON reporter
  a. i think it should include the test_case_name because BENCHMARK_CAPTURE essentially defines multiple benchmarks. it's arguable that it shouldn't use "/" as the separator.

Matthdonau: alright will change that

Issue #1416, opened 2022-06-20
[BUG] cannot find benchmark/version.h


hsdk123:
**Describe the bug**

```
FAILED: _deps/google_benchmark-build/src/CMakeFiles/benchmark.dir/reporter.cc.obj 
  C:\PROGRA~2\MICROS~4\2019\COMMUN~1\VC\Tools\MSVC\1428~1.293\bin\Hostx64\x64\cl.exe  /nologo /TP -DAL_LIBTYPE_STATIC -DHAVE_STD_REGEX -DHAVE_STEADY_CLOCK -D_CRT_SECURE_NO_WARNINGS -I_deps\google_benchmark-src\include -I_deps\google_benchmark-src\src -I_deps\google_benchmark-build\include /DWIN32 /D_WINDOWS /GR /EHsc /W4 /Zi /O2 /Ob1 /DNDEBUG -MT /utf-8 -std:c++17 /showIncludes /Fo_deps\google_benchmark-build\src\CMakeFiles\benchmark.dir\reporter.cc.obj /Fd_deps\google_benchmark-build\src\CMakeFiles\benchmark.dir\benchmark.pdb /FS -c _deps\google_benchmark-src\src\reporter.cc
C:\Users\daego\Desktop\Projects\Light.vn-core\core\out\build\x64-Release-Cored\_deps\google_benchmark-src\src\reporter.cc(23): fatal error C1083: Cannot open include file: 'benchmark/version.h': No such file or directory
```

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Windows
  - Compiler and version: Visual Studio 2019 16.8.2

**To reproduce**

I include google benchmarks through cmake's fetchcontent
```
set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "" FORCE)
FetchContent_Declare(
  google_benchmark
  GIT_REPOSITORY https://github.com/google/benchmark.git
  GIT_TAG        main
)
FetchContent_MakeAvailable(google_benchmark)
```

**Expected behavior**

**Screenshots**

**Additional context**

It looks this bug is caused by
https://github.com/google/benchmark/commit/efadf67a124528389b7faa5ed75a358528dbae22

dmah42:
i think this is no longer an issue as we don't create a version.h any more.

Issue #1004, opened 2020-07-17
[FR] Add a version number


krzikalla:
I tried to figure out the version of my installed google/benchmark. It is either 1.5.0 or 1.5.1, but I am unsure.
I'd like to have a version.h file or something similar, which gives me the version at a glance.

akagamy55:
Close:(#1004)

akagamy55:
Close:([FR]#1004)

dominichamon:
it would be _really_ nice if when you ran a benchmark binary the version of the library was part of the intro metadata.

PR #1417, opened 2022-06-20
Revert "Add possibility to ask for libbenchmark version number (#1004)"

dominichamon:
Reverts google/benchmark#1403

due to two reported bugs:

version status has nothing useful (not a git checkout) and
#1416


dominichamon:
@Matthdonau sorry for reverting your work.  feel free to revert the revert and make the improvements necessary to make this work:

1. instead of exiting when we don't get a git version, we should return the default (l74-ish of the python script).
2. figure out why cmake embedding fails... maybe we don't export the version.h correctly?

Matthdonau:
Regarding 1: I think I don't understand. What you describe is what should happen. if no git is available or we are not in a git checkout the default project version (currently 1.6.1) is used.
Regarding 2: @dominichamon I suspect the problem is that the version.h header is generated in CMAKE_BINARY_DIR while PROJECT_BINARY_DIR would be correct

dominichamon:
> Regarding 1: I think I don't understand. What you describe is what should happen. if no git is available or we are not in a git checkout the default project version (currently 1.6.1) is used.

l71 .. we exit. we don't return the default. this is being tickled when using snapshot zip releases I think.

> Regarding 2: @dominichamon I suspect the problem is that the version.h header is generated in CMAKE_BINARY_DIR while PROJECT_BINARY_DIR would be correct

ah maybe. we should do whatever the other generated header is doing as that seems to work.

Matthdonau:
I made a PR. The problem is: If there is no volatile-status.txt there is no default version to default to. Is this really the issue here or is it the issue I fixed in the PR?

PR #1418, opened 2022-06-20
Add possibility to ask for libbenchmark version number #1004

Matthdonau:


Matthdonau:
This should (to the best of my knowledge) fix the two bugs mentioned in #1417. I was able to reproduce both issues and fix them at least for my testing. However I cannot test with the setup of the two bug reporter. Therefore I cant be 100% sure if it fixes their issues as well. Sorry for introducing the bugs. I'll understand every scepticism merging this again.

dominichamon:
no skepticism at all! we have a lot of platforms and a lot of build configurations to manage.

config/get_git_version.py
```diff
@@ -0,0 +1,101 @@
...
+                        git_version = key_value[1].strip()
+                    if key == args.is_dirty_name:
+                        is_dirty = key_value[1].strip()
+                    if key == args.default_version:
+                        default_version = key_value[1].strip()
+    except:
+        # In case volatile-status cannot be read, exit with an error
+        sys.exit("Cannot open volatile-status.txt")
+
+    if git_version == "" or is_dirty == "" or default_version == "":
```
dominichamon, line 75: i think this will still have the bug. if we didn't manage to read a git_version or is_dirty then we want to return the default version.

however, if default_version is "" then we have an API contract failure and should sys.exit.

so (earlier)

```
if default_version == "":
  sys.exit("default_version must be set")
```

and here

```
if git_version == "" or is_dirty == "":
  git_version = default_version
else:
  git_version = normalize_version(git_version, is_dirty)
```

i think.

Matthdonau: I cant think of a case where this happens. In workspace_status.py if git cant be called or we are not in a git checkout the git_version gets set to 0.0.0 or v.0.0.0 and git is_dirty gets set to true while the default version is set correctly.
This means either all of the fields are empty or none. I just or'ed them in case anyone changes workspace_status.py.

The only case where we do not have a git_version or is_dirty is when workspace_status.py isn't called . If this can happen there is no default_version we can set to either

dominichamon: @tkoeppe was the person who reported the issue to me. i think they were using a snapshot zip and then building it, so they didn't have a git checkout but were building with bazel.

did you change this behaviour in this new version?

dominichamon: the build didn't start as the python script exited without outputing anything useful that bazel could use.

Matthdonau: Do you have at any chance a log for this. Because if the sys.exit in the python file were triggered there should be output to stderr. namely:
Cannot open volatile-status.txt or No usable entry in volatile-status.txt
If I change the if clauses so the always catch for testing purposes I get:
```
INFO: Analyzed target //:benchmark (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /Users/mad/Desktop/Verwaltung/Programs/gbench/benchmark/working2/benchmark/BUILD.bazel:21:24: Action include/benchmark/version.h failed: (Exit 1): get_git_version failed: error executing command bazel-out/host/bin/get_git_version --header bazel-out/darwin-fastbuild/bin/include/benchmark/version.h --header_input include/version.h.in --volatile_file bazel-out/volatile-status.txt ... (remaining 6 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox
No usable entry in volatile-status.txt
Target //:benchmark failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.015s, Critical Path: 0.54s
INFO: 2 processes: 2 internal.
FAILED: Build did NOT complete successfully
```

Matthdonau: if you encountered that I probably know how to fix that. Otherwise I don't know what happened there

dominichamon: > Do you have at any chance a log for this. Because if the sys.exit in the python file were triggered there should be output to stderr. namely: Cannot open volatile-status.txt or No usable entry in volatile-status.txt If I change the if clauses so the always catch for testing purposes I get:
> 
> ```
> INFO: Analyzed target //:benchmark (0 packages loaded, 0 targets configured).
> INFO: Found 1 target...
> ERROR: /Users/mad/Desktop/Verwaltung/Programs/gbench/benchmark/working2/benchmark/BUILD.bazel:21:24: Action include/benchmark/version.h failed: (Exit 1): get_git_version failed: error executing command bazel-out/host/bin/get_git_version --header bazel-out/darwin-fastbuild/bin/include/benchmark/version.h --header_input include/version.h.in --volatile_file bazel-out/volatile-status.txt ... (remaining 6 arguments skipped)
> 
> Use --sandbox_debug to see verbose messages from the sandbox
> No usable entry in volatile-status.txt
> Target //:benchmark failed to build
> Use --verbose_failures to see the command lines of failed build steps.
> INFO: Elapsed time: 1.015s, Critical Path: 0.54s
> INFO: 2 processes: 2 internal.
> FAILED: Build did NOT complete successfully
> ```

it was the "No usable entry" output.  this is why i think the issue was that we weren't in a git checkout and the previous logic would just bail if it couldn't get a git version.

Matthdonau: ah okay. perfect. That means there are ways of building the lib with bazel which ignore(?) the .bazelrc. That's fixable pretty easily. Give me a sec

config/get_git_version.py
```diff
@@ -0,0 +1,100 @@
...
+                    key_value = entry.split(' ', 1)
+                    key = key_value[0].strip()
+                    if key == args.version_variable_name:
+                        git_version = key_value[1].strip()
+                    if key == args.is_dirty_name:
+                        is_dirty = key_value[1].strip()
+    except:
+        # In case volatile-status cannot be read, use the default version
+        git_version = args.default_version
+        is_dirty = "TRUE"
```
dominichamon, line 71: i _think_ in this case it doesn't matter if is_dirty is true or false, but the logging in normalize_version might be surprising if it's unzipped from a snapshot or something.

config/get_git_version.py
```diff
@@ -0,0 +1,100 @@
...
+                        help='variablename of the boolean communicating if the workspace has no local changes')
+    parser.add_argument('--default_version',
+                        required=True,
+                        help='default for version which should be used in case git was not executable.')
+
+    args = parser.parse_args()
+
+    # Read volatile-status.txt file
+    git_version = ""
+    is_dirty = ""
```
dominichamon, line 57: any reason why this isn't a boolean?

`is_dirty = None` and then just use `True` and `False` throughout?

config/get_git_version.py
```diff
@@ -0,0 +1,100 @@
...
+            for entry in f.read().split("\n"):
+                if entry:
+                    key_value = entry.split(' ', 1)
+                    key = key_value[0].strip()
+                    if key == args.version_variable_name:
+                        git_version = key_value[1].strip()
+                    if key == args.is_dirty_name:
+                        is_dirty = key_value[1].strip()
+    except:
+        # In case volatile-status cannot be read, use the default version
```
dominichamon, line 69: it might be worth logging here to stdout or stderr warning the user that we couldn't determine the git version appropriately.

dominichamon: it's ok either way, i guess. consistency is good, but it's also a different build tool so.. i'm flexible.

config/get_git_version.py
```diff
@@ -64,22 +68,28 @@ def main():
                     if key == args.version_variable_name:
                         git_version = key_value[1].strip()
                     if key == args.is_dirty_name:
-                        is_dirty = key_value[1].strip()
+                        if key_value[1].strip() == "TRUE":
```
dominichamon, line 71: you can simplify this to:

```
is_dirty = (key_value[1].strip() == "TRUE")
```

workspace_status.py
```diff
@@ -39,23 +39,21 @@ def get_git_dirty(path):
...
         if out.decode() != "":
             return "TRUE"
         else:
             return "FALSE"
 
     except:
-        # Be pessimistic. In case git is not available
-        # assume the repository to be dirty.
-        return "TRUE"
+        return "FAlSE"
```
dominichamon, line 56: typo: FALSE not FAlSE

PR #1425, opened 2022-07-03
Pass the default help string into custom help printer

yurikhan:
Improves #1329.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -292,7 +292,8 @@ namespace benchmark {
 class BenchmarkReporter;
 
 BENCHMARK_EXPORT void Initialize(int* argc, char** argv,
-                                 void (*HelperPrinterf)() = NULL);
+                                 void (*HelperPrinterf)(FILE*,
+                                                        const char*) = NULL);
 BENCHMARK_EXPORT void Shutdown();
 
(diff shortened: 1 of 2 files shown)


include/benchmark/benchmark.h
```diff
@@ -292,7 +292,8 @@ namespace benchmark {
 class BenchmarkReporter;
 
 BENCHMARK_EXPORT void Initialize(int* argc, char** argv,
-                                 void (*HelperPrinterf)() = NULL);
+                                 void (*HelperPrinterf)(FILE*,
+                                                        const char*) = NULL);
```
yurikhan, line 296: This is a breaking change to the API. If needed, I think we can overload `Initialize` to accept either an argumentless function or a two-argument version. Please say so if this is necessary.

LebedevRI:
Hmm yeah, the current implementation is indeed not what i would expect either :/
I think the proper solution would be to expose the `internal::DefaultHelperPrintf()`, and let the user call it when needed.

yurikhan:
@LebedevRI You mean like this?

LebedevRI:
> @LebedevRI You mean like this?

Yep!

yurikhan:
Not sure what the sanitizer checks are trying to tell me.

PR #1435, opened 2022-07-19
Stop generating the export header and just check it in

dominichamon:
This also removes the dual licensing requirement.

As far as i can tell there is no behavioural change here so I'm not sure why it needed to be generated in the first place...


oontvoo:
Thanks! The windows CI failed? (not clear to me why it's failing tho - the logs dont have much info)

dominichamon:
> Thanks! The windows CI failed? (not clear to me why it's failing tho - the logs dont have much info)

the windows release CI is quite flaky... i don't have a windows machine to test on but i think it's due to the benchmark running at a very different speed to what's expected and the regex test fails.

src/CMakeLists.txt
```diff
@@ -58,6 +55,10 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
   target_link_libraries(benchmark PRIVATE kstat)
 endif()
 
+if (NOT BUILD_SHARED_LIBS)
+  add_definitions(-DBENCHMARK_STATIC_DEFINE)
```
LebedevRI, line 59: These should be `targed_compile_definitions(benchmark PRIVATE -DBENCHMARK_STATIC_DEFINE)`

test/CMakeLists.txt
```diff
@@ -24,6 +24,10 @@ if( NOT uppercase_CMAKE_BUILD_TYPE STREQUAL "DEBUG" )
   endforeach()
 endif()
 
+if (NOT BUILD_SHARED_LIBS)
+  add_definitions(-DBENCHMARK_STATIC_DEFINE)
```
LebedevRI, line 28: Same

dominichamon: not sure about this one as there's multiple targets all of which need the definition.

sergiud:
> [...] I'm not sure why it needed to be generated in the first place...

The reason is separation of concerns. In particular, it does not make sense to check-in artifacts that can be generated. Additionally, CMake users can profit from improvements to `GenerateExportHeader` without relying on benchmark updates.

Now there's an automatically generated header file which is part of the code base that will either never be updated or will be patched manually. This approach does not sound very convincing.

dominichamon:
1. we don't only support cmake so extra effort was required for other build systems, and critically this broke some important integrations
2. it's not automatically generated and shouldn't need to change. if you go through the contents it's a pretty straightforward bit of logic ensuring that the right attributes are set for the right builds. i'll be very surprised if this needs to change outside of specific platform/compiler support.

sergiud:
1. If there are Bazel issues with the export header, I don't see why changing the CMake export is necessary. This PR introduced not only unnecessary complexity and copy-pasted code but also seemingly broke the build in some configurations. See #1450, #1441, #1436. Why not provide a dedicated export header for Bazel builds only instead?
2. I disagree: `GenerateExportHeader` output does not have to be static, e.g., because GNU style attributes could be replaced by C++11 style attributes depending on the compiler's capabilities. There is no good reason why benchmark should decide what functionality outside of library's domain is best for its users.

dominichamon:
1. i changed it because it broke other configurations that couldn't handle the generation of the header.
2. so cmake should make that decision instead? i don't understand your point about deciding functionality: this is about ensuring the symbols are available in the library in different build configurations, not about introducing functionality outside its domain.

sergiud:
1. I don't understand. What configurations did not work in CMake?
2. Of course, this is what's a build system is for: to provide common abstractions.

dominichamon:
not CMake ones.

_if_ we want to bring back generating the header then it needs to work across all the systems we need to support. unfortunately some of these systems are not available to be tested by CI bots and only show up on integration.

i think that means we need to have a static header for all _but_ cmake, which means we may as well also use that static header for cmake to avoid maintaining two systems, one of which might change under our noses if cmake decides to change their implementation.

bstordrup:
I had failing build on a project in Visual Studio 2022 where I used a build of a fork of benchmark (latest version).

As soon as I reverted this pull request, rebuild and deployed benchmark off my local benchmark repository, the previously failing build started to succeed.

Seems like generating the export.h file is vital for the link process in Visual Studio.

bstordrup:
I got my failing build working in a different approach, but it required a change in the steps to build the benchmark library off my local fork.

I reverted all changes, and then I build the library using the following steps:

- `cmake -E chdir "build" cmake -DBENCHMARK_DOWNLOAD_DEPENDENCIES=on -DBUILD_SHARED_LIBS=on -DCMAKE_BUILD_TYPE=Release ../`
- `cmake --build "build" --config Release`
- `cmake --build "build" --config Release --target install`

Using the `-DBUILD_SHARED_LIBS=on` is required to get the `export.h` functioning correct when it comes to the `BENCHMARK_EXPORT` define when you want to use it as a shared lib.

PR #1439, opened 2022-07-21
simplified code

maochongxin:


dominichamon:
i don't understand how this is simpler or better. can you add more details to why you think it's useful?

maochongxin:
> i don't understand how this is simpler or better. can you add more details to why you think it's useful?

Thanks for reply, I think here, it is more appropriate to use the size() method directly than the `size` variable itself, the new `size` variable is more appropriate to represent the size of the new buffer.

dominichamon:
fair enough :)  thanks!

PR #1444, opened 2022-07-25
Remove stray comment and added missing header

oontvoo:
- The export.h is no longer generated, so removed the comment.
- Added export.h to benchmark_main


oontvoo:
@dominichamon

dominichamon (approved):

PR #1448, opened 2022-07-26
Fixed linker issue

oontvoo:
Since GetAbortHandler is no longer inline, we're now getting a missing definition from the linker in our internal build.
Added the weak-attr to resolve this.

Error:
```
ld: error: undefined reference due to --no-allow-shlib-undefined
```

### DIFF
--- a/src/check.h
+++ b/src/check.h
@@ -9,12 +9,18 @@
 #include "internal_macros.h"
 #include "log.h"
 
+#if defined(_WIN32)
+#define WEAK_ATTR __declspec(selectany)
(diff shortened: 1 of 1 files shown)


oontvoo:
@dominichamon

dominichamon:
why would this be weak? we provide an implementation in check.cc that should be in the library...

oontvoo:
Ah, actually never mind - Yeah the real fix should be add that new file to BUILD. (it's not currently - so the linker thought definition was missing!).

PR #1449, opened 2022-07-27
Fixed build issues on window

oontvoo:
- Added missing dlimport/export attributes in function definitions. (They are needed in both decls and defs)
- Removed dlimport/dlexprt attribute in private field. (global_context is not exported anywhere).

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -417,7 +417,7 @@ class Benchmark;
 class BenchmarkImp;
 class BenchmarkFamilies;
 
-BENCHMARK_EXPORT std::map<std::string, std::string>* GetGlobalContext();
+BENCHMARK_EXPORT std::map<std::string, std::string>*& GetGlobalContext();
 
 BENCHMARK_EXPORT
 void UseCharPointer(char const volatile*);
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -417,6 +417,8 @@ class Benchmark;
 class BenchmarkImp;
 class BenchmarkFamilies;
 
+BENCHMARK_EXPORT std::map<std::string, std::string>* GetGlobalContext();
+
 BENCHMARK_EXPORT
 void UseCharPointer(char const volatile*);
 
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "third_party/benchmark/include/benchmark/benchmark.h"
+#include "benchmark/benchmark.h"
 
 #include "benchmark_api_internal.h"
 #include "benchmark_runner.h"
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "benchmark/benchmark.h"
+#include "third_party/benchmark/include/benchmark/benchmark.h"
 
 #include "benchmark_api_internal.h"
 #include "benchmark_runner.h"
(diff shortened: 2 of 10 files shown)


oontvoo:
@dominichamon : more fixes :)

src/benchmark.cc
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "benchmark/benchmark.h"
+#include "third_party/benchmark/include/benchmark/benchmark.h"
```
dominichamon, line 15: this change is a mistake I think

src/benchmark.cc
```diff
@@ -678,9 +678,9 @@ void PrintDefaultHelp() {
 }
 
 void Initialize(int* argc, char** argv, void (*HelperPrintf)()) {
-  internal::HelperPrintf = HelperPrintf;
   internal::ParseCommandLineFlags(argc, argv);
   internal::LogLevel() = FLAGS_v;
+  internal::HelperPrintf = HelperPrintf;
```
dominichamon, line 683: this should be fixed at HEAD

dominichamon (changes requested):

dominichamon:
`<artificial>:(.text+0x2e7e): undefined reference to `benchmark::internal::global_context[abi:cxx11]'` suggests we do need something on `global_context`

oontvoo:
> global_context

Oh, turns out - there are a few places in the code where this variable was forward-declared then used :( 
I'll move the decl to the header then - IMO, it's not very clean to use forward decl like this.

include/benchmark/benchmark.h
```diff
@@ -417,6 +417,8 @@ class Benchmark;
 class BenchmarkImp;
 class BenchmarkFamilies;
 
+BENCHMARK_EXPORT std::map<std::string, std::string>* GetGlobalContext();
```
dominichamon: this was very deliberately _not_ intended to be exposed as part of the public API.

oontvoo: I see that - that's why it's put in the `internal` namespace. Effectively, this is as "private" as the `global_context` field we had previously. Other callers were still able to access the "hidden" field by forward declaring it, so I don't think adding a getter here makes it any worse.

dominichamon: i had hoped it would be more difficult or more obvious that someone was  doing something they shouldn't be, but the `internal` namespace _should_ ideally give that signal too.

one thing i'm curious about in general: what build issues does this fix that aren't showing up on the CI bots and why aren't they showing up on the CI bots?

oontvoo: Previously, when we were getting undefined symbol error for the global_context when it's not marked with the export attribute. But adding the attribute causes "duplicate symbol definition" because of all the forward decl. (On the other hand, we cannot remove the forward decls because the field is used by other files).

So it seemed adding a properly exported getter function here and removing all the forward decls is the cleanest way to fix this. 

I don't know why the CI bots don't report the error. It's possible that the bots only tests the scenario where STATIC is set (ie., the import/export attribute is no-op)?

dominichamon: the solution is fine. i'm just baffled because we have windows-2019 and windows-2022, debug and release, static and shared on the CI bots.

what configuration did you see the build errors in?

oontvoo: (i'll send you the sponge links)

dominichamon: gotcha. for posterity: it's windows+bazel+clang (mostly) which we don't cover in CI.

dominichamon (approved):

sergiud:
In general, benchmark export too many `internal` symbols already. Consider compiling the internal parts as object files that are reused for tests only. This avoids exposing internal symbols in the library which reduces its size. See, for example, how this is solved in [Ceres solver](https://github.com/ceres-solver/ceres-solver/blob/68c53bb39552cd4abfd6381df08638285f7386b3/internal/ceres/CMakeLists.txt#L169-L173).

dominichamon:
> In general, benchmark exports too many `internal` symbols already. Consider compiling the internal parts as object files that are reused for tests only. This avoids exposing internal symbols in the library which reduces its size. See, for example, how this is solved in [Ceres solver](https://github.com/ceres-solver/ceres-solver/blob/68c53bb39552cd4abfd6381df08638285f7386b3/internal/ceres/CMakeLists.txt#L169-L173).

that's really clean.

i'm going to merge this to resolve the current issues but a longer term plan to move things to internal library sounds great.

PR #1452, opened 2022-08-03
fix #1446 by removing the address operator

### LINKED ISSUE #1446: [BUG] Aliased benchmark function does not compile with 1.6.1.
**Describe the bug**
Aliased benchmark function does not compile with 1.6.1.

**System**
  - OS: Linux
  - Compiler and version: clang 13.0.0

**To reproduce**
Creating an alias of a benchmark function and forwarding the alias to the BENCHMARK macro results in a compile error.
```cpp
static auto templated_test_double = templated_test<std::complex<double>>;
BENCHMARK(templated_test_double);
```

Error: "no known conversion from 'void (**)(benchmark::State &)' to 'benchmark::internal::Function *'"

Example: https://godbolt.org/z/3aq4T1as9

**Expected behavior**
Should compile like in version 1.5.0

**Additional context**
Seems to be related to #1262. Is the *&* required in front of `__VA_ARGS__`?
```
  BENCHMARK_PRIVATE_DECLARE(_benchmark_) =                           \
      (::benchmark::internal::RegisterBenchmarkInternal(             \
          new ::benchmark::internal::FunctionBenchmark(#__VA_ARGS__, \
                                                       &__VA_ARGS__)))
```

buster3:
Try to fix #1446 by removing the address operator.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1372,7 +1372,7 @@ class Fixture : public internal::Benchmark {
   BENCHMARK_PRIVATE_DECLARE(_benchmark_) =                           \
       (::benchmark::internal::RegisterBenchmarkInternal(             \
           new ::benchmark::internal::FunctionBenchmark(#__VA_ARGS__, \
-                                                       &__VA_ARGS__)))
+                                                       __VA_ARGS__)))
 #else
 #define BENCHMARK(n)                                     \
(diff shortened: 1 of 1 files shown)


dominichamon:
would you mind also adding the example from https://godbolt.org/z/3aq4T1as9 as a test?

dmah42:
#1538 is this + test

PR #1455, opened 2022-08-03
Explicitly cast int literals to int8_t in tests to silence implicit-conversion warnings

oontvoo:
Error came from:
```
: error: implicit conversion loses integer precision: 'const int' to 'const signed char' [-Werror,-Wimplicit-int-conversion]
```


oontvoo:
@dominichamon

dominichamon (approved):

Issue #1462, opened 2022-08-08
[BUG] Configuration failure with clang and cmake 3.24 on windows


Maetveis:
Benchmark fails to configure when using clang++ (non cl) using CMake 3.24.
  - OS: Windows 11
  - Compiler and version: Clang 15 (gcc compatible driver)
    - Earlier versions most likely fail too
  - Cmake version 3.24.0
  - ninja version: 1.11.0

**To reproduce**
1. sync to commit 974cd5a5c5a78e76ebc50961f4dbf3bf6d4ade4e (main at the time of writing)
2. In a developer terminal try configuring with `cmake -GNinja -DCMAKE_CXX_COMPILER=clang++ -S . -B build`
3. See error
```plaintext
(...)
-- Performing Test HAVE_STD_REGEX -- failed to compile: Change Dir: C:/Users/mesza/Desktop/Develpment/rocRAND/build/googlebenchmark-src/build/CMakeFiles/CMakeTmp

Run Build Command(s):C:/ProgramData/chocolatey/bin/ninja.exe cmTC_be128 && [1/2] Building CXX object CMakeFiles/cmTC_be128.dir/std_regex.cpp.obj
FAILED: CMakeFiles/cmTC_be128.dir/std_regex.cpp.obj
CLANG_~1.EXE   -std=c++11  -Wall  -Wextra  -Wshadow  -Wfloat-equal  -pedantic  -pedantic-errors  -Wshorten-64-to-32  -fstrict-aliasing  -Wno-deprecated-declarations  -Wstrict-aliasing  -Wthread-safety  -g -Xclang -gcodeview -O0 -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -MD -MT CMakeFiles/cmTC_be128.dir/std_regex.cpp.obj -MF CMakeFiles\cmTC_be128.dir\std_regex.cpp.obj.d -o CMakeFiles/cmTC_be128.dir/std_regex.cpp.obj -c C:/Users/mesza/Desktop/Develpment/rocRAND/build/googlebenchmark-src/cmake/std_regex.cpp
In file included from C:/Users/mesza/Desktop/Develpment/rocRAND/build/googlebenchmark-src/cmake/std_regex.cpp:1:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.32.31326\include\regex:12:
[...]

PR #1464, opened 2022-08-08
Try removing attempt to set the C++ standard

### LINKED ISSUE #1460: [BUG] requirement of C++11 is incompatible with Google Test
Google Benchmark insists on being compiled with the standard not exceeding C++11
https://github.com/google/benchmark/blob/974cd5a5c5a78e76ebc50961f4dbf3bf6d4ade4e/CMakeLists.txt#L174

At the same time, Google Benchmark requires Google Test, which is incompatible with C++11 
https://github.com/google/googletest/blob/5b909beeec178f338be997830b6c31a80cda7a93/googletest/include/gtest/internal/gtest-internal.h#L635
```
  typedef ::std::map<std::string, CodeLocation, std::less<>> RegisteredTestsMap;
```
`std::less` without template parameters doesn't compile in C++11.

dmah42:
Fixes #1460 #1462

Let's see what the buildbots make of this.


LebedevRI:
No, this is wrong.

dominichamon:
> No, this is wrong.

it's just a draft because i want to see what happens... the compiler used should be free to set it's standard. we already had a fallback to c++0x so the only issue i think would be if someone tries to use a compiler that doesn't support c++11.

if we want to require that then setting `CMAKE_CXX_STANDARD` and maybe `REQUIRED` is the way to do that.. but i just want to see how necessary it is.

LebedevRI:
There is absolutely no requirement that a supported compiler version actually happens
to have the right (the one you expect, and everyone expects different things) default standard.

dominichamon:
first test complete: OSX falls back too far beyond C++11.

next up: setting the standard.

LebedevRI:
https://cmake.org/cmake/help/latest/policy/CMP0067.html might be relevant.

dominichamon:
> https://cmake.org/cmake/help/latest/policy/CMP0067.html might be relevant.

nice :)

CMakeLists.txt
```diff
@@ -137,6 +138,9 @@ if (BENCHMARK_BUILD_32_BITS)
   add_required_cxx_compiler_flag(-m32)
 endif()
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
LebedevRI, line 142: Probably also want `CMAKE_CXX_EXTENSIONS OFF`

dominichamon: yeah.. i'm also eyeing up the check where we:

```
set(EXTRA_CXX_FLAGS "")
if (WIN32 AND "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
  # Clang on Windows fails to compile the regex feature check under C++11
  set(EXTRA_CXX_FLAGS "-DCMAKE_CXX_STANDARD=14")
endif()
```

and wondering how that will end up interacting with this change.

dominichamon:
> > https://cmake.org/cmake/help/latest/policy/CMP0067.html might be relevant.
> 
> nice :)

cmake 3.8+ only .. ubuntu 16.04 is still on 3.5.1: https://launchpad.net/ubuntu/xenial/+source/cmake

dominichamon:
`try_compile` also doesn't support setting the standard in 3.5.1.

so our older platform support requires explicit `--std` flag setting, i think... and for consistency, it suggests we need to do the same in the rest of the project.

until we can drop older platforms, i don't think we can embrace updated cmake norms...

any other ideas, @LebedevRI ?

LebedevRI:
cmake 3.5.1 sounds *really* ancient.
Even debian old-old-stable has 3.7.
I think 3.13 should be an unambiguous dependency.

dominichamon:
ubuntu bionic (the oldest LTS ubuntu we support) is 3.10.2 (https://packages.ubuntu.com/bionic/cmake) so i think we can get past 3.8.

it might mean dropping the 14.04 and 16.04 bots but they are _really_ old at this point and we do have  a policy on dependencies for a reason.

LebedevRI:
Sounds reasonable?

CMakeLists.txt
```diff
@@ -137,6 +139,10 @@ if (BENCHMARK_BUILD_32_BITS)
...
```
sergiud, line 143: This is not meaningful. MSVC, for instance, does not provide exclusive C++11 support, only C++14 and later: https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170. CMake will probably decay to the newer standard anyway.

dominichamon:
> Sounds reasonable?

image

seems like we might just have had our hand forced...

CMakeLists.txt
powderluv: yes FWIW we need this check for clang on windows.

LebedevRI:
@dmah42 what's the status here?

dmah42:
> @dmah42 what's the status here?

i don't remember where i got to. let me see if i can resurrect it.

dmah42:
it builds and passes tests but I'm not sure  it's the right thing to do.  I would appreciate an ack.

LebedevRI:
> it builds and passes tests but I'm not sure it's the right thing to do. I would appreciate an ack.

Given that we don't want to just happen to build under whatever the standard is default
for the compiler that is used, but specifically with C++11, this is the right thing to do.

LebedevRI (approved):
LG

dmah42:
merged.. now let's see if anyone finds an issue on any unusual platforms with it.

PR #1468, opened 2022-08-10
bump cmake dep and docs

dominichamon:


LebedevRI (approved):
Sounds good to me.
Thanks.

PR #1472, opened 2022-08-20
Enable aarch64 Linux wheel builds, use cibuildwheel action directly

nicholasjng:
This commit enables arm64 Linux wheel builds for Python. It also changes the build procedure on Linux using
cibuildwheel in GitHub Actions. Instead of the more granular, verbose approach that was used until now, we opt for the GitHub Action released by cibuildwheel directly.

We also change the Bazel install procedure in the manylinux Docker container image. Previously, Bazel was installed from an added RHEL repo, since that is the recommended official way of installing Bazel on CentOS platforms.
However, the last successful build available for manylinux2014 has been Bazel 4, which is showing its age with the release of Bazel 6 coming up as of this commit.

After this change, prebuilt Bazel binaries are downloaded using wget directly from the Bazel GitHub release page. Since Bazel is built for both x86 and arm64 on Linux, we immediately gain wheel build support for these architectures. However, since the architecture of the manylinux image is aarch64 instead of arm64, a shell script was added that normalizes aarch64 to arm64, and installs the correct arm64 Bazel binary if necessary.

### DIFF
--- a/.github/install_bazel.sh
+++ b/.github/install_bazel.sh
@@ -1,0 +1,13 @@
+if ! bazel version; then
+  arch=$(uname -m)
+  if [ "$arch" == "aarch64" ]; then
+    arch="arm64"
+  fi
+  echo "Installing wget and downloading $arch Bazel binary from GitHub releases."
+  yum install -y wget
+  wget "https://github.com/bazelbuild/bazel/releases/download/5.2.0/bazel-5.2.0-linux-$arch" -O /usr/local/bin/bazel
+  chmod +x /usr/local/bin/bazel
+else
+  # bazel is installed for the correct architecture
+  exit 0
+fi
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -39,27 +39,23 @@ jobs:
       - name: Check out Google Benchmark
         uses: actions/checkout@v3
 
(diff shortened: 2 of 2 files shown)


nicholasjng:
As always in PRs like this, a successful wheel building run with this change can be found here:

https://github.com/nicholasjng/benchmark/actions/runs/2895922501

dominichamon (approved):

.github/install_bazel.sh
```diff
@@ -0,0 +1,13 @@
+if ! bazel version; then
```
dominichamon, line 1: https://github.com/google/benchmark/blob/main/.github/workflows/bazel.yml assumes it's available and it seems to be.

why do we need this?

nicholasjng: Linux wheel build commands run inside a Docker container, which does not have access to the runner's installed tools.

nicholasjng: As for why the script checks for the install even if it is never available on vanilla manylinux images: Other emulated Linux architectures (ppc64le etc.) do not have Bazel binaries ready for download, and would have to be bootstrapped (generally at image build time).
If we wanted to build ppc, we would probably use a custom manylinux-based Docker image with Bazel available - and the install check would then decay gracefully into an immediate zero exit.

PR #1481, opened 2022-09-03
Discuss sources of variance in the user guide

matta:
It took me a long time to figure some of this stuff out, so it would be great to include it in the guide, assuming it is correct enough!

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1301,8 +1301,15 @@ Linux workstation are:
 
 1. Use the performance governer as [discussed
 above](#disabling-cpu-frequency-scaling).
+1. Disable processor boosting by:
+   ```sh
+   echo 0 | sudo tee /sys/devices/system/cpu/cpufreq/boost
+   ```
+   See the Linux kernel's
+   [boost.txt](https://www.kernel.org/doc/Documentation/cpu-freq/boost.txt)
+   for more information.
 2. Set the benchmark program's task affinity to a fixed cpu.  For example:
-   ```bash
+   ```sh
    taskset -c 0 ./mybenchmark
    ```
 3. Disabling Hyperthreading/SMT.  This can be done in the Bios or using the
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -62,7 +62,7 @@
 
 [Disabling CPU Frequency Scaling](#disabling-cpu-frequency-scaling)
 
-[Reducing Variance in Benchmarks](#reducing-variance)
+[Reducing Variance in Benchmarks](reducing_variance.md)
 
 <a name="output-formats" />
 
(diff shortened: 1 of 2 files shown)


docs/user_guide.md
```diff
@@ -1266,12 +1269,54 @@ cpupower frequency-info -o proc
...
-good place to start looking for options.
+<a name="reducing-variance" />
+
+## Reducing Variance in Benchmarks
+
+The Linux CPU frequency governor [discussed
+above](#disabling-cpu-frequency-scaling) is not the only source of noise in
+benchmarks.  Some, but not all, of the sources of variance include:
+
+1. On multi-core machines not all CPUs run the same speed, so running a
```
LebedevRI, line 1280: Multi-socket?

matta: Here I was thinking about things like TurboBoost, which is not limited to multi-socket.  Also, on my AMD Ryzen 9 5900X at least one CPU always seems to be running below the usual amount even though the performance governor is being used. Could this be thermal throttling? I'm not sure.

Then there are the various "boost" modes of Intel and AMD chips. Which reminds me, disabling that reduced variance on my system significantly, so I now mention it.

```sh
% cat /proc/cpuinfo | grep 'cpu MHz'                                                                                                                              /sys/devices/system/cpu/cpu0
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3442.683
cpu MHz		: 3700.000
cpu MHz		: 3402.804
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
cpu MHz		: 3700.000
```

LebedevRI: Ok, then this should be something like `On multi-core machines not all CPUs/CPU cores/CPU threads run the same speed<...>`

dominichamon:
could you pull this out of the user guide into its own document? i think it's important enough to get that but also isn't something someone who just wants to know how to use/run the thing needs to know right away.

docs/user_guide.md
matta: Ok, went with that wording (though in the same commit I moved the content to a new document).

matta:
> could you pull this out of the user guide into its own document? i think it's important enough to get that but also isn't something someone who just wants to know how to use/run the thing needs to know right away.

Done. I also moved the performance governor discussion to the new document, but a small forwarding paragraph in the user guide in case people have links there, or find the specific error message issued by the benchmark code there.

matta:
Oopse, reopening.  I'm not very familiar with github and didn't realize what "close with comment" did.  ;-)

docs/user_guide.md
LebedevRI: Thanks.

dominichamon:
this is great. i think we'll also move the "random interleaving" into this reducing variance doc too as it is meant precisely for that.

let me know if you want to do that too or if not i'll just merge this.

dominichamon (approved):

matta:
> let me know if you want to do that too or if not i'll just merge this.

My preference is to merge it now and then make further changes.

dominichamon:
thank you!

PR #1485, opened 2022-09-08
Support --benchmarks_filter in the compare.py 'benchmarks' command

### LINKED ISSUE #1484: [FR] Support compare.py benchmarks a.json b.json --benchmark_filter=blah

I've got an a.json and a b.json that took a while to collect due to lots of benchmarks and lots of repetitions.  I was surprised that

```
compare.py benchmarks a.json b.json --benchmark_filter=blah
```

Ignored `--benchmarks_filter=` and always displayed all the results.

The "benchmarks" sub-command honors `--benchmarks_filter=` and reports as if only those benchmarks were run.

Post processing the benchmark .json before running compare.py is an inconvenient workaround.

matta:
Previously compare.py ignored the --benchmarks_filter argument when loading JSON.  This defeated any workflow when a single run of the benchmark was run, followed by multiple "subset reports" run against it with the 'benchmarks' command.

Concretely this came up with the simple case:

 compare.py benchmarks a.json b.json --benchmarks_filter=BM_Example

This has no practical impact on the 'filters' and
'benchmarkfiltered' comand, which do their thing at a later stage.

Fixes #1484

### DIFF
--- a/tools/compare.py
+++ b/tools/compare.py
@@ -9,25 +9,28 @@ import argparse
 from argparse import ArgumentParser
 import json
 import sys
+import os
 import gbench
 from gbench import util, report
-from gbench.util import *
 
 
 def check_inputs(in1, in2, flags):
     """
     Perform checking on the user provided inputs and diagnose any abnormalities
     """
-    in1_kind, in1_err = classify_input_file(in1)
-    in2_kind, in2_err = classify_input_file(in2)
-    output_file = find_benchmark_flag('--benchmark_out=', flags)
-    output_type = find_benchmark_flag('--benchmark_out_format=', flags)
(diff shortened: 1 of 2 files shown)


LebedevRI:
I'm pretty sure this is `benchmarksfiltered` mode of `compare.py` already.

matta:
Thanks, I've confirmed that invocations like this work:

```
compare.py benchmarksfiltered ./results/clang.json '/10$' ./results/gcc.json '/10$'
```

LebedevRI:
Perfect, thank you for checking.

matta:
Yeah, so I’ll pass each json and the same Reyes twice? I can live with that!

Issue #1487, opened 2022-09-10
[windows] Terminal background colour


gvanem:
Hello folks, my first issue.

The function `ColorPrintf()` ass-u-mes that all users of this library have terminals with a black background colour.
I.e. using it in [Microsoft's STL](https://github.com/microsoft/STL.git) library and the `benchmark-std_copy.exe` program, the output like like this:

![STL-1](https://user-images.githubusercontent.com/945271/189484264-1b7ae7b6-8003-4ac8-ae32-0010261bd4c3.png)

Not very nice IMHO.

With my blue terminal background I'd like it to look like:
![STL-2](https://user-images.githubusercontent.com/945271/189484329-e720fd3b-138c-4b1f-a3e6-c212b473b5ab.png)

which I fixed with:
```diff
--- a/src/colorprint.cc  2022-09-10 14:58:12
+++ b/src/colorprint.cc 2022-09-10 14:54:32
@@ -136,13 +136,16 @@
   CONSOLE_SCREEN_BUFFER_INFO buffer_info;
   GetConsoleScreenBufferInfo(stdout_handle, &buffer_info);
   const WORD old_color_attrs = buffer_info.wAttributes;
+  WORD  new_attr;

   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
   // printed but has not yet reached the console.
   fflush(stdout);
+  new_attr = (buffer_info.wAttributes & ~7);
+  new_attr &= ~8;    // Since 'wAttributes' could have been hi-intensity at startup.
   SetConsoleTextAttribute(stdout_handle,
-                          GetPlatformColorCode(color) | FOREGROUND_INTENSITY);
+                          new_attr | GetPlatformColorCode(color) | FOREGROUND_INTENSITY);
   vprintf(fmt, args);

   fflush(stdout);
```

PS. I use [JPsoft's](https://jpsoft.com) 4NT as the shell.

**System**
  - Win-10 
  - Compiler and version: MSVC ver. 19.34.31721 for x64

dominichamon:
could you create a pull request for this?

gvanem:
I'm not very familiar with git pull request. But my attempt: https://github.com/google/benchmark/pull/1488

imrichardcole:
Taking a look

imrichardcole:
To confirm why this is happening (PR incoming) the issue can be captured like so:

[...]

PR #1488, opened 2022-09-12
[windows] Fix for background colours besides black

gvanem:
Fix for issue https://github.com/google/benchmark/issues/1487.

### DIFF
--- a/src/colorprint.cc
+++ b/src/colorprint.cc
@@ -136,13 +136,16 @@ void ColorPrintf(std::ostream& out, LogColor color, const char* fmt,
   CONSOLE_SCREEN_BUFFER_INFO buffer_info;
   GetConsoleScreenBufferInfo(stdout_handle, &buffer_info);
   const WORD old_color_attrs = buffer_info.wAttributes;
-
+  WORD  new_attr;
+  
   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
   // printed but has not yet reached the console.
   fflush(stdout);
+  new_attr = (buffer_info.wAttributes & ~7);
+  new_attr &= ~8;    // Since 'wAttributes' could have been hi-intensity at startup.
   SetConsoleTextAttribute(stdout_handle,
-                          GetPlatformColorCode(color) | FOREGROUND_INTENSITY);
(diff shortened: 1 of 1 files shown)


src/colorprint.cc
```diff
@@ -136,13 +136,16 @@ void ColorPrintf(std::ostream& out, LogColor color, const char* fmt,
...
   const WORD old_color_attrs = buffer_info.wAttributes;
-
+  WORD  new_attr;
+  
   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
   // printed but has not yet reached the console.
   fflush(stdout);
+  new_attr = (buffer_info.wAttributes & ~7);
+  new_attr &= ~8;    // Since 'wAttributes' could have been hi-intensity at startup.
```
LebedevRI, line 146: What are 7 and 8 here? Can we avoid magic numbers?

LebedevRI: Is https://docs.microsoft.com/en-us/windows/console/console-screen-buffers#character-attributes lying and there are no macros for these magic numbers?

gvanem: I give up.

dominichamon: the comment suggests there's a bug here. `BACKGROUND_INTENSITY` is 0x0080 i think. `FOREGROUND_INTENSITY` is 0x0008.

same with the masking of ~7 if you're trying to mask off the backgrounds.. not sure how this works (i am not a windows dev though).

LebedevRI: Let me rephraze: can you not just replace `8` with `BACKGROUND_INTENSITY` and `7` with `FOREGROUND_INTENSITY`?

LebedevRI (changes requested):
(as per inline comment)

src/colorprint.cc
LebedevRI, line 146: @gvanem To clarify/reiterate, what happens if you literally apply the following change:
```suggestion
  new_attr = (buffer_info.wAttributes & ~FOREGROUND_INTENSITY);
  new_attr &= ~BACKGROUND_INTENSITY;    // Since 'wAttributes' could have been hi-intensity at startup.
```
Does it not compile?
The problem is that no one will have a clue what those `7` and `8` really meant,
and the docs imply that there are named macros, so we should use them.

PR #1489, opened 2022-09-13
Bump pybind11 version to enable Python 3.11 wheel builds

### LINKED ISSUE #1476: [BUG] Python 3.11 wheel builds fail due to outdated pybind11
**Describe the bug**
During work on my latest wheel-related PR #1472 I tried building wheels for the upcoming release of Python 3.11 (which is now supported by cibuildwheel). This fails due to a change in Python's C API.

For the GitHub Actions log, check here: https://github.com/nicholasjng/benchmark/runs/7925860405

The issue could be as simple as bumping pybind11 to the latest available version that supports Python 3.11, but this might break legacy wheel builds (basically, for any older Python version that does not have the latest pybind wheels available).

nicholasjng:
This commit bumps the pybind11 version to 2.10.0, which is the first pybind version coming with Python 3.11 support. This change is necessary to facilitate wheel builds for Python 3.11 and upward, as changes to Python internals in 3.11 broke compatibility with older pybind11 versions.

Fixes #1476.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Build wheels on ${{ matrix.os }} using cibuildwheel
         uses: pypa/cibuildwheel@v2.9.0
         env:
-          CIBW_BUILD: 'cp37-* cp38-* cp39-* cp310-*'
+          CIBW_BUILD: 'cp37-* cp38-* cp39-* cp310-* cp311-*'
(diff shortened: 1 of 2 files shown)


dominichamon (approved):

nicholasjng:
Successful run: https://github.com/nicholasjng/benchmark/actions/runs/3020551599

dominichamon:
thanks!

PR #1493, opened 2022-09-27
Deprecate constant reference API to DoNotOptimize.

ckennelly:
The compiler assume that a constant reference, even though escaped via asm
volatile, is unchanged.  The const-ref interface is deprecated to discourage
new uses of it, as subtle compiler optimizations (invariant hoisting, etc.) can
occur.

Within microbenchmarks for Abseil's hashtables, BM_FindMiss_Hot
(https://github.com/google/fleetbench/blob/c0eaa90671d6cc99eb065864e74f0175bee24a5d/fleetbench/swissmap/hot_swissmap_benchmark.cc#L48)
has a `const uint32_t key` is passed to to the lookup of a hashtable.
With the `key` marked `const`, LLVM hoists part of the lookup
calculation outside of the loop.

With the `const` removed, this hoisting does not occur.


LebedevRI:
I think this missing some code examples.

dominichamon:
do any of our tests now trigger this deprecation message?

LebedevRI:
> > I think the description is missing some code examples.
> 
> Done.

Thanks, but i would like to see a more self-contained example on godbolt where said hoisting can be observed.

LebedevRI:
> > > I think the description is missing some code examples.
> > 
> > 
> > Done.
> 
> Thanks, but i would like to see a more self-contained example on godbolt where said hoisting can be observed.

I'm asking because in general const-ness of the pointer does not give any breathing room to the compiler.
So there is something else going on, and i'm not sure if this will be affecting genuine cases too.

dominichamon:
> > > I think the description is missing some code examples.
> > 
> > 
> > Done.
> 
> Thanks, but i would like to see a more self-contained example on godbolt where said hoisting can be observed.

https://gcc.godbolt.org/z/EG4PdT9G6  is about as minimal as i can get it on short order that shows the effect with the current code.

LebedevRI:
> 

Thanks! I haven't tried running that yet, but what is the observation there?
Is it really just the different performance, or is `RandomNonexistent()` call only called once? Or `set.find(key)`?

LebedevRI:
Also, we have a bit of documentation problem. There, `DoNotOptimize()` is being used
to literally completely obscure the value, make it completely opaque from compiler,
yet all of the comments say that it is designed to stop the compiler from eliding said value,
but i've never seen any docs that say that it should also obscure it's value.

In the `const` variant, the `value` variable is taken by value by `DoNotOptimize()`,
so SROA successfully promotes it from stack to register,
and in non-`const` variant that obviously does not happen.

What happens if you do `DoNotOptimize(&key)` with the `const` variant, does it fix it?
Perhaps we need a new set of magic functions `ObscureValue()` instead.

dominichamon:
> > 
> 
> Thanks! I haven't tried running that yet, but what is the observation there? Is it really just the different performance, or is `RandomNonexistent()` call only called once? Or `set.find(key)`?

it's the hoisting of the `RandomNonExistent` as per the original comment on the PR.  it's unclear how much this specifically is a performance issue, but it indicates a broader class of "missed optimizations" due to the implementation of `DoNotOptimize`.

i should note, we've applied this patch internally already as it had significant impact across our internal benchmarks.

dmah42 (approved):

bgaifullin:
This patch breaks using DoNotOptimize with rvalue reference (see #1584)

PR #1495, opened 2022-10-03
Do not depend on unversioned python binary

MatzeB:
Some linux distributions no longer provide `python` binary and require usage of `python3` instead. This changes the scripts here and uses cmake `find_package(Python3` when running python.


LebedevRI:
Is the CMake change nessesary given the rest of the diff?
Put differently, would dropping it still solve your issue?

PR #1496, opened 2022-10-03
Do not depend on unversioned python binary

MatzeB:
Some linux distributions no longer provide `python` binary and require usage of `python3` instead. This changes the scripts here and uses cmake `find_package(Python3` when running python.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,6 +332,7 @@ include_directories(${PROJECT_SOURCE_DIR}/include)
 add_subdirectory(src)
 
 if (BENCHMARK_ENABLE_TESTING)
+  find_package(Python3 3.6 REQUIRED COMPONENTS Interpreter)
   enable_testing()
   if (BENCHMARK_ENABLE_GTEST_TESTS AND
       NOT (TARGET gtest AND TARGET gtest_main AND
--- a/test/AssemblyTests.cmake
+++ b/test/AssemblyTests.cmake
@@ -47,7 +47,7 @@ macro(add_filecheck_test name)
   set_target_properties(${name} PROPERTIES COMPILE_FLAGS "-S ${ASM_TEST_FLAGS}")
   set(ASM_OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/${name}.s")
   add_custom_target(copy_${name} ALL
-      COMMAND ${PROJECT_SOURCE_DIR}/tools/strip_asm.py
+      COMMAND ${Python3_EXECUTABLE} ${PROJECT_SOURCE_DIR}/tools/strip_asm.py
         $<TARGET_OBJECTS:${name}>
         ${ASM_OUTPUT_FILE}
       BYPRODUCTS ${ASM_OUTPUT_FILE})
--- a/tools/compare.py
+++ b/tools/compare.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import unittest
 """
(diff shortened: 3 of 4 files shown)


LebedevRI:
Is the CMake change nessesary given the rest of the diff?
Put differently, would dropping it still solve your issue?

_Originally posted by @LebedevRI in https://github.com/google/benchmark/issues/1495#issuecomment-1265896845_

MatzeB:
> Is the CMake change nessesary given the rest of the diff? Put differently, would dropping it still solve your issue?

Well either the cmake changes or the she-bang changes fixes things for me on CentOS8. I thought going through `find_package(Python3` was just good cmake style, but happy to drop it from the change.

LebedevRI:
I think it's not right to have the shebang **AND** do that CMake dance.
I would guess the shebang thing should be enough,
but then we indeed probably want to check that python interpreter is present at CMake time.
But then if we run those scripts via interpreter, does the shebang help with anything?

Anyways let's just merge as is i guess.

LebedevRI:
(please resolve the CLA issue, though.)

CMakeLists.txt
```diff
@@ -332,6 +332,7 @@ include_directories(${PROJECT_SOURCE_DIR}/include)
 add_subdirectory(src)
 
 if (BENCHMARK_ENABLE_TESTING)
+  find_package(Python3 3.6 REQUIRED COMPONENTS Interpreter)
```
dominichamon, line 335: i read a bit about find_package but couldn't find my answer: i know this isn't an EXACT match but what versions of python3 will this match? should it be a version range?

LebedevRI:
CLA still seems stuck, should have it unstuck?

dominichamon:
i checked manually and @MatzeB hasn't signed the Google CLA as far as i can tell.

MatzeB:
I am told CLA should work now. And when I go to `https://cla.developers.google.com/clas` I see "You are covered by the following corporate agreements:" "

Google Corporate CLA | Facebook, Inc. | Apr 29, 2016 12:42 PDT

[Google Corporate CLA](https://cla.developers.google.com/about/google-corporate)	Facebook, Inc.	Apr 29, 2016 12:42 PDT


Not sure why it's not working, does it still require the old xxx@fb.com mails instead of xxx@meta.com?

LebedevRI:
Let's see if reopening PR helps.

dominichamon:
i just forced a rescan on my end

dominichamon (approved):

joshmcorreia:
This merge forced my CI/CD pipelines that use googlebenchmark to fail because our images don't have Python installed, even though googlebenchmark has never required Python before.

LebedevRI:
> This merge forced my CI/CD pipelines that use googlebenchmark to fail because our images don't have Python installed, even though googlebenchmark has never required Python before.

Looks like you need to specify `-DBENCHMARK_ENABLE_TESTING=OFF`.

joshmcorreia:
> Looks like you need to specify `-DBENCHMARK_ENABLE_TESTING=OFF`.

We want it on, and have had no issues with it in the past...

LebedevRI:
Then i guess perhaps this change should be changed to use `find_program()` instead of requiring `python3-dev` package to be present, that would profile cmake files.

LebedevRI:
> That would be great, this is currently blocking all of my company's pipelines so we can't release anything. If you could revert this change while coming up with an update it would be greatly appreciated.

I would love to, but apparently i can not.

joshmcorreia:
Thanks for the revert!

PR #1497, opened 2022-10-04
Support for QuRT OS (Hexagon RTOS)

Maratyszcza:
This change makes possible to build and use the Google Benchmark framework on Hexagon DSPs

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1534,8 +1534,15 @@ class Fixture : public internal::Benchmark {
 #endif
 
 // Helper macro to create a main routine in a test that runs the benchmarks
+// Note the workaround for Hexagon simulator passing argc != 0, argv = NULL.
 #define BENCHMARK_MAIN()                                                \
   int main(int argc, char** argv) {                                     \
+    char arg0_default[] = "benchmark";                                  \
+    char* args_default = arg0_default;                                  \
+    if (!argv) {                                                        \
+      argc = 1;                                                         \
+      argv = &args_default;                                             \
+    }                                                                   \
     ::benchmark::Initialize(&argc, argv);                               \
     if (::benchmark::ReportUnrecognizedArguments(argc, argv)) return 1; \
     ::benchmark::RunSpecifiedBenchmarks();                              \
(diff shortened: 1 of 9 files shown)


include/benchmark/benchmark.h
```diff
@@ -1534,8 +1534,15 @@ class Fixture : public internal::Benchmark {
 #endif
 
 // Helper macro to create a main routine in a test that runs the benchmarks
+// Note the workaround for Hexagon simulator passing argc != 0, argv = NULL.
 #define BENCHMARK_MAIN()                                                \
   int main(int argc, char** argv) {                                     \
+    char arg0_default[] = "benchmark";                                  \
```
dominichamon, line 1540: can we wrap this in `BENCHMARK_OS_QURT`?  it's non-standard behaviour so i don't really want it leaking out everywhere.

src/timers.cc
```diff
@@ -38,6 +38,9 @@
 #include <mach/mach_port.h>
 #include <mach/thread_act.h>
 #endif
+#if defined(BENCHMARK_OS_QURT)
+#include <qurt/qurt.h>
```
dominichamon, line 42: is there a difference between `<qurt.h>` and `<qurt/qurt.h>`? you use both in this change.

dominichamon (commented):
thanks for the new platform support!  i left a couple of comments as i'm not familiar with the platform.

include/benchmark/benchmark.h
Maratyszcza: This is inside a macro definition, so `#ifdef` can't be used here. I can replicate the whole `#define BENCHMARK_MAIN()` definition inside `#ifdef` as a work-around, but it feels less elegant. What would you prefer?

src/timers.cc
Maratyszcza: Good catch! Due to how include paths are setup, both refer to the same file, but `<qurt.h>` is the canonical one. Fixed.

include/benchmark/benchmark.h
dominichamon: oh you're right, of course. blech. having the whole thing repeated is bad, and having this chunk in here that's only for QURT is bad.

i guess this is the least bad :/

dominichamon (approved):

PR #1502, opened 2022-10-18
Add information for supported Python versions to setup.py

nicholasjng:
Adds qualifiers for Python 3.9-3.11 indicating them being supported in the Python bindings building.

Support for Python 3.6 was removed, so the indicator for Python 3.6 was removed.


dominichamon:
what does "support for Python 3.6 was removed" mean? i don't think we've explicitly set out to not support 3.6 have we?

nicholasjng:
True. This was meant to say that release wheel builds for Python 3.6 have been stopped - the project can still be built for Python 3.6 locally, though. What do you think?

dominichamon:
ah if it only affects the wheels, that's fine.

dominichamon (approved):

PR #1506, opened 2022-11-10
Removed deprecated function

oontvoo:


dominichamon:
should i do a release before we land this?

oontvoo:
Yeah, I think that would make sense given this may break existing code. (Internally, there is only one user of this ... not sure about external users). Thanks!

dominichamon:
v1.7.1 created

dominichamon (approved):

dominichamon:
thanks for the cleanup

PR #1508, opened 2022-11-11
Return option benchmark_perf_counters to help output

### LINKED ISSUE #1507: [BUG] default help does not mention perf_counters
With a simple main.cc` that only contains
```
BENCHMARK_MAIN();
```
the help output does not mention the option `--benchmark_perf_counters`, but the cli-option is accepted and works.

The help text got lost in c5636440403c9790ee17d217ed511ce7467048e4, not sure if by accident or intention.

MaximSmolskiy:
Fix #1507

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -676,7 +676,9 @@ void PrintDefaultHelp() {
           "          [--benchmark_out_format=<json|console|csv>]\n"
           "          [--benchmark_color={auto|true|false}]\n"
           "          [--benchmark_counters_tabular={true|false}]\n"
+#if defined HAVE_LIBPFM
           "          [--benchmark_perf_counters=<counter>,...]\n"
+#endif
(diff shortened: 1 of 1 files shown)


src/benchmark.cc
```diff
@@ -676,6 +676,7 @@ void PrintDefaultHelp() {
           "          [--benchmark_out_format=<json|console|csv>]\n"
           "          [--benchmark_color={auto|true|false}]\n"
           "          [--benchmark_counters_tabular={true|false}]\n"
+          "          [--benchmark_perf_counters=<counter>,...]\n"
```
LebedevRI, line 679: Should be guarded by `HAVE_LIBPFM`

MaximSmolskiy: Added

LebedevRI (approved):
lg

PR #1511, opened 2022-11-14
Include the benchmark's family-name in State

oontvoo:
For compat with internal library, where State::name() returns the benchmark's family name.


dominichamon:
can you add tests please?

nicholasjng:
Would you consider adding this to the Python API as well? That would require one fewer parses of the benchmark name into family and arguments :)

dmah42:
perfcounter failures are unrelated (seems like an apt server is down)

dmah42:
merging this as is, but i agree Python bindings would be nice to have.

PR #1523, opened 2023-01-09
Fix Clang-Tidy warnings related to modernize-use-override

SunBlack:
Currently we have the problem that we have the static code analysis of MSVC enabled, but some warnings cannot be marked as errors, because they come into our code via macros, where the definition of external/system includes does not help. One issue is [C26435](https://learn.microsoft.com/en-us/cpp/code-quality/c26435) (`Use single virtual specification`).

For this MR I checked all warnings related:
- Clang-Tidy: [modernize-use-override](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-override.html)
- MSVC: [C26433](https://learn.microsoft.com/en-us/cpp/code-quality/c26433), [C26435](https://learn.microsoft.com/en-us/cpp/code-quality/c26435) & [C26436](https://learn.microsoft.com/en-us/cpp/code-quality/c26436)

Furthermore I replaced all `BENCHMARK_OVERRIDE` by `override` within all `cc` files (as the readme states that C++11 is required to compile this framework and there was already sometimes used `override` without the macro) - the header files are still using `BENCHMARK_OVERRIDE` to keep C++03 compatibility.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1242,7 +1242,7 @@ class BENCHMARK_EXPORT FunctionBenchmark : public Benchmark {
   FunctionBenchmark(const char* name, Function* func)
       : Benchmark(name), func_(func) {}
 
-  virtual void Run(State& st) BENCHMARK_OVERRIDE;
+  void Run(State& st) BENCHMARK_OVERRIDE;
 
  private:
   Function* func_;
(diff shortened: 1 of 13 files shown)


dmah42 (approved):

dmah42:
thanks!

PR #1525, opened 2023-01-11
Allow specifying number of iterations via --benchmark_min_time.

oontvoo:
Make the flag accept two new suffixes:
 + <integer>x: number of iterations
 + <floag>s: minimum number of seconds.

This matches the internal benchmark API.


oontvoo:
@dmah42 Let me know if we should rename the flag in this patch too? Thanks!

dmah42:
aside from the broken build, this is also a breaking change.  before it's merged we should do a release and then release right after it's merged with clear documentation about the behaviour change.

i haven't checked the code yet.. but does it also accept the old behaviour (ie if the "s" is missing")?

oontvoo:
> but does it also accept the old behaviour (ie if the "s" is missing")?

Yes - if suffixes are omitted, then we'd assume it's the number of seconds for backward compat

src/benchmark_runner.cc
```diff
@@ -140,23 +141,81 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
   manager->NotifyThreadComplete();
 }
 
+std::variant<int, double> ParseBenchMinTime(const std::string& value) {
```
dmah42, line 144: https://github.com/google/benchmark/blob/main/CMakeLists.txt#L141 <-- C++11 is the only requirement (except under MSVC).

no std::variant until C++17.

src/benchmark_runner.cc
```diff
@@ -140,23 +141,81 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+}
+
+double GetMinTime(const benchmark::internal::BenchmarkInstance& b) {
+  if (!IsZero(b.min_time())) return b.min_time();
+
+  std::variant<int, double> iters_or_time =
+      ParseBenchMinTime(FLAGS_benchmark_min_time);
+
+  // If the flag was used to specify number of iters, then return 0 for time.
+  if (iters_or_time.index() == 0) return 0;
```
dmah42, line 184: this check is wrong. someone could pass in "0x" expecting it to be a minimum of 0 iterations.

oontvoo: I'm not sure I understand why it's wrong.

If someone passes "<anything>x", then we interpret the number as the number of iterations.
Whether it's zero or not doesn't make a difference. (ie., doesn't change the fact that they are specifying the number of iterations and not the time limit). 

This check is just checking to see whether the value of the flag is for iterations or time. (`.index() == 0` is to check which member of the variant is active.  0 is for the first member `int` and `1` is for second member `double`)

oontvoo:
adding tests - please hold off on the review. Will ping when ready. Thanks!

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+
+  if (value.empty()) {
+    ret.tag = BenchTimeType::TIME;
+    ret.time = 0.0;
+    return ret;
+  }
+
+  if (value.back() == 'x') {
+    std::string num_iters_str = value.substr(0, value.length() - 1);
+    int num_iters = std::atoi(num_iters_str.c_str());
```
dmah42, line 162: if we have `strtonum` we can get an error, but i'm not sure we do.  at least `strtol` provides some range protection (atoi does no such thing).

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+
+    // std::atoi doesn't provide useful error messages, so we do some
+    // sanity checks.
```
dmah42, line 165: please avoid the word "sanity check"

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+
+  std::string min_time_str;
+  if (value.back() != 's') {
+    BM_VLOG(2) << "Value passed to --benchmark_min_time should have a suffix. "
+                  "Eg., `30s` for 30-seconds.";
+    min_time_str = value;
+  } else {
+    min_time_str = value.substr(0, value.length() - 1);
+  }
+  double min_time = std::atof(min_time_str.c_str());
```
dmah42, line 184: use strtod and check errno?

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+
+  BM_CHECK(min_time > 0 &&
+           !(min_time == 0 && (min_time_str != "0" || min_time_str != "0.0")))
```
dmah42, line 187: this is dangerous. "00.0", "0.", ".0" are all valid representations of 0.0.

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+      << "`. Expected --benchmark_min_time=<float>x.";
+
+  ret.tag = BenchTimeType::TIME;
+  ret.time = min_time;
+
+  return ret;
+}
+
+double GetMinTime(const benchmark::internal::BenchmarkInstance& b) {
+  if (!IsZero(b.min_time())) return b.min_time();
```
dmah42, line 198: so here we accept 0 is an invalid value for time, but we accept it as a valid time on the command line.  maybe we can simplify your error checks by just not allowing 0 as a valid minimum time.

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+  BenchTimeType iters_or_time = ParseBenchMinTime(FLAGS_benchmark_min_time);
+  // If the flag was used to specify number of iters, then return 0 for time.
+  if (iters_or_time.tag == BenchTimeType::ITERS) return 0;
```
dmah42, line 201: use a switch so we don't miss it if someone adds a new tag.

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+
+  return iters_or_time.time;
+}
+
+bool BenchMinTimeHasIters() {
```
dmah42, line 206: why does this go back to the flag if we already parsed it to determine the BenchTimeType?

src/benchmark_runner.cc
```diff
@@ -140,23 +140,102 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+         FLAGS_benchmark_min_time.back() == 'x';
+}
+
+IterationCount GetIters(const benchmark::internal::BenchmarkInstance& b) {
+  if (b.iterations() != 0) return b.iterations();
+
+  // We've already checked that this flag is currently used to pass
+  // iters but do a sanity check anyway.
+  BenchTimeType parsed = ParseBenchMinTime(FLAGS_benchmark_min_time);
+  assert(parsed.tag == BenchTimeType::ITERS);
```
dmah42, line 217: use benchmark CHECK please

src/benchmark_runner.cc
```diff
@@ -173,18 +171,25 @@ BenchTimeType ParseBenchMinTime(const std::string& value) {
     return ret;
   }
 
-  std::string min_time_str;
-  if (value.back() != 's') {
+  const char* time_str = value.c_str();
+  bool has_suffix = value.back() == 's';
+  if (!has_suffix) {
     BM_VLOG(2) << "Value passed to --benchmark_min_time should have a suffix. "
```
dmah42, line 183: if this is something we want users to take action on, we need to log at a higher level than VLOG(2).

src/benchmark_runner.cc
```diff
@@ -194,27 +199,23 @@ BenchTimeType ParseBenchMinTime(const std::string& value) {
   return ret;
 }
 
-double GetMinTime(const benchmark::internal::BenchmarkInstance& b) {
+double GetMinTime(const benchmark::internal::BenchmarkInstance& b,
+                  const BenchTimeType& iters_or_time) {
   if (!IsZero(b.min_time())) return b.min_time();
-  BenchTimeType iters_or_time = ParseBenchMinTime(FLAGS_benchmark_min_time);
   // If the flag was used to specify number of iters, then return 0 for time.
   if (iters_or_time.tag == BenchTimeType::ITERS) return 0;
```
dmah42, line 214: the current default minimum time is 0.5. shouldn't we keep that?

src/benchmark_runner.cc
```diff
@@ -257,6 +261,25 @@ BenchmarkRunner::BenchmarkRunner(
   }
 }
 
+void BenchmarkRunner::UpdateReport(RunResults& run_results) {
```
dmah42: i don't understand why this is necessary. can you add a comment explaining it please?

oontvoo: The BenchmarkName from the RunResult object was created based on the BenchmarkInstance's BenchmarkName.
Hence its min_time could be zero (if the user never calls `Benchmark::MinTime(...))`). 

We want to update that to the value set in the command-line flag, if user specified the flag.

The logic in this is also a bit convoluted:
 - The default value of BenchmarkInstance's min_time is 0.0 
 - The default value of command-line flag min_time is 0.5
 - We update the name to BenchmarkRunner's min_time only if it's neither of the default values. (*)

(*) This is still not very good reasoning. Maybe we want a way to really know whether user overrides the default value or not and only print the value in that case. (If they override the default value with a default, that's fine too - we'd still print it. Eg., if they set --benchmark_min_time=0.0, we should still print it)

WDYT?

src/benchmark_runner.cc
oontvoo: Ah, good point

src/benchmark_runner.cc
dmah42: i think we should be more consistent. whether the benchmark min time was set by the macro or by the commandline flag, it should be captured somewhere that doesn't require posthoc rewriting of history.

ie, what happens if we store `b.min_time` from the macro _or_ the flag? then all the benchmark naming should just work, right?

dmah42: also, do we want to add a suffix for iterations if the user overrode the iterations on the command line?

oontvoo: Ooh, I see - hmm. I've only added this part so that we can test the flag's setting 
(since there's no way to pull the min_time field out of BenchmarkRunner from a test) ...

any suggestion?

Otherwise, maybe can just rewrite the test to test that the flag was parsed correctly?

dmah42: maybe just add the min time to the benchmark runner so we can access it for testing? :)

oontvoo: Done. 

(ended up added a hook to report the config from BenchmarkRunner to the BenchmarkReporter interface, because from a test, we dont have access to BenchmarkRunner objects)

src/benchmark.cc
```diff
@@ -65,12 +65,21 @@ BM_DEFINE_bool(benchmark_list_tests, false);
...
+//
+// For backward compatibility, the `s` suffix may be omitted, in which case,
+// the specified number is interpreted as the number of seconds.
+//
+// For cpu-time based tests, this is the lower bound
 // on the total cpu time used by all threads that make up the test.  For
 // real-time based tests, this is the lower bound on the elapsed time of the
 // benchmark execution, regardless of number of threads.
-BM_DEFINE_double(benchmark_min_time, 0.5);
+BM_DEFINE_string(benchmark_min_time, "0.5s");
```
dmah42, line 82: should this use `kDefaultMinTime`?

oontvoo: Yes. Done!

include/benchmark/benchmark.h
```diff
@@ -280,6 +280,10 @@ BENCHMARK(BM_test)->Unit(benchmark::kMillisecond);
 namespace benchmark {
 class BenchmarkReporter;
 
+// Default number of minimum benchmark running time in seconds.
+const double kDefaultMinTime = 0.5;
+const char kDefaultMinTimeStr[] = "0.5s";
```
dmah42, line 285: i was hoping to not have this declared twice.  is there no way to set the default based on `kDefaultMinTime` directly?

or extract the double default from the string version?

oontvoo: kind of yes :) I've moved the double field to benchmark_runner.cc (where it's used) and parse the string field there.

dmah42 (commented):
@LebedevRI should also look at this

oontvoo:
Hi @LebedevRI, any thought on this? Thanks!

src/benchmark_runner.cc
```diff
@@ -140,23 +145,110 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+    ret.tag = BenchTimeType::ITERS;
+    ret.iters = num_iters;
+    return ret;
+  }
+
+  const char* time_str = value.c_str();
+  bool has_suffix = value.back() == 's';
+  if (!has_suffix) {
+    BM_VLOG(0) << "Value passed to --benchmark_min_time should have a suffix. "
+                  "Eg., `30s` for 30-seconds.";
```
LebedevRI, line 184: This makes it look like other suffixes (`ms`?) are supported

src/benchmark_runner.cc
```diff
@@ -140,23 +145,110 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+  }
+
+  char* p_end;
+  // Reset error before it's changed by strtod.
+  const int old_errno = errno;
+  errno = 0;
+  double min_time = std::strtod(time_str, &p_end);
+  const int new_errno = errno;
+  errno = old_errno;
```
LebedevRI, line 193: Why do we care to restore the `errno`?

src/benchmark_runner.cc
oontvoo: Yes, we could extend the suffixes support later - but for the first iteration, let's keep it simple. (And I'd like to get the interface in place first)

src/benchmark_runner.cc
oontvoo: just thought it's generally good to not pollute errno - can remove this if you'd prefer

oontvoo:
Hi, if there's no further objection, can we please get this rolling? Thanks!

dmah42 (approved):

dmah42:
i'd like @LebedevRI to approve too

src/benchmark_runner.cc
```diff
@@ -140,23 +145,100 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
...
+  }
+
+  char* p_end;
+  // Reset errno before it's changed by strtod.
+  errno = 0;
+  double min_time = std::strtod(time_str, &p_end);
+
+  // After a successfull parse, p_end should point to the suffix 's'
+  // or the end of the string, if the suffix was omitted.
+  BM_CHECK(errno == 0 && p_end != nullptr &&
```
LebedevRI, line 189: This check is present in release builds, right?

test/benchmark_min_time_flag_time_test.cc
```diff
@@ -0,0 +1,90 @@
...
+#include <cstring>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "benchmark/benchmark.h"
+
+// Tests that we can specify the min time with
+// --benchmark_min_time=<NUM> (no suffix needed) OR
+// --benchmark_min_time=<NUM>s
```
LebedevRI, line 14: Is there a negative test for `1ms` / `1z` / `1hs`?

LebedevRI:
Consider this rubber-stamped.

LebedevRI:
(also, what about warm-up option, should it also be affected?)

src/benchmark_runner.cc
chfast: Why not std::stod()?

oontvoo:
> (also, what about warm-up option, should it also be affected?)

No, this will not affect the warm-up option.
(We could expand that flag, if it's useful? l)

src/benchmark_runner.cc
oontvoo: because it didn't have a very good error-reporting interface (see previous iteration of this patch where stod was used)

src/benchmark_runner.cc
dmah42: no, it acts like `assert` and is disabled in `NDEBUG` builds

test/benchmark_min_time_flag_time_test.cc
oontvoo: Yes - just added a set of death tests for these

dmah42:
some valid linker errors: `min_time_parse_gtest.cc:(.text+0x6ba): undefined reference to `benchmark::internal::ParseBenchMinTime(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'`

oontvoo:
> > > some valid linker errors: `min_time_parse_gtest.cc:(.text+0x6ba): undefined reference to `benchmark::internal::ParseBenchMinTime(std::__cxx11::basic_string<char, std::char_traits, std::allocator > const&)'`
> > 
> > 
> > Hmm, any idea why the definition is not seen?
> > (The function is defined in benchmark_runner.cc - and moreover,t he same code built in other configurations .... what's difference about these "shared" runs? thinking )
> 
> Ok - now the tests all PASSED in Debug builds but FAILED in Release builds (due to "failed to die"). I must be missing something here but the `#ifndef NDEBUG` should have prevented the tests from running in Release mode - why are they still running? (and hence failing?)

Ah, found it! `TEST_BENCHMARK_LIBRARY_HAS_NO_ASSERTIONS` is the needed guard

dmah42 (approved):

dmah42:
any more comments here before we rebase and squash and merge?

LebedevRI:
> Consider this rubber-stamped.

dmah42:
wonderful. thank you.

Issue #1521, opened 2022-12-15
[FR] Consider switching to `nanobind` Python bindings


nicholasjng:
**Is your feature request related to a problem? Please describe.**
Not a problem, more an idea: For Python >=3.8 bindings using C++17, the [nanobind](https://github.com/wjakob/nanobind) project (from the same hand as `pybind11`) promises binary size reductions and runtime performance increases (~40% in their benchmarks) over `pybind11` for frequently re-called functions in a loop.

For a benchmark library, runtime performance benefits like these could be interesting - especially since the quoted runtime performance increase for "calling functions repeatedly in a loop" fits the bill exactly in this case.

**Describe the solution you'd like**
Implement the GBM Python bindings in nanobind instead, measure the performance / binding size differences, decide where to go from there after interpreting the results.

**Describe alternatives you've considered**
Staying with the current thing.

**Additional context**
I want to do this as a personal experiment, to learn how to implement C++ Python bindings. If something comes out of it other than the nanobind-specific results (better/any docstrings, type hints etc), they could be added to the existing bindings, too.

Not sure how stable `nanobind` is right now, but since it has `pybind11` as its direct predecessor project, it might already work sufficiently for this scope.

nicholasjng:
So I ported over GBM to nanobind, since the necessary tools used are now fully available in `nanobind`. The results (all on macOS 12.6.2, Apple M1 arm64, Python 3.10.9):

## Binding sizes
Measured as wheel size on disk.

- using `pybind11`: 252KB
- using `nanobind`: 189KB **(-25%)**

## Compile / wheel build times
Measured with `time python setup.py bdist_wheel`, with a preceding `bazel clean` (so that build caches are removed, but dependency downloads stay, and thus do not influence the time).

pybind11: python setup.py bdist_wheel  0.18s user 0.06s system 3% cpu 7.145 total
nanobind: python setup.py bdist_wheel   0.17s user 0.06s system 4% cpu 5.063 total **(-29%)**

## Example script timings / iterations
Created with `python bindings/python/google_benchmark/example.py`.

With `pybind11`:

```
➜ python bindings/python/google_benchmark/example.py
Unable to determine clock rate from sysctl: hw.cpufrequency: No such file or directory
This does not affect benchmark measurements, only the metadata output.
2023-01-10T20:41:19+01:00
Running bindings/python/google_benchmark/example.py
Run on (10 X 24.1213 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
[...]

PR #1526, opened 2023-01-11
Switch bindings implementation to `nanobind`

nicholasjng:
Title is self-explanatory. This uses `nanobind` to build Python bindings for Google Benchmark instead of `pybind11`, with the following consequences:

- Python 3.7 support is dropped, since nanobind needs PEP590 vectorcalls, first added in Python 3.8.
- `nanobind` is checked in as a bazel dep, currently at HEAD instead of a tag, because some features used to generate the GBM binding code have not made it into a release yet.
- The minimum MacOS version was bumped up to 10.14 (Mojave), to get full C++17 support.

For performance considerations and changes in contrast to the current implementation, see the discussion #1521.

(For reviewers: I am not sure of the options I gave in the nanobind BUILD file - I looked at what compile commands CMake generated in release mode for nanobind and went on from there.)

The bindings now also contain the state name from #1511.


bindings/python/nanobind.BUILD
```diff
@@ -0,0 +1,41 @@
...
+        "src/tensor.cpp",
+        "src/trampoline.cpp",
+    ],
+    copts = [
+        "-fexceptions",
+        "-std=c++17", # requires C++17
+        "-Os",  # size optimization option for nanobind
+        "-flto", # enable LTO
+    ],
+    linkopts = ["-undefined suppress", "-flat_namespace", "-flto"],
```
dmah42, line 37: seems like -flat_namespace is not recognised by gcc (see test failure)

dmah42 (dismissed):

nicholasjng:
Hmm, I guess I need to pass the `copts` to nanobind via a `select` as well. The `cxxopt` for C++17 is different as well on Windows, I will get to it asap.

nicholasjng:
I fixed the MSVC breakage, which was due to (my) user error. Should be good to go now, binding test matrix is all green, other failures are due a downed apt mirror presumably.

If you want, I can restructure the PR and split it up (e.g. carving out the CI parts), to make it easier to revert to `pybind11` in the event, without touching the other changes here. Let me know what you think.

dmah42 (approved):

PR #1528, opened 2023-01-13
Fix Clang-Tidy warnings readability-else-after-return

SunBlack:
See [readability-else-after-return](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html).


dmah42 (approved):

dmah42:
thanks!

PR #1529, opened 2023-01-13
[doc]Add quick instructions on how to enable sans checks

oontvoo:


dmah42 (approved):

dmah42:
is there anywhere in `docs` that should also be updated?

also do you know why they're currently failing? 😬

oontvoo:
> is there anywhere in `docs` that should also be updated?

I'd looked under `docs/` but not seen anything related ... (If it's helpful, we could also add a new file entirely for instructions on building and testing benchmarks libraries (aimed for benchmarks contributors rather than users?)

> also do you know why they're currently failing? grimacing

Sadly, no - I couldn't reproduce the failures locally ...

dmah42:
> > is there anywhere in `docs` that should also be updated?
> 
> I'd looked under `docs/` but not seen anything related ... (If it's helpful, we could also add a new file entirely for instructions on building and testing benchmarks libraries (aimed for benchmarks contributors rather than users?)

maybe https://google.github.io/benchmark/platform_specific_build_instructions.html

https://google.github.io/benchmark/user_guide.html should have some useful tips for being a developer but it is _massive_.

> 
> > also do you know why they're currently failing? grimacing
> 
> Sadly, no - I couldn't reproduce the failures locally ...

hm.

PR #1532, opened 2023-01-18
Introduce profiler interface to benchmark framework.

minglotus-6:
- Users of benchmark library provides implementations (e.g., defines how counters are collected and reported) and registers profiler once.
- The profiler is init'ed once per 'RunSpecifiedBenchmarks' and injected in a way to skip capturing states for 'Benchmark::Setup' or 'Benchmark::TearDown'. The profiler calls 'Finalize' to do post-processing.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -391,6 +391,30 @@ class MemoryManager {
 BENCHMARK_EXPORT
 void RegisterMemoryManager(MemoryManager* memory_manager);
 
+class BENCHMARK_EXPORT Profiler {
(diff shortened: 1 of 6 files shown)


LebedevRI:
This seems to miss some motivational works. Why is that something we want? What problem does it solve?

minglotus-6:
> > Going to close and send another PR.
> 
> That's not how PR's work :)

Hi Roman, 
    Sorry I missed your comment about motivation before closing it :( Going to follow up on https://github.com/google/benchmark/pull/1533

PR #1535, opened 2023-02-02
Modernize setup.py, extend Python bindings CI

nicholasjng:
distutils is deprecated and will be removed in Python 3.12, so this commit modernizes the Python bindings `setup.py` file in order to future-proof the code.

On top of this, type hints were added for all of the convenience functions to make static type checking adoption easier in the future, if desired.

A context manager was added to temporarily write the Python include path to the Bazel WORKSPACE file - but unlike previously, the WORKSPACE file is reverted to its previous state after the build to not produce changes on every rebuild.

Lastly, the Python bindings test matrix was extended to all major platforms to create a more complete picture of the current state of the bindings, especially with regards to upcoming wheel builds.

### DIFF
--- a/.github/workflows/test_bindings.yml
+++ b/.github/workflows/test_bindings.yml
@@ -8,17 +8,21 @@ on:
 
 jobs:
   python_bindings:
-    runs-on: ubuntu-latest
+    name: Test GBM Python bindings on ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


dmah42 (approved):

dmah42:
thanks

PR #1537, opened 2023-02-06
remove best effort support for ubuntu 18.04

dmah42:
as of 2022-01-12, github removed support for ubuntu-18.04.  as such, we are removing it from github actions and no longer will support it actively.

it was already "best effort" but time marches on.

### DIFF
--- a/.github/workflows/build-and-test-perfcounters.yml
+++ b/.github/workflows/build-and-test-perfcounters.yml
@@ -14,20 +14,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # ubuntu-18.04 is deprecated but included for best-effort
-        os: [ubuntu-22.04, ubuntu-20.04, ubuntu-18.04]
+        os: [ubuntu-22.04, ubuntu-20.04]
         build_type: ['Release', 'Debug']
     steps:
     - uses: actions/checkout@v2
 
     - name: install libpfm
       run: sudo apt -y install libpfm4-dev
-
-    - name: setup cmake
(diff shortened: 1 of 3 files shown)


LebedevRI:
While i don't have an opinion on what we should do regarding supporting that particular distro,
i feel like i should point out that it is not relevant what linux distributions are natively supported
by github runners, because there's such thing as a docker, with idea being:
 use the freshest distro in `runs-on:` field, and then just pick the distro you want
(from whatever the source, e.g. https://hub.docker.com/_/ubuntu), and specify it in `container: image:`
That is what i have always been doing, at least.

See https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

dmah42:
> While i don't have an opinion on what we should do regarding supporting that particular distro, i feel like i should point out that it is not relevant what linux distributions are natively supported by github runners, because there's such thing as a docker, with idea being: use the freshest distro in `runs-on:` field, and then just pick the distro you want (from whatever the source, e.g. https://hub.docker.com/_/ubuntu), and specify it in `container: image:` That is what i have always been doing, at least.
> 
> See https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

yeah, we could. i'm ok with dropping support for this particular platform though.

dmah42:
maybe we can use containers for the sanitizer builds... or at least a shared cache so we don't keep rebuilding the world for each one.

dmah42:
sanitizer builds are timing out.. that's next on my list.

PR #1539, opened 2023-02-07
replace complexity with simplicity for faster libc++ setup

dmah42:
instead of building libc++ from source, let's just install the latest version.  we may get lower coverage but it should save ~50m off our github sanitizer actions.


.github/workflows/sanitizer.yml
```diff
@@ -45,6 +45,11 @@ jobs:
         echo "EXTRA_FLAGS=-g -O2 -fno-omit-frame-pointer -fsanitize=thread -fno-sanitize-recover=all" >> $GITHUB_ENV
         echo "LIBCXX_SANITIZER=Thread" >> $GITHUB_ENV
 
+    - name: fine-tune asan options
+      # in clang+asan we get an error from std::regex. ignore it.
+      if: matrix.sanitizer == 'asan' && matrix.compiler == 'clang'
+        echo "ASAN_OPTIONS=alloc_dealloc_mismatch=0" >> $GITHUB_ENV
```
nicholasjng, line 51: You are probably missing a `run:` here, judging from the GitHub Actions log.

nicholasjng:
Looks great, very nice speedups! You could probably bump `actions/checkout` to latest (v4 IIRC) in the sanitizer action to get rid of the warnings.

Issue #1509, opened 2022-11-11
[FR] Consider adopting the same support policies as other google C++ projects


coryan:
Other google C++ projects are adopting (or have adopted) the support policies outlined at

https://opensource.google/documentation/policies/cplusplus-support

(see this [matrix](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md) for details).

Please consider adopting the same policies in this project.  For example, this project requires CMake >= 3.16.3, which makes it hard to use from other projects.

devjgm:
+1 I'd like to see google/benchmark adopt these same policies

dominichamon:
i don't think we're far off.

https://github.com/google/benchmark/blob/main/docs/dependencies.md has justifications for the cmake support, which i don't think are unreasonable given how widely available 3.16 is at this point.

if someone wants to go through the matrix and see if we're aligned or not, i'm open to that :)

devjgm:
> https://github.com/google/benchmark/blob/main/docs/dependencies.md has justifications for the cmake support, which i don't think are unreasonable given how widely available 3.16 is at this point.

I think that documented policy should be revisited in light of [the C++ support policy][cpp-policy]. I think this feature request is for Google Benchmark to forego its own support policy and instead adopt the common [the C++ support policy][cpp-policy], unless its existing policy offered even more support, which it doesn't.

[cpp-policy]: https://opensource.google/documentation/policies/cplusplus-support

dmah42:
in #1550 i'm creating a CI bot to check cmake 3.10.0 support and i've hit a potential snag: cmake 3.10.0 only supports MSVC generators for VS 15 (2017) and lower.  we currently support VS 16 (2019) and higher.

it may be that we can support these older VS versions but i'm not a windows developer so i don't know for sure.  i will continue to experiment, but this may be a fly in the cmake ointment.

dmah42:
per a comment on the PR from @coryan : we can assume per policy that we're supporting the newer Windows environments only and they have newer cmake available.

dmah42:
i think we're there.  the only nuance we have is that we don't require modern flavours of C++ to use the library, only to build it.  so the header should remain free of any language features that require C++14 or newer (iirc).

so i think we can close this as complete.  feel free to reopen it if i've missed something.

PR #1544, opened 2023-02-16
cleanup: support CMake >= 3.10

coryan:
This aligns the project with the CMake support policies in:

https://opensource.google/documentation/policies/cplusplus-support

I also simplied the management of CMake policies. Most of the overriden policies (anything <= CMP0067) are enabled by default when you require CMake >= 3.10.  But it is easier to just declare that you will accept newer policies when they are available using the `...3.22` notation.

Part of the work for #1509

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,18 +1,5 @@
-cmake_minimum_required (VERSION 3.16.3)
-
-foreach(p
-    CMP0048 # OK to clear PROJECT_VERSION on project()
-    CMP0054 # CMake 3.1
-    CMP0056 # export EXE_LINKER_FLAGS to try_run
-    CMP0057 # Support no if() IN_LIST operator
-    CMP0063 # Honor visibility properties for all targets
-    CMP0067 # Honor language standard in try_compile() source file signature
-    CMP0077 # Allow option() overrides in importing projects
-    )
-  if(POLICY ${p})
-    cmake_policy(SET ${p} NEW)
-  endif()
-endforeach()
+# Require CMake 3.10. If available, use the policies up to CMake 3.22.
+cmake_minimum_required (VERSION 3.10...3.22)
 
 project (benchmark VERSION 1.7.1 LANGUAGES CXX)
 
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -1,7 +1,7 @@
 # Build tool dependency policy
 
-We follow the [Foundational C++ support policy] for our build tools. In
-particular the "Build Systems" [section].
+We follow the [Foundational C++ support policy](https://opensource.google/documentation/policies/cplusplus-support) for our build tools. In
+particular the ["Build Systems" section](https://opensource.google/documentation/policies/cplusplus-support#build-systems).
 
 ## CMake
 
@@ -11,6 +11,3 @@ distributions include newer versions, for example:
 * Ubuntu 20.04 provides CMake 3.16.3
 * Debian 11.4 provides CMake 3.18.4
 * Ubuntu 22.04 provides CMake 3.22.1
-
-[Foundational C++ support policy]: https://opensource.google/documentation/policies/cplusplus-support
-[Foundational C++ support policy]: https://opensource.google/documentation/policies/cplusplus-support#build-systems
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -1,18 +1,16 @@
 # Build tool dependency policy
 
-To ensure the broadest compatibility when building the benchmark library, but
-still allow forward progress, we require any build tooling to be available for:
+We follow the [Foundational C++ support policy] for our build tools. In
+particular the "Build Systems" [section].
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,18 +1,5 @@
...
-    CMP0063 # Honor visibility properties for all targets
-    CMP0067 # Honor language standard in try_compile() source file signature
-    CMP0077 # Allow option() overrides in importing projects
-    )
-  if(POLICY ${p})
-    cmake_policy(SET ${p} NEW)
-  endif()
-endforeach()
+# Require CMake 3.10. If available, use the policies up to CMake 3.22.
+cmake_minimum_required (VERSION 3.10...3.22)
```
dmah42, line 2: https://google.github.io/benchmark/dependencies.html is pretty clear about why we support 3.16.3

coryan: I guess I am asking you to reconsider.  A number of other Google projects are perfectly capable of making forward progress (AFAICT that is the *why* in that document) while supporting CMake >= 3.10.  And it is not like this project was using any CMake > 3.16 features.  In fact, it seems this project uses more CMake < 3.0 idioms than CMake >= 3 idioms.

In any case, this is your project.  If you cannot support CMake 3.10, we will just have to keep using the older versions.

devjgm: @dmah42 I also think this policy may be worth reconsidering. Abseil, Protobuf, and gRPC (among others) have published a [C++ support policy](https://opensource.google/documentation/policies/cplusplus-support) that allows us to move forward while also allowing each project to remain compatible with eachother. More and more Google projects are adopting these policies. I think it would be good if Google benchmark considering following suite.

coryan:
Closing for now.  If you change your mind let me know and I will gladly send the PR again.

CMakeLists.txt
dmah42: i didn't intend this to be the end of the conversation.  it was a little brusque (sorry) but my intent was to ensure that we understand the implications of opening up to older cmake versions.

i remember we had to go past 3.7 to support some features at some point, but we may have just skipped right over 3.10.

devjgm: Thanks @dmah42 ! I'll let @coryan follow up w/ the PR, but I wanted to also point out that at the top of https://opensource.google/documentation/policies/cplusplus-support there's a link to [this table](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md), which lists the specific versions of things that we support. It makes it a bit easier to see which versions the policies implies. In this case, we can see that the table is showing CMake >= 3.10, which is where Carlos was getting these versions.

dmah42:
seems like it works so sure.. can we update the dependencies.md doc to point to the C++ support policy (or the README to point to it instead of the dependencies doc, if it's completely subsumed)?

CMakeLists.txt
dmah42: yep, i saw. i think we might want to just point at this rather than track our own dependencies, as long as we're true to it (as in my other comment).

coryan:
I looked into the CI failure, and I am not sure how it could be related to this change.  But do let me know if I should look further or you have suggestions for fixes.

coryan:
Also, if you want any other changes in the PR do let me know too.

EricWF:
I think this change makes sense, and I'm OK going appropriately backwards in our support here *because* the support policy of these other projects should be our support policy as well. 

Could this PR add a pointer to the support policy of the other projects mentioned in the discussion, preferably Googletests', as well as a statement affirming that we plan to follow said policy as well.

coryan:
@EricWF I do not believe googletest has changed their README file to state what is their support policy (whether that means they have not adopted the OSS Foundations policy or not, I could not say).  I think [Abseil has](https://github.com/abseil/abseil-cpp#support), and we both know those teams are closely related.  Is that enough?

If so, I am happy to edit [/docs/dependencies.md](https://github.com/google/benchmark/blob/main/docs/dependencies.md), or any other `.md` file here that you think is appropriate.

dmah42:
> @EricWF I do not believe googletest has changed their README file to state what is their support policy (whether that means they have not adopted the OSS Foundations policy or not, I could not say). I think [Abseil has](https://github.com/abseil/abseil-cpp#support), and we both know those teams are closely related. Is that enough?
> 
> If so, I am happy to edit [/docs/dependencies.md](https://github.com/google/benchmark/blob/main/docs/dependencies.md), or any other `.md` file here that you think is appropriate.

yes please update docs/dependencies.md to point to the policy for CMake. thanks.

the CI failure is not you.

coryan:
> yes please update docs/dependencies.md to point to the policy for CMake. thanks.

Done. And rebased.

dmah42 (approved):

dmah42:
thanks!

LebedevRI:
Is there a CI entry to ensure that cmake-3.10 actually works for us?

PR #1549, opened 2023-02-19
Fix CPU frequency estimation on riscv

dtcxzyw:
I used to run llvm-test-suite on the riscv64 platform, but I found that the CPU frequency given by the benchmark framework was incorrect. Unfortunately, neither the ```cpufreq``` interface nor ```BogoMIPS``` is implemented on the linux-riscv64 platform, it just uses 
```rdcycle``` and ```nanosleep``` to estimate CPU frequency.

However, cycle counters are per-core on RISC-V, and what “sleep” means is also not standardized by the RISC-V specifications.
On my machine, calling ```nanosleep``` doesn't work because the cycle counter doesn't increase during sleeping. Furthermore, it does not prevent thread migration during timing.

Please also refer to https://github.com/riscv/riscv-isa-manual/issues/140 and the RISC-V Instruction Set Manual.

I use the pseudo-random number generator given by ```<random>``` instead of calling ```SleepForMilliseconds``` as the load and temporarily pin the current thread to a core during timing. This patch works fine on my machine.

It has been tested on the following platforms:
+ Linux 5.19.0-1012-generic 13~22.04.1-Ubuntu SMP Thu Jan 12 15:34:31 UTC 2023 riscv64 + gcc 11.3.0 riscv64-linux-gnu + SiFive u74-mc
+ Linux 5.15.0-57-generic 63-Ubuntu SMP Thu Nov 24 13:43:17 UTC 2022 x86_64 + gcc 11.3.0 x86_64-linux-gnu/clang 14.0.0
+ Windows 10 22H2 19045.2604 + MSVC 19.29.30147.0

If it is OK, I can also drop bogoMIPS and just use the estimation to close https://github.com/google/benchmark/issues/312.
Since ```SleepForMilliseconds``` is no longer used, ```sleep.h``` and ```sleep.cc``` can be removed from the framework.

### CODE REVIEW

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -34,6 +34,34 @@ if (HAVE_LIBPFM)
   target_compile_definitions(benchmark PRIVATE -DHAVE_LIBPFM)
 endif()
 
+# pthread affinity, if available
+include(CheckCXXSourceCompiles)
```
dmah42, line 38: could we extract this into a separate file in [cmake/](https://github.com/google/benchmark/tree/main/cmake) ?

src/sysinfo.cc
```diff
@@ -46,6 +46,12 @@
 #if defined(BENCHMARK_OS_QURT)
 #include <qurt.h>
 #endif
+#if defined(BENCHMARK_HAS_PTHREAD_AFFINITY)
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
```
dmah42, line 51: is this necessary?  if it's just for pthread can we revert the definition afterwards please?  i'd like to avoid people accidentally depending on non-standard extensions or non-portable functions.

src/sysinfo.cc
```diff
@@ -544,6 +553,80 @@ int GetNumCPUs() {
...
+      ret = pthread_setaffinity_np(self, sizeof(affinity), &affinity);
+      return ret == 0;
+#elif defined(BENCHMARK_OS_WINDOWS_WIN32)
+      self = GetCurrentThread();
+      DWORD_PTR mask = static_cast<DWORD_PTR>(1) << GetCurrentProcessorNumber();
+      previous_affinity = SetThreadAffinityMask(self, mask);
+      return previous_affinity != 0;
+#else
+      return false;
+#endif
```
dmah42, line 591: `#endif  // BENCHMARK_HAS_PTHREAD_AFFINITY`

or similar please, to make this easier to reason about.

dtcxzyw:
> https://github.com/google/benchmark/actions/runs/4217605263/jobs/7332317281#step:6:24 .. looks like a bit cleverer #defines are needed to avoid the lambda capture that's unnecessary.

It has been resolved. It looks like only clang++ complains about this. By the way, I think clang is actually used for the workflow ```macos-latest.**.**.g++``` because g++ doesn't recognize the option ```-Wunused-lambda-capture``` at all.

src/sysinfo.cc
```diff
@@ -20,6 +20,15 @@
 #include <versionhelpers.h>
 #include <windows.h>
 
+// WINAPI_FAMILY_PARTITION is defined in winapifamily.h
```
dmah42, line 23: why did you move this?  it is only currently used here, but we try to keep all the platform definitions in one place so they're easy to find if we need to change something.

dmah42:
looks good.. just that one thing i'd like you to move back please, then i can merge it :)

src/sysinfo.cc
dtcxzyw: I checked it again on win10 and found that the macro ```BENCHMARK_OS_WINDOWS_WIN32``` is not defined. This is because the original code used the macro ```WINAPI_FAMILY_PARTITION``` before it was defined in ```winapifamily.h```. The alternative is to move it back and include ```winapifamily.h``` before the use.

dtcxzyw:
> looks good.. just that one thing i'd like you to move back please, then check CI, then i can merge it :)

I think ```sleep.h``` and ```sleep.cc``` should be removed before merging since the function ```SleepForMilliseconds``` is no longer used.

dmah42:
> > looks good.. just that one thing i'd like you to move back please, then check CI, then i can merge it :)
> 
> I think `sleep.h` and `sleep.cc` should be removed before merging since the function `SleepForMilliseconds` is no longer used.

sure. if they're unused let's do that. i'll wait :)

dmah42 (approved):

dmah42:
thank you!

PR #1548, opened 2023-02-17
ci: add cmake test for 3.10.0

devjgm:
Related to https://github.com/google/benchmark/pull/1544

I followed the pattern of the bazel.yml workflow and copied
the basic cmake commands from the top-level README.md.

PR #1550, opened 2023-02-21
add compiler to build-and-test and create min-cmake CI bot

dmah42:
Related to #1544 
Inspired by #1548 

I noticed when adding the new min-cmake CI bot that we didn't actually set the compiler correctly for g++ and clang++ any more.  let's see how much this breaks.

### DIFF
--- a/.github/workflows/build-and-test-min-cmake.yml
+++ b/.github/workflows/build-and-test-min-cmake.yml
@@ -55,17 +55,13 @@ jobs:
       fail-fast: false
       matrix:
         msvc:
-          - VS-16-2019
-          - VS-17-2022
+          - VS-15-2017
         arch:
           - x64
         include:
-          - msvc: VS-16-2019
-            os: windows-2019
-            generator: 'Visual Studio 16 2019'
-          - msvc: VS-17-2022
-            os: windows-2022
-            generator: 'Visual Studio 17 2022'
+          - msvc: VS-15-2017
+            os: windows-2017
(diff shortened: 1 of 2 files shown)


coryan:
You can use Ninja as a generator on Windows too.  It may require downloading and installing Ninja for windows.  It has pre-compiled binaries:

https://github.com/ninja-build/ninja/releases

You may also consider a less full matrix in your testing.  In other projects we do not test all the different combinations of OS, distro, build tools, compiler versions, compiler settings, etc.  We test the oldest and newest version of each dimension in at least one build.  Also consider that on Windows it is unlikely that folks are going to use CMake==3.10.  That is more realistic for a Linux distro that ships with an specific version of CMake.

dmah42:
> You can use Ninja as a generator on Windows too. It may require downloading and installing Ninja for windows. It has pre-compiled binaries:
> 
> https://github.com/ninja-build/ninja/releases
> 

i can. our users do not.

> You may also consider a less full matrix in your testing. In other projects we do not test all the different combinations of OS, distro, build tools, compiler versions, compiler settings, etc. 

i'm not sure that reducing the coverage of our CI bots is the correct response to "this doesn't work in some CI bots".

> We test the oldest and newest version of each dimension in at least one build. Also consider that on Windows it is unlikely that folks are going to use CMake==3.10. That is more realistic for a Linux distro that ships with an specific version of CMake.

that's fair.  as the policy says only that the newest version of Windows is supported i can drop that section from this workflow.

PR #1552, opened 2023-02-28
Implement unlimited number of performance counters

HFTrader:
Linux performance counters will limit the number of hardware counters per reading group. For that reason the implementation of PerfCounters is limited to 3. However if only software counters are added, there is no reason to limit the counters. For hardware counters, we create multiple groups and store a vector or leaders in the PerfCounters object. When reading, there is an extra time waste by iterating through all the group leaders. However this should be the same performance as with today. Reading is done by groups and it had to be heavily adjusted with the logic being moved to PerfCounterValues. I created a test for x86-64 and took care of filtering out the events in case it runs in a platform that does not support those counters - the test will not fail. The current tests were already failing (ReOpenExistingCounters, CreateExistingMeasurements and MultiThreaded) on the main branch and they continue to fail after this implementation - I did not fix those not to conflate all here.

### DIFF
--- a/src/perf_counters.h
+++ b/src/perf_counters.h
@@ -17,6 +17,7 @@
 
 #include <array>
 #include <cstdint>
+#include <cstring>
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


dmah42:
Fixes #1553

src/perf_counters.h
```diff
@@ -66,7 +70,38 @@ class PerfCounterValues {
             sizeof(uint64_t) * (kPadding + nr_counters_)};
   }
 
-  static constexpr size_t kPadding = 1;
+  // This reading is complex and as the goal of this class is to
+  // abstract away the intrincacies of the reading process, this is
+  // a better place for it
+  size_t Read(const std::vector<int>& leaders) {
```
dmah42, line 76: any reason why the implementation is in the header?

HFTrader: No particular reason. I was just trying to follow the flow of your code. I can definitely move there.

HFTrader:
I have considered moving the values_ std::array into std::vector but I did not want to change much at once. std::vector would create a new indirection and that might be undesired.

dmah42:
windows release failures are going to be unrelated.. they're a bit flaky.

dmah42:
wonderful, thank you!

PR #1555, opened 2023-03-02
Fix failing perf gtests

HFTrader:
This is more of a *tentative* pull request to understand better the architecture, as I think the changes in this commit are more profound than they should be for just a "test fix".

I cannot understand why the PerfCounters counters_ object inside PerfCounterMeasurement was made static. There is really no reason for that from a perf_event point of view. I am here thinking that the reason for the static is some higher level process/thread issue that I do not familiar with. If so, kindly disregard this pull request.

Once you make that object non-static, the entire architecture simplifies a lot. I created a shared pointer on the off-chance that the static had to do with copy/move semantics and the way the top level class was created.

I could not create that shared pointer with std::make_shared because the relevant PerfCounters constructor is private and std::forward is obviously not a friend class. So we got an extra int malloc there. But I suspect this will have no impact since these objects are so infrequently created.

I did not want to replace the shared_ptr with a unique_ptr since the gain in performance would be really nil and the danger of introducing some move semantics glitch, much higher. Interesting 
enough, what you had there was effectively an intrusive pointer. 

The nice thing is that this shared pointer also acts as a std::optional - in the sense you can query it through its bool operator.

All tests were passing then - apart from the multithreaded in release mode. I had to make sure it stayed there computing for an arbitrary 0.25 seconds until the tests passed. Apparently it was going too fast - or even perhaps the compiler was optimizing away the entire squared sum. It is passing consistently now. I changed a bit the name of the variables there so I could understand better. Hope you do not mind.

If the basics of this pull request are sound, there is more to clean up like do away with the NoCounters() method which only returns an empty shared_ptr at the moment.

Please let me know. Thank you.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -221,6 +221,7 @@ BenchTimeType ParseBenchMinTime(const std::string& value) {
 
 BenchmarkRunner::BenchmarkRunner(
     const benchmark::internal::BenchmarkInstance& b_,
+    const std::shared_ptr<PerfCountersMeasurement>& pcm_,
     BenchmarkReporter::PerFamilyRunReports* reports_for_family_)
     : b(b_),
       reports_for_family(reports_for_family_),
@@ -239,10 +240,7 @@ BenchmarkRunner::BenchmarkRunner(
       iters(has_explicit_iteration_count
                 ? ComputeIters(b_, parsed_benchtime_flag)
                 : 1),
-      perf_counters_measurement(StrSplit(FLAGS_benchmark_perf_counters, ',')),
-      perf_counters_measurement_ptr(perf_counters_measurement.IsValid()
-                                        ? &perf_counters_measurement
-                                        : nullptr) {
+      perf_counters_measurement_ptr(pcm_) {
   run_results.display_report_aggregates_only =
       (FLAGS_benchmark_report_aggregates_only ||
        FLAGS_benchmark_display_aggregates_only);
@@ -255,7 +253,7 @@ BenchmarkRunner::BenchmarkRunner(
     run_results.file_report_aggregates_only =
         (b.aggregation_report_mode() & internal::ARM_FileReportAggregatesOnly);
     BM_CHECK(FLAGS_benchmark_perf_counters.empty() ||
-             perf_counters_measurement.IsValid())
+             perf_counters_measurement_ptr->IsValid())
         << "Perf counters were requested but could not be set up.";
   }
 }
@@ -269,12 +267,12 @@ BenchmarkRunner::IterationResults BenchmarkRunner::DoNIterations() {
   // Run all but one thread in separate threads
   for (std::size_t ti = 0; ti < pool.size(); ++ti) {
     pool[ti] = std::thread(&RunInThread, &b, iters, static_cast<int>(ti + 1),
-                           manager.get(), perf_counters_measurement_ptr);
+                           manager.get(), perf_counters_measurement_ptr.get());
   }
   // And run one thread here directly.
   // (If we were asked to run just one thread, we don't create new threads.)
   // Yes, we need to do this here *after* we start the separate threads.
-  RunInThread(&b, iters, 0, manager.get(), perf_counters_measurement_ptr);
+  RunInThread(&b, iters, 0, manager.get(), perf_counters_measurement_ptr.get());
 
   // The main thread has finished. Now let's wait for the other threads.
   manager->WaitForAllThreads();
@@ -465,7 +463,7 @@ void BenchmarkRunner::DoOneRepetition() {
     manager.reset(new internal::ThreadManager(1));
     b.Setup();
     RunInThread(&b, memory_iterations, 0, manager.get(),
-                perf_counters_measurement_ptr);
+                perf_counters_measurement_ptr.get());
     manager->WaitForAllThreads();
     manager.reset();
     b.Teardown();
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -221,7 +221,7 @@ BenchTimeType ParseBenchMinTime(const std::string& value) {
 
 BenchmarkRunner::BenchmarkRunner(
     const benchmark::internal::BenchmarkInstance& b_,
-    const std::shared_ptr<PerfCountersMeasurement>& pcm_,
+    PerfCountersMeasurement* pcm_,
     BenchmarkReporter::PerFamilyRunReports* reports_for_family_)
     : b(b_),
       reports_for_family(reports_for_family_),
@@ -267,12 +267,12 @@ BenchmarkRunner::IterationResults BenchmarkRunner::DoNIterations() {
   // Run all but one thread in separate threads
   for (std::size_t ti = 0; ti < pool.size(); ++ti) {
     pool[ti] = std::thread(&RunInThread, &b, iters, static_cast<int>(ti + 1),
-                           manager.get(), perf_counters_measurement_ptr.get());
+                           manager.get(), perf_counters_measurement_ptr);
   }
   // And run one thread here directly.
   // (If we were asked to run just one thread, we don't create new threads.)
   // Yes, we need to do this here *after* we start the separate threads.
-  RunInThread(&b, iters, 0, manager.get(), perf_counters_measurement_ptr.get());
+  RunInThread(&b, iters, 0, manager.get(), perf_counters_measurement_ptr);
 
   // The main thread has finished. Now let's wait for the other threads.
   manager->WaitForAllThreads();
@@ -463,7 +463,7 @@ void BenchmarkRunner::DoOneRepetition() {
     manager.reset(new internal::ThreadManager(1));
     b.Setup();
     RunInThread(&b, memory_iterations, 0, manager.get(),
-                perf_counters_measurement_ptr.get());
+                perf_counters_measurement_ptr);
     manager->WaitForAllThreads();
     manager.reset();
     b.Teardown();
--- a/src/benchmark_runner.h
+++ b/src/benchmark_runner.h
@@ -58,7 +58,7 @@ BenchTimeType ParseBenchMinTime(const std::string& value);
 class BenchmarkRunner {
  public:
   BenchmarkRunner(const benchmark::internal::BenchmarkInstance& b_,
-                  const std::shared_ptr<PerfCountersMeasurement>& pmc_,
+                  PerfCountersMeasurement* pmc_,
                   BenchmarkReporter::PerFamilyRunReports* reports_for_family);
 
   int GetNumRepeats() const { return repeats; }
@@ -104,7 +104,7 @@ class BenchmarkRunner {
   // So only the first repetition has to find/calculate it,
   // the other repetitions will just use that precomputed iteration count.
 
-  std::shared_ptr<PerfCountersMeasurement> perf_counters_measurement_ptr;
+  PerfCountersMeasurement* perf_counters_measurement_ptr;
 
   struct IterationResults {
     internal::ThreadManager::Result results;
--- a/src/benchmark_runner.h
+++ b/src/benchmark_runner.h
@@ -104,7 +104,7 @@ class BenchmarkRunner {
   // So only the first repetition has to find/calculate it,
   // the other repetitions will just use that precomputed iteration count.
 
-  PerfCountersMeasurement* perf_counters_measurement_ptr;
+  PerfCountersMeasurement* const perf_counters_measurement_ptr = nullptr;
 
   struct IterationResults {
     internal::ThreadManager::Result results;
--- a/src/benchmark_runner.h
+++ b/src/benchmark_runner.h
@@ -58,6 +58,7 @@ BenchTimeType ParseBenchMinTime(const std::string& value);
 class BenchmarkRunner {
  public:
   BenchmarkRunner(const benchmark::internal::BenchmarkInstance& b_,
+                  const std::shared_ptr<PerfCountersMeasurement>& pmc_,
                   BenchmarkReporter::PerFamilyRunReports* reports_for_family);
 
   int GetNumRepeats() const { return repeats; }
@@ -103,8 +104,7 @@ class BenchmarkRunner {
   // So only the first repetition has to find/calculate it,
   // the other repetitions will just use that precomputed iteration count.
 
-  PerfCountersMeasurement perf_counters_measurement;
-  PerfCountersMeasurement* const perf_counters_measurement_ptr;
+  std::shared_ptr<PerfCountersMeasurement> perf_counters_measurement_ptr;
 
   struct IterationResults {
     internal::ThreadManager::Result results;
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -71,7 +71,7 @@ bool PerfCounters::IsCounterSupported(const std::string& name) {
   return (ret == PFM_SUCCESS);
 }
 
-PerfCounters PerfCounters::Create(
+std::shared_ptr<PerfCounters> PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
   if (counter_names.empty()) {
     return NoCounters();
@@ -159,8 +159,8 @@ PerfCounters PerfCounters::Create(
     }
   }
 
-  return PerfCounters(counter_names, std::move(counter_ids),
-                      std::move(leader_ids));
+  return std::shared_ptr<PerfCounters>(new PerfCounters(
+      counter_names, std::move(counter_ids), std::move(leader_ids)));
 }
 
 void PerfCounters::CloseCounters() const {
@@ -183,7 +183,7 @@ bool PerfCounters::Initialize() { return false; }
 
 bool PerfCounters::IsCounterSupported(const std::string&) { return false; }
 
-PerfCounters PerfCounters::Create(
+std::shared_ptr<PerfCounters> PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
   if (!counter_names.empty()) {
     GetErrorLogInstance() << "Performance counters not supported.";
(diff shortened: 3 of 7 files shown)


dmah42:
@mtrofin should have an opinion here as he wrote it this way iirc.

i don't think your changes are way off, but i will point you to https://google.github.io/styleguide/cppguide.html#Ownership_and_Smart_Pointers which governs our use of `shared_ptr` in particular.

specifically: "Do not design your code to use shared ownership without a very good reason. One such reason is to avoid expensive copy operations, but you should only do this if the performance benefits are significant, and the underlying object is immutable (i.e., std::shared_ptr<const Foo>). "

ie, i think we probably do want a `unique_ptr` here and access outside the owner through weak pointers, however i'm open to being told that this is a special enough case that it matters.

src/perf_counters.cc
```diff
@@ -71,7 +71,7 @@ bool PerfCounters::IsCounterSupported(const std::string& name) {
   return (ret == PFM_SUCCESS);
 }
 
-PerfCounters PerfCounters::Create(
+std::shared_ptr<PerfCounters> PerfCounters::Create(
```
dmah42, line 74: i would expect (maybe naively) that we only create the perf counters once.  why, then, would we need a shared pointer here?  either the created perf counter object is owned by this object _or_ the caller (which should be singular).

test/perf_counters_gtest.cc
```diff
@@ -85,110 +81,135 @@ TEST(PerfCountersTest, Read2Counters) {
...
+    measurements.clear();
+    counter.Start();
+    EXPECT_TRUE(counter.Stop(measurements));
+  }
 }
 
+using benchmark::ChronoClockNow;
+
 size_t do_work() {
+  const double kMinimumElapsedSeconds = 0.25;
```
dmah42, line 158: i'm not a fan of explicit waits in tests as they tend to get flaky and surprising (and obviously slow).

the _right_ way to do this i think is to use `DoNotOptimize` on `res`.

HFTrader: You got another case yesterday that I commented where the compiler was smart enough to collapse the loop even though the DoNotOptimize worked. That happened here too. The compiler is just outsmarting our assumptions. In this case, that was the reason for the tests failing in release mode.

src/perf_counters.cc
HFTrader: I understand that. However what we had before with a reference counter was effectively a poor man's intrusive shared pointer like boost::intrusive_ptr. That was the reason I used std::shared_ptr. However that might have been only because PerfCounters was static and shared across the board with several threads. If we use std::unique_ptr then what we are saying is that this base object is never supposed to be shared, which is a huge assumption I did not want to make at this point. But I can put it there and test.

mtrofin: The intention is not to share counters, in the absence of a scenario. Unless we have a scenario for google/benchmarks, I'd suggest we maintain the invariant.

test/perf_counters_gtest.cc
mtrofin: try `__attribute__((optnone))` on `doWork`. It's `__attribute__((optimize(0))` in gcc. The test shouldn't run on windows anyway, so the msvc toolchain alternative doesn't apply here.

HFTrader: Are you guys okay defining BENCHMARK_DONT_OPTIMIZE in include/benchmark.h along with BENCHMARK_ALWAYS_INLINE and BENCHMARK_NOEXCEPT just to have the compiler-specific defines all in one place?

src/perf_counters.cc
HFTrader: The way it was, it was written to share counters, even with a specific reference counter in there. As a suggestion, I changed PerfCounters to be a regular member. Have a look at the commit just pushed and see if you like. It simplifies things quite a bit.

test/perf_counters_gtest.cc
```diff
@@ -97,98 +99,138 @@ TEST(PerfCountersTest, Read2Counters) {
...
+    counter.Start();
+    EXPECT_TRUE(counter.Stop(measurements));
+  }
 }
 
-size_t do_work() {
-  size_t res = 0;
-  for (size_t i = 0; i < 100000000; ++i) res += i * i;
-  return res;
+// This probably should go into include/benchmark.h together with
```
dmah42, line 166: i think it's ok to make this part of the public API.. it fits alongside DoNotOptimize as a method-level version.

src/perf_counters.cc
mtrofin: Sorry, I misspoke. You are right that the `PerfCounters` object is shared. What I meant to say is that we want one instance only, to be shared by all Measurements. IIUC, this PR would create a new `PerfCounters` for each `PerfCountersMeasurement`.

We could get rid of the ref count like you suggesting by keeping `counters_` static, and making sure it's inited only once. But, one drawback of using a pointer there is that now measurements would have to be through an indirection. That could add noise to measurements - dcache misses, for instance - and, given the purpose of this code, I'd argue it's best to avoid such noise as much as possible.

HFTrader: Ohhh ok now I am getting the reason why these objects were made static. It makes sense. Although I would like to propose this approach: why not create a single PerfCountersMeasurement object and pass it to each BenchmarkRunner's constructor? That would be a bit more explicit than a static object.

mtrofin: a `PerfCountersMeasurement` keeps track of the start/stop counter values - "start" and "stop" being values snapshotted at the beginning/end of some region of code of interest (==benchmarked code). Arguably, we'd want separate instances of `PerfCountersMeasurement` for separate experiments - they can share, though, the same source of info (i.e. the `PerfCounters` object)

HFTrader: I cannot think of a use case for having overlapping measurements of the same counters. If we have multiple threads running separate tests on the same process, that is outright wrong because the samples will contain contributions from all threads. If we have several tests running sequentially then having separate values for each one might affect adversely (say dTLB) than having a single, shared PerfCountersMeasurement object.

mtrofin: Right, I'm not suggesting either of those. It may be more of an aesthetic consideration - you want to make a measurement, you make a `PCM` object... but I can see how what you're suggesting could let us assert that, for example, only one client makes measurements, and I like that. Maybe it'd be easier to see how the code looked like?

test/perf_counters_gtest.cc
HFTrader: Having trouble with MSVC though... it does not accept single-function attributes like gcc and clang so we would be forced to use two OFF/ON macros like [here](https://godbolt.org/z/e5Erhzf6Y). However that requires the standard `_Pragma()` form to be used which is somewhat new and might break old compilers.

mtrofin: the test shouldn't even run on msvc, because of that `!PerfCounters::kSupported` test, so maybe on msvc don't add any attribute just to make the compiler happy?

src/perf_counters.cc
HFTrader: Yes, Sir! I have added another commit to this patch. Please have a look. I tried to describe as much as possible in the commit but please let me know your thoughts. All tests are passing but I am afraid of some side effects I'm not taking in account for my lack of deep understanding of the project.

test/perf_counters_gtest.cc
HFTrader: True. Ends up it seems that it uses the same syntax of the Intel compiler. Not surprised by that actually.

src/benchmark_runner.cc
```diff
@@ -221,6 +221,7 @@ BenchTimeType ParseBenchMinTime(const std::string& value) {
 
 BenchmarkRunner::BenchmarkRunner(
     const benchmark::internal::BenchmarkInstance& b_,
+    const std::shared_ptr<PerfCountersMeasurement>& pcm_,
```
dmah42, line 224: i'm still uncomfortable with the use of shared_ptr for this. we should be able to cleanly describe the lifetime and ownership of the PerfCountersMeasurement object, and shared_ptr elides that.

test/perf_counters_gtest.cc
```diff
@@ -37,26 +39,25 @@ TEST(PerfCountersTest, NegativeTest) {
     return;
   }
   EXPECT_TRUE(PerfCounters::Initialize());
-  EXPECT_FALSE(PerfCounters::Create({}).IsValid());
-  EXPECT_FALSE(PerfCounters::Create({""}).IsValid());
-  EXPECT_FALSE(PerfCounters::Create({"not a counter name"}).IsValid());
+  EXPECT_TRUE(PerfCounters::Create({}).IsValid());
+  EXPECT_TRUE(PerfCounters::Create({""}).IsValid());
+  EXPECT_TRUE(PerfCounters::Create({"not a counter name"}).IsValid());
```
dmah42, line 44: why would this be a valid PerfCounters?

test/perf_counters_gtest.cc
```diff
@@ -97,98 +98,136 @@ TEST(PerfCountersTest, Read2Counters) {
...
+  std::vector<std::string> kMetrics({kGenericPerfEvent1});
+  std::vector<PerfCounters> counters(6);
+  for (auto& counter : counters) {
+    counter = PerfCounters::Create(kMetrics);
+  }
   PerfCounterValues values(1);
   EXPECT_TRUE(counters[0].Snapshot(&values));
-  EXPECT_FALSE(counters[4].Snapshot(&values));
-  EXPECT_FALSE(counters[5].Snapshot(&values));
+  EXPECT_TRUE(counters[4].Snapshot(&values));
```
dmah42, line 114: i'm not sure what this is testing now if these are just all valid

src/benchmark_runner.cc
HFTrader: Absolutely. You can see in RunBenchmark (benchmark.cc:351) that the object is created and then passed to N BenchmarkRunner objects (line 360). That is the reason why in the previous code we had a reference counter. So basically we are going from a hand-made intrusive pointer implementation to a proper STL shared_ptr, not much of a change. 

We could pass the pointer raw but there is an off-chance that in the future one of those objects linger beyond the scope of the counter and creates a segfault that would be hard to pinpoint. I think that's the main reason why we use smart pointers in the first place. But it is your code, I can definitely remove it - although I would not advise doing so.

test/perf_counters_gtest.cc
HFTrader: In fact, one of the results of these changes is that we do not need that `IsValid()` method anymore. It was there to prevent access to a null pointer. With the new filtering logic that went in yesterday, we are just dropping the bad entries and the object is always in a good state afterwards. Semantically it looks wrong and I think we can drop that IsValid method altogether since the object now is always valid anyways.

test/perf_counters_gtest.cc
HFTrader: Yes, I agree. This is one of the tests that were previously failing. On older Intel architectures there was a limit of 3 PMCs while in newer ones there are 6 PMCs so these tests were not returning false as expected. And new hardware PMCs are being introduced with every new launch [see the AMD Zen4](https://www.phoronix.com/news/LInux-6.2-AMD-Zen-4-Events) so there is really not much sense in keeping these tests as they are, in fact.

test/perf_counters_gtest.cc
mtrofin: Maybe replace with `HasAllRequestedCounters` or something like that?

test/perf_counters_gtest.cc
HFTrader: True. This entire PR is hanging now on a green light on the core changes. There is really a substantial amount of clean up to be done if true, including comments.

test/perf_counters_gtest.cc
HFTrader: Then we would have to keep a bool around in the object to reflect the fact that the object has now less counters than was requested. Apart from the tests, there is no real use for that from the code point of view. 

We could yes compare the new number of counters with the expected number of counters and check that some were indeed dropped?

src/benchmark_runner.cc
```diff
@@ -255,7 +253,7 @@ BenchmarkRunner::BenchmarkRunner(
     run_results.file_report_aggregates_only =
         (b.aggregation_report_mode() & internal::ARM_FileReportAggregatesOnly);
     BM_CHECK(FLAGS_benchmark_perf_counters.empty() ||
-             perf_counters_measurement.IsValid())
+             perf_counters_measurement_ptr->IsValid())
```
mtrofin, line 256: We should add a check that if perf counters are requested, the thread pool has 0 size. I thought we did that check but I don't see it anymore. Anyway - as you pointed out yesterday, it'd be whacky otherwise because each observation would see effects of all threads (unless the counters were set up per thread, which wasn't and isn't the case)

src/benchmark_runner.cc
dmah42: we know when all the runner objects complete, so that should be the scope of the unique_ptr.. the runners should get weak pointers.

i understand the ease that comes with shared pointers (and i agree the existing ref counting was worse than using them) but i would like to try to stick to something that requires us to understand the correctness of the code.

test/perf_counters_gtest.cc
dmah42: it looks like we ended up using _Pragma instead of __attribute__ .. i don't know enough about this area but why is that? __attribute__ should be well-supported by the compilers we care about.

src/benchmark_runner.cc
HFTrader: I could see the case where someone will look at the combined number of cycles or branch misses on a multi-threaded test, although the utility as you said is questionable. But it would be up to the user to interpret it.  Imho we should not restrict a feature if it does not cause the software to reach an invalid state. Maybe a warning?

src/benchmark_runner.cc
HFTrader: If the runners get weak pointers then they will have to check the pointer every time they access it, which is an extra step. Im fine with that. But just so we are in the same page, weak pointers - as in [std::weak_ptr](https://en.cppreference.com/w/cpp/memory/weak_ptr) - have to be created from existing std::shared_ptr, they cannot be created from std::unique_ptr. We would have to pass a raw pointer from a unique_ptr.

mtrofin: Why even give weak_ptr, since it;'d be a bug if access was attempted outside the scope of the unique_ptr, a naked ptr would suffice (and also avoid extra checks)

test/perf_counters_gtest.cc
mtrofin: Ah, because you already warn the user if some counter names "didn't make it"? Then yes, for the tests only we can just do what you suggested.

test/perf_counters_gtest.cc
HFTrader: I would file that in the bad try shelf. That was because MSVC (always you Microsoft!) does not have an equivalent attribute but it has a pragma, so my solution was to make a pair of ON/OFF macros. We could have both or return to the previous attribute macro with the caveat that it will not work on MSVC.

src/benchmark_runner.cc
dmah42: yes sorry, i meant raw/naked pointer.  i still had shared_ptr in my head.

HFTrader: That's okay with me. I just think that we are going back to the old C style with cowboy pointers while we have modern features to prevent a crash. But this is your code, I will make the change.

test/perf_counters_gtest.cc
HFTrader: Correction - MSVC does not have a do-not-optimize attribute, it has a line-level `#pragma` only.

HFTrader: Committed a revert change. Please have a look. Thank you.

src/benchmark_runner.cc
HFTrader: @dmah42 Alright the commit is in! Please have a look.

src/benchmark_runner.cc
HFTrader: It is now printing a meaty warning if you select threads and perf counters. 
```
$ test/benchmark_test --benchmark_perf_counters=cycles
2023-03-03T12:45:32-06:00
Running test/benchmark_test
Run on (24 X 4568.16 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x24)
  L1 Instruction 32 KiB (x24)
  L2 Unified 512 KiB (x24)
  L3 Unified 16384 KiB (x8)
Load Average: 4.38, 4.32, 4.24
***WARNING*** There are 80 benchmarks with threads and 1 performance counters were requested. Beware counters will reflect the combined usage across all threads.
---------------------------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------------------
BM_Factorial                                  0.000 ns        0.000 ns   1000000000 cycles=330n 40320
BM_Factorial/real_time                        0.000 ns        0.000 ns   1000000000 cycles=322n 40320
BM_CalculatePiRange/1                          1.32 ns         1.32 ns    531367024 cycles=5.00025 0
BM_CalculatePiRange/8                          10.5 ns         10.5 ns     66417872 cycles=40.002 3.28374
BM_CalculatePiRange/64                         84.3 ns         84.3 ns      8298636 cycles=320.015 3.15746
```

test/perf_counters_gtest.cc
HFTrader: So in the current PR and in another one I placed just now to fix an edge case, it compares the number of current, accepted PMCs with the number of requested. As in 

```c++
  EXPECT_EQ(PerfCounters::Create({}).num_counters(), 0);
  EXPECT_EQ(PerfCounters::Create({""}).num_counters(), 0);
  EXPECT_EQ(PerfCounters::Create({"not a counter name"}).num_counters(), 0);
```

test/perf_counters_gtest.cc
HFTrader: I removed the tests for `N>3` since it would be impossible to make any reasonable assumption regarding the current hardware without explicitly querying libpfm.

src/benchmark_runner.cc
HFTrader: Out of complete curiosity, what is the general concern with shared pointers? I dont use them either but I use boost::intrusive_ptr quite extensively.

src/perf_counters.cc
```diff
@@ -47,8 +47,9 @@ size_t PerfCounterValues::Read(const std::vector<int>& leaders) {
       size -= data_bytes;
     } else {
       int err = errno;
-      GetErrorLogInstance() << "Error reading lead " << lead << " errno:" << err
-                            << " " << ::strerror(err) << "\n";
+      GetErrorLogInstance()
+          << "PerfCounterValues: error reading fd:" << lead << " errno:" << err
```
mtrofin, line 51: this is unrelated, correct? may be best for a separate small CL.

src/perf_counters.h
```diff
@@ -90,10 +90,13 @@ class BENCHMARK_EXPORT PerfCounters final {
   // True iff this platform supports performance counters.
   static const bool kSupported;
 
-  bool IsValid() const { return !counter_names_.empty(); }
+  bool IsValid() const { return counter_ids_.size() == counter_names_.size(); }
```
mtrofin, line 93: is this ever not the case?

mtrofin (commented):
I mostly have nits at this point, my only non-nit comment is that this patch should probably be split - one chunk doing the refactoring, the other the "fix failing tests" bit (probably a third for that IIUC unrelated change in error reporting, line 50 in `perf_counters.cc`)

src/benchmark_runner.h
```diff
@@ -103,8 +104,7 @@ class BenchmarkRunner {
   // So only the first repetition has to find/calculate it,
   // the other repetitions will just use that precomputed iteration count.
 
-  PerfCountersMeasurement perf_counters_measurement;
-  PerfCountersMeasurement* const perf_counters_measurement_ptr;
+  PerfCountersMeasurement* perf_counters_measurement_ptr;
```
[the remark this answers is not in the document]
HFTrader: Yes as I said in the beginning, this patch was mostly an experimentation. It started with fixing flags but it evolved to something else. I can divide neatly into several PRs, sure.

src/perf_counters.cc
HFTrader: Unrelated. It was throwing that at some point and I felt it needed more language.

HFTrader: I mean, it is related to another PR approved earlier this week.

src/perf_counters.h
HFTrader: To be fair, this was just me trying to justify the existence of `IsValid()` at this point.

HFTrader:
This pull request has been replaced by #1559

src/benchmark_runner.cc
dmah42: to ensure that lifetimes are well-understood through the design of the code.  shared_ptr allows you to get away with thinking less about lifetimes and ownership, trusting that  things will just work out.  we prefer to avoid that _unless_ there are performance benefits and the object is immutable (to limit some classes of bugs).

src/perf_counters.h
dmah42: if we don't need the IsValid we can drop it.

PR #1557, opened 2023-03-03
Fix edge case while filtering performance counters

HFTrader:
Even though we filter out bad performance counter names, it might still be the case where a PMC cannot be allocated due to system limits (eg perf_event_open returns -1 for any reason). In this case, the current implementation will drop the entire performance counter set.

Instead, filter out the failing PMC and warn the user. This was the behavior previously subject to a PR but this case slipped in, inherited.

The test for this fix is reflected in test/perf_counters_gtest.cpp under PerfCountersTest::NegativeTest.

### DIFF
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -83,28 +83,41 @@ PerfCounters PerfCounters::Create(
         << PerfCounterValues::kMaxCounters << "\n";
     return NoCounters();
   }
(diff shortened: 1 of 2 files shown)


HFTrader:
I'm confused why the merge conflicts. I branched out of main and it does not seem to have been updated. Please disregard.

HFTrader:
Bad merge. Please disregard.

PR #1558, opened 2023-03-03
Fix edge case perfcounters create

HFTrader:
When performance counters are added, they are checked for validity of
their names. However even if the counters are valid, they can still be
rejected (fail to open fd or fail to enable). When that happens, the
current behavior will reject all counters instead of just dropping the
problematic ones.

In this change we keep pushing through adding only the valid PMCs. This was
the behavior intended with a previous merged PR but this edge condition
slipped through.

Sorry for the mess, this branch was not synced with main when I pushed the 
first commit so I ended up with two commits on top of each other.


HFTrader:
This pull request has been replaced by #1559

PR #1559, opened 2023-03-04
Refactoring of PerfCounters infrastructure

HFTrader:
The main feature in this pull request is the removal of the static sharing of PerfCounters and instead creating them at the top `RunBenchmarks()`  function where all benchmark runners are created. A single PerfCountersMeasurement object is created and then shared with all the new BenchmarkRunners objects, one per existing benchmark.

Other features conflated here in this PR are:
- Added BENCHMARK_DONT_OPTIMIZE macro in global scope
- Removal of the `IsValid()` query, being replaced by checking the number of remaining counters after validity tests
- Refactoring of all GTests to reflect the changes and new semantics
- extra comments throughout the new code to clarify intent
- fix a harmless issue where the Create() method would invalidate the entire set of counters if one of them failed to open after being validated

It was truly impossible to separate all those features in different PRs as requested since they are so interdependent on each other so I'm just pushing them together and asking for forgiveness.

This PR comes replacing PRs 1555 and 1558 which have been closed.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -348,20 +348,43 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
 
     size_t num_repetitions_total = 0;
 
+    // This perfcounters object needs to be created before the runners vector
+    // below so it outlasts their lifetime.
+    PerfCountersMeasurement perfcounters(
+        StrSplit(FLAGS_benchmark_perf_counters, ','));
+
+    // Vector of benchmarks to run
     std::vector<internal::BenchmarkRunner> runners;
     runners.reserve(benchmarks.size());
+
+    // Count the number of benchmarks with threads to warn the user in case
+    // performance counters are used.
+    int benchmarks_with_threads = 0;
+
+    // Loop through all benchmarks
     for (const BenchmarkInstance& benchmark : benchmarks) {
       BenchmarkReporter::PerFamilyRunReports* reports_for_family = nullptr;
       if (benchmark.complexity() != oNone)
         reports_for_family = &per_family_reports[benchmark.family_index()];
-
-      runners.emplace_back(benchmark, reports_for_family);
+      benchmarks_with_threads += (benchmark.threads() > 0);
+      runners.emplace_back(benchmark, &perfcounters, reports_for_family);
       int num_repeats_of_this_instance = runners.back().GetNumRepeats();
       num_repetitions_total += num_repeats_of_this_instance;
       if (reports_for_family)
         reports_for_family->num_runs_total += num_repeats_of_this_instance;
     }
     assert(runners.size() == benchmarks.size() && "Unexpected runner count.");
+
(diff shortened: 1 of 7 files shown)


HFTrader:
The Bazel failures dont seem related to code. What to do?

dmah42 (approved):

dmah42:
rerunning the osx one, that looks like a local failure.  the windows one seems like some timings changed away from expectations.  that _shouldn't_ be related to this PR.  i'l try rerunning that too.

src/benchmark.cc
```diff
@@ -348,21 +348,44 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
...
+    // This perfcounters object needs to be created before the runners vector
+    // below so it outlasts their lifetime.
+    PerfCountersMeasurement perfcounters(
+        StrSplit(FLAGS_benchmark_perf_counters, ','));
+
+    // Vector of benchmarks to run
     std::vector<internal::BenchmarkRunner> runners;
     runners.reserve(benchmarks.size());
+
+    // Count the number of benchmarks with threads to warn the user in case
```
mtrofin, line 360: nit: I think this is number of threads that will run benchmarks. The benchmarks could have threads as part of their internal implementation otherwise no problem.

mtrofin:
lgtm, some nits

src/benchmark.cc
HFTrader: That is true. However I don't think we have control over that, or even if we could possibly know how many threads a process had. That said, I don't think it matters if you disclose that the usage will be aggregated per process and test.

HFTrader:
> rerunning the osx one, that looks like a local failure. the windows one seems like some timings changed away from expectations. that _shouldn't_ be related to this PR. i'l try rerunning that too.

Maybe because some tests are now running in full while before they were being collapsed by the compiler?

src/benchmark.cc
HFTrader: Maybe I am not understanding what you are saying?

dmah42: i think the wording is correct. this is the number of benchmarks that have requested to run threaded.

dmah42:
thanks for all the iterations :D

src/benchmark.cc
```diff
@@ -348,21 +348,44 @@ void RunBenchmarks(const std::vector<BenchmarkInstance>& benchmarks,
...
+    int benchmarks_with_threads = 0;
+
+    // Loop through all benchmarks
     for (const BenchmarkInstance& benchmark : benchmarks) {
       BenchmarkReporter::PerFamilyRunReports* reports_for_family = nullptr;
       if (benchmark.complexity() != oNone)
         reports_for_family = &per_family_reports[benchmark.family_index()];
-
-      runners.emplace_back(benchmark, reports_for_family);
+      benchmarks_with_threads += (benchmark.threads() > 0);
```
jmr, line 369: It looks like this is always true. Should it be `threads() > 1`?

PR #1560, opened 2023-03-04
Fixed parentheses error thrown by MinGW gcc compiler

fangjunzhou:
# Original Error Message

image


LebedevRI:
Looks like this was resolved at some point:
https://github.com/google/benchmark/blob/54e4327190b6c06aeadc3186ed17566fd8da2a61/src/benchmark_runner.cc#L210

Issue #963, opened 2020-05-07
[FR] state.SkipWithMessage


NAThompson:
**Is your feature request related to a problem? Please describe.**

There are circumstances where a benchmark needs to be skipped, but this is not due to an error. In this case, it might be nice to have a `state.SkipWithMessage`.

**Describe the solution you'd like**

Add the following to `benchmark.cc`:

```cpp
void State::SkipWithMessage(const char* msg) {
  CHECK(msg);
  error_occurred_ = false;
  {
    MutexLock l(manager_->GetBenchmarkMutex());
    if (manager_->results.has_error_ == false) {
      // uh wait not quick correct; it's not an error it's info:
      manager_->results.error_message_ = msg;
      manager_->results.has_error_ = false;
    }
  }
  total_iterations_ = 0;
  if (timer_->running()) timer_->StopTimer();
}
```

Of course I'd prepare the PR if this is a desired feature.

dominichamon:
It sounds reasonable enough.. do you have any more details on when you'd want to skip a benchmark at runtime dynamically (instead of using the command line filter)?

NAThompson:
@dominichamon : This particular case is when user-supplied datasets are parametrized as inputs to `google/benchmarks`. The data flows through a lot of potential benchmarks; but not all of these benchmarks are relevant to the data provided.

gatorfax:
i have another use case for this.

we are using google benchmark for GPU-based microbenchmarks.  very often a specific GPU device does not support a feature we are writing a test for, e.g. raytracing.

today, we use the `SkipWithError` but with a custom error message that we have to regex in our testing automation to report it as a `skip` rather than an `error`.
it "works" but this would be a good QoL fix.

dmah42:
sounds reasonable. PRs welcome :)

PR #1564, opened 2023-03-06
[FR] state.SkipWithMessage #963

gatorfax:
https://github.com/google/benchmark/issues/963


src/console_reporter.cc
```diff
@@ -135,9 +135,14 @@ void ConsoleReporter::PrintRunData(const Run& result) {
...
 
-  if (result.error_occurred) {
-    printer(Out, COLOR_RED, "ERROR OCCURRED: \'%s\'",
-            result.error_message.c_str());
+  if (result.skipped) {
+    if (result.skip_is_error) {
+      printer(Out, COLOR_RED, "ERROR OCCURRED: \'%s\'",
+              result.skip_message.c_str());
+    } else {
+      printer(Out, COLOR_WHITE, "SKIPPED: \'%s\'",
```
gatorfax, line 143: the message color was not important to me.
i picked white so that it still stood out from the surrounding text and was not RED.

src/json_reporter.cc
```diff
@@ -254,9 +254,14 @@ void JSONReporter::PrintRunData(Run const& run) {
       BENCHMARK_UNREACHABLE();
     }()) << ",\n";
   }
-  if (run.error_occurred) {
-    out << indent << FormatKV("error_occurred", run.error_occurred) << ",\n";
-    out << indent << FormatKV("error_message", run.error_message) << ",\n";
+  if (run.skipped) {
+    if (run.skip_is_error) {
```
gatorfax, line 258: this could be a if..else if.. statement to remove the nesting if's but i feel like the logic is easier to follow this way

include/benchmark/benchmark.h
```diff
@@ -866,7 +890,8 @@ class BENCHMARK_EXPORT State {
  private:
   bool started_;
   bool finished_;
-  bool error_occurred_;
+  bool skipped_;
+  bool skip_is_error_;
```
gatorfax, line 894: my first draft kept the name(s) as `error_ocurred_` but then i had a bunch of `if skipped | error` statements.  it was much cleaner when i refactored to make `error` a subset of `skipped`

dmah42: do you think this would be cleaner as an enum? 

```
enum Skipped {
  NotSkipped,
  SkippedWithMessage,
  SkippedWithError,
};
```

or does it just make things more complex?

dmah42: the way it is is slightly fragile if someone sets "skip_is_error_" without setting "skipped_".

dmah42:
the resolved comments explaining your thinking are really helpful, thank you.

include/benchmark/benchmark.h
gatorfax: i agree.  i will change it to an enum.

include/benchmark/benchmark.h
```diff
@@ -655,6 +655,17 @@ enum AggregationReportMode
       ARM_FileReportAggregatesOnly | ARM_DisplayReportAggregatesOnly
 };
 
+enum Skipped
+#if defined(BENCHMARK_HAS_CXX11)
+    : unsigned
+#else
```
dmah42, line 661: nit: we can drop the #else

gatorfax: sure, i just copied the pattern from the previous enum.

dmah42 (approved):

dmah42:
thanks!

PR #1567, opened 2023-03-07
Convert uses of `const char*` to `std::string`

dmah42:


HFTrader:
I'm in love with std::string_view but I guess C++17 is not yet standard?

bindings/python/google_benchmark/benchmark.cc
```diff
@@ -161,12 +153,12 @@ NB_MODULE(_benchmark, m) {
...
-                    &State::SetBytesProcessed)
+                   &State::SetBytesProcessed)
       .def_prop_rw("complexity_n", &State::complexity_length_n,
-                    &State::SetComplexityN)
+                   &State::SetComplexityN)
       .def_prop_rw("items_processed", &State::items_processed,
-                    &State::SetItemsProcessed)
-      .def("set_label", (void (State::*)(const char*)) & State::SetLabel)
+                   &State::SetItemsProcessed)
+      .def("set_label", (void(State::*)(const char*)) & State::SetLabel)
```
nicholasjng, line 161: This would need to bechanged to accept a `const std::string&` too, or does it work with implicit conversion?

dmah42: hm. interesting that the bindings CI bot didn't fail.

dmah42:
> I'm in love with std::string_view but I guess C++17 is not yet standard?

https://opensource.google/documentation/policies/cplusplus-support

if all supported compilers use C++17 as _default_ then we can drop C++14.

having said that, we've tried to keep the public header as backwards compatible as possible, even if we build the rest of the library with newer standards.

PR #1568, opened 2023-03-07
Fix examples in user guide using deprecated `DoNotOptimize`-API

### LINKED ISSUE #1566: [BUG] Examples from User Guide trigger deprecation warning for const-reference `DoNotOptimize`
A few examples from the User Guide trigger the warning introduced with https://github.com/google/benchmark/pull/1493 for the const-reference version of `DoNotOptimize`:
* https://godbolt.org/z/eqGbT7jzT
* https://godbolt.org/z/n5bbvE869
* https://godbolt.org/z/z7d9e6sP7

As far as I understand, the intended solution is to introduce temporaries and pass those to `DoNotOptimize` like in https://github.com/google/benchmark/pull/1562? If so, I propose to change the examples accordingly to show users how to use `DoNotOptimize` properly without getting the warning.

mjacobse:
The const-reference API to `DoNotOptimize` was deprecated with #1493. Some examples in the user guide are using exactly that deprecated interface. This fixes that by passing non-const lvalues instead. Fixes #1566

I tried to provide a brief comment explaining why a copy is passed to `DoNotOptimize`. I opted to do so only in the section that addresses the `DoNotOptimize` feature directly, to not clutter the examples that highlight other aspects too much. I do wonder whether a more elaborate explanation in the "Preventing Optimization" section might be warranted, but perhaps that is for another PR?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,6 +43,7 @@ Jussi Knuuttila <jussi.knuuttila@gmail.com>
 Kaito Udagawa <umireon@gmail.com>
 Kishan Kumar <kumar.kishan@outlook.com>
 Lei Xu <eddyxu@gmail.com>
+Marcel Jacobse <mjacobse@uni-bremen.de>
 Matt Clarkson <mattyclarkson@gmail.com>
 Maxim Vafin <maxvafin@gmail.com>
 MongoDB Inc.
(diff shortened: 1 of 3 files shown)


dmah42 (approved):

dmah42:
thanks!

PR #1569, opened 2023-03-08
Fix build break with nvc++ when -Werror is ON

### LINKED ISSUE #1556: [BUG] Release build fails with nvhpc-22.5 (nvc++)
**Describe the bug**
I have google benchmark (84c71faa8126e4eedc2bb520352615cb4484d6ad) as dependency in my cmake project. When I trigger a Release build with `-DCMAKE_BUILD_TYPE=Release` using nvhpc-22.5, the build fails because of:

```
"[xxxxx]/master_benchmarks/ext/google-benchmark/src/timers.cc", line 86: error: function "benchmark::<unnamed>::MakeTime(const rusage &)" was declared but never referenced
  double MakeTime(struct rusage const& ru) {
         ^


"[xxxx]/master_benchmarks/ext/google-benchmark/src/benchmark.cc", line 201: error: offsetof applied to a type other than a standard-layout class
                           ^
```

Currently I use `set(BENCHMARK_ENABLE_WERROR OFF)` to force the build to pass with warnings.

**System**
  - OS: `Linux 4.19.0-18-amd64 #1 SMP Debian 4.19.208-1 (2021-09-29) x86_64 GNU/Linux`
  - Compiler and version: `nvc++ 22.5-0 64-bit target on x86-64 Linux -tp zen`

**To reproduce**
1. Use google benchmark (84c71faa8126e4eedc2bb520352615cb4484d6ad) as dependency in cmake
2. run `cmake -DCMAKE_MAKE_PROGRAM=ninja -DCMAKE_CXX_COMPILER=nvc++ -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-stdpar=multicore" -S . -B ./cmake-build-debug-nvidia`
3. run build ( I use the command `cmake --build cmake-build-debug-nvidia --target master_benchmarks -j 8`)

**Expected behavior**
I expected the build to run without any errors or warnings (since it does so with other compilers e.g. g++)

**Screenshots**
image

image

HFTrader:
Build breaks when -Werror is turned on because of unhandled cases of inocuous/pedantic warnings. Adopted the same solution as for Intel PGI compiler - just disable -Werror manually, unless BENCHMARK_FORCE_WERROR is enabled. Fixes #1556.


HFTrader:
Just realized I fixed through CMake but did not think about Bazel.

dmah42:
> Just realized I fixed through CMake but did not think about Bazel.

i think we don't have as many flags in bazel so it's probably fine.

dmah42 (approved):

PR #1573, opened 2023-03-10
Address warnings on NVIDIA nvc++

HFTrader:
Types of warnings were being generated:
1. Deprecated warnings - solved by defining the relevant BENCHMARK_* macros for nvc++ and adding pragma suppress on a couple of .cc files
2. Setup/TearDown const vs non-const partial override - solved by adding non-const version. Alternatively I could add a respective suppress but I found not providing an override was dangerous?
3. Static but not referenced - added diagnostic suppress for that file
Related to #1556


test/args_product_test.cc
```diff
@@ -31,6 +31,15 @@ class ArgsProductFixture : public ::benchmark::Fixture {
 
     actualValues.insert(ranges);
   }
+  // Added for completeness - compiler warns about partial override
+  void SetUp(::benchmark::State& state) override {
```
dmah42, line 35: but there's a default implementation that calls the const version.. ie, there is no partial override here.  

i don't really want to suggest to users (who use tests as a template) that they need to override both versions.

maybe we need to storm forward into the promised deprecation of the const versions.

dmah42 (approved):

PR #1578, opened 2023-03-17
Simplify CycleClock a bit by reducing duplication

brad0:
There are three copies of the same chunk of code for different processors. Simplify things by having one copy for the various processors and place it at the bottom.


HFTrader:
I feel like the leaving of each platform in its own bucket was on purpose. if you have an issue and look into that file it is immediately clear what is going on. Playing devil's advocate, it is a lot of risk for little gain. I would say there is a loss: the comments for mips were now together with the other platforms.  And there is no gain either in performance or compile time. (Apologies I am not even an approver)

dmah42:
yes, this was deliberate. it allows platforms to migrate to alternatives without significant refactoring.

i'm going to close this.  i appreciate it isn't obvious why this was deliberate and your cleanup attempt is still very much appreciated.

PR #1579, opened 2023-03-17
Correct libpfm dep for integrators.

shiqing117:
This change corrects the label when the macro is called by another repository.


dmah42 (approved):

dmah42:
thanks!

PR #1587, opened 2023-04-18
Remove not needed msan implementation from workflows 

Zzzabiyaka:
Hi team, 

As msan is not currently used and probably is not expected to be used in the near future I'm suggesting to remove this build step what will make the pipeline log easier to read and will remove one unneded step.

### DIFF
--- a/.github/workflows/sanitizer.yml
+++ b/.github/workflows/sanitizer.yml
@@ -21,12 +21,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: configure msan env
-      if: matrix.sanitizer == 'msan'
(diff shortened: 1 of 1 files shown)


dmah42:
there's a TODO a few lines above.  it is intended to be used once someone has time to look at why it's currently failing.

LebedevRI:
The libc++ that is being used for msan build, is it built with msan sanitization itself? (can you point me at said failure?)

dmah42:
> The libc++ that is being used for msan build, is it built with msan sanitization itself? (can you point me at said failure?)

https://github.com/google/benchmark/pull/1589

PR #1589, opened 2023-04-27
reenable msan sanitizer workflow

dmah42:


LebedevRI:
Yup, i do not see where you custom-build libc++. That is really necessary for MSan.
Roughly, you need
```
$ git clone --depth=1 https://github.com/llvm/llvm-project
$ mkdir <llvm-build> && cd <llvm-build>
$ cmake -GNinja ../llvm-project/runtimes/ -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_USE_SANITIZER=MemoryWithOrigins
$ ninja cxx cxxabi unwind
$ cd benchmark-build-dir
$ CC=clang CXX=clang++ CXXFLAGS="-fsanitize=memory -fsanitize-memory-track-origins -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -stdlib=libc++ -L<llvm-build>/lib -lc++abi -Isystem<llvm-build>/include -Isystem<llvm-build>/include/c++/v1 -Wl,-rpath,<llvm-build>/lib" cmake <...> ..
```

dmah42:
something like https://github.com/google/benchmark/blob/main/.github/libcxx-setup.sh ?

we used to have this, but with it the sanitizer workflow timed out because the libc++ build is way too slow.  maybe it's worth trying just for msan.

LebedevRI:
> something like https://github.com/google/benchmark/blob/main/.github/libcxx-setup.sh ?

Close, but my snippet is the more modern version of that incantation.

> we used to have this, but with it the sanitizer workflow timed out because the libc++ build is way too slow. maybe it's worth trying just for msan.

Unless the whole LLVM gets built accidentally, i don't believe that should be slow at all, maybe 10 extra minutes.

dmah42:
note: LLVM build is at 30% after 12m.

LebedevRI:
> msan debug failure is the same as when we didn't build libc++ with msan (unless i messed up the linking somehow)

You have noticed that the libc++ build itself failed, right?

dmah42:
> > msan debug failure is the same as when we didn't build libc++ with msan (unless i messed up the linking somehow)
> 
> You have noticed that the libc++ build itself failed, right?

gah, no i didn't.  i guess the script should return an error status.

.github/libcxx-setup.sh
```diff
@@ -1,27 +1,29 @@
 #!/usr/bin/env bash
 
+set -e
+
 # Checkout LLVM sources
-#git clone --depth=1 https://github.com/llvm/llvm-project.git llvm-project
-#
+git clone --depth=1 https://github.com/llvm/llvm-project.git llvm-project
```
LebedevRI, line 6: One more thing to note, it is rather issue-prone to use libc++ from trunk,
e.g. because nowadays they only guarantee building libc++ with clang
from the same revision or so, so building with older clang may fail,
and i'm not sure that will be caught at cmake time.
So it would be best to checkout some specific tag.

.github/libcxx-setup.sh
```diff
@@ -1,27 +1,29 @@
...
+cd ./llvm-project/runtimes/
+cmake -DCMAKE_C_COMPILER=${CC}                  \
+      -DCMAKE_CXX_COMPILER=${CXX}               \
+      -DCMAKE_BUILD_TYPE=RelWithDebInfo         \
+      -DCMAKE_INSTALL_PREFIX=/usr               \
+      -DLIBCXX_ABI_UNSTABLE=OFF                 \
+      -DLLVM_USE_SANITIZER=${LIBCXX_SANITIZER}  \
+      -DLLVM_BUILD_32_BITS=${BUILD_32_BITS}     \
+      -DLLVM_ENABLE_RUNTIMES='libcxx;libcxxabi;libunwind' \
+      -S llvm -B llvm-build -G "Unix Makefiles"
```
LebedevRI, line 23: Right, i guess it's `-S .` now.

dmah42: yep. working on it.  can't test locally so i'm iterating here :(

dmah42:
ok it builds much faster (yay) but still fails with the same error in `sysinfo`.

LebedevRI:
> ok it builds much faster (yay) but still fails with the same error in `sysinfo`.

Nice!
I'd suggest to drop the install stuff, and explicitly add `-stdlib=libc++ -L<llvm-build>/lib -lc++abi -Isystem<llvm-build>/include -Isystem<llvm-build>/include/c++/v1 -Wl,-rpath,<llvm-build>/lib` to CXXFLAGS.

dmah42:
welp. msan passes, now asan is failing.  it looks like an odr violation in libcxx so i tried suppression but failed. concerned that it is odr violation but coming from something in our code (the output tests maybe?)

LebedevRI:
Could you try passing `-DBUILD_SHARED_LIBS=OFF` (!) for sanitizer build jobs?

.github/workflows/sanitizer.yml
```diff
@@ -46,36 +45,27 @@ jobs:
...
-        echo "CC=gcc-10" >> $GITHUB_ENV
-        echo "CXX=g++-10" >> $GITHUB_ENV
-
-    - name: install llvm stuff
-      if: matrix.compiler == 'clang'
+    - name: build libc++
       run: |
         "${GITHUB_WORKSPACE}/.github/libcxx-setup.sh"
-        echo "EXTRA_CXX_FLAGS=\"-stdlib=libc++\"" >> $GITHUB_ENV
+        echo "EXTRA_CXX_FLAGS=-stdlib=libc++ -L ${GITHUB_WORKSPACE}/llvm-build/lib -lc++abi -Isystem${GITHUB_WORKSPACE}/llvm-build/include -Isystem${GITHUB_WORKSPACE}/llvm-build/include/c++/v1 -Wl,-rpath,${GITHUB_WORKSPACE}/llvm-build/lib" >> $GITHUB_ENV
```
LebedevRI, line 67: Ok, last guess:
Prepend `-nostdlibc++` before `-stdlib=libc++` here.

.github/workflows/sanitizer.yml
```diff
@@ -46,36 +45,27 @@ jobs:
...
-        echo "CC=gcc-10" >> $GITHUB_ENV
-        echo "CXX=g++-10" >> $GITHUB_ENV
-
-    - name: install llvm stuff
-      if: matrix.compiler == 'clang'
+    - name: build libc++
       run: |
         "${GITHUB_WORKSPACE}/.github/libcxx-setup.sh"
-        echo "EXTRA_CXX_FLAGS=\"-stdlib=libc++\"" >> $GITHUB_ENV
+        echo "EXTRA_CXX_FLAGS=-nostdlibc++ -stdlib=libc++ -L ${GITHUB_WORKSPACE}/llvm-build/lib -lc++abi -Isystem${GITHUB_WORKSPACE}/llvm-build/include -Isystem${GITHUB_WORKSPACE}/llvm-build/include/c++/v1 -Wl,-rpath,${GITHUB_WORKSPACE}/llvm-build/lib" >> $GITHUB_ENV
```
LebedevRI, line 67: Err, i guess it's apparently `-nostdlib++`, without the `c` letter.
Also, are you sure about the space after `-L`?

LebedevRI:
Ok, final guess: add `-fsanitize-address-use-odr-indicator` to *global* CXXFLAGS, when building both the libc++ and benchmark.

LebedevRI:
@dmah42 sorry for so much back-n-forth here.

LebedevRI:
> Ok, final guess: add `-fsanitize-address-use-odr-indicator` to _global_ CXXFLAGS, when building both the libc++ and benchmark.

.github/workflows/sanitizer.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - name: configure msan env
       if: matrix.sanitizer == 'msan'
       run: |
-        echo "EXTRA_FLAGS=-g -O2 -fno-omit-frame-pointer -fsanitize=memory -fsanitize-memory-track-origins" >> $GITHUB_ENV
+        echo "EXTRA_FLAGS=-g -O2 -fno-omit-frame-pointer -fsanitize=memory -fsanitize-memory-track-origins -fsanitize-address-use-odr-indicator" >> $GITHUB_ENV
```
LebedevRI, line 26: It should be for `if: matrix.sanitizer == 'asan'` block, though

dmah42: ugh. i forgot what i was doing over the weekend. sorry.

dmah42: this is the default for non-windows so i can drop it anyway

.github/workflows/sanitizer.yml
```diff
@@ -89,6 +79,7 @@ jobs:
         -DBENCHMARK_ENABLE_ASSEMBLY_TESTS=OFF
         -DBENCHMARK_ENABLE_LIBPFM=OFF
         -DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON
+        -DBUILD_SHARED_LIBS=OFF
```
LebedevRI, line 82: Not needed anymore

LebedevRI (approved):
Yay, ship it! (probably should be squash-merged)
Thank you.

LebedevRI (approved):

dmah42:
thanks for all the help!

PR #1594, opened 2023-05-10
upgrade python requirements to latest

dmah42:
i was having issues building //tools/... against the current requirements with python 3.10. upgrading them seems to have helped.


macandy13 (approved):
LGTM, thanks for fixing this!

PR #1596, opened 2023-05-10
Fix code triggering -Wsign-conversion

oontvoo:


oontvoo:
Not sure why some workflows failed with only "This job failed" message and no logs ... :thinking:

dmah42:
should we enable this warning in the build?

oontvoo:
> should we enable this warning in the build?

Yes - good point . let me give it a try

PR #1598, opened 2023-05-11
Update Python version to PY3, as indicated by the actual source file.

macandy13:
The code compiled in this binary is actually Python 3 code (see https://github.com/google/benchmark/blob/main/tools/compare.py).


dmah42:
thanks!

PR #1606, opened 2023-05-29
Add project name to CMake status message.

paulcaprioli:
When building `benchmark` as part of a larger CMake project, the meaning of the following status message is not clear:
```
-- Version: 1.8.0
```
This patch identifies Google benchmark as the project to which the version refers:
```
-- Google benchmark version: 1.8.0
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,7 +110,7 @@ else()
   set(VERSION "${GIT_VERSION}")
 endif()
 # Tell the user what versions we are using
-message(STATUS "Version: ${VERSION}")
+message(STATUS "Google benchmark version: ${VERSION}")
 
 # The version of the libraries
 set(GENERIC_LIB_VERSION ${VERSION})


LebedevRI:
I think there was some way to prefix all cmake on-screen output for sub-projects with a custom prefix?

But regardless, i think this probably makes sense, at least it's rather unintrusive.

paulcaprioli:
Thanks, Roman, I was unaware of `cmake --log-context`, which is useful if a larger project does something like this:
```
list(APPEND CMAKE_MESSAGE_CONTEXT "benchmark")
add_subdirectory(benchmark)
list(POP_BACK CMAKE_MESSAGE_CONTEXT)
```

I still think this PR makes sense, but I don't think I want to bother with the CLA process for such a trivial patch.   If someone who has completed it wishes to add the words "Google benchmark" to the message string, I think that would be nice.  I'm not a lawyer, and this is not legal advice, but I would venture to guess that Google has the right to add its name to the message string without my signature on a CLA.

dmah42:
thanks, i created a similar patch.

paulcaprioli:
If Google Benchmark is included in a project that uses git, then git naturally retrieves the git version of the enclosing project.  So, for example, one might see the following messages printed with the latest benchmark code:

```
git version: v0.1.0-32-g302c23a7-dirty normalized to 0.1.0.32
Google Benchmark version: 0.1.0.32
```

Now, maybe I should not be including a copy of Google Benchmark directly in my project, and I am OK if this is not something you want to support.

On the other hand, perhaps you might wish to change:

```
message(STATUS "Google Benchmark version: ${VERSION}")
```
to
```
message(STATUS "Google Benchmark version: ${benchmark_VERSION}")
```
since the git version is reported on the previous line anyway.

Regardless, I wish to thank Google and the various volunteers for creating and maintaining this project.

--Paul

LebedevRI:
The problem is as long it is determined from the git tag,
it will always be detected incorrectly in that situation,
regardless of how you name that variable.

paulcaprioli:
My suggestion for that status message is to ignore the information retrieved from the git tag.
In this project's CMakelists.txt file, the git tag is set in `GIT_VERSION` and, if nonzero, copied into `VERSION`.
The variable `benchmark_VERSION` is set at the top of the file by the `project()` function:
```
project (benchmark VERSION 1.8.3 LANGUAGES CXX)
```
Some short documentation is found at:
```
https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_VERSION.html
```

Issue #1006, opened 2020-07-21
[FR] human readable name of benchmark


karthikeyann:
https://github.com/google/benchmark#running-a-subset-of-benchmarks
As shown in the readme.md, 
https://github.com/google/benchmark/blame/master/README.md#L424

The benchmark name contains the state.range values as human readable format. (eg. `32k`)
But the current code does not provide any switches or automatic way to format these numbers to human readable format. 
```
Benchmark              Time           CPU Iterations
----------------------------------------------------
BM_memcpy/32          11 ns         11 ns   79545455
BM_memcpy/32k       2181 ns       2185 ns     324074
```

**Describe the solution you'd like**
Provide a function to enable human readable names for benchmark names. 
for powers of 2 and powers of 10.

PR #1607, opened 2023-06-07
Add `--benchmark_human_readable` flag

DiegoKrupitza:
Allows used to add a command line flag called `--benchmark_human_readable`. By adding this flag the arguments passed to benchmarks are formated in a human friendly format. This means that numbers that are the power of 2 are formatted as `2^x` (e.g., 64 will be `2^6`). For numbers that are the power of 10 a different formatting style is used. Numbers 0-999 no formatting is used. For numbers 1000-999999 the format `k` is used (e.g., `32000` -> `32k`). This also works for millions and billions. For numbers greater than 999 billion no special formatting is used. 
The design is rather simple allowing to by easily extendable.

Closes: https://github.com/google/benchmark/issues/1006


src/benchmark_api_internal.cc
```diff
@@ -43,7 +46,20 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
       }
     }
 
-    name_.args += StrFormat("%" PRId64, arg);
+    // formatting args either in default mode or in human-readable
+    auto formattedArgument = StrFormat("%" PRId64, arg);
+
+    if (FLAGS_benchmark_human_readable) {
+      if ((arg & (arg - 1)) == 0 && (arg > 1)) {
```
dmah42, line 53: can i suggest an `IsPowerOfTwo` function to allow this logic to read easier?

src/benchmark_api_internal.cc
```diff
@@ -43,7 +46,20 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
...
+        // power of two human format
+        formattedArgument = StrFormat("2^%.0f", std::log2(arg));
```
dmah42, line 55: `Base2HumanReadableFormat` could sit alongside `Base10HumanReadableFormat`

src/benchmark_api_internal.cc
```diff
@@ -43,7 +46,20 @@ BenchmarkInstance::BenchmarkInstance(Benchmark* benchmark, int family_idx,
       }
     }
 
-    name_.args += StrFormat("%" PRId64, arg);
+    // formatting args either in default mode or in human-readable
+    auto formattedArgument = StrFormat("%" PRId64, arg);
+
+    if (FLAGS_benchmark_human_readable) {
+      if ((arg & (arg - 1)) == 0 && (arg > 1)) {
+        // power of two human format
```
dmah42, line 54: this comment (and the one on line 57) add nothing to the code.. please remove.

src/benchmark_api_internal.cc
```diff
@@ -114,5 +130,55 @@ void BenchmarkInstance::Teardown() const {
...
+const std::array<Base10HumanReadableFormat, 5>
+    Base10HumanReadableFormat::ranges = {
+        Base10HumanReadableFormat{1, ""},             // from 0 to 10^3-1
+        Base10HumanReadableFormat{1000, "k"},         // from 10^3 to 10^6-1
+        Base10HumanReadableFormat{1000000, "m"},      // from 10^6 to 10^9-1
+        Base10HumanReadableFormat{1000000000, "bn"},  // from 10^9 to 10^12-1
+        Base10HumanReadableFormat{1000000000000, ""}  // above show as full
+};
+
+std::string Base10HumanReadableFormat::get(const int64_t& arg) {
```
dmah42, line 143: what's wrong with https://github.com/google/benchmark/blob/main/src/string_util.cc#L121 ?

src/benchmark_api_internal.h
```diff
@@ -10,10 +12,29 @@
 
 #include "benchmark/benchmark.h"
 #include "commandlineflags.h"
+#include "string_util.h"
 
 namespace benchmark {
 namespace internal {
 
+struct Base10HumanReadableFormat {
```
dmah42, line 20: this doesn't belong in the public interface of the internal api.  there are existing string format files that it might belong to.

src/benchmark_api_internal.cc
DiegoKrupitza: Oh I did not realise there is already a function for that. I will try to adapt this method such we can be used it for base10 (and have the correct abbreviations)

DiegoKrupitza:
I squashed all the commits into a single one such that there is only one commit to merge.

dmah42:
> I squashed all the commits into a single one such that there is only one commit to merge.

thanks.. i generally "squash and merge" anyway, but this means less editing of commit messages for me :D

src/benchmark_api_internal.h
```diff
@@ -10,6 +12,7 @@
 
 #include "benchmark/benchmark.h"
 #include "commandlineflags.h"
+#include "string_util.h"
```
dmah42, line 15: are these headers needed in this file?  i don't think there's a dependency being added to the interface.

src/string_util.cc
```diff
@@ -44,7 +51,8 @@ void ToExponentAndMantissa(double val, double thresh, int precision,
   // in 'precision' digits.
   const double adjusted_threshold =
       std::max(thresh, 1.0 / std::pow(10.0, precision));
-  const double big_threshold = adjusted_threshold * one_k;
+  const double big_threshold =
+      (adjusted_threshold * one_k) - inclusiveBigThreshhold;
```
dmah42, line 55: you appear to be subtracting a boolean from a double.. that's surprising and maybe you could make this explicit instead of relying on an implicit conversion.

src/string_util.h
```diff
@@ -65,6 +65,35 @@ using std::stoul;  // NOLINT(misc-unused-using-decls)
 #endif
 // NOLINTEND
 
+/**
+ * Check whether the given value is a power of two or not.
+ * @param val the value to check
+ */
+bool IsPowerOfTwo(const int64_t& val);
```
dmah42, line 72: do these need to be exposed in the API?  i think only `FormatHumanReadable` is accessed by a different implementation file.

test/human_readable_gtest.cc
```diff
@@ -0,0 +1,70 @@
...
+TEST(HumanReadableTest, base2) {
+  for (int i = 2; i < 10; ++i) {
+    const auto res = benchmark::Base2HumanReadableFormat(1 << i);
+
+    const std::string expected = "2^" + std::to_string(i);
+    EXPECT_STREQ(res.c_str(), expected.c_str());
+  }
+}
+
+TEST(HumanReadableTest, base10) {
```
dmah42, line 19: i'd prefer if this one test was split into a test per case.  having multiple test cases in a single test makes it harder to debug should one of the test cases fail.

src/string_util.h
DiegoKrupitza: Yes this method doesnt need to be exposed. Exposing Base10 and Base2 is needed for tests and may be needed by other functions in the future.

src/string_util.cc
DiegoKrupitza: I used this neat little hack to make the code a litte bit more concise. But I can make it more explicit and add a little bit of docu why this trick is needed.

DexterIV:
im not a big fan of "human readable format". Id consider binary output to be not human readable but not base10 numbers. Asked chatgpt what it "thinks" about that and here is the answer: 
`The current name "Human Readable Format" seems appropriate for describing the functionality of the option. However, if you are looking for an alternative name that captures the essence of the feature, you could consider something like "Formatted Number Display" or "Enhanced Numeric Representation."`
just a thing to consider :)

dmah42:
> im not a big fan of "human readable format". Id consider binary output to be not human readable but not base10 numbers. Asked chatgpt what it "thinks" about that and here is the answer: `The current name "Human Readable Format" seems appropriate for describing the functionality of the option. However, if you are looking for an alternative name that captures the essence of the feature, you could consider something like "Formatted Number Display" or "Enhanced Numeric Representation."` just a thing to consider :)

i think it's a pretty well-defined name at this point.  `df -h`, for instance:
```
       -h, --human-readable
              print sizes in powers of 1024 (e.g., 1023M)
```

or `du -h`:
```
       -h, --human-readable
              print sizes in human readable format (e.g., 1K 234M 2G)
```

src/benchmark_api_internal.h
```diff
@@ -1,6 +1,8 @@
 #ifndef BENCHMARK_API_INTERNAL_H
 #define BENCHMARK_API_INTERNAL_H
 
+#include <array>
```
dmah42, line 4: still seeing these two headers. i'm sure they don't need to be added and i'd rather minimise any dependencies.

src/string_util.cc
```diff
@@ -32,7 +38,8 @@ static const int64_t kUnitsSize = arraysize(kBigSIUnits);
 
 void ToExponentAndMantissa(double val, double thresh, int precision,
                            double one_k, std::string* mantissa,
-                           int64_t* exponent) {
+                           int64_t* exponent,
+                           bool inclusiveBigThreshhold = false) {
```
dmah42, line 42: seems like we only set this to true when we're doing base-10.  it might be worth having a comment why this is necessary for base-10.  or even better, renaming the variable to indicate that directly.

PR #1608, opened 2023-06-12
Fix pass rvalue to DoNotOptimize

### LINKED ISSUE #1584: [BUG] `DoNotOptimize(T &&)` should not be deprecated
**Describe the bug**

In https://github.com/google/benchmark/pull/1493, the `T const &` overload of `DoNotOptimize` was deprecated because the compiler is still able to make assumptions about the value of the variable. However, it is unclear whether this concern would apply to an rvalue reference overload.

**To reproduce**

```cpp
int f() {
    return 5;
}
void g() {
    benchmark::DoNotOptimize(f());
}
```

**Expected behavior**

No warning is emitted.

bgaifullin:
Fixes #1584 1584

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -465,7 +465,7 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 }
 
 template <class Tp>
-inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
+inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp&& value) {
 #if defined(__clang__)
   asm volatile("" : "+r,m"(value) : : "memory");
 #else
@@ -501,7 +501,7 @@ template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE
     typename std::enable_if<std::is_trivially_copyable<Tp>::value &&
                             (sizeof(Tp) <= sizeof(Tp*))>::type
-    DoNotOptimize(Tp& value) {
+    DoNotOptimize(Tp&& value) {
   asm volatile("" : "+m,r"(value) : : "memory");
 }
 
@@ -509,7 +509,7 @@ template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE
     typename std::enable_if<!std::is_trivially_copyable<Tp>::value ||
                             (sizeof(Tp) > sizeof(Tp*))>::type
-    DoNotOptimize(Tp& value) {
+    DoNotOptimize(Tp&& value) {
   asm volatile("" : "+m"(value) : : "memory");
 }
 
@@ -526,7 +526,7 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 }
 
 template <class Tp>
-inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
+inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp&& value) {
   asm volatile("" : "+m"(value) : : "memory");
 }
 #endif


include/benchmark/benchmark.h
```diff
@@ -465,7 +465,7 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 }
 
 template <class Tp>
-inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
+inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp&& value) {
```
dmah42, line 468: i believe this should be inside the defined(BENCHMARK_HAS_CXX11) section below

bgaifullin: fixed

include/benchmark/benchmark.h
```diff
@@ -465,7 +472,7 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 }
 
 template <class Tp>
-inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
```
LebedevRI, line 468: But this currently takes a simple reference?
https://godbolt.org/z/z3GP44jnG

bgaifullin: The magic is in using `template <typename T>`
compiler collapses extra '&'.  if T is alread 'int&', 'T&&' - 'int&&&' transforms into 'int&'. This is universal reference.
see https://godbolt.org/z/eEfxfa8E6

bgaifullin:
windows build failed with error
```
Visual C++ build tools seems to be installed at C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC 
But Bazel can't find the following tools: 
    VCVARSALL.BAT, cl.exe, link.exe, lib.exe, ml64.exe 
for x64 target architecture 
```

seems like it does not related to current PR

dmah42:
> windows build failed with error
> 
> ```
> Visual C++ build tools seems to be installed at C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC 
> But Bazel can't find the following tools: 
>     VCVARSALL.BAT, cl.exe, link.exe, lib.exe, ml64.exe 
> for x64 target architecture 
> ```
> 
> seems like it does not related to current PR

that looks suspiciously like github changed something in the windows image.

dmah42:
> > windows build failed with error
> > ```
> > Visual C++ build tools seems to be installed at C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC 
> > But Bazel can't find the following tools: 
> >     VCVARSALL.BAT, cl.exe, link.exe, lib.exe, ml64.exe 
> > for x64 target architecture 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > seems like it does not related to current PR
> 
> that looks suspiciously like github changed something in the windows image.

https://github.com/actions/runner-images/issues/7662 may be related

dmah42:
> @dmah42 How to restart checks (seems like issue for windows has been resolved) ?

i've restarted them.

dmah42:
still failing, but i'm comfortable that a) it's an issue we can't do much about and b) unlikely to be an issue based on your PR.

include/benchmark/benchmark.h
```diff
@@ -465,7 +465,13 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
...
+inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(
+#ifdef BENCHMARK_HAS_CXX11
```
dmah42, line 469: we have a BENCHMARK_HAS_CXX11 block below.  can we not simplify this somehow so this is folded into the version at line 510?

bgaifullin: I think about this, but in case of reordering blocks, the behaviour will be changed.

on other hand, if the BENCHMARK_HAS_CXX11 related code is moved to separate block - it causes code duplication (

dmah42 (dismissed):

LebedevRI:
@bgaifullin thank you!
While we are at it, could you please add a test so this does not get re-broken, please?

bgaifullin:
The windows build still failing with error about missing compiler binary. ((

LebedevRI (approved):

LebedevRI:
@dmah42 can the failing CI runs be un-required?

PR #1609, opened 2023-06-12
Use std::function in ComputeStatistics()

### LINKED ISSUE #1604: [FR] Aggregated statistics for all threads in a benchmark
First Issue. I would like to report the speedup (T(1) / T(n)) when using multiple threads with ThreadRange(). There exists ComputeStatistics(), but it only contains values for one specific thread. It seems there is no way to get all values for a single named benchmark. Please correct me if I´m wrong.

Ideally, there should be a method similiar to ComputeStatistics() that aggregates all values for all threads.

Maybe access to the State can be provided in ComputeStatistics(). This way, at least each run can be identified by its thread_index and its name. Values can then be stored externally.

Tasemo:
Uses std::function instead of function pointers for the callback provided with ComputeStatistics(), if C++11 is used. This allows for greater flexibility as the user can use closures, member functions, bind expressions, etc. This should be a non API breaking change as std::function implicitly stores function pointers as well.

Closes #1604. Instead of a dedicated function, all the values can now be aggregated by the user. This should be the easier change.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -63,6 +63,7 @@ Shuo Chen <chenshuo@chenshuo.com>
 Staffan Tjernstrom <staffantj@gmail.com>
 Steinar H. Gunderson <sgunderson@bigfoot.com>
 Stripe, Inc.
+Tim Oelkers <tim.oelkers@web.de>
 Tobias Schmidt <tobias.schmidt@in.tum.de>
 Yixuan Qiu <yixuanq@gmail.com>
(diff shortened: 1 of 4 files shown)


LebedevRI:
I'm not sure this is that simple.

If library is used (not compiled!) by C++03 code,
object `compute_` in `Statistics` type would be a pointer,
but once `Statistics` is passed into the library's internals,
it will silently become a different `Statistics` type,
one where `compute_` object is suddenly a `std::function`.
But `std::function` was never constructed?

PR #1610, opened 2023-06-13
Fix a typo in regex choice

Chilledheart:
BENCHMARK_HAVE_STD_REGEX is not used but HAVE_STD_REGEX like the other two choices, i.e. HAVE_GNU_POSIX_REGEX and HAVE_POSIX_REGEX.


dmah42:
oh wow, great catch thank you!

dmah42 (approved):

PR #1615, opened 2023-06-23
Add support for bzlmod (excluding Python bindings)

macandy13:
This is the first part of migrating google_benchmark to support bzlmod, the new module dependency management system. This part of the migration focuses on migrating all targets in //:all, //tools:all. It excludes the //bindings/.. targets for now, as those are depending on some changes in rules_python.

Overarching issue: https://github.com/bazelbuild/bazel-central-registry/issues/384


MODULE.bazel
```diff
@@ -0,0 +1,34 @@
+module(name = "com_github_google_benchmark", version="1.8.0")
+
+bazel_dep(name = "bazel_skylib", version = "1.4.1")
+bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "rules_foreign_cc", version = "0.9.0")
+bazel_dep(name = "rules_cc", version = "0.0.6")
+bazel_dep(name = "rules_python", version = "0.23.1")
+bazel_dep(name = "googletest", version = "1.12.1", repo_name = "com_google_googletest")
+bazel_dep(name = "libpfm", version = "4.11.0")
```
dmah42, line 9: this is an optional dependency. is there any way to mark that?

WORKSPACE
```diff
@@ -11,8 +11,8 @@ rules_foreign_cc_dependencies()
 load("@rules_python//python:pip.bzl", pip3_install="pip_install")
 
 pip3_install(
-   name = "py_deps",
-   requirements = "//:requirements.txt",
+   name = "tools_pip_deps",
+   requirements = "//tools:requirements.txt",
```
dmah42, line 15: i believe the "standard" for pip is to have the requirements in the root.  is that not the case?  and why tools and not bindings/python?

WORKSPACE.bzlmod
```diff
@@ -0,0 +1,2 @@
+# This file marks the root of the Bazel workspace.
```
dmah42, line 1: i thought `WORKSPACE` marked the root of the Bazel workspace.  why do we need a second WORKSPACE file?

MODULE.bazel
macandy13: I don't think there is a way to express optional dependencies in the new system. But from what I found from my experiments is that the dependency will only be downloaded into the bazel external repo cache when it is actually needed, and that is only the case when the pfm flag is set.

WORKSPACE
macandy13: Yeah, but in this repo there are multiple `requirements.txt` files in various directories - and the interesting thing is that the requirements in `tools/requirements` seems to be unused, but the pip requirements in the root dir are only applicable for the targets in `tools`.

WORKSPACE.bzlmod
macandy13: This is for bzlmod, so that it replaces the existing WORKSPACE file when bzlmod is enabled, so it doesn't confuse the build system. 

See also https://docs.google.com/document/d/1JtXIVnXyFZ4bmbiBCr5gsTH4-opZAFf5DMMb-54kES0/edit#heading=h.y054fjub9max

MODULE.bazel
dmah42: ah, i wasn't sure if it was something that would look for a local matching version (installed by the OS package manager) or that it would download.  if it's the latter, then yes that's safer, even if it does download it.

WORKSPACE
dmah42: ah you're right. i haven't paid much attention to the python bits. thanks.

dmah42 (approved):

dmah42:
is there anything we need to do with the github action bots?

macandy13:
Sorry for the delay, was ooo for the last days. 

Good point about the github actions. I think we should at least have a single environment in which we run with the `--enable_bzlmod` flag being set so this won't break in the future. Will take a look at it soon.

PR #1617, opened 2023-06-30
Add a CI test for the new bzlmod integration

macandy13:
Enhanced the already existing bazel test to also run with the `--enable_bzlmod` flag.

Technically it shouldn't be necessary to run these additional test on all operating systems, but this will make the eventual switch to exclusively use bzlmod (slightly) easier.

The one thing that is unclear is if there should be another test for the `--define pfm=1` flag, which uses more libraries. This is missing for the base case and is currently only exercises in the cmake CI workflows, so it might be worth a separate PR to introduce those tests (for both variants then).

Related issue: https://github.com/bazelbuild/bazel-central-registry/issues/384

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -33,3 +33,32 @@ jobs:
     - name: test
       run: |
         bazel test --test_output=all //test/...
+
+  job:
+    name: bazel_with_bzlmod
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-2022]
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: mount bazel cache
+      uses: actions/cache@v3
+      env:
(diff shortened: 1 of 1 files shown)


dmah42:
agreed, we should probably test pfm, but not in this PR.

i wonder if there's a way to simplify this by having a matrix-level setting for bzlmod and enabling or disabling the flag based on that?

macandy13:
Good point, I will dig a bit deeper into the syntax and update the pr.

dominic ***@***.***> schrieb am Fr., 30. Juni 2023, 10:59:

macandy13:
Ok, I think I have a nice solution now (although the "&& ||" trick for emulating a ternary condition is definitely debatable ;). Also tried to incorporate pfm but got hit by the failing test with pfm-enabled, so I'll probably keep that for another PR where the test would be fixed as well.

PTAL.

dmah42 (approved):

dmah42:
i need to update the "expected tests" to cover these too.  merging without waiting for the non-existant tests.

PR #1622, opened 2023-07-04
Fix passing non-const lvalue refs to DoNotOptimize

### LINKED ISSUE #1619: [BUG] DoNotOptimize `const &` overload chosen for non-const lvalue references
**Describe the bug**
After df9a99d998d9f038a53a970a77ce31c8c2a882ce, `const &` overloads are chosen where previously the `&` overload was chosen. This leads to deprecation warnings where IMO there should be none. `const &` takes precedence over the new `&&` overload.

Godbolt: https://godbolt.org/z/5s6sKT7dc

<details><summary>Click to show code from godbolt example</summary>

```cpp

#include <algorithm>
#include <vector>

static void test(benchmark::State & state)
{
    std::vector<int> vec(100);

    for (auto _ : state)
    {
        std::fill(vec.begin(), vec.end(), 1);
        benchmark::DoNotOptimize(vec);
    }
}
BENCHMARK(test);

BENCHMARK_MAIN();
```

```
<source>: In function 'void test(benchmark::State&)':
<source>:13:33: error: 'typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type benchmark::DoNotOptimize(const Tp&) [with Tp = std::vector<int>; typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type = void]' is deprecated: The const-ref version of this method can permit undesired compiler optimizations in benchmarks [-Werror=deprecated-declarations]
   13 |         benchmark::DoNotOptimize(vec);
      |         ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
In file included from <source>:1:
/opt/compiler-explorer/libs/google-benchmark/trunk/include/benchmark/benchmark.h:502:5: note: declared here
  502 |     DoNotOptimize(Tp const& value) {
      |     ^~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

</details>

**System**
  - OS: Ubuntu 22.04.2 LTS
  - Compiler and version: gcc version 12.3.0 (locally), 13.1.0 (godbolt)

**To reproduce**
See bug description.

**Expected behavior**
IMO, it should still work. There probably needs to be both `&` and `&&` overloads. It does work if I add additional `&` overloads.

eseiler:
Resolves #1619 
I re-added the `&` overloads that were changed to `&&` in df9a99d998d9f038a53a970a77ce31c8c2a882ce. New tests do not seem necessary because the existing tests already emitted the deprecation warnings.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -122,6 +122,11 @@ compile_benchmark_test(skip_with_error_test)
 add_test(NAME skip_with_error_test COMMAND skip_with_error_test --benchmark_min_time=0.01s)
 
 compile_benchmark_test(donotoptimize_test)
+# Enable errors for deprecated deprecations (DoNotOptimize(Tp const& value)).
+check_cxx_compiler_flag(-Werror=deprecated-declarations BENCHMARK_HAS_DEPRECATED_DECLARATIONS_FLAG)
+if (BENCHMARK_HAS_DEPRECATED_DECLARATIONS_FLAG)
+  target_compile_options (donotoptimize_test PRIVATE "-Werror=deprecated-declarations")
+endif()
 # Some of the issues with DoNotOptimize only occur when optimization is enabled
 check_cxx_compiler_flag(-O3 BENCHMARK_HAS_O3_FLAG)
 if (BENCHMARK_HAS_O3_FLAG)
--- a/cmake/GoogleTest.cmake
+++ b/cmake/GoogleTest.cmake
@@ -29,18 +29,24 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 
 include(${GOOGLETEST_PREFIX}/googletest-paths.cmake)
 
-# googletest doesn't seem to want to stay build warning clean so let's not hurt ourselves.
-if (MSVC)
-  add_compile_options(/wd4244 /wd4722)
-else()
-  add_compile_options(-w)
-endif()
-
 # Add googletest directly to our build. This defines
 # the gtest and gtest_main targets.
 add_subdirectory(${GOOGLETEST_SOURCE_DIR}
                  ${GOOGLETEST_BINARY_DIR}
                  EXCLUDE_FROM_ALL)
+
+# googletest doesn't seem to want to stay build warning clean so let's not hurt ourselves.
+if (MSVC)
+  target_compile_options(gtest PRIVATE "/wd4244" "/wd4722")
+  target_compile_options(gtest_main PRIVATE "/wd4244" "/wd4722")
+  target_compile_options(gmock PRIVATE "/wd4244" "/wd4722")
+  target_compile_options(gmock_main PRIVATE "/wd4244" "/wd4722")
+else()
+  target_compile_options(gtest PRIVATE "-w")
+  target_compile_options(gtest_main PRIVATE "-w")
+  target_compile_options(gmock PRIVATE "-w")
+  target_compile_options(gmock_main PRIVATE "-w")
+endif()
 
 if(NOT DEFINED GTEST_COMPILE_COMMANDS)
     set(GTEST_COMPILE_COMMANDS ON)
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -465,19 +465,24 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 }
 
 template <class Tp>
-inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(
-#ifdef BENCHMARK_HAS_CXX11
-    Tp&& value
-#else
-    Tp& value
-#endif
-) {
+inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
 #if defined(__clang__)
   asm volatile("" : "+r,m"(value) : : "memory");
 #else
   asm volatile("" : "+m,r"(value) : : "memory");
 #endif
 }
+
+#ifdef BENCHMARK_HAS_CXX11
+template <class Tp>
+inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp&& value) {
+#if defined(__clang__)
+  asm volatile("" : "+r,m"(value) : : "memory");
+#else
+  asm volatile("" : "+m,r"(value) : : "memory");
+#endif
+}
+#endif
 #elif defined(BENCHMARK_HAS_CXX11) && (__GNUC__ >= 5)
 // Workaround for a bug with full argument copy overhead with GCC.
 // See: #1340 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105519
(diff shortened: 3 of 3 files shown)


LebedevRI:
Thank you!
This certainly needs tests, since we are clearly missing them.

LebedevRI:
> This certainly needs tests, since we are clearly missing them.

What i mean is that we should have noticed those new deprecation warnings.
(Is CI building without `-Werror`?) So `-Werror` should be always enabled for those tests.

eseiler:
I'm not familiar with Bazel, and don't know where and how compiler flags are set. For CMake, the deprecations warnings are disabled:

https://github.com/google/benchmark/blob/408f9e06676ffdc226adf652e013aa15861589ad/CMakeLists.txt#L192

LebedevRI:
> I'm not familiar with Bazel, and don't know where and how compiler flags are set.

Yep, no idea whatsoever either.

> For CMake, the deprecations warnings are disabled:

Well, my point is, we should have *some* test, which is always compiled with `-Wdeprecated -Werror`,
that would have caught this bug.

LebedevRI:
TBN, not sure we can globally enable `-Wdeprecated -Werror`,
because that should also break tests for methods that are deprecated...

eseiler:
> Well, my point is, we should have _some_ test, which is always compiled with `-Wdeprecated -Werror`, that would have caught this bug.

I see. We would need some way to explicitly build the library with those flags. Setting compile flags for a test does not affect the library (where the error would be emitted from). For now, I just tried not disabling those warnings. Works locally - on main an error is emitted, on this PR not. The flags were apparently disabled during some refactoring https://github.com/google/benchmark/blame/main/CMakeLists.txt#L192.
Maybe they are not necessary anymore.

LebedevRI:
> > Well, my point is, we should have _some_ test, which is always compiled with `-Wdeprecated -Werror`, that would have caught this bug.
> 
> I see. We would need some way to explicitly build the library with those flags. Setting compile flags for a test does not affect the library (where the error would be emitted from).

That statement makes no sense, isn't the exact opposite is true?

eseiler:
> That statement makes no sense, isn't the exact opposite is true?

Yes, it's nonsense :)

If found that `target_compile_options (donotoptimize_test PRIVATE "-Wdeprecated-declarations")` works.

However, 
https://github.com/google/benchmark/blob/408f9e06676ffdc226adf652e013aa15861589ad/cmake/GoogleTest.cmake#L36
Causes tests to be compiled with `-w`, so nothing will be emitted.

I changed it locally to 
```
  target_compile_options(gtest PRIVATE "-w")
  target_compile_options(gtest_main PRIVATE "-w")
  target_compile_options(gmock PRIVATE "-w")
  target_compile_options(gmock_main PRIVATE "-w")
```
which will cause the `donotoptimize_test` to emit deprecation warnings.

LebedevRI:
> However,
> 
> https://github.com/google/benchmark/blob/408f9e06676ffdc226adf652e013aa15861589ad/cmake/GoogleTest.cmake#L36
> 
> 
> Causes tests to be compiled with `-w`, so nothing will be emitted.

Hm, are you sure? That is only for building the googletest library itself,
it should not propagate to the tests that link to that library.

eseiler:
> Hm, are you sure? That is only for building the googletest library itself, it should not propagate to the tests that link to that library.

On main, CMakeLists.txt does this:
```cmake
if (BENCHMARK_ENABLE_TESTING)
  enable_testing()
  if (BENCHMARK_ENABLE_GTEST_TESTS AND
      NOT (TARGET gtest AND TARGET gtest_main AND
           TARGET gmock AND TARGET gmock_main))
    if (BENCHMARK_USE_BUNDLED_GTEST)
      include(GoogleTest)
    else()
      # ...
    endif()
  endif()
  add_subdirectory(test)
endif()
```

`include(GoogleTest)` shoudl refer to `GoogleTest.cmake`. It will set `-w` which then is in effect for everything that follows. So, `-w` is set for `add_subdirectory(test)`.
`add_compile_options(-w)` applies to the directory, which is in this case the root directory, because the include is called from the root CMakeLists.txt.

Instead of changing the `add_compile_options(-w)`, I can also just move the `add_subdirectory(test)` up such that the test directory is added before the `-w` flag is set.

test/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ compile_benchmark_test(skip_with_error_test)
 add_test(NAME skip_with_error_test COMMAND skip_with_error_test --benchmark_min_time=0.01s)
 
 compile_benchmark_test(donotoptimize_test)
+target_compile_options (donotoptimize_test PRIVATE "-Wdeprecated-declarations")
```
LebedevRI, line 125: We really want `-Werror` too.
But, this all needs to first check with `check_cxx_compiler_flag()`.

LebedevRI:
Huh, okay, that does appear broken indeed :|
Let's try to go with `target_compile_options()`?

test/CMakeLists.txt
eseiler: Done. I did both at once with `-Werror=deprecated-declarations`. If it's preferable to do `-Werror` and `-Wdeprecated-declarations` separately, I can also do so.

CMakeLists.txt
```diff
@@ -321,6 +321,8 @@ add_subdirectory(src)
 
 if (BENCHMARK_ENABLE_TESTING)
   enable_testing()
+  # The test directory must be added before include(GoogleTest). GoogleTest.cmake sets -w globally.
+  add_subdirectory(test)
```
LebedevRI, line 325: Does CMake support delayed target parsing?
In the sense, tests in `test/` link to `gtest`/`gtest_main`/`gmock`/`gmock_main` targets,
but said targets won't exist yet, because we now add tests before creating them.
Does that work in CMake? I've literally never seen that before.

eseiler: It does work for `target_link_libraries`. It does not work for modifying the target. E.g., adding `target_compile_options (gtest PRIVATE "-w")` in `test/CMakeLists.txt` will fail because the `gtest` target is not yet defined.

Alternatively, the `-w` could only be set for the gtest/gmock targets.

LebedevRI: I think we should fix the `-w` bug while we are at it, globally, not just for our particular problem.
Could you please try changing the current `add_compile_options(-w)` logic to the `target_compile_options()`?

LebedevRI: > Could you please try changing the current add_compile_options(-w) logic to the target_compile_options()?

(i'm **hoping** it would just work, and not uncover hundreds of unintentionally-silenced warnings...)

LebedevRI:
(if this doesn't get merged in next ~24 hours we should revert https://github.com/google/benchmark/pull/1608 / https://github.com/google/benchmark/issues/1584.)

CMakeLists.txt
eseiler: With the current change, the warnings are already unsilenced. I didn't get any warnings for the whole test suite locally. So far so good :)

I'll only be able to incorporate the changes this afternoon (which is in 6 hours for me).

dmah42:
> (if this doesn't get merged in next ~24 hours we should revert #1608 / #1584.)

i don't see why we can't merge this as soon as the bots are happy.

CMakeLists.txt
eseiler: Done

eseiler:
Formatting should be fixed

LebedevRI:
@eseiler so i'm looking at the `main` branch, and if i do the `cmake/GoogleTest.cmake` change locally,
no test fails with those false-positive deprecation warnings. So we are missing a test for the bug this is fixing...

eseiler:
Mhm, when I remove the benchmark.h changes, there are multiple errors in the `donotoptimize` test on my machine.

I guess I'll have to double check

LebedevRI:
> 

I just tried that, and i don't see any (clang16 + release build mode)

eseiler:
Apparently, clang prefers the `&&` overload anyway.
I get the error with GCC, but not with clang.

So, for clang the `&` overload could be removed. But the same overload is also used for Intel compiler; and I don't know whether the Intel compiler needs the `&` overload.

LebedevRI:
@eseiler thank you!

I've verified that this addresses false-positives i was seeing,
so since `1.8.1` was a dud, so after this is merged,
it might be a good idea to expedite `.2`.

dmah42:
thanks

dmah42:
(i'll release 1.8.2 tomorrow my time.. in about 12 hours)

LebedevRI:
@eseiler @dmah42 thank you!

Issue #1491, opened 2022-09-20
Null hypothesis test discrepancy


vicentebolea:
https://github.com/google/benchmark/blob/49aa374da96199d64fd3de9673b6f405bbc3de3e/tools/gbench/report.py#L212

As far as I understand, in google benchmark `compare.py` script we set:

- The null hypothesis as the baseline and the contender to have the same results.
- The alternative hypothesis is the contrary, that they are different.

In a null hypothesis we expect:

A null hypothesis to be rejected when the `pvalue` of the contender results in the baseline distribution is less than the alpha value (`alpha = 1-confidence level`).

However, in the source code we determine that the results are different when `pvalue >= alpha`. This is the reverse of a what we are looking for.

On a similar question for performance level, we want one side hypothesis test, since getting less execution time than the average should not be considered a failure.

I wonder if I am missing something here, I also wonder why using `mannwhitneyu` instead of a normal distribution since `mannwhitneyu` results are good for small numbers with a larger N.

LebedevRI:
Presumably you have read #593?

vicentebolea:
@LebedevRI thanks for the prompt reply. I just went through the thread, great read. And I agree with your initial idea of using tdist. However, it still does not answer of the reason why in the code we reject the null hypothesis on pval >= alpha. Even in the command help string it says:

```
 if the calculated p-value is
                        below this value, then the result is said to be
                        statistically significant and the null hypothesis is
                        rejected. (default: 0.0500)
```

Not that I am not an expert in data analysis and probability statistics, I just want to understand this.

LebedevRI:
That test is there to determine whether or not we can confidently say that two measurement sets
aren't actually the same measurement set. We get large p-value when we *can not* say that,
and small value when we can say that. So we want small p-value, and that is exactly what that line does.

vicentebolea:
@LebedevRI 

Here is where I am confused, we set the line as red (failure) when the large pvalue (results are same) and green when results are different. It is counter-intuitive since if your benchmarks are much slower or much faster you will get a green result then. I think that green should mean that the results are statistically similar to the baseline.

LebedevRI:
> Here is where I am confused, we set the line as red (failure) when the large pvalue (results are same) and green when results are different. It is counter-intuitive since if your benchmarks are much slower or much faster you will get a green result then.

Correct. That's the whole point.

> I think that green should mean that the results are statistically similar to the baseline.

Do you believe the documentation is backwards to what we do?

[...]

PR #1624, opened 2023-07-06
Update tools.md for missing color meaning issue #1491

varshneydevansh:
## Issue > 
Null hypothesis test discrepancy #1491

## Solution Proposed >  
documentation needs to be enhanced

## Changes made >
Added the lines to the tools.md file which explains the following -

  1. Clarify the meaning of the colors in the output.
  2. Defining the 'success' and 'failure' explicitly.
  3. Clarify the interpretation of the p-value and the null hypothesis.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -241,6 +241,8 @@ A statistically significant difference is determined by a **p-value**, which is 
   1. If the p-value is less than the chosen significance level (alpha), we reject the null hypothesis and conclude the benchmarks are significantly different.
   2. If the p-value is greater than or equal to alpha, we fail to reject the null hypothesis and treat the two benchmarks as similar.
 
+Also please note that **even if** we determine that there **is** a statistically-significant difference between the two measurements, it does not _necessarily_ mean that the actual benchmarks that were measured **are** different, or vice versa, even if we determine that there is **no** statistically-significant difference between the two measurements, it does not necessarily mean that the actual benchmarks that were measured **are not** different.
+
 
 
 ### U test
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -216,8 +216,8 @@ Let's first try to see what the different columns represent in the above `compar
 
 In the comparison section, the relative differences in both time and CPU time are displayed for each input size.
 
-When comparing benchmarks, `compare.py` uses statistical tests to determine whether the performance has significantly changed. 
-This is communicated through color coding:
+When comparing benchmarks, `compare.py` uses statistical tests to determine whether there is a statistically-significant difference between the measurements being compared.
+The result of said the statistical test is additionally communicated through color coding:
 ```diff
 + Green:
 ```
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -188,6 +188,34 @@ As you can note, the values in `Time` and `CPU` columns are calculated as `(new 
 
 ## Note: Interpreting the output
 
+Performance measurements are an art, and performance comparisons are doubly so.
+Results are often noisy and don't necessarily have large absolute differences to them,
+so just by visual inspection, it is not at all apparent if two measurements are actually showing
+a performance change or not. It is even more confusing with multiple benchmark repetitions.
+
+Thankfully, what we can do, is use statistical tests on the results
+to determine whether the performance has statistically significantly changed.
+`compare.py` uses [Mann–Whitney U test](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test),
+with a null hypothesis being that there's no difference in performance.
+
+
+Let's first try to see what the different columns represent in the above `compare.py` benchmarking output:
+
+  1. **Benchmark:** The name of the function being benchmarked, along with the size of the input (after the slash).
+
+  2. **Time:** The average time per operation, across all iterations.
+
+  3. **CPU:** The average CPU time per operation, across all iterations.
+
+  4. **Iterations:** The number of iterations the benchmark was run to get a stable estimate.
+
+  5. **Time Old and Time New:** These represent the average time it takes for a function to run in two different scenarios or versions. For example, you might be comparing how fast a function runs before and after you make some changes to it.
+
+  6. **CPU Old and CPU New:** These show the average amount of CPU time that the function uses in two different scenarios or versions. This is similar to Time Old and Time New, but focuses on CPU usage instead of overall time.
+
+
+In the comparison section, the relative differences in both time and CPU time are displayed for each input size.
+
 When comparing benchmarks, `compare.py` uses statistical tests to determine whether the performance has significantly changed. 
 This is communicated through color coding:
 ```diff
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -186,9 +186,22 @@ Benchmark                               Time             CPU      Time Old      
 This is a mix of the previous two modes, two (potentially different) benchmark binaries are run, and a different filter is applied to each one.
 As you can note, the values in `Time` and `CPU` columns are calculated as `(new - old) / |old|`.
 
+### Note: Interpreting the output
 
-4. The below output is a summary of a benchmark comparison with statistics
-provided for a multi-threaded process.
+Performance measurements are an art, and performance comparisons are doubly so.
+Results are often noisy and don't necessarily have large absolute differences to
+them, so just by visual inspection, it is not at all apparent if two
+measurements are actually showing a performance change or not. It is even more
+confusing with multiple benchmark repetitions.
+
+Thankfully, what we can do, is use statistical tests on the results to determine
+whether the performance has statistically-significantly changed. `compare.py`
+uses [Mann–Whitney U
+test](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test), with a null
+hypothesis being that there's no difference in performance.
+ 
+**The below output is a summary of a benchmark comparison with statistics
+provided for a multi-threaded process.**
 ```
 Benchmark                                               Time        CPU    Time Old      Time New       CPU Old       CPU New
 -----------------------------------------------------------------------------------------------------------------------------
@@ -199,7 +212,7 @@ benchmark/threads:1/process_time/real_time_stddev    +0.3974    +0.3933         
 benchmark/threads:1/process_time/real_time_cv        +0.6329    +0.6280           0             0             0             0
 OVERALL_GEOMEAN                                      -0.1442    -0.1442           0             0             0             0
 ```
-
+--------------------------------------------
 Here's a breakdown of each row:
 
 **benchmark/threads:1/process_time/real_time_pvalue**: This shows the _p-value_ for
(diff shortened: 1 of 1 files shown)


docs/tools.md
```diff
@@ -186,6 +186,35 @@ Benchmark                               Time             CPU      Time Old
 This is a mix of the previous two modes, two (potentially different) benchmark binaries are run, and a different filter is applied to each one.
 As you can note, the values in `Time` and `CPU` columns are calculated as `(new - old) / |old|`.
 
+## Note: Interpreting the output
+
```
LebedevRI, line 190: This could use a snippet of `compare.py` output that contains said `pvalue` output.

docs/tools.md
```diff
@@ -186,6 +186,35 @@ Benchmark                               Time             CPU      Time Old
...
+When comparing benchmarks, `compare.py` uses statistical tests to determine whether the performance has significantly changed.
```
LebedevRI, line 191: I think this should start with a more human-facing explanation first.
Something to the effect of:

```
Performance measurements are an art, and performance comparisons are doubly so.
Results are often noisy, and don't necessarily have large absolute difference between them,
so just by visual inspection, it is not at all apparent if two measurements are actually showing
a performance change or not. It is even more confusing with multiple benchmark repetitions.

Thankfully, what we can do, is use statistical tests on the results
to determine whether the performance has statistically-significantly changed.
`compare.py` uses [Mann–Whitney U test](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test),
with a null hypothesis being that there's no difference in performance.
```
and streamline the rest.

LebedevRI (commented):
Thank you for looking into this!
I'm not sure how to best write this, but i do have some thoughts.

docs/tools.md
```diff
@@ -186,6 +186,63 @@ Benchmark                               Time             CPU      Time Old
...
+  4. **Iterations:** The number of iterations the benchmark was run to get a stable estimate.
+
+  5. **Time Old and Time New:** These represent the average time it takes for a function to run in two different scenarios or versions. For example, you might be comparing how fast a function runs before and after you make some changes to it.
+
+  6. **CPU Old and CPU New:** These show the average amount of CPU time that the function uses in two different scenarios or versions. This is similar to Time Old and Time New, but focuses on CPU usage instead of overall time.
+
+
+In the comparison section, the relative differences in both time and CPU time are displayed for each input size.
+
+When comparing benchmarks, `compare.py` uses statistical tests to determine whether the performance has significantly changed.
```
LebedevRI, line 219:
```
When comparing benchmarks, `compare.py` uses statistical tests to determine whether there is a statistically-significant difference between the measurements being compared.
The result of said statistical test is additionally communicated through color coding:
```

docs/tools.md
```diff
@@ -186,6 +186,63 @@ Benchmark                               Time             CPU      Time Old
...
+    'Success' means 'successfully finding a difference'.
+    'Failure' means 'failing to find a difference'.
+```
+A statistically significant difference is determined by a **p-value**, which is a measure of the probability that the observed difference could have occurred just by random chance. A smaller p-value indicates stronger evidence against the null hypothesis. 
+
+**Therefore:**
+  1. If the p-value is less than the chosen significance level (alpha), we reject the null hypothesis and conclude the benchmarks are significantly different.
+  2. If the p-value is greater than or equal to alpha, we fail to reject the null hypothesis and treat the two benchmarks as similar.
+
+
```
LebedevRI, line 244:
```
Also please note that **even if** we determine that there **is** a statistically-significant difference
between the two measurements, it does not _necessarily_ mean that the actual benchmarks
that were measured **are** different, or vice versa, even if we determine that
there is **no** statistically-significant difference between the two measurements,
it does not necessarily mean that the actual benchmarks that were measured **are not** different.
```

docs/tools.md
```diff
@@ -186,6 +186,65 @@ Benchmark                               Time             CPU      Time Old
...
 
+## Note: Interpreting the output
+
+Performance measurements are an art, and performance comparisons are doubly so.
+Results are often noisy and don't necessarily have large absolute differences to them,
+so just by visual inspection, it is not at all apparent if two measurements are actually showing
+a performance change or not. It is even more confusing with multiple benchmark repetitions.
+
+Thankfully, what we can do, is use statistical tests on the results
+to determine whether the performance has statistically significantly changed.
```
LebedevRI, line 197: statistically-significantly

docs/tools.md
```diff
@@ -186,6 +186,65 @@ Benchmark                               Time             CPU      Time Old
...
+  But remember, in this context:
+
+    'Success' means 'successfully finding a difference'.
+    'Failure' means 'failing to find a difference'.
+```
+A statistically significant difference is determined by a **p-value**, which is a measure of the probability that the observed difference could have occurred just by random chance. A smaller p-value indicates stronger evidence against the null hypothesis. 
+
+**Therefore:**
+  1. If the p-value is less than the chosen significance level (alpha), we reject the null hypothesis and conclude the benchmarks are significantly different.
+  2. If the p-value is greater than or equal to alpha, we fail to reject the null hypothesis and treat the two benchmarks as similar.
```
LebedevRI, line 242: I think this should go before the section that talks about the color coding.

varshneydevansh (commented):
Thanks for your guidance, Roman. The starting paragraph felt so interesting while reading, and the changes you suggested made this more well versed.

It was the first time that I came across about benchmarking, learned so much in the process.

If anything, still felt inconsistent or incomplete, I would love to make those changes.

docs/tools.md
```diff
@@ -186,6 +186,68 @@ Benchmark                               Time             CPU      Time Old
...
+
+
+A statistically-significant difference is determined by a **p-value**, which is a measure of the probability that the observed difference could have occurred just by random chance. A smaller p-value indicates stronger evidence against the null hypothesis. 
+
+**Therefore:**
+  1. If the p-value is less than the chosen significance level (alpha), we reject the null hypothesis and conclude the benchmarks are significantly different.
+  2. If the p-value is greater than or equal to alpha, we fail to reject the null hypothesis and treat the two benchmarks as similar.
+
+
+When comparing benchmarks, `compare.py` uses statistical tests to determine whether there is a statistically-significant difference between the measurements being compared.
```
LebedevRI, line 226: It seems like this sentence should not be here (just delete it)

docs/tools.md
```diff
@@ -186,6 +186,68 @@ Benchmark                               Time             CPU      Time Old
...
+This might seem counter-intuitive if you're expecting 'green' to mean 'improved performance' and 'red' to mean 'worsened performance'. 
+```bash
+  But remember, in this context:
+
+    'Success' means 'successfully finding a difference'.
+    'Failure' means 'failing to find a difference'.
+```
+
+
+Also, please note that **even if** we determine that there **is** a statistically-significant difference between the two measurements, it does not _necessarily_ mean that the actual benchmarks that were measured **are** different, or vice versa, even if we determine that there is **no** statistically-significant difference between the two measurements, it does not necessarily mean that the actual benchmarks that were measured **are not** different.
```
LebedevRI, line 247: I believe, rest of the docs are linewrapped (w/ line width of 80),
so please format all the changed lines in this diff.

docs/tools.md
LebedevRI: Eh, i think that could be useful, but it looks out of place there.
What i was thinking of was along the lines of `Example output` from above,
but one that does actually contain `pvalue`

docs/tools.md
varshneydevansh: I deleted the sentence.

When comparing benchmarks, `compare.py` uses statistical tests to determine whether there is a statistically-significant difference between the measurements being compared.

LebedevRI:
@dmah42 someone more fluent in english will need to proofread the wording,
i can only mostly help with the content here.

dmah42:
> @dmah42 someone more fluent in english will need to proofread the wording, i can only mostly help with the content here.

seems good to me.  feel free to merge it whenever you're happy with the content.

docs/tools.md
varshneydevansh: So, should I add an example like this? This is an hypothetical example.
```bash
Comparing BM_memcpy (from ./a.out) to BM_copy (from ./a.out)
Benchmark                               Time             CPU      Time Old      Time New       CPU Old       CPU New       P-value
--------------------------------------------------------------------------------------------------------------------------------------
[BM_memcpy vs. BM_copy]/8            +5.1649         +5.1637            37           230            37           230        0.001
[BM_memcpy vs. BM_copy]/64          +21.4352        +21.4374            74          1654            74          1653        0.0001
[BM_memcpy vs. BM_copy]/512        +143.6022       +143.5865            91         13122            91         13120        0.00001
```

LebedevRI: Let's add 
```
Benchmark                                                              Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------------
benchmark/threads:1/process_time/real_time_pvalue                    0.0000          0.0000      U Test, Repetitions: 27 vs 27
benchmark/threads:1/process_time/real_time_mean                     -0.1442         -0.1442            90            77            90            77
benchmark/threads:1/process_time/real_time_median                   -0.1444         -0.1444            90            77            90            77
benchmark/threads:1/process_time/real_time_stddev                   +0.3974         +0.3933             0             0             0             0
benchmark/threads:1/process_time/real_time_cv                       +0.6329         +0.6280             0             0             0             0
OVERALL_GEOMEAN                                                     -0.1442         -0.1442             0             0             0             0

```

varshneydevansh: Hi Roman,

I added the provided example in addition, I also added an explanation summary of each output line.

https://github.com/google/benchmark/pull/1624/commits/0f64e48fcd8bace2a26d918d4b5c99ebb8258224

docs/tools.md
```diff
@@ -186,6 +186,146 @@ Benchmark                               Time             CPU      Time Old
 This is a mix of the previous two modes, two (potentially different) benchmark binaries are run, and a different filter is applied to each one.
 As you can note, the values in `Time` and `CPU` columns are calculated as `(new - old) / |old|`.
 
+
+4. The below output is a summary of a benchmark comparison with statistics
```
LebedevRI, line 190: It doesn't make sense to label this as `4.`, this is not a mode of operation.

docs/tools.md
```diff
@@ -186,6 +186,146 @@ Benchmark                               Time             CPU      Time Old
...
+Results are often noisy and don't necessarily have large absolute differences to
+them, so just by visual inspection, it is not at all apparent if two
+measurements are actually showing a performance change or not. It is even more
+confusing with multiple benchmark repetitions.
+
+Thankfully, what we can do, is use statistical tests on the results to determine
+whether the performance has statistically-significantly changed. `compare.py`
+uses [Mann–Whitney U
+test](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test), with a null
+hypothesis being that there's no difference in performance.
```
LebedevRI, line 250: Move this block to be before `4. The below output is a summary of a benchmark comparison with statistics`

LebedevRI (approved):
@varshneydevansh thank you!
I suppose this could use some more editorial work,
but this is certainly much better than what we have now :)

varshneydevansh:
Thank you so much @LebedevRI for guiding me. I recently encountered benchmarking for the first time and began learning about it. Then I realized that this is something where, alongside learning, I can even contribute.

I learned a lot from this process, not just about benchmarking, but how to organize thoughts in a way that makes it easier to understand for others.

Regards.

Issue #1583, opened 2023-04-03
[FR] Additional Python feature plans


nicholasjng:
I think the following points might benefit GBM's Python distribution. They are all optional, but with varying degrees of usefulness:

- [ ] Adding a modern build system-compliant (PEP517/518) `pyproject.toml`.
- [ ] Including all C++ source and header files into the wheel, along with the `cmake` directory; that way, Python people can include GBM as a build-time dependency and e.g. roll their own bindings. `nanobind` does this, specifically for integrating with `scikit-build`, for Bazel, it should work the same way.
- [ ] Adding a `benchmark.parametrize` decorator similar to `pytest.mark.parameterize` or `absl.testing.parameterized`, so that users don't have to `register` each benchmark by hand for each set of keyword arguments in the cartesian product.
- [ ] Making flag parsing with `absl` optional by opening up the flag parser as a configurable element of the benchmarking workflow.
- [ ] Consolidating all per-benchmark arguments in a single decorator to avoid excessive decorator stacking.
- [ ] Adding better IDE support by providing docstrings to the bound Python objects.

I will tackle them as I get the time, especially the registration QoL improvements might be worth it in my opinion.

PR #1625, opened 2023-07-09
Add pyproject.toml file for PEP518 compliance

nicholasjng:
The newly created `pyproject.toml` contains all static metadata as well as the readme and version as dynamic arguments, to be read by setuptools during the build.

What is left in the `setup.py` for now is the custom Bazel extension class, since that is not properly supported yet.

Part of #1583.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,50 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "google_benchmark"
+description = "A library to benchmark code snippets."
+requires-python = ">=3.8"
+license = {file = "LICENSE"}
+keywords = ["benchmark"]
+
+authors = [
+    {name = "Google", email = "benchmark-discuss@googlegroups.com"},
+]
+
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Topic :: Software Development :: Testing",
+    "Topic :: System :: Benchmark",
+]
+
+dynamic = ["readme", "version"]
+
+dependencies = [
+    "absl-py>=0.7.1",
+]
+
+[project.urls]
+Homepage = "https://github.com/google/benchmark"
+Documentation = "https://github.com/google/benchmark/tree/main/docs"
+Repository = "https://github.com/google/benchmark.git"
+Discord = "https://discord.gg/cz7UX7wKC2"
+
+[tool.setuptools]
+package-dir = {"" = "bindings/python"}
+zip-safe = false
+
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+
+[project]
+name = "google_benchmark"
+description = "A library to benchmark code snippets."
+requires-python = ">=3.8"
+license = {file = "LICENSE"}
+keywords = ["benchmark"]
+
+authors = [
+    {name = "Google", email = "benchmark-py@google.com"},
```
dmah42, line 13: why isn't this the existing benchmark discussion group?

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "google_benchmark"
+description = "A library to benchmark code snippets."
+requires-python = ">=3.8"
+license = {file = "LICENSE"}
+keywords = ["benchmark"]
+
```
dmah42, line 11: maybe also readme pointing to the github readme?

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+dynamic = ["readme", "version"]
+
+dependencies = [
+    "absl-py>=0.7.1",
+]
+
+[project.urls]
+Homepage = "https://github.com/google/benchmark"
+Documentation = "https://github.com/google/benchmark/tree/main/docs"
+Repository = "https://github.com/google/benchmark.git"
```
dmah42, line 38: we could add Discord too now: https://discord.gg/cz7UX7wKC2

pyproject.toml
nicholasjng: This is meant to be more of a direct point of contact to the "authors" specifically, and it needs to be an email address.

See e.g. https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#authors-maintainers

pyproject.toml
nicholasjng: The readme is dynamic, see the last `[tool.setuptools]` block. Not sure what happens if it is specified twice, but I can also add it to the `project` part.

pyproject.toml
dmah42: benchmark-discuss@googlegroups.com :)

nicholasjng: Updated the email in 7fdfc4626e8de14130a0923e71b784f79ec6df93.

pyproject.toml
nicholasjng: Added the Discord link in 7fdfc4626e8de14130a0923e71b784f79ec6df93.

pyproject.toml
dmah42: the project readme should be a link to the existing markdown file in github aiui

nicholasjng: Hmm, I don't think so: Here in the docs, it says that if it is a string, it needs to be a path relative to the `pyproject.toml`: https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#readme

However, also in that docs section, it appears that from an `.md` extension, it automatically assumes `content-type = text/markdown`, so the dynamic part can go away and be replaced with the static `readme = "README.md"` field. What do you think?

dmah42: i probably read it wrong then :)

dmah42 (approved):

PR #1626, opened 2023-07-09
Bump `nanobind` to stable v1.4.0 tag

nicholasjng:
This seems to reduce binding sizes even further, with a wheel size of 175KB on my local machine (macOS 13.4.1).

### DIFF
--- a/bindings/python/nanobind.BUILD
+++ b/bindings/python/nanobind.BUILD
@@ -31,6 +31,7 @@ cc_library(
         "src/nb_internals.cpp",
         "src/nb_internals.h",
         "src/nb_ndarray.cpp",
+        "src/nb_static_property.cpp",
         "src/nb_type.cpp",
         "src/trampoline.cpp",
     ],
--- a/bazel/benchmark_deps.bzl
+++ b/bazel/benchmark_deps.bzl
@@ -48,8 +48,7 @@ def benchmark_deps():
         new_git_repository(
             name = "nanobind",
             remote = "https://github.com/wjakob/nanobind.git",
-            commit = "1ffbfe836c9dac599496a170274ee0075094a607", # v0.2.0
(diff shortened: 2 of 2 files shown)


bindings/python/nanobind.BUILD
```diff
@@ -31,6 +31,7 @@ cc_library(
         "src/nb_internals.cpp",
         "src/nb_internals.h",
         "src/nb_ndarray.cpp",
+        "src/nb_static_property.cpp",
```
dmah42, line 34: should this be sourced as an external library from somewhere (github release tarball?) so we don't need to have a parallel build of the library maintained here?

nicholasjng: It can be used as a build-time CMake dependency via `scikit-build`, but this requires changing the package build to use `scikit-build` - that could be an avenue.

Otherwise, there is an [open JAX PR](https://github.com/google/jax/pull/16100/files#diff-f74002fc431c40bce15860cef736ba876710da3be319c8c281195aade6ae3e88) that ports some existing pybind11 code to nanobind - there they have a leaner build file, but still explicit configs (I think that is sensible too tbh)

In the end, however, we are pulling straight from github in Bazel too, so this might already be what you intended? (The build file is necessary to mark it as a Bazel dep)

dmah42: i thought similar to how we depend on googletest, we could depend on nanobind.  but maybe it's a different type of dependency if we don't directly link to it?  or maybe nanobind doesn't support bazel directly?

nicholasjng: Exactly, `nanobind` only supports CMake. So I'm afraid we need the full build file setup regardless. But at least the link above shows that it can still be simplified, especially if we don't particularly care about pinning the submodules.

Issue #1627, opened 2023-07-10
[BUG] PFM_ROOT is not used when ENABLE_LIBPFM=ON


chipot:
**Describe the bug**
The usual way to point cmake to a dependency via `<Name>_ROOT` variable or via `CMAKE_PREFIX_PATH` is not working.

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Linux
  - Compiler and version:  GCC 13.0

**To reproduce**
1. pull main branch
2. build/install libpfm4 anywhere except in standard system path
3. cmake -D BENCHMARK_ENABLE_LIBPFM=ON -D PFM_ROOT=/the/path/to/libpfm4/
4. cmake configure succeeds, but did not find libpfm, and thus the support is disabled.

**Expected behavior**
I expected to be able to inject PFM_ROOT to where I installed libpfm, but could not. Looking into it more, the custom `FindPFM.cmake` module is not doing the right thing.

PR #1628, opened 2023-07-10
pfm: Use a more standard CMake approach for finding libpfm

chipot:
This PR is trying to address https://github.com/google/benchmark/issues/1627

I reworked `FindPFM.cmake` to use `find_library` and `find_path` instead of `check_library_exists` and `check_include_file`. The main problem with these two is that they expect CFLAGS and LDFLAGS to be set correctly before the call.

The module now creates an `IMPORTED` target `PFM::libpfm` and we can link to it as expected.

### DIFF
--- a/cmake/Modules/FindPFM.cmake
+++ b/cmake/Modules/FindPFM.cmake
@@ -1,26 +1,22 @@
 # If successful, the following variables will be defined:
-# HAVE_LIBPFM.
-# Set BENCHMARK_ENABLE_LIBPFM to 0 to disable, regardless of libpfm presence.
-include(CheckIncludeFile)
-include(CheckLibraryExists)
-include(FeatureSummary)
-enable_language(C)
+# PFM_FOUND.
+# PFM_LIBRARIES
+# PFM_INCLUDE_DIRS
+# the following target will be defined:
+# PFM::libpfm
 
-set_package_properties(PFM PROPERTIES
-                       URL http://perfmon2.sourceforge.net/
-                       DESCRIPTION "a helper library to develop monitoring tools"
-                       PURPOSE "Used to program specific performance monitoring events")
+include(FindPackageHandleStandardArgs)
 
-check_library_exists(libpfm.a pfm_initialize "" HAVE_LIBPFM_INITIALIZE)
-if(HAVE_LIBPFM_INITIALIZE)
-  check_include_file(perfmon/perf_event.h HAVE_PERFMON_PERF_EVENT_H)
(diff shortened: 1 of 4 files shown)


dmah42:
fantastic!  i wonder if this is why the CI bots fail the pfm tests... cc @macandy13

dmah42:
CI failures are unrelated

dmah42 (approved):

dmah42:
thanks!

cmake/Modules/FindPFM.cmake
```diff
@@ -1,26 +1,22 @@
...
+# PFM_FOUND.
+# PFM_LIBRARIES
+# PFM_INCLUDE_DIRS
+# the following target will be defined:
+# PFM::libpfm
 
-set_package_properties(PFM PROPERTIES
-                       URL http://perfmon2.sourceforge.net/
-                       DESCRIPTION "a helper library to develop monitoring tools"
-                       PURPOSE "Used to program specific performance monitoring events")
```
LebedevRI, line 12: These should remain

dmah42: added back with a followup commit

cmake/Modules/FindPFM.cmake
```diff
@@ -1,26 +1,22 @@
...
-else()
-  message("Perf Counters support requested, but was unable to find libpfm.")
+find_library(PFM_LIBRARY NAMES pfm)
+find_path(PFM_INCLUDE_DIR NAMES perfmon/pfmlib.h)
+
+find_package_handle_standard_args(PFM REQUIRED_VARS PFM_LIBRARY PFM_INCLUDE_DIR)
+
+if (PFM_FOUND AND NOT TARGET PFM::libpfm)
+    add_library(PFM::libpfm UNKNOWN IMPORTED)
+    set_target_properties(PFM::libpfm PROPERTIES
```
LebedevRI, line 17: Is there a convention on the names of imported targets?
If not, i think `PFM::PFM` might be better-ish?

dmah42:
i was suggesting that the current version of this code that doesn't do a great job of finding pfm might be the reason why the CI machines can't run the tests.  maybe with this change they can?

macandy13:
When I tried the tests, the problem was not that libpfm wasn't found (I was using bazel for doing this), it was that the underlying VM does not report any performance counters at all. The pfm tests actually expect that the system at least reports a few standard ones.

Do you have a link to the error you were seeing?

dmah42:
i didn't check any errors, i was just being optimistic.

PR #1633, opened 2023-07-14
Update bzlmod support to new rules_python extension API

jsharpe:
The bzlmod extension API in rules_python before version 0.24.0 was marked as experimental and has changed.
Its expected that the 0.24.0 API should now be more stable as feature parity has been achieved with the non-bzlmod variant.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,7 +4,7 @@ bazel_dep(name = "bazel_skylib", version = "1.4.1")
 bazel_dep(name = "platforms", version = "0.0.6")
 bazel_dep(name = "rules_foreign_cc", version = "0.9.0")
 bazel_dep(name = "rules_cc", version = "0.0.6")
-bazel_dep(name = "rules_python", version = "0.23.1")
+bazel_dep(name = "rules_python", version = "0.24.0")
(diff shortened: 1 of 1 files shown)


MODULE.bazel
```diff
@@ -16,19 +16,9 @@ bazel_dep(name = "libpfm", version = "4.11.0")
...
-)
-use_repo(interpreter, "interpreter")
-
 pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
 pip.parse(
-    name="tools_pip_deps",
-    incompatible_generate_aliases = True,
-    python_interpreter_target="@interpreter//:python",
+    hub_name="tools_pip_deps",
+    python_version = "3.9",
```
jsharpe, line 22: It's possible this line shouldn't be here; am in discussion with the rules_python authors around this design decision as this is problematic when the main repo declares a different default python_version.

dmah42: i see.. so this can be different to the `python.toolchain` declared above?

jsharpe: See https://github.com/bazelbuild/rules_python/issues/1320

dmah42 (approved):

PR #1643, opened 2023-08-09
[perf-counters] Fix pause/resume

mtrofin:
Using `state.PauseTiming() / state.ResumeTiming()` was broken.

Thanks @virajbshah for the the repro testcase.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -229,7 +229,6 @@ void State::PauseTiming() {
     for (const auto& name_and_measurement : measurements) {
       auto name = name_and_measurement.first;
       auto measurement = name_and_measurement.second;
-      BM_CHECK_EQ(std::fpclassify(double{counters[name]}), FP_ZERO);
       counters[name] += Counter(measurement, Counter::kAvgIterations);
     }
   }
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -230,7 +230,7 @@ void State::PauseTiming() {
       auto name = name_and_measurement.first;
       auto measurement = name_and_measurement.second;
       BM_CHECK_EQ(std::fpclassify(double{counters[name]}), FP_ZERO);
-      counters[name] = Counter(measurement, Counter::kAvgIterations);
+      counters[name] += Counter(measurement, Counter::kAvgIterations);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/benchmark.cc
```diff
@@ -230,7 +230,7 @@ void State::PauseTiming() {
       auto name = name_and_measurement.first;
       auto measurement = name_and_measurement.second;
       BM_CHECK_EQ(std::fpclassify(double{counters[name]}), FP_ZERO);
```
LebedevRI, line 232: I don't believe this assertion remains valid after this change.

mtrofin: thanks for the catch - in a debug build it trips, indeed.

test/perf_counters_test.cc
```diff
@@ -21,17 +21,62 @@ static void BM_Simple(benchmark::State& state) {
...
+
+ADD_CASES(TC_JSONOut, {{"\"name\": \"BM_WithPauseResume\",$"}});
+
 static void CheckSimple(Results const& e) {
   CHECK_COUNTER_VALUE(e, double, "CYCLES", GT, 0);
   CHECK_COUNTER_VALUE(e, double, "BRANCHES", GT, 0.0);
 }
+
+static void CheckInstrCount(Results const& e) {
+  CHECK_COUNTER_VALUE(e, double, "INSTRUCTIONS", GT, kIters);
```
dmah42, line 68: with pause and resume, shouldn't this be GT 2*kIters?  i don't know how the INSTRUCTIONS perf counters work so i could be wrong.

mtrofin: Yes - the current test checks against the original regression, but I see what you mean. Ideally we'd check that the value in the "with pause resume" is < 2 * "value in the without pause resume". Is there a recommended way to do that, other than capturing the value in the "without pause resume" in some global and relying on execution order?

dmah42: would `LT, 1.5*kIters` not be enough at least for an upper bound?

mtrofin: each iteration executes a bit of instructions, and it depends on opt level how many (and architecture)

dmah42: we can't rely on execution order.  we can guarantee that both have run when `RunOutputTests` completes, so maybe we can store the number of instructions for both BMs and compare them?

mtrofin: done.

LebedevRI (approved):
The test is rather fragile, but seems fine in general.
Thank you.

dmah42:
thanks!

Issue #1505, opened 2022-11-10
[BUG] build fails with IntelLLVM 2022.2.x compiler


damageboy:
Attempting to build google benchmark fails with the latest intel compiler in release mode

The OS is ubuntu 22.04, Intel LLVM compiler 2022.2, 

**To reproduce**
```bash
❯ cmake -DCMAKE_BUILD_TYPE=Release .. -G Ninja -DBENCHMARK_ENABLE_TESTING=OFF
-- The CXX compiler identification is IntelLLVM 2022.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/intel/oneapi/compiler/2022.2.1/linux/bin/icpx - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Failed to find LLVM FileCheck
-- Found Git: /usr/bin/git (found version "2.38.1") 
-- git version: v1.7.0-25-g398a8ac2 normalized to 1.7.0.25
-- Version: 1.7.0.25
-- Looking for shm_open in rt
-- Looking for shm_open in rt - found
-- Performing Test HAVE_CXX_FLAG_WALL
-- Performing Test HAVE_CXX_FLAG_WALL - Success
-- Performing Test HAVE_CXX_FLAG_WEXTRA
-- Performing Test HAVE_CXX_FLAG_WEXTRA - Success
-- Performing Test HAVE_CXX_FLAG_WSHADOW
-- Performing Test HAVE_CXX_FLAG_WSHADOW - Success
-- Performing Test HAVE_CXX_FLAG_WFLOAT_EQUAL
-- Performing Test HAVE_CXX_FLAG_WFLOAT_EQUAL - Success
-- Performing Test HAVE_CXX_FLAG_WERROR
-- Performing Test HAVE_CXX_FLAG_WERROR - Success
-- Performing Test HAVE_CXX_FLAG_WSUGGEST_OVERRIDE
-- Performing Test HAVE_CXX_FLAG_WSUGGEST_OVERRIDE - Success
-- Performing Test HAVE_CXX_FLAG_PEDANTIC
-- Performing Test HAVE_CXX_FLAG_PEDANTIC - Success
-- Performing Test HAVE_CXX_FLAG_PEDANTIC_ERRORS
-- Performing Test HAVE_CXX_FLAG_PEDANTIC_ERRORS - Success
[...]

PR #1644, opened 2023-08-10
Fix IntelLLVM compiler error

joannapng:
### **In response to issue #1505**  
The IntelLLVM Compiler has the `-ffast-math` flag enabled by default, which in turn, turns on the `-ffinite-math-only` flag, which assumes that arguments and results are not NaNs or Infs. Use the `-fno-finite-math-only` flag to suppress the errors.


dmah42:
failing windows release test is unrelated.

dmah42 (approved):

dmah42:
we should consider adding intel compilers to the CI bots i think.

dmah42:
thanks!

PR #1652, opened 2023-08-21
State: Initialize counters with kAvgIteration in constructor

jmr:
Previously, `counters` was updated in `PauseTiming()` with
`counters[name] += Counter(measurement, kAvgIteration)`.

The first `counters[name]` call inserts a counter with no flags.

There is no `operator+=` for `Counter`, so the insertion is done
by converting the `Counter` to a `double`, then constructing a
`Counter` to insert from the `double`, which drops the flags.

Pre-insert the `Counter` with the correct flags, then only
update `Counter::value`.

Introduced in https://github.com/google/benchmark/commit/1c64a36c5b8ee75d462b3fe7a9d020c66a2a1094 ([perf-counters] Fix pause/resume (https://github.com/google/benchmark/pull/1643)).

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -179,6 +179,16 @@ State::State(std::string name, IterationCount max_iters,
   BM_CHECK_LT(thread_index_, threads_)
       << "thread_index must be less than threads";
 
+  // Add counters with correct flag now.  If added with `counters[name]` in
+  // `PauseTiming`, a new `Counter` will be inserted the first time, which
+  // won't have the flag.  Inserting them now also reduces the allocations
+  // during the benchmark.
+  if (perf_counters_measurement_) {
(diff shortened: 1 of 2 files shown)


jmr:
@mtrofin

jmr:
I'm not sure why most of the sanitizer tests are failing.

The history on my branch is getting out of control.  This should be squashed-and-merged when it's ready.

jmr:
Can you trigger the presubmits on HEAD? I wonder if something changed with the sanitizer setup.

dmah42:
> Can you trigger the presubmits on HEAD? I wonder if something changed with the sanitizer setup.

yeah they're failing for changes to documents

mtrofin:
Thanks for fixing this. I forgot how `Counter` worked and assumed they are collected "in total" by design. Thanks!

PR #1653, opened 2023-08-21
perf-counters: Make tests pass on Android

jmr:
Tested on Pixel 3 and Pixel 6.  Reduce test to the intersection of
what passes on all platforms.

Pixel 6 doesn't support BRANCHES and only supports two perf
counters.


jmr:
Sanitizers failing here, too.

dmah42:
> Sanitizers failing here, too.

hm.  i wonder if something in the infra broke.  github does change their images sometimes without warning.

PR #1656, opened 2023-08-24
perf_counters: Initialize once only when needed

jmr:
This works around some performance problems running Android under QEMU. Calling `pfm_initialize` was very slow, and was called during dynamic initialization (before `main` or when loaded as a shared library). This happened whenever benchmark was linked, even if no benchmarks were run.

Instead, call `pfm_initialize` at most once, and only when one of:
1. `PerfCounters::Initialize` is called
2. `PerfCounters::Create` is called with a non-empty counter list
3. `PerfCounters::IsCounterSupported` is called

The return value of the first `pfm_initialize()` is saved and returned from all subsequent `PerfCounters::Initialize` calls.

### DIFF
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -61,7 +61,9 @@ const bool PerfCounters::kSupported = true;
 // initialization was successful
 static bool InitLibPfmOnce() {
   // Function-scope static gets initialized only once on first call.
-  static bool success = []() { return pfm_initialize() == PFM_SUCCESS; }();
+  static const bool success = []() {
+    return pfm_initialize() == PFM_SUCCESS;
+  }();
   return success;
 }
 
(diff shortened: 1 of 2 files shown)


jmr:
@mtrofin

src/perf_counters.cc
```diff
@@ -57,9 +57,20 @@ size_t PerfCounterValues::Read(const std::vector<int>& leaders) {
 
 const bool PerfCounters::kSupported = true;
 
-bool PerfCounters::Initialize() { return pfm_initialize() == PFM_SUCCESS; }
+// Initializes libpfm only on the first call.  Returns whether that single
+// initialization was successful
+static bool InitLibPfmOnce() {
```
jmr, line 62: I could also delete `InitLibPfmOnce` and put this directly in `Initialize`.

I thought it was clearer to have the other calls use this name, though.

dmah42: i think having it directly in `Initialize` and having the other methods call `Initialize` is clearer.

dmah42: i think that's fine.  the callsites are blindly trying to initialize and if it already is, then it's a no-op.

jmr: Updated.

dmah42 (approved):

dmah42:
thanks!

PR #1657, opened 2023-08-24
test: Use gtest_main only when needed 

jmr:
There are two types of tests.  `*_gtest.cc` files use `gtest` and
`gtest_main`.  `*_test.cc` files define their own main.

Only depend on `gtest`/`gtest_main` when needed.  This is similar
to what `CMakeLists.txt` does.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -69,7 +69,7 @@ cc_library(
     for test_src in glob(["*_gtest.cc"])
 ]
 
-# Tests that do not gtest.  These have their own `main` defined.
+# Tests that do not use gtest.  These have their own `main` defined.
 [
     cc_test(
         name = test_src[:-len(".cc")],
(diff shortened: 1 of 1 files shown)


test/BUILD
```diff
@@ -49,29 +49,48 @@ cc_library(
     ],
 )
 
+# Tests that use gtest.  These rely on `gtest_main`.
 [
     cc_test(
         name = test_src[:-len(".cc")],
         size = "small",
         srcs = [test_src],
-        args = TEST_ARGS + PER_SRC_TEST_ARGS.get(test_src, []),
```
dmah42, line 57: we still need TEST_ARGS or we get NaNs sometimes.

dmah42:
if you'd asked, i would have sworn that i had this in here already....  so weird.  thank you!

dmah42 (approved):

PR #1658, opened 2023-08-24
StatisticsMedian: Fix bug

jmr:
Previously, this could return the wrong result when there was an even number of elements.

There were two `nth_element` calls.  The second call could change elements in `[center2, end])`, which was where `center` pointed.  Therefore, `*center` sometimes had the wrong value after the second `nth_element` call.

Rewrite to use `max_element` instead of the second call to `nth_element`.  This avoids modifying the vector.

### DIFF
--- a/src/statistics.cc
+++ b/src/statistics.cc
@@ -42,13 +42,12 @@ double StatisticsMedian(const std::vector<double>& v) {
   auto center = copy.begin() + v.size() / 2;
   std::nth_element(copy.begin(), center, copy.end());
 
-  // did we have an odd number of samples?
-  // if yes, then center is the median
-  // it no, then we are looking for the average between center and the value
-  // before
+  // Did we have an odd number of samples?  If yes, then center is the median.
+  // If not, then we are looking for the average between center and the value
+  // before.  Instead of resorting, we just look for the max value before it.
+  // (Since `copy` is partially sorted.)
   if (v.size() % 2 == 1) return *center;
-  auto center2 = copy.begin() + v.size() / 2 - 1;
-  std::nth_element(copy.begin(), center2, copy.end());
+  auto center2 = std::max_element(copy.begin(), center);
   return (*center + *center2) / 2.0;
 }
 


LebedevRI:
Please add tests.

LebedevRI:
> > Please add tests.
> 
> The tests already exist. They failed sometimes. That's how I found it.

Could you show the failure? Define `sometimes`? Was that failure non-deterministic? I've never seen it...
This should be tested in https://github.com/google/benchmark/blob/dfc8a92abc88a9d630a9f8e01c678fedde4c3090/test/statistics_gtest.cc#L15-L19.

src/statistics.cc
```diff
@@ -42,13 +42,12 @@ double StatisticsMedian(const std::vector<double>& v) {
...
-  // it no, then we are looking for the average between center and the value
-  // before
+  // Did we have an odd number of samples?  If yes, then center is the median.
+  // If not, then we are looking for the average between center and the value
+  // before.  Instead of resorting, we just look for the max value before it.
+  // (Since `copy` is partially sorted.)
   if (v.size() % 2 == 1) return *center;
-  auto center2 = copy.begin() + v.size() / 2 - 1;
-  std::nth_element(copy.begin(), center2, copy.end());
+  auto center2 = std::max_element(copy.begin(), center);
```
dmah42, line 50: could this also just be `center - 1`?  if `copy` is sorted up to `center`, then the value before is just `(center - 1)` no?

and `center` can`t be `copy.begin()` because we guarantee `v.size() >= 2`.

jmr:
> Could you show the failure? 

```
[ RUN      ] StatisticsTest.Median
third_party/benchmark/test/statistics_gtest.cc:17: Failure
Expected equality of these values:
  benchmark::StatisticsMedian({1, 2, 3, 4})
    Which is: 3
  2.5
```

> Define `sometimes`?  Was that failure non-deterministic? 

1/2 of the time, non-deterministically.

> I've never seen it... 

If you use `libc++` with appropriate debug flags, you will see it.  It will randomize `[begin, nth)` and `(nth, end)`:

https://github.com/llvm/llvm-project/blob/5ec13535235d07eafd64058551bc495f87c283b1/libcxx/include/__algorithm/nth_element.h#L237

A longer sequence will decrease the probability of success factorially.

> This should be tested in
> 
> https://github.com/google/benchmark/blob/dfc8a92abc88a9d630a9f8e01c678fedde4c3090/test/statistics_gtest.cc#L15-L19
> 
> .

It is.

src/statistics.cc
jmr: That's what `partial_sort` does, not `nth_element`.  `nth_element(begin, nth, end)` just ensures `*i <= *nth` for `begin <= i < nth` and `*i >= *nth` for `nth < i < end`.

LebedevRI:
Ok, so the **actual** bug here is iterator invalidation, effectively.
We should dereference `center` before calling `std::nth_element()` again, and then everything would still work.
This was really not at all obvious from the patch description, maybe improve it a bit?

But this is still an improvement, so okay.

jmr:
> Ok, so the **actual** bug here is iterator invalidation, effectively. 

No, the iterator is still valid.  The problem is an assumption that `nth_element(begin, nth, end)` doesn't affect `(nth, end)`.

LebedevRI:
> > Ok, so the **actual** bug here is iterator invalidation, effectively.
> 
> No, the iterator is still valid. The problem is an assumption that `nth_element(begin, nth, end)` doesn't affect `(nth, end)`.

I did say "effectively" - if you consider that the `center` is an iterator
pointing to the median-ish element of the `copy`.

dmah42 (approved):

dmah42:
thanks!

PR #1668, opened 2023-09-20
Increase the kMaxIterations limit

### LINKED ISSUE #1663: [BUG] --benchmark_min_time flag not working as intended

andreas-abel:
This pull request fixes #1663. Note that as a result of this change, the columns in the console output can become misaligned if the actual iteration count is too high. This will be dealt with separately (see also #1664).

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -64,7 +64,7 @@ MemoryManager* memory_manager = nullptr;
 
 namespace {
 
-static constexpr IterationCount kMaxIterations = 1000000000;
+static constexpr IterationCount kMaxIterations = 1000000000000;
 const double kDefaultMinTime =
     std::strtod(::benchmark::kDefaultMinTimeStr, /*p_end*/ nullptr);
 


LebedevRI (approved):
LG, thank you!

dmah42 (approved):

dmah42:
approved, but there's more lint errors i think.

andreas-abel:
It looks like the other lint error comes from this PR: https://github.com/google/benchmark/pull/1681

I think it should be fixed there.

dmah42:
> It looks like the other lint error actually comes from this PR: #1681
> 
> I think it should be fixed there.

that's been merged and didn't have a lint issue.. maybe try updating this branch to head?

andreas-abel:
Actually, it looks like the other one does have this issue, even though it was merged: https://github.com/google/benchmark/actions/runs/6535161131/job/17744142279

dmah42:
oh woops.  i thought i had lint set as a required CI.  i'll fix that separately then.

LebedevRI:
> oh woops. i thought i had lint set as a required CI. i'll fix that separately then.

Oops...

LebedevRI:
> thank you!

Issue #1597, opened 2023-05-11
[BUG] cxx03_test Fails Due to C++ Version


joshdorsey:
**Describe the bug**
A clean build of the google benchmark library as described [here](https://github.com/google/benchmark#installation) fails with an error after the last command: "C++11 or greater detected. Should be C++03."

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Windows 10
  - Compiler and version:
  ```$ clang++ --version
clang version 13.0.1
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin
```
 - CMake Version: 3.20.1
 - Ninja Version: 1.10.2

**To reproduce**
1. Checkout the repo to 318dd44
2. `cmake -E make_directory "build"`
3. `cmake -E chdir "build" cmake -DBENCHMARK_DOWNLOAD_DEPENDENCIES=on -GNinja -DCMAKE_BUILD_TYPE=Release ../`
4. `cmake --build "build" --config Release`
5. Uh-oh!
```
FAILED: test/CMakeFiles/cxx03_test.dir/cxx03_test.cc.obj
C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE -DBENCHMARK_STATIC_DEFINE -DHAVE_STD_REGEX -DHAVE_STEADY_CLOCK -DHAVE_THREAD_SAFETY_ATTRIBUTES -DTEST_BENCHMARK_LIBRARY_HAS_NO_ASSERTIONS -I../include -Wall  -Wextra  -Wshadow  -Wfloat-equal  -Werror  -pedantic  -pedantic-errors  -Wshorten-64-to-32  -fstrict-aliasing  -Wno-deprecated-declarations  -Wno-deprecated  -Wstrict-aliasing  -Wthread-safety -O3  -D_DLL -D_MT -Xclang --dependent-lib=msvcrt -fvisibility-inlines-hidden   -UNDEBUG -w -std=c++14 -MD -MT test/CMakeFiles/cxx03_test.dir/cxx03_test.cc.obj -MF test\CMakeFiles\cxx03_test.dir\cxx03_test.cc.obj.d -o test/CMakeFiles/cxx03_test.dir/cxx03_test.cc.obj -c ../test/cxx03_test.cc
../test/cxx03_test.cc:8:2: error: C++11 or greater detected. Should be C++03.
#error C++11 or greater detected. Should be C++03.
 ^
[...]

PR #1669, opened 2023-09-26
Audit MSVC references in cmake files to consider clang++

Maetveis:
There are three major compilers on Windows targeting the MSVC ABI (i.e. linking with microsofts STL etc.):
  - `MSVC`
  - `clang-cl` aka clang with the MSVC compatible CLI
  - `clang++` aka clang with gcc compatible CLI

The cmake variable `MSVC` is only set for the first two as it defined in terms of the CLI interface provided:

> Set to true when the compiler is some version of Microsoft Visual
> C++ or another compiler simulating the Visual C++ cl command-line syntax.

(from cmake docs)

For many of the checks in the library its the ABI that matters not the cmdline, so check `CMAKE_CXX_SIMULATE_ID` too, if it is `MSVC` the current compiler is targeting the MSVC ABI. This handles `clang++`

Fixes: #1597

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ if(BENCHMARK_FORCE_WERROR)
   set(BENCHMARK_ENABLE_WERROR ON)
 endif(BENCHMARK_FORCE_WERROR)
 
-if(NOT MSVC)
+if(NOT (MSVC OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC"))
   option(BENCHMARK_BUILD_32_BITS "Build a 32 bit version of the library." OFF)
 else()
   set(BENCHMARK_BUILD_32_BITS OFF CACHE BOOL "Build a 32 bit version of the library - unsupported when using MSVC)" FORCE)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ if(BENCHMARK_FORCE_WERROR)
   set(BENCHMARK_ENABLE_WERROR ON)
 endif(BENCHMARK_FORCE_WERROR)
 
-if(NOT MSVC)
+if(NOT (MSVC OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC"))
```
Maetveis, line 24: I'm not sure about this one, is this unsupported generally on windows (exluding mingw), or is it a specific problem with MSVC (and clang-cl)?

CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ option(BENCHMARK_ENABLE_LIBPFM "Enable performance counters provided by libpfm"
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 
-if(MSVC)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
Maetveis, line 48: I'm assuming here that the below workaround is only needed for the "real" MSVC and not `clang-cl` or `clang++` as they can natively cross compile and don't have separate ARM variants AFAIK.

CMakeLists.txt
dmah42: i don't know, but i'm quite comfortable with windows-like being only 64-bit at this point.

CMakeLists.txt
dmah42: that's my assumption too.

Maetveis:
I don't think the sanitizer builds failing are relevant, this shouldn't change anything on Linux and all windows jobs pass.

dmah42:
> I don't think the sanitizer builds failing are relevant, this shouldn't change anything on Linux and all windows jobs pass.

agreed. filed https://github.com/google/benchmark/issues/1670

dmah42 (approved):

PR #1682, opened 2023-10-19
Set -Wno-unused-variable for tests

oontvoo:
We  used assert() a lot in tests and that can cause build breakages in some of the opt builds (since assert() are removed)

it's not practical to sprinkle "(void)" everywhere so I think setting this warning option is the best option for now.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -18,6 +18,9 @@ TEST_COPTS = [
     #    "-Wshorten-64-to-32",
     "-Wfloat-equal",
     "-fstrict-aliasing",
+    ## assert() are used a lot in tests upstream, which may be optimised out leading to
+    ## unused-variable warning.
+    "-Wno-unused-variable",
 ]
 
 # Some of the issues with DoNotOptimize only occur when optimization is enabled


dmah42:
do you know why we're not seeing these as issues in the CI bots?  release mode would strip the asserts, and we have both -Wall and -Werror enabled so we should be getting these warnings/errors...

LebedevRI:
Because `-Wall` is a very small subset of all the compiler-provided warnings.
You'd want `-Weverything`-clean, with explicit opt-outs.

dmah42:
> Because `-Wall` is a very small subset of all the compiler-provided warnings. You'd want `-Weverything`-clean, with explicit opt-outs.

ah yes, the nuanced distinction between "all" and "everything" (everywhere all at once, presumably).

dmah42 (approved):

LebedevRI:
Is this missing the CMake part?

oontvoo:
> Is this missing the CMake part?

Good point - will send a followup PR.

Thanks both!

dmah42:
> > Is this missing the CMake part?
> 
> Good point - will send a followup PR.
> 
> Thanks both!

#1683 i think does it.

test/BUILD
```diff
@@ -18,6 +18,9 @@ TEST_COPTS = [
     #    "-Wshorten-64-to-32",
     "-Wfloat-equal",
     "-fstrict-aliasing",
+    ## assert() are used a lot in tests upstream, which may be optimised out leading to
```
jmr, line 21: What "upstream" is this referring to?

I ran into the same problem with unused variables, but only needed one BENCHMARK_MAYBE_UNUSED to fix it.

oontvoo: I was referring to the test code in this repo (which tends to use assert() rather than GTEST's assertion utilities)

PR #1685, opened 2023-10-24
Reapply size optimizations for clang & MSVC, LTO for Mac+Linux

nicholasjng:
Working towards cross-platform optimal nanobind building configurations.

Re: LTO:

The Windows case (the option name is "/GL") is more complicated, since there, the compiler options also need to be passed to the linker if LTO is enabled.

Since we are gating the linker options on platform at the moment instead of compiler, we need to implement a Bazel boolean flag for the case `"Platform == MacOS && Compiler == AnyOf(gcc, clang)"`.

I also applied `buildifier`, which formats Bazel files (BUILD, WORKSPACE, .bzl) nicely - we could integrate this in CI if there's interest.

### DIFF
--- a/bindings/python/nanobind.BUILD
+++ b/bindings/python/nanobind.BUILD
@@ -2,18 +2,38 @@ licenses(["notice"])
 
 package(default_visibility = ["//visibility:public"])
 
-filegroup(
-    name = "symboltable",
-    srcs = ["cmake/darwin-ld-cpython.sym"],
+config_setting(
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

dmah42:
a buildifier linter would be great for CI.

nicholasjng:
Ha, Windows Python CI is green now post-merge, did you bet on that to work?

https://github.com/google/benchmark/actions/runs/6639541298

PR #1687, opened 2023-10-26
[bindings] Add LTO builds on Windows+MSVC

nicholasjng:
Gates the MSVC switches behind an `@bazel_skylib:selects` statement.

This is a first experiment from best guesses and studying the MSVC docs.

### DIFF
--- a/bindings/python/nanobind.BUILD
+++ b/bindings/python/nanobind.BUILD
@@ -44,8 +44,8 @@ cc_library(
         "include",
     ],
     linkopts = select({
-        ":winplusmsvc": ["/LTGC"],  # allows other compilers on Windows than just MSVC.
-        "@platforms//os:macos": ["-Wl,@$(location :cmake/darwin-ld-cpython.sym)"],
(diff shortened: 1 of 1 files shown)


nicholasjng:
We should compare the perf numbers from the Python test for this build to those of some of the previous jobs - I read a forum post today about how MSVC LTO can in some cases tank performance.

bindings/python/nanobind.BUILD
```diff
@@ -33,6 +44,7 @@ cc_library(
         "include",
     ],
     linkopts = select({
+        ":winplusmsvc": ["/LTGC"],  # allows other compilers on Windows than just MSVC.
```
dmah42, line 47: so clang on windows takes /LTGC as a parameter?

nicholasjng: Ah sorry, I guess that came from the comment. The `:winplusmsvc` condition only applies for Windows + MSVC, so for Windows + clang, it falls through to `//conditions:default`, which is no options.

I can update the comment if it's too misleading.

dmah42 (approved):

PR #1690, opened 2023-10-31
Add `setuptools_scm` for dynamic zero-config Python versioning

nicholasjng:
This removes the need for manually bumping versions in the Python bindings.

For the wheel uploads, the correct semver version is inferred in the case of tagged commits, which is exactly the case in GitHub CI.

The docs were updated to reflect the changes in the release workflow.

We also add a runtime-useable version, which is the best practice mentioned in the `setuptools_scm` docs, see https://setuptools-scm.readthedocs.io/en/latest/usage/#version-at-runtime.


dmah42 (approved):

dmah42:
much better. thank you!

PR #1691, opened 2023-11-01
Fix unit tests compilation by non-gnu/msvc compilers with c++11 support.

ivafanas:
donotoptimize_test.cc could not be compiled under non-gnu / non-msvc compilers, because only deprecated version of DoNotOptimize is available for these compilers. Tests are compiled with -Werror. Patch fixes test compilation by providing non-deprecated version of DoNotOptimize for compilers with c++11 standard support.


dmah42 (approved):

dmah42:
thanks!

ivafanas:
Thank you too!

Issue #1699, opened 2023-11-11
[BUG] Missing `\n` at the End of Console Output


itzmeanjan:
:wave: 

It looks like we're missing a `\n` at the end of this message.

https://github.com/google/benchmark/blob/159eb2d0ffb85b86e00ec1f983d72e72009ec387/src/perf_counters.cc#L257

Without `\n`, output looks like :point_down:.

![Screenshot from 2023-11-11 21-34-48](https://github.com/google/benchmark/assets/45074836/13efc9be-d7d0-4e55-8c63-d65dde363e14)

I discovered this issue while benchmarking with libPFM on a Raspberry Pi 4b, more @ https://github.com/itzmeanjan/kyber/blob/6f8dc9070abe8dfbe0a8e400151b061ff0ef4ad0/README.md?plain=1#L251.

dmah42:
patches welcome! :)

itzmeanjan:
@dmah42  Sent PR https://github.com/google/benchmark/pull/1700

PR #1700, opened 2023-11-13
Add Missing `\n` at the End of Error Log String

itzmeanjan:
Closes https://github.com/google/benchmark/issues/1699

### DIFF
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -254,7 +254,7 @@ bool PerfCounters::IsCounterSupported(const std::string&) { return false; }
 PerfCounters PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
   if (!counter_names.empty()) {
-    GetErrorLogInstance() << "Performance counters not supported.";
+    GetErrorLogInstance() << "Performance counters not supported.\n";
   }
   return NoCounters();
 }


src/perf_counters.cc
```diff
@@ -254,7 +254,7 @@ bool PerfCounters::IsCounterSupported(const std::string&) { return false; }
 PerfCounters PerfCounters::Create(
     const std::vector<std::string>& counter_names) {
   if (!counter_names.empty()) {
-    GetErrorLogInstance() << "Performance counters not supported.";
+    GetErrorLogInstance() << "Performance counters not supported.\n";
```
dmah42, line 257: nit: `std::endl` is the C++ standard line ending output character.

dmah42 (approved):

src/perf_counters.cc
dmah42: let me know if you want to change this or if i should just merge it.

dmah42:
decided we can live with the `\n` :)

src/perf_counters.cc
itzmeanjan: Just seeing this and I would just stick to `\n` to avoid a STDOUT flush, more @ https://en.cppreference.com/w/cpp/io/manip/endl.

itzmeanjan:
Thanks for the review and merge @dmah42 :clinking_glasses:

PR #1701, opened 2023-11-15
Fix CMakeLists.txt for perf_counters_test

illbegood:
perf_counters_test fails because `benchmark/test/perf_counters_test.cc` has an INSTRUCTIONS counter. And doesn't have the BRANCHES one (since e441a8cb112a3c13629749ec8c3d65b2d170b10e).


dmah42 (approved):

dmah42:
thanks for spotting that

PR #1704, opened 2023-11-23
Support windows MSYS2 environments

LebedevRI:


LebedevRI:
Hurray, understanding that weird `Exit code 0xc0000135` was fun, that unblocks me elsewhere.

@dmah42 all yours, pls merge :)

include/benchmark/benchmark.h
```diff
@@ -746,13 +746,13 @@ class BENCHMARK_EXPORT State {
   // have been called previously.
   //
   // NOTE: KeepRunning may not be used after calling either of these functions.
-  BENCHMARK_ALWAYS_INLINE StateIterator begin();
-  BENCHMARK_ALWAYS_INLINE StateIterator end();
+  inline BENCHMARK_ALWAYS_INLINE StateIterator begin();
```
dmah42, line 749: should we just have a definition for `BENCHMARK_ALWAYS_INLINE` for WINDOWS && !CLANG or something?

test/CMakeLists.txt
```diff
@@ -64,30 +64,38 @@ macro(compile_output_test name)
           ${BENCHMARK_CXX_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
 endmacro(compile_output_test)
 
+macro(benchmark_add_test)
+  add_test(${ARGV})
+  if(WIN32 AND BUILD_SHARED_LIBS)
+    cmake_parse_arguments(TEST "" "NAME" "" ${ARGN})
+    set_tests_properties(${TEST_NAME} PROPERTIES ENVIRONMENT_MODIFICATION "PATH=path_list_prepend:$<TARGET_FILE_DIR:benchmark::benchmark>")
```
dmah42, line 71: is this something users of the library have had to do or are you introducing it with this change? (i don't use windows platforms so i have no idea...)

test/CMakeLists.txt
```diff
@@ -232,7 +240,12 @@ if (BENCHMARK_ENABLE_GTEST_TESTS)
 
   macro(add_gtest name)
     compile_gtest(${name})
-    add_test(NAME ${name} COMMAND ${name})
+    benchmark_add_test(NAME ${name} COMMAND ${name})
+    if(WIN32 AND BUILD_SHARED_LIBS)
+      set_tests_properties(${name} PROPERTIES
```
dmah42, line 245: this feels like something CMake should be taking care of with the dependencies, no?

test/CMakeLists.txt
LebedevRI: This is exactly the same thing you were previously doing in
https://github.com/google/benchmark/blob/c8ef1ee99eca565f6a58a9f815fba5ef1f210783/.github/workflows/build-and-test.yml#L105-L109

LebedevRI: (i'm just moving it into the right place)

test/CMakeLists.txt
LebedevRI: I don't know, this is just how things work on windows i guess.
I agree that it's very broken, and i don't use windows myself too.

include/benchmark/benchmark.h
LebedevRI: The issue is that the *definition* was marked as `inline`, while *declaration* was not.
In windows, that effectively results in two different function declarations.
So i don't think this is `BENCHMARK_ALWAYS_INLINE`-specific in any way,
and thus i don't see why `BENCHMARK_ALWAYS_INLINE` should be altered.

dmah42: it reads a bit surprisingly to have an "always inline" attribute but then have to supplement it with "inline".  i assumed you needed this because on some platforms "always inline" defines to "".

test/CMakeLists.txt
dmah42: i don't remember adding that.. anyway, it's fine, just seems like an odd windows only thing for cmake to need.

dmah42 (approved):

LebedevRI:
@dmah42 thank you!

Might it be time for `.4` release?

include/benchmark/benchmark.h
LebedevRI: As far as i understand it, `BENCHMARK_ALWAYS_INLINE` is an optimization hint
that does not contain the same binding semantical meaning (on the symbol visibility)
as the  `inline` one.

PR #1705, opened 2023-12-01
fix(cmakelists.txt): enforce CMake to find PFM or fail when BENCHMARK…

### LINKED ISSUE #1702: [FR] Enforce CMake to find PFM or fail when `BENCHMARK_ENABLE_LIBPFM` is `ON`
Currently, setting `BENCHMARK_ENABLE_LIBPFM` to `ON` only *tries* to find `PFM`, but I think it should *find it or fail* since the user wants it.

https://github.com/google/benchmark/blob/93a96a26a6ba36894d7fa6f43513708c595ccb5f/CMakeLists.txt#L314-L316

Does the `find_package` simply need the `REQUIRED` keyword?

ananta:
Enforce CMake to find PFM or fail when `BENCHMARK_ENABLE_LIBPFM` is `ON`

fix #1702

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,7 +312,7 @@ find_package(Threads REQUIRED)
 cxx_feature_check(PTHREAD_AFFINITY)
 
 if (BENCHMARK_ENABLE_LIBPFM)
-  find_package(PFM)
+  find_package(PFM REQUIRED)
 endif()
 
 # Set up directories


dmah42 (approved):

LebedevRI:
(@dmah42 i'm guessing this was simply forgot to be actually merged while waiting for ci)

PR #1706, opened 2023-12-05
Fix cmake endmacro label.

pkestene:


LebedevRI:
Looks like this was fixed in https://github.com/google/benchmark/pull/1715
Thank you!

PR #1709, opened 2023-12-06
[NFC] `complexity_n` is not of `IterationCount` type

LebedevRI:
There is no bug here, but it gave me a scare the other day.
It is not incorrect to use `IterationCount` here,
since it's just an `int64_t` either way,
but it's wildly confusing. Let's not do that.


dmah42:
nice catch.  i completely missed that.

dmah42 (approved):

LebedevRI:
@dmah42 thank you.

PR #1710, opened 2023-12-07
Fix editable install by unsetting `build_ext.copy_extensions_to_source`

nicholasjng:
This method was the culprit for the recent editable install breakage, since it just tries to copy the generated extension file without checking its existence.

Since the `BazelExtension` uses a non-standard location to store the build artifacts, calling the copy method fails the build since the extension is not found in the expected location.

But, since we already copy the file into the source tree as part of the `BazelExtension.bazel_build` method, it's fine - the extension appears in the right place, and the egg info is generated correctly as well.

This method also does not affect the general install, so it solves the editable problem without regressing the fixed install.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,14 @@ class BuildBazelExtension(build_ext.build_ext):
         # explicitly call `bazel shutdown` for graceful exit
         self.spawn(["bazel", "shutdown"])
 
+    def copy_extensions_to_source(self):
+        """
+        Copy generated extensions into the source tree.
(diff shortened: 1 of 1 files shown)


nicholasjng:
To check that this works, try to `python -m pip install -e .` from this branch.

What I would like to test before the merge (because I have run into this before) is whether the `shutil.copyfile` API actually overwrites the `.so` instead of changing its update time - if this is not the case, it leads to catastrophic failure on macOS due to its kernel protections.

dmah42 (approved):

PR #1711, opened 2023-12-07
Upgrade core bazel dependencies

nicholasjng:
Bumps `rules_foreign_cc` to v0.10.1 (October 2023), `bazel_skylib` to v1.5.0 (November 2023), `rules_python` to v0.27.1 (December 2023), and `nanobind` to v1.8.0 (November 2023).

Also syncs GoogleTest to v1.12.1 (the last C++11 supporting version) to be the same as in MODULE.bazel.

Since the latest `rules_python` changed its setup calling convention, that is updated also in the WORKSPACE file.

FYI: I removed absl (`com_google_absl` in the Bazel deps), since it appears to be currently unused.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -3,12 +3,12 @@ module(
     version = "1.8.3",
 )
 
-bazel_dep(name = "bazel_skylib", version = "1.4.2")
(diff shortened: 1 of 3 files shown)


nicholasjng:
This is going to time out again I fear.

PR #1713, opened 2023-12-18
Check out repo at depth 0 for Python tests, bump Python and PyPI actions

nicholasjng:
The reason for this is that `setuptools-scm` installs a version relative to the last release tag - if no tag is found, the default version is taken to be v0.1.0. This was the case in GitHub Actions, where only the PR branch is checked out.

Also unpins build system requirements in the `pyproject.toml`.

The sdist build system was changed to `build` from `python setup.py sdist` for forward compatibility - `build` is superior in every way, and the advertised solution by both cibuildwheel and PyPA itself.

Bump `actions/setup-python` to v5, `pypa/gh-action-pypi-publish` to v1.8.11, and `docker/setup-qemu-action` to v3.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -16,16 +16,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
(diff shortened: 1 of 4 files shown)


nicholasjng:
@dmah42 I noticed while working on this that some setup might need to be done in the GBM PyPI account to get trusted publishing to work - the Actions setup is only one of the required parts, the account setup being the other.

dmah42 (approved):

Issue #1716, opened 2023-12-20
[FR] Define HOST_NAME_MAX for z/OS


abhina-sree:
**Is your feature request related to a problem? Please describe.**

We hit this warning on z/OS #warning "HOST_NAME_MAX not defined. using 64" because we do not define HOST_NAME_MAX. 

**Describe the solution you'd like**

We will define it to _POSIX_HOST_NAME_MAX

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

abhina-sree:
This PR is merged in.

PR #1717, opened 2023-12-20
define HOST_NAME_MAX for z/OS

abhina-sree:
This patch defines HOST_NAME_MAX for z/OS to fix the following [issue](https://github.com/google/benchmark/issues/1716)


dmah42 (approved):

dmah42:
thanks!

Issue #1718, opened 2023-12-20
[FR] Add z/OS case to ThreadCPUUsage to avoid error


abhina-sree:
**Is your feature request related to a problem? Please describe.**

We are hitting the following error "Per-thread timing is not available on your system." because we do not support CLOCK_THREAD_CPUTIME_ID.

**Describe the solution you'd like**

I will implement a solution similar to SOLARIS that calls ProcessCPUUsage instead

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

LebedevRI:
This has been fixed by https://github.com/google/benchmark/pull/1719.

PR #1719, opened 2023-12-20
Fix per-thread timing error on z/OS

abhina-sree:
This patch calls ProcessCPUUsage() for z/OS because we do not support CLOCK_THREAD_CPUTIME_ID. This fixes the following [issue](https://github.com/google/benchmark/issues/1718)


dmah42 (approved):

dmah42:
thanks!

Issue #1720, opened 2023-12-22
[BUG] fail to import gbench into compare.py with bazel and python3.12 


dhmemi:
**Describe the bug**
Execute `tools/compare.py` with bazel and python3.12 will get an error:
```bash
$ bazelisk run @com_google_benchmark//tools:compare benchmarks ./baseline.json ./benchmark.json
Traceback (most recent call last):
  File "C:\Users\admin\AppData\Local\Temp\Bazel.runfiles_2c5wqe9z\runfiles\com_google_benchmark\tools\compare.py", line 13, in <module>
    import gbench
ModuleNotFoundError: No module named 'gbench'
```

**System**
Which OS, compiler, and compiler version are you using:
  - OS: 
  - Compiler and version: 

**To reproduce**
To reproduce this bug:
1. Import rules_python and register python3.12:
```starlark

http_archive(
    name = "rules_python",
    sha256 = "e85ae30de33625a63eca7fc40a94fea845e641888e52f32b6beea91e8b1b2793",
    strip_prefix = "rules_python-0.27.1",
    url = "https://github.com/bazelbuild/rules_python/releases/download/0.27.1/rules_python-0.27.1.tar.gz",
)

load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")

py_repositories()

python_register_toolchains(
    name = "python3_12",
    ignore_root_user_error = True,
    python_version = "3.12.0",
)

load("@python3_12//:defs.bzl", "interpreter")
load("@rules_python//python:pip.bzl", "pip_parse")
pip_parse(
    name = "tools_pip_deps",
    python_interpreter_target = interpreter,
    requirements_lock = "//:requirements-build.txt",
)

load("@tools_pip_deps//:requirements.bzl", install_deps_other = "install_deps")

# Call it to define repos for your requirements.
install_deps_other()

```

2. Import google benchmark in the WORKSPACE file:
```starlark
http_archive(
    name = "com_google_benchmark",
    sha256 = "0094b77c01e9973876f1831f76cbae43fdaf09b424e467dc4171abaa6884e455",
    strip_prefix = "benchmark-1.8.2",
    url = "https://github.com/google/benchmark/archive/refs/tags/v1.8.2.zip",
)
```
3. run command: `bazelisk run @com_google_benchmark//tools:compare benchmarks ./baseline.json ./benchmark.json` 

**Additional context**
One of the way to fix that may be add  `imports` field to `//tools/compare` target：
```starlark
[...]

PR #1721, opened 2023-12-22
fix: fail to import gbench in bazel and python3.12

dhmemi:
See #1720


dmah42:
you might have a tabs vs spaces issue in the change you made.

dmah42 (approved):

dmah42:
thanks!

PR #1723, opened 2023-12-28
Change Fixture to use non-const SetUp and TearDown in example

FantasqueX:
Const SetUp and TearDown were deprecated in https://github.com/google/benchmark/pull/285


FantasqueX:
Gently ping~

dmah42 (approved):

dmah42:
thank you!

PR #1724, opened 2024-01-02
Fix division by zero for low frequency timers for CV statistics

ivafanas:
`clock_gettime` resolution is implementation defined:
https://linux.die.net/man/3/clock_gettime

`double StatisticsCV(const std::vector<double>& v)` function gets vector of zeroes on system with low frequency timers in `user_counters_tabular_test` test `BM_Counters_Tabular/repeats:2/threads:1_cv` case. Division result is `nan` and test fails due to regexp mismatch.

Additionally, division by zero (even FP zero) is a UB in C++ and should be avoided.


dmah42 (approved):

dmah42:
nice :) thanks.

Issue #1725, opened 2024-01-03
[FR] Large File Support


oToToT:
**Is your feature request related to a problem? Please describe.**

Google benchmark looks to be using libc functions like `fseeko`, `ftello`, `fopen`. (IIUC, std::ifstream may results in using these functions.)
On some "small" platforms, it might result in EOVERFLOW when dealing with large files.

**Describe the solution you'd like**
Compiles with these three flags enabled:
  `-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE`.

this is the portable/POSIX method for transparently using 64bit types everywhere.  for details on each define, see:
    # [http://www.gnu.org/s/libc/manual/html_node/Feature-Test-Macros.html](https://www.google.com/url?q=http://www.gnu.org/s/libc/manual/html_node/Feature-Test-Macros.html&sa=D&source=buganizer&usg=AOvVaw20qmkaHe9EbXx97C__k15p)
    # _LARGEFILE_SOURCE: enable support for new LFS funcs (ftello/etc...)
    # _LARGEFILE64_SOURCE: enable support for 64bit variants (off64_t/fseeko64/etc...)
    # _FILE_OFFSET_BITS: default to 64bit variants (off_t is defined as off64_t)

**Describe alternatives you've considered**

N/A

**Additional context**

https://issuetracker.google.com/issues/201531268

oToToT:
(I can work on this.)

dmah42:
have you seen this issue anywhere?  `ifstream` is only used in context of opening sysinfo files ("/proc/cpuinfo" and similar) none of which should be "large".  `ofstream` is used to write out files and i suppose these could become large for larger projects, but maybe not on smaller platforms (as they couldn't run such large projects, i suppose).

oToToT:
Thanks @vapier for clearer information and sorry for my initial vague arguments.

IIUC, for 64-bit `time_t` support (to prevent [Y2K38](https://en.wikipedia.org/wiki/Year_2038_problem)), we only need to add `-D_TIME_BITS=64` to the compile flag if LFS is already there.
If we want to do so, I can adjust #1726 to include it (and rename it to "Enable Large-file Support + 64-bit `time_t`").

vapier:
let's keep the two separate for now.  LFS has been around for a long time and should be easy enough.  64-bit time_t is a newer project and still semi-influx.

PR #1726, opened 2024-01-03
Enable Large-file Support

oToToT:
This should fix https://github.com/google/benchmark/issues/1725

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -51,6 +51,10 @@ cc_library(
     }),
     defines = [
         "BENCHMARK_STATIC_DEFINE",
+	# Large-file Support
+	"_FILE_OFFSET_BITS=64",
+	"_LARGEFILE64_SOURCE",
+	"_LARGEFILE_SOURCE",
     ] + select({
         ":perfcounters": ["HAVE_LIBPFM"],
         "//conditions:default": [],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -51,10 +51,10 @@ cc_library(
     }),
     defines = [
         "BENCHMARK_STATIC_DEFINE",
-	# Large-file Support
-	"_FILE_OFFSET_BITS=64",
-	"_LARGEFILE64_SOURCE",
-	"_LARGEFILE_SOURCE",
+        # Turn on Large-file Support
+        "_FILE_OFFSET_BITS=64",
+        "_LARGEFILE64_SOURCE",
+        "_LARGEFILE_SOURCE",
     ] + select({
         ":perfcounters": ["HAVE_LIBPFM"],
         "//conditions:default": [],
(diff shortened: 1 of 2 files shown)


dmah42:
i'm not sure how necessary this is (see my comment on #1725 ) but if we do need it please also add these definitions to the bazel build.

dmah42 (approved):

dmah42:
i don't know what the pre-commit failure means or why it's an issue that the bazel file has changed in an expected way.  maybe do what it says (`pre-commit run --all-files`).

perhaps some file is being generated based on the BUILD.bazel?

oToToT:
Looks like I accidentally use tabs instead of whitespaces as indent. It should be fixed now.

dmah42 (approved):

dsharlet:
Is there a way to disable this change? Our project updated to a newer version of googlebenchmark, and this change breaks on some platforms, e.g.: https://github.com/google/XNNPACK/actions/runs/13079108015/job/36498319833

dsharlet:
I share @dmah42 's confusion about why this would be necessary, I doubt the benchmark framework attempts to work with such large files that this matters.

dmah42:
should the flags be conditionally included on 64-bit platforms only?

oToToT:
> should the flags be conditionally included on 64-bit platforms only?

IIUC, LFS is mainly for 32-bit platforms [1], so it has no sense to only include it on 64-bit platforms. (e.g. `ftell` should already be returning 64-bit values on 64-bit platforms)

It seems like XNNPACK is using r23b ndk [2]. Is it possible to raise it to r24? From my random searching, update the target to r24 should make LFS available. (I haven't tested it locally, so please correct me if I'm wrong.)

cc @vapier for vis

[1] https://en.wikipedia.org/wiki/Large-file_support
[2] https://github.com/google/XNNPACK/blob/c467d1a1b79db7c2686536720989cf524308eb4e/.github/workflows/build.yml#L277-L282

dsharlet:
Even with the latest NDK available, this same problem exists: https://github.com/google/XNNPACK/actions/runs/13122183940/job/36610509007?pr=7770

I really think we just need a way to disable this, and failing that, consider reverting it (I don't understand the motivation for this change).

oToToT:
> Even with the latest NDK available, this same problem exists: https://github.com/google/XNNPACK/actions/runs/13122183940/job/36610509007?pr=7770
> 
> I really think we just need a way to disable this, and failing that, consider reverting it (I don't understand the motivation for this change).

I think it works now after you have correctly updated both NDK version and compile flags.

dmah42:
my understanding of the motivation is that on some platforms (OSes) the inode can be 64-bit, meaning that without this we can get hard-to-find errors on I/O.  if there's a way to limit this change to those platforms, then let's do that.  otherwise, i'm happy to revert it until we see those errors in the wild.

PRs welcome for either.

oToToT:
Do we still need to revert it even XNNPACK has already solved the issue themselves?

https://github.com/google/XNNPACK/commit/411f9dea4e79c121f8adb2fa8eecabd7fd616555

dmah42:
if they were the only folks with an issue then no i don't think we do.

PR #1727, opened 2024-01-03
CycleClock: use RDTIME instead of RDCYCLE on RISC-V

aurel32:
Starting with Linux 6.6 [1], RDCYCLE is a privileged instruction on RISC-V and can't be used directly from userland. There is a sysctl option to change that as a transition period, but it will eventually disappear.

Use RDTIME instead, which while less accurate has the advantage of being synchronized between CPU (and thus monotonic) and of constant frequency.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cc4c07c89aada16229084eeb93895c95b7eabaa3

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -189,15 +189,16 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
 #endif
   return tsc;
 #elif defined(__riscv)  // RISC-V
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

dmah42:
thank you!

PR #1732, opened 2024-01-05
Issue1731 created console does not receive output

### LINKED ISSUE #1731: [BUG] Console created through AllocConsole does not receive colored output
**Describe the bug**
I use `benchmark.dll` inside a non-console application because I want to benchmark different implementations of ways to solve the same problem. And it is not possible to extract the code into a separate project, as the benchmarked code calls several other functions.

Therefore, I rolled my own BenchmarkMain method that does the same as BENCHMARK_MAIN macro, but also creates a console with `AllocConsole`, creates `std::ostream` instances for `STD_OUTPUT_HANDLE` and `STD_ERROR_HANDLE`, creates a `ConsoleReporter` instance and calls `RunSpecificBenchmarks` with this displayReporter.

The console did receive output, but it reported that Color printing was only supported on Windows - which makes no sense as the console being written to is on Windows.

Along the way, the `std::cout.rdbuf()` is saved, and replaced with the the same `.rdbuf()` that is used for the `std::ostream out` instance.

**System**
  - OS: Windows 11 23H2 (build 22631.2861)
  - Compiler and version: Visual C++ 2022 (Visual Studio 2022 17.8.3), cl.exe version 19.38.33133 for x64

**To reproduce**
Inside a non-console application, add the following function:

```
int BenchmarkMain (int argc, char **argv)
{
    char arg0_default[] = "benchmark";
    char* args_default = arg0_default;
    if (!argv) {
      argc = 1;
      argv = &args_default;
    }

    if (!argv)
    {
        argc = 2;
        argv = (char **) args_default;
    }

    ::benchmark::Initialize (&argc, argv);
    if (::benchmark::ReportUnrecognizedArguments (argc, argv))
    {
        return 1;
    }

    // Create console reporter
    AllocConsole ();
    HANDLE stdOutHandle = GetStdHandle (STD_OUTPUT_HANDLE);
    HANDLE stdErrHandle = GetStdHandle (STD_ERROR_HANDLE);


    // Get file descriptors
    int fdOut = _open_osfhandle ((intptr_t) stdOutHandle, _O_TEXT);
    int fdErr = _open_osfhandle ((intptr_t) stdErrHandle, _O_TEXT);

    // Get FILE*
    FILE *fpOut = _fdopen (fdOut, "w");
    FILE *fpErr = _fdopen (fdErr, "w");

    // Create ofstream
    std::ofstream outStream (fpOut);
    std::ofstream errStream (fpErr);

    // Create ostream
    std::ostream out (outStream.rdbuf ());
    std::ostream err (errStream.rdbuf ());

    auto savedBuf = std::cout.rdbuf ();
    std::cout.rdbuf (outStream.rdbuf ());

    auto displayReporter = std::make_unique<::benchmark::ConsoleReporter> ();
    displayReporter->SetOutputStream (&out);
    displayReporter->SetErrorStream (&err);

    ::benchmark::RunSpecifiedBenchmarks (displayReporter.get());

    // Restore std::cout.rdbuf() - otherwise an exception is thrown at exit.
    std::cout.rdbuf (savedBuf);

    return 0;
}
```

Make sure that the function is being called when the application starts - I control this with a command line parameter.

Note that the output written to the console is without any color formatting.

**Expected behavior**
Output from benchmarking should get color-enabled output in the created console window

**Additional context**
The application, where I call the benchmark library, is a MFC Windows application created with use of an external library (CodeJock).

bstordrup:
Changed to allow coloring output on a console created via `AllocConsole` from a non-console project on Windows.

Fixes #1731

### DIFF
--- a/src/colorprint.cc
+++ b/src/colorprint.cc
@@ -140,12 +140,12 @@ void ColorPrintf(std::ostream& out, LogColor color, const char* fmt,
   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
   // printed but has not yet reached the console.
-  fflush(stdout);
+  out.flush();
   SetConsoleTextAttribute(stdout_handle,
                           GetPlatformColorCode(color) | FOREGROUND_INTENSITY);
-  vprintf(fmt, args);
+  out << FormatString(fmt, args);
 
-  fflush(stdout);
+  out.flush();
   // Restores the text color.
   SetConsoleTextAttribute(stdout_handle, old_color_attrs);
 #else
--- a/src/console_reporter.cc
+++ b/src/console_reporter.cc
@@ -42,11 +42,15 @@ bool ConsoleReporter::ReportContext(const Context& context) {
   PrintBasicContext(&GetErrorStream(), context);
 
 #ifdef BENCHMARK_OS_WINDOWS
-  if ((output_options_ & OO_Color) && &std::cout != &GetOutputStream()) {
-    GetErrorStream()
-        << "Color printing is only supported for stdout on windows."
-           " Disabling color printing\n";
-    output_options_ = static_cast<OutputOptions>(output_options_ & ~OO_Color);
+  if ((output_options_ & OO_Color)) {
+    auto stdOutBuf = std::cout.rdbuf();
+    auto outStreamBuf = GetOutputStream().rdbuf();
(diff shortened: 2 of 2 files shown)


dmah42 (approved):

dmah42:
thanks for catching this.

bstordrup:
@dmah42, you are so welcome :-)

PR #1733, opened 2024-01-06
tools/compare: don't actually discard valid (but zero) `pvalue`

LebedevRI:
This is a bit embarrassing.

The test added in middle commit does not produce a pvalue,
because the pvalue is `0`. Duh!

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -315,7 +315,7 @@ def get_difference_report(json1, json2, utest=False):
             have_optimal_repetitions, cpu_pvalue, time_pvalue = calc_utest(
                 timings_cpu, timings_time
             )
-            if cpu_pvalue and time_pvalue:
+            if cpu_pvalue is not None and time_pvalue is not None:
                 utest_results = {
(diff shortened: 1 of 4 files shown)


tools/gbench/report.py
```diff
@@ -1490,6 +1490,91 @@ def test_json_diff_report_pretty_printing(self):
...
+
+            testInputs = os.path.join(
+                os.path.dirname(os.path.realpath(__file__)), "Inputs"
+            )
+            testOutput1 = os.path.join(testInputs, "test5_run0.json")
+            testOutput2 = os.path.join(testInputs, "test5_run1.json")
+            with open(testOutput1, "r") as f:
+                json1 = json.load(f)
+                json1["benchmarks"] = [json1["benchmarks"][0]
+                                       for i in range(1000)]
```
LebedevRI, line 1509: I assume we don't want a `.json` with 10k lines? :)

dmah42 (approved):

dmah42:
good catch. thanks!

LebedevRI:
@dmah42 thank you!

PR #1735, opened 2024-01-09
Issue 1734: Streams not flushed if not running actual benchmarks

bstordrup:
If the call paths in RunSpecifiedBenchmarks do not go into internal::RunBenchmarks, the Out and Err streams are flushed before exiting.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -617,8 +617,6 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* display_reporter,
   if (FLAGS_benchmark_list_tests) {
     for (auto const& benchmark : benchmarks)
       Out << benchmark.name().str() << "\n";
-    Out.flush();
-    Err.flush();
   } else {
     internal::RunBenchmarks(benchmarks, display_reporter, file_reporter);
   }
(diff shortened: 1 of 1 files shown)


LebedevRI:
I suppose this makes sense, we can't quite be sure that `std::endl` is consistently used.
But, let's streamline this: let's flush *both* streams, but only before `exit`/`return`.

bstordrup:
Added some more calls to Out.flush() and Err.flush()

bstordrup:
I noted that there is a complexity test failing, but not sure why.

src/benchmark.cc
```diff
@@ -597,20 +601,30 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* display_reporter,
...
+    Out.flush();
+    Err.flush();
     return 0;
   }
 
   if (FLAGS_benchmark_list_tests) {
     for (auto const& benchmark : benchmarks)
       Out << benchmark.name().str() << "\n";
+    Out.flush();
+    Err.flush();
```
LebedevRI, line 621: These are not needed

bstordrup: Removed

LebedevRI (approved):
LGTM, thank you!

bstordrup:
@LebedevRI, you,re welcome... Just started to use the library - it's really cool.

LebedevRI:
I'm not sure why those `complexity_benchmark` failures happen. Looks like there's some non-determinizm.

PR #1737, opened 2024-01-12
Allow for collection of more PMUs than are physically available.

DivineOb:
Normally we can only collect as many PMUs as we have physical PMU counters because they are managed as a single pinned group and extra requested PMUs will not be collected. With this patch we instead collect each PMU as a separate group once this limit is hit using PMU multiplexing.

Also fix a number of typos.

### DIFF
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -86,6 +86,12 @@ PerfCounters PerfCounters::Create(
     Initialize();
   }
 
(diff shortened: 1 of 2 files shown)


LebedevRI:
Could you please split the typos into a preparatory PR?

Would https://github.com/google/benchmark/pull/1380 address your usecase?

PR #1738, opened 2024-01-12
Enable vectors in `DoNotOptimize`

goldsteinn:
Just a QOL change to make benchmarks vector code easier.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -461,15 +461,15 @@ BENCHMARK_DEPRECATED_MSG(
     "The const-ref version of this method can permit "
     "undesired compiler optimizations in benchmarks")
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
-  asm volatile("" : : "r,m"(value) : "memory");
+  asm volatile("" : : "r,m,v"(value) : "memory");
 }
 
 template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
 #if defined(__clang__)
-  asm volatile("" : "+r,m"(value) : : "memory");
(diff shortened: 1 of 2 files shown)


LebedevRI:
Tests?

goldsteinn:
> Tests?

Do you want aarch64 + x86_64, or is just x86_64 enough.

LebedevRI:
> > Tests?
> 
> Do you want aarch64 + x86_64, or is just x86_64 enough.

I don't really know what kind of input variable type this targets
so i can't tell what exactly should be tested.
Can this be generically reached via clang vector types e.g.?

goldsteinn:
> > > Tests?
> > 
> > 
> > Do you want aarch64 + x86_64, or is just x86_64 enough.
> 
> I don't really know what kind of input variable type this targets so i can't tell what exactly should be tested. Can this be generically reached via clang vector types e.g.?

Done, tests only compiled for gcc/clang.

LebedevRI:
Some generic questions:
1. Has that asm constraint always been supported in every compiler version ever released?
2. What happens when it's used on other compilers? (MSVC?)
3. Does the order matter? Should it be `r,v,m` (i.e. mem being last)?

goldsteinn:
> Some generic questions:
> 
> 1. Has that asm constraint always been supported in every compiler version ever released?
> 2. What happens when it's used on other compilers? (MSVC?)
> 3. Does the order matter? Should it be `r,v,m` (i.e. mem being last)?

Yeah, this patch is a bad idea. Its not super portable AFAICT (across compilers or arch within).

LebedevRI:
Aha, so clang 3.9 didn't know that: https://godbolt.org/z/eP9zbjb4P
And even then, it's support is target dependent: https://godbolt.org/z/eP9zbjb4P
This, indeed, makes me uneasy about this feature, unfortunately...

Issue #1707, opened 2023-12-05
[BUG] QNX compilation failure on v1.8.3. Patch suggestion.


Arech:
**Describe the bug**
There are some trivial compilation errors due `-Werror` and old-style C-casts in QNX specific block of `src/sysinfo.cc`
```
FAILED: src/CMakeFiles/benchmark.dir/sysinfo.cc.o
qnx-sdp/qnx710/host/linux/x86_64/usr/bin/ntoaarch64-g++ --sysroot=qnx-sdp/qnx710/target/qnx7 -DBENCHMARK_STATIC_DEFINE -DHAVE_STD_REGEX -DHAVE_STEADY_CLOCK -D_QNX_SOURCE -I/home/user/vcpkg/buildtrees/benchmark/src/v1.8.3-eb6e61b079.clean/include -I/home/user/vcpkg/buildtrees/benchmark/src/v1.8.3-eb6e61b079.clean/src -fexceptions -D_XOPEN_SOURCE=600 -D_POSIX_C_SOURCE=200112L -D_QNX_SOURCE -fexceptions -D_XOPEN_SOURCE=600 -D_POSIX_C_SOURCE=200112L -D_QNX_SOURCE  -Wall  -Wextra  -Wshadow  -Wfloat-equal  -Wold-style-cast  -Werror  -Wsuggest-override  -pedantic  -pedantic-errors  -fstrict-aliasing  -Wno-deprecated-declarations  -Wno-deprecated  -Wstrict-aliasing -g -std=c++11 -fvisibility=hidden -fvisibility-inlines-hidden -MD -MT src/CMakeFiles/benchmark.dir/sysinfo.cc.o -MF src/CMakeFiles/benchmark.dir/sysinfo.cc.o.d -o src/CMakeFiles/benchmark.dir/sysinfo.cc.o -c /home/user/vcpkg/buildtrees/benchmark/src/v1.8.3-eb6e61b079.clean/src/sysinfo.cc
/home/user/vcpkg/buildtrees/benchmark/src/v1.8.3-eb6e61b079.clean/src/sysinfo.cc: In function 'double benchmark::{anonymous}::GetCPUCyclesPerSecond(benchmark::CPUInfo::Scaling)':
/home/user/vcpkg/buildtrees/benchmark/src/v1.8.3-eb6e61b079.clean/src/sysinfo.cc:774:69: error: use of old-style cast to 'int64_t' {aka 'long int'} [-Werror=old-style-cast]
[...]

PR #1739, opened 2024-01-16
Fix C-style typecasting in QNX-specific code

Arech:
C-style typecasting breaks the build due to `-Werror=old-style-cast` which should remain in place.

Resolves https://github.com/google/benchmark/issues/1707

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -777,8 +777,9 @@ double GetCPUCyclesPerSecond(CPUInfo::Scaling scaling) {
   kstat_close(kc);
   return clock_hz;
 #elif defined(BENCHMARK_OS_QNX)
-  return static_cast<double>((int64_t)(SYSPAGE_ENTRY(cpuinfo)->speed) *
-                             (int64_t)(1000 * 1000));
+  return static_cast<double>(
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

Arech:
`complexity_benchmark` is failing with a pretty cryptic error to me.
```
66: BM_Complexity_O_N/2048         D:/a/benchmark/benchmark/test/output_test_helper.cc:116: CheckCase: Check `TC.match_rule != MR_Next' failed. Expected line "BM_Complexity_O_N_RMS                              7 %            25 %    " to match regex "^%rms_name %rms %rms[ ]*$"
66:     actual regex string "^BM_Complexity_O1_RMS [ ]*[0-9]+ % [ ]*[0-9]+ %[ ]*$"
66:     started matching near: BM_Complexity_O_N_RMS                              7 %            25 %    
66:                 11826 ns        15625 ns         1000
```
https://github.com/google/benchmark/actions/runs/7540451018/job/20525988749?pr=1739#step:6:5222

I don't think it might ever be affected by the PR. Can anyone please suggest what to do?

Arech:
@dmah42 ah, so it won't break the merge process then? Ok, then it looks like my role is over here and I presume it should merge automatically. Thanks for a quick review!

dmah42:
thank you!

PR #1740, opened 2024-01-18
Bump pre-commit dependencies

nicholasjng:
Bumps `mypy` to v1.8.0, and `ruff` to v0.1.13.

Also fix a mypy error in `tools.gbench.util` - the condition behaves the same as before, but in the new mypy version, the old condition results in an unreachable code error for the final `return False` statement.

This is most likely a bug in mypy's reachability analysis, but the fix is easy enough here to circumvent it.

### DIFF
--- a/tools/gbench/util.py
+++ b/tools/gbench/util.py
@@ -131,9 +131,7 @@ def load_benchmark_results(fname, benchmark_filter):
         if benchmark_filter is None:
             return True
         name = benchmark.get("run_name", None) or benchmark["name"]
-        if re.search(benchmark_filter, name):
-            return True
-        return False
+        return re.search(benchmark_filter, name) is not None
 
     with open(fname, "r") as f:
         results = json.load(f)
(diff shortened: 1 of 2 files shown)


tools/gbench/util.py
```diff
@@ -131,7 +131,7 @@ def benchmark_wanted(benchmark):
         if benchmark_filter is None:
             return True
         name = benchmark.get("run_name", None) or benchmark["name"]
-        if re.search(benchmark_filter, name):
+        if re.search(benchmark_filter, name) is not None:
```
dmah42, line 134: how about just

`return re.search(benchmark_filter, name) is not None`

nicholasjng: Good call! Done in [6eece2b](https://github.com/google/benchmark/pull/1740/commits/6eece2bc2465facb62b69ab1285ee7634be23cab).

dmah42 (approved):

PR #1744, opened 2024-01-25
Add min rel accuracy stopping criterion

maartenarnst:


src/thread_timer.h
```diff
@@ -75,9 +83,11 @@ class ThreadTimer {
 
   // Accumulated time so far (does not contain current slice if running_)
   double real_time_used_ = 0;
+
   double cpu_time_used_ = 0;
   // Manually set iteration time. User sets this with SetIterationTime(seconds).
   double manual_time_used_ = 0;
+  double manual_time_used2_ = 0;
```
LebedevRI, line 90: So we need to track every timer as-is *and* as as a sum-of-squares?

src/benchmark_runner.cc
```diff
@@ -341,13 +337,13 @@ bool BenchmarkRunner::ShouldReportIterationResults(
   // or because an error was reported.
   return i.results.skipped_ ||
          i.iters >= kMaxIterations ||  // Too many iterations already.
-         i.seconds >=
-             GetMinTimeToApply() ||  // The elapsed time is large enough.
+         (((i.seconds >= GetMinTimeToApply()) &&
+          (b.use_manual_time() && !IsZero(GetMinRelAccuracy()) && (std::sqrt(i.seconds2 / i.iters - std::pow(i.seconds / i.iters, 2.)) / (i.seconds / i.iters) / sqrt(i.iters) <= GetMinRelAccuracy()))) || // The relative accuracy is enough.
```
LebedevRI, line 341: Err, i think this needs a bit of refactoring...

src/benchmark_runner.cc
```diff
@@ -322,17 +326,9 @@ IterationCount BenchmarkRunner::PredictNumItersNeeded(
...
-  // So what seems to be the sufficiently-large iteration count? Round up.
-  const IterationCount max_next_iters = static_cast<IterationCount>(
-      std::llround(std::max(multiplier * static_cast<double>(i.iters),
-                            static_cast<double>(i.iters) + 1.0)));
-  // But we do have *some* limits though..
-  const IterationCount next_iters = std::min(max_next_iters, kMaxIterations);
-
-  BM_VLOG(3) << "Next iters: " << next_iters << ", " << multiplier << "\n";
-  return next_iters;  // round up before conversion to integer.
-}
```
LebedevRI, line 335: I do not understand the changes to this function.
Does this even compile? `multiplier` is a local variable, and it seems unused now.

src/thread_timer.h
maartenarnst: Yes, the idea is to look at the time durations as independent identically distributed samples of a random variable. And then use second-order statistics to estimate mean, variance, and accuracy of the estimate of the mean. So for this, we need indeed to keep track of the sum of the squares of the time durations.

src/benchmark_runner.cc
```diff
@@ -139,6 +139,7 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
     results.cpu_time_used += timer.cpu_time_used();
     results.real_time_used += timer.real_time_used();
     results.manual_time_used += timer.manual_time_used();
+    results.manual_time_used2 += timer.manual_time_used2();
```
romintomasetti, line 142:
```suggestion
    results.manual_time_used_pow2 += timer.manual_time_used2();
```
I think it's a better name.

romintomasetti: (same remark for other `<...>2` that would become `<...>_pow2`)

LebedevRI:
As you probably noticed, non-manual timers don't produce per-iteration times
(I'm somewhat planning on looking into that, but i don't expect the default to change),
so this fundamentally only works for manual time.

Hphu2004 (approved):

include/benchmark/benchmark.h
```diff
@@ -1190,11 +1194,21 @@ class BENCHMARK_EXPORT Benchmark {
...
+  // Set the minimum amount of time to use to determine the required number
+  // of iterations when running this benchmark. This option overrides
+  // the `benchmark_min_time` flag.
   // REQUIRES: `t > 0` and `Iterations` has not been called on this benchmark.
   Benchmark* MinTime(double t);
 
+  // Set the minimum relative accuracy to use to determine the required number
+  // of iterations when running this benchmark. This option overrides
+  // the `benchmark_min_rel_accuracy` flag.
+  // REQUIRES: `t > 0`, `Iterations` has not been called on this benchmark, and
```
romintomasetti, line 1206:
```suggestion
  // REQUIRES: `r > 0`, `Iterations` has not been called on this benchmark, and
```

include/benchmark/benchmark.h
```diff
@@ -1190,11 +1194,21 @@ class BENCHMARK_EXPORT Benchmark {
...
+  // time is measured manually, i.e., `UseManualTime` has been called on this
+  // benchmark and each benchmark iteration should call SetIterationTime(seconds)
```
romintomasetti, line 1208:
```suggestion
  // benchmark and each benchmark iteration should call `SetIterationTime(seconds)`
```

src/benchmark.cc
```diff
@@ -81,6 +81,18 @@ BM_DEFINE_string(benchmark_filter, "");
...
+// long to run the benchmark.
+//
+// REQUIRES: time is measured manually.
+//
+// Manual timers provide per-iteration times. The relative accuracy is
+// measured as the standard deviation of these per-iteration times divided by
+// the mean and the square root of the number of iterations. The benchmark is
+// run until the specified minimum time or number of iterations is reached
+// and the measured relative accuracy meets the specified requirement.
+BM_DEFINE_double(benchmark_min_rel_accuracy, 0.0);
```
romintomasetti, line 94:
```suggestion
// Manual timers provide per-iteration times. The relative accuracy is
// measured as the standard deviation of these per-iteration times divided by
// the mean and the square root of the number of iterations. The benchmark is
// run until the following conditions are fulfilled:
//   1. the specified minimum time or number of iterations is reached
//   2. the measured relative accuracy meets the specified requirement
BM_DEFINE_double(benchmark_min_rel_accuracy, 0.0);
```
(I find your formulation prone to misunderstandings)

include/benchmark/benchmark.h
maartenarnst: OK, done.

include/benchmark/benchmark.h
maartenarnst: OK, done.

src/benchmark.cc
maartenarnst: Thanks and done. Better indeed.

src/benchmark_runner.cc
maartenarnst: OK, done.

maartenarnst:
Hi @LebedevRI. In the mean time, I have elaborated the PR more. It is now in a better state for discussion. 

Questions/issues that @romintomasetti and I are thinking of are:

- Is this feature of controlling the number of iterations through a statistical criterion of interest to you? Our main motivation is that it's not always easy to set a "min time" in advance to ensure that the result has converged. Even if it's ultimately not retained as a stopping criterion, reporting a measure of statistical accuracy may be of interest.

- I agree that as it stands, the feature is limited to benchmarks that use manual timing. On our side, we use `benchmark` for GPU kernels, mainly with `Kokkos`, an HPC library that itself also uses `benchmark` for its performance tests.  We use manual timing, as you also suggest in your user guide. This use case may be sufficiently widespread as a motivation. (It may be worth exploring whether calls to the iterator for the `state` might be a way of obtaining per-iteration times for benchmarks that don't use manual timing).  

- How should the new stopping criterion (statistical relative accuracy) interact with the existing stopping criterion (time or number of iterations)? In the current state of the PR, `ShouldReportIterationResults` uses an "AND" logic. It will return true if both criteria are reached. For both criteria, there is an opt-out/opt-in by setting the threshold to zero. So the current behaviour of `benchmark` is maintained, and this PR is mainly adding functionality.

- Do you agree with the current way of implementing the feature? The current implementation keeps track of the sum of the squares of the per-iteration times. An alternative could be to store all the per-iteration times. An advantage would be that such an alternative implementation would allow your existing code for computing statistics to be reused. It would also be more flexible for future extensions, e.g. looking at statistics other than the mean and variance, such as detecting outliers. Drawbacks may be the increased storage requirement and needs to dynamically resize the storage space.

Many thanks in advance!

LebedevRI:
> * Is this feature of controlling the number of iterations through a statistical criterion of interest to you? Our main motivation is that it's not always easy to set a "min time" in advance to ensure that the result has converged. Even if it's ultimately not retained as a stopping criterion, reporting a measure of statistical accuracy may be of interest.

I don't have an answer to this question. Following replies assume "yes", but i'm undecided.

> * I agree that as it stands, the feature is limited to benchmarks that use manual timing. On our side, we use `benchmark` for GPU kernels, mainly with `Kokkos`, an HPC library that itself also uses `benchmark` for its 
performance tests.  We use manual timing, as you also suggest in your user guide.

Yes, that use of manual time for GPU kernels is indeed very reasonable.

> This use case may be sufficiently widespread as a motivation.

> (It may be worth exploring whether calls to the iterator for the `state` might be a way of obtaining per-iteration times for benchmarks that don't use manual timing).

True per-iteration times are, a complicated topic. 
The current library has been very intentionally designed
to prohibit that, and while, as i have said,
i was already planning on looking into that,
i'm not sure how that would look like. 

> * How should the new stopping criterion (statistical relative accuracy) interact with the existing stopping criterion (time or number of iterations)? In the current state of the PR, `ShouldReportIterationResults` uses an "AND" logic. It will return true if both criteria are reached. For both criteria, there is an opt-out/opt-in by setting the threshold to zero. So the current behaviour of `benchmark` is maintained, and this PR is mainly adding functionality.

That's my biggest problem honestly. The existing lack of max-time limit
for manual time already seems problematic. What if the requested accuracy
is more precise than what is actually possible on a given machine?
(run out of time to reply here, may elaborate later)

> * Do you agree with the current way of implementing the feature? The current implementation keeps track of the sum of the squares of the per-iteration times. An alternative could be to store all the per-iteration times. An advantage would be that such an alternative implementation would allow your existing code for computing statistics to be reused. It would also be more flexible for future extensions, e.g. looking at statistics other than the mean and variance, such as detecting outliers. Drawbacks may be the increased storage requirement and needs to dynamically resize the storage space.

This feature does not need to *store* per-iteration times,
it only needs to compute standard deviation of all of the per-iteration times,
which can be computed as a "live stdev" the way you currently do it,
so i certainly agree with only just a new accumulator,
and would not agree with an alternative design.

As i have said, true per-iteration times is a complicated topic, and storing them doubly so.

LebedevRI:
Also, `BenchmarkRunner::DoOneRepetition()` is wrong now,
we now can't re-use iteration count from one repetition to the others.
(now == with this accuracy check).

maartenarnst:
> Also, `BenchmarkRunner::DoOneRepetition()` is wrong now,
> we now can't re-use iteration count from one repetition to the others.
> (now == with this accuracy check).

I'm not sure to understand entirely what you mean.

The condition 

- https://github.com/google/benchmark/blob/30a37e1b0bfbec07bbb44f8cac8c162ef9f5f9ed/src/benchmark_runner.cc#L432-L434

seems to be such that even after the changes in this PR, the behavior is that the first repetition sets the number of iterations and the next repetitions then reuse that number. 

An issue may be that subsequent repetitions don't reach the required relative accuracy for the reused number of iterations. But isn't this issue also present for the current stopping criterion based on the time?

LebedevRI:
> The condition
> 
> * https://github.com/google/benchmark/blob/30a37e1b0bfbec07bbb44f8cac8c162ef9f5f9ed/src/benchmark_runner.cc#L432-L434
> 
> seems to be such that even after the changes in this PR, the behavior is that the first repetition sets the number of iterations and the next repetitions then reuse that number.
> 
> An issue may be that subsequent repetitions don't reach the required relative accuracy for the reused number of iterations.

Precisely, yes. Presumably that is not what you want if you explicitly request convergence-based run-time.

> But isn't this issue also present for the current stopping criterion based on the time?

Sure.

LebedevRI:
(To be noted, i'm just pointing out that failure mode,
perhaps you consider such unexpected outcome to be acceptable.)

maartenarnst:
Hi @LebedevRI.

In the mean time, I've thought a bit more about it.

An answer may be to look at it as two steps. This PR could be a first step focused on introducing a new convergence criterion. And exploring changes to the behavior for multiple repetitions could be a follow-up.

If I do think now about the behavior for multiple repetitions, I'm not sure I would call the issue a failure mode.

From a statistical point of view, doing the repetitions using a fixed number of iterations may be quite sensible. An advantage would be that basic versions of theoretical results from statistics would give some support to the approach. Such as basic versions of the central limit theorem to characterize theoretically the variability of a partial sum of a fixed number of random variables.  

The current behavior for multiple repetitions, and keeping it unchanged for the new convergence criterion, also appears quite clear and easy to understand. The first repetition fixes the number of iterations. It can be done via a maximum number of iterations, a minimum number of time to apply, and/or, now, a minimum relative accuracy. Then, in subsequent repetitions, the fixed number of iterations is reused. Perhaps doing the experiment of trying to write a piece of doc about the new convergence criterion could be a good next step.

Those would be my two cents, coming more from a background of statistics.

PR #1746, opened 2024-01-29
set library version in bazel

dmah42:
note if bzlmod is disabled the version will be an empty string

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -748,9 +748,7 @@ int InitializeStreams() {
 
 }  // end namespace internal
 
-std::string GetBenchmarkVersiom() {
-  return {BENCHMARK_VERSION};
-}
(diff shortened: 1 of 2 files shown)


LebedevRI:
Also drop the `hello, bazel` fallback from code?

src/benchmark.cc
```diff
@@ -752,7 +752,7 @@ std::string GetBenchmarkVersiom() {
 #if defined(BENCHMARK_VERSION)
   return {BENCHMARK_VERSION};
 #else
-  return "hello, bazel!";
```
LebedevRI, line 755:
```suggestion
std::string GetBenchmarkVersiom() {
  return {BENCHMARK_VERSION};
}
```

LebedevRI (approved):
Seems fine to me.
Thank you

LebedevRI (approved):

PR #1753, opened 2024-02-07
CycleClock: Add support for Alpha architecture

thesamesam:
This uses the native RPCC instruction, which comes with a caveat.  The
PCC (Program Cycle Register) tick counter on Alpha is a 32-bit unsigned
integer, which means that it overflows every ~4-ish seconds, depending
on clock speed.  Therefore this is only safe to use for small
measurements.  This is the same implementation used in fftw.


src/cycleclock.h
```diff
@@ -218,6 +218,14 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   uint64_t pcycle;
   asm volatile("%0 = C15:14" : "=r"(pcycle));
   return static_cast<double>(pcycle);
+#elif defined(__alpha__)
+  // The PCC (process cycle counter) register is an unsigned 32-bit integer,
+  // which means it overflows every ~4sec depending on CPU frequency.  This
+  // tick count will only be reliable for short intervals.  This implementation
+  // is the same as what fftw uses.
+  uint64_t cc;
```
dmah42, line 226: two thoughts: if we know it's 32-bits why do we read into a 64-bit and mask?

also if it's only 4 seconds i'm not sure it's going to be very useful. is there no alternative?

src/sysinfo.cc
```diff
@@ -550,7 +550,9 @@ int GetNumCPUs() {
             "CPU ID assignments in /proc/cpuinfo seem messed up."
             " This is usually caused by a bad BIOS.\n");
   }
-  return num_cpus;
+  // If we are unable to identify the number of CPUs, default to one rather
+  // than zero.
```
LebedevRI, line 554: This shouldn't be a global fix-up.

src/cycleclock.h
LebedevRI: Does `gettimeofday()` (or an alternative) work on that arch? There's precedent for using that.

matoro: > two thoughts: if we know it's 32-bits why do we read into a 64-bit and mask?
> 
> also if it's only 4 seconds i'm not sure it's going to be very useful. is there no alternative?

Hi, I helped work on this.  It's read into a 64-bit because the register is 64-bits wide, the lower 32 bits are the actual counter, while the high 32 bits are "operating system dependent".  Linux does not do anything with them today.

> Does `gettimeofday()` (or an alternative) work on that arch? There's precedent for using that.

It works, the problem is that clock precision is too low.  You can see this for example in the python bindings:

```
In [2]: time.clock_getres(time.CLOCK_MONOTONIC)
Out[2]: 0.000976563
```

This has been the cause of many a PR of mine, as on alpha two calls to `gettimeofday()` less than ~1ms apart will return the same timestamp.

That said, the `rpcc` technique seems to sufficient to pass the test suite.  I could include both options and make it configurable at compile-time perhaps?

src/sysinfo.cc
matoro: Can you elaborate what you mean?  It should continue to default to zero and there should be code to parse this nonstandard `/proc/cpuinfo`?  Or something else?

src/cycleclock.h
LebedevRI: I don't think it makes sense to make this configurable.
It should just work right (by the definition of "right",
that is expected by the users of that architecture),
or not at all.

matoro: > I don't think it makes sense to make this configurable. It should just work right (by the definition of "right", that is expected by the users of that architecture), or not at all.

Then `rpcc` is the way to go.  There isn't any method of measuring cycle counts for arbitrary periods on the architecture, so this is the best compromise.

src/sysinfo.cc
matoro: > Presumably no other architecture so far has needed this fix-up, so if this fix-up is needed for alpha, it should not be retroactively added to other architectures at the same time.
> 
> If `/proc/cpuinfo` is different there, then you certainly should actually parse it in an alpha-specific way.

I believe there should be some sort of error condition then, rather than returning zero as if it was a valid number of CPUs.  The calling code does not seem to consider this an error condition, and just blindly keeps going.  For example, when returning zero CPUs, later code subtracts 1 from this (resulting in -1), then tries to cast to a `size_t` and allocate a vector of beyond `SIZE_MAX` length.  So, not being able to parse `/proc/cpuinfo` does not seem to have been considered as a valid error condition.

Would you be okay with having this throw a runtime error if `/proc/cpuinfo` is unable to be parsed?

LebedevRI: Once again, does that happen on Alpha, or on an other architecture?

matoro: > Once again, does that happen on Alpha, or on an other architecture?

It happens on Alpha, but it would also happen on any architecture that uses a different format for `/proc/cpuinfo`.  For example, I just checked, and SPARC also uses a different format and runs into the same problem.

Yes, parsing for their formats should be added, but why not bail out when you don't have enough information rather than returning data you know is bad, especially if the callers don't check it?

LebedevRI: For example because it hides bugs. By the same logic,
`#error You need to define CycleTimer for your OS and CPU`
should not be there, but it should default to standard wall clock.
Either the architecture is supported, or it is not.

matoro: > For example because it hides bugs. By the same logic, `#error You need to define CycleTimer for your OS and CPU` should not be there, but it should default to standard wall clock. Either the architecture is supported, or it is not.

Right, that's what I'm proposing.  Instead of defaulting to 0 CPUs or 1 CPU, throw an error instead (though it has to be a runtime error rather than compile-time).  Is that an OK solution?

LebedevRI: I suppose
```
  if (num_cpus == 0) {
    std::cerr << "Failed to detect CPU count\n";
    return -1;
  }
```
would make sense here, yes.
But not soft fix-up like it is now.

matoro: > I suppose
> 
> ```
>   if (num_cpus == 0) {
>     std::cerr << "Failed to detect CPU count\n";
>     return -1;
>   }
> ```
> 
> would make sense here, yes. But not soft fix-up like it is now.

This won't work either, because callers don't check for errors.  I think it will have to be an exception.

LebedevRI: That is how errors are dealt with elsewhere in the function.

src/cycleclock.h
dmah42: the test suite is optimised to run quickly so that's not surprising.

i don't know if we have real world users that have per iteration timings of >4s but it wouldn't surprise me if we did.  i think until there's a 64-bit high precision timer on arch, gettimeofday is going to be the better option.

src/sysinfo.cc
dmah42: we don't support exceptions.

matoro: I just tested, and returning -1 results in the same issues as returning 0.  Here is an example of what it looks like with -1:

```
Running /var/tmp/portage/dev-cpp/benchmark-1.8.3/work/benchmark-1.8.3_build/test/benchmark_test
Run on (-1 X 1987.32 MHz CPU )
Load Average: 1.09, 1.04, 0.90
terminate called after throwing an instance of 'std::length_error'
  what():  cannot create std::vector larger than max_size()
<end of output>
Test time =   0.05 sec
Test Failed.
```

I see that it's used immediately in `CPUInfo` constructor without error checking.  I'll have to experiment and see what is a reasonable way to propagate such an error up.

src/cycleclock.h
LebedevRI: s/per-iteration/per-repetition/, which is much worse

src/sysinfo.cc
LebedevRI: FWIW, personally, i just wouldn't bother.

matoro: Then how about `std::abort()` (or, I guess the abort handler stuff)?  It looks like there's just no good way for callers to handle errors here and there are no circumstances under which you can meaningfully work with zero CPUs.

dmah42: yeah we could.. the `CHECK` macros we have get compiled out in release builds but maybe we should have one that doesn't get compiled out (or just `CallAbortHandler` directly).

src/sysinfo.cc
```diff
@@ -510,10 +510,13 @@ int GetNumCPUs() {
   int max_id = -1;
   std::ifstream f("/proc/cpuinfo");
   if (!f.is_open()) {
-    std::cerr << "failed to open /proc/cpuinfo\n";
-    return -1;
+    PrintErrorAndDie("Failed to open /proc/cpuinfo");
```
LebedevRI, line 513: White there, could you please cleanup rest of this function, Ctrl+F for `stderr`?

LebedevRI:
(generally looks fine to me, thank you)

src/sysinfo.cc
```diff
@@ -478,25 +478,27 @@ int GetNumCPUs() {
 #ifdef BENCHMARK_HAS_SYSCTL
   int num_cpu = -1;
   if (GetSysctl("hw.ncpu", &num_cpu)) return num_cpu;
-  fprintf(stderr, "Err: %s\n", strerror(errno));
-  std::exit(EXIT_FAILURE);
+  PrintErrorAndDie("Err: ", strerror(errno));
```
LebedevRI, line 481: This becoming too messy :)
Please split this PR into two:
1. The actual changes to support the architecture
2. All of the PrintErrorAndDie stuff

matoro: Sorry, a little confused, isn't that what you asked for above?

LebedevRI: I guess i thought it would look nicer?
Concretely, i'm getting stuck on how asymmetrical this all is,
sometimes there is a check for ncpu=0, sometimes there is not.

My next comment would be to extract the `if ncpu<1 abort` logic
into a wrapper function.

I just don't think it makes sense to do all that cleanup in this PR.

matoro: OK, will dropping the last commit be sufficient for this PR then?

LebedevRI: "sysinfo: support parsing /proc/cpuinfo on Alpha" also has some `PrintErrorAndDie` stuff.

matoro: IMO that is immediately and directly relevant to this PR.  I spent a lot of time investigating and trying to understand why the tests were failing, looking for issues with the cycle timer, when it just turned out that failing to parse `/proc/cpuinfo` silently returns garbage data that results in all sorts of cryptic errors down the line later.  I believe handling this in a sensible way is super important in order to make porting to this and other platforms a reasonable process.

LebedevRI: Hey, i'm only talking about how to get this diff to the state where i'd be happy to merge it,
and splitting it into two separate chunks will do that.
We spent more time arguing about that than it would have taken to get both of those PR's merged.

LebedevRI (approved):
Ship it!

LebedevRI:
@thesamesam thank you!

thesamesam:
Thank you all!

PR #1756, opened 2024-02-14
sysinfo.cc: Always abort on GetNumCPUs failure

thesamesam:
Defines a wrapper function, CheckNumCPUs, which enforces that GetNumCPUs never returns fewer than one CPU.  There is no reasonable way to continue if we are unable to identify the number of CPUs.

This is a followup to https://github.com/google/benchmark/pull/1753.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -474,12 +474,11 @@ std::string GetSystemName() {
 #endif  // Catch-all POSIX block.
 }
 
-int GetNumCPUs() {
+int GetNumCPUsImpl() {
 #ifdef BENCHMARK_HAS_SYSCTL
   int num_cpu = -1;
   if (GetSysctl("hw.ncpu", &num_cpu)) return num_cpu;
-  fprintf(stderr, "Err: %s\n", strerror(errno));
-  std::exit(EXIT_FAILURE);
+  PrintErrorAndDie("Err: ", strerror(errno));
 #elif defined(BENCHMARK_OS_WINDOWS)
   SYSTEM_INFO sysinfo;
   // Use memset as opposed to = {} to avoid GCC missing initializer false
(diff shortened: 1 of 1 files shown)


src/sysinfo.cc
```diff
@@ -474,44 +474,51 @@ std::string GetSystemName() {
...
+int CheckNumCPUs(const int num_cpus) {
+  if (num_cpus < 1) {
+    PrintErrorAndDie(
+        "Unable to extract number of CPUs.  If your platform uses "
+        "/proc/cpuinfo, custom support may need to be added.");
+  }
+  return num_cpus;
+}
+
 int GetNumCPUs() {
```
LebedevRI, line 486:
```
int GetNumCPUs() {
  int num_cpus = GetNumCPUsImpl();
  if (num_cpus < 1) {
    PrintErrorAndDie(
        "Unable to extract number of CPUs.  If your platform uses "
        "/proc/cpuinfo, custom support may need to be added.");
  }
  return num_cpus;
}

int GetNumCPUsImpl() {
```

LebedevRI: (And place `GetNumCPUsImpl()` before `GetNumCPUs()` i guess)

src/sysinfo.cc
```diff
@@ -474,12 +474,11 @@ std::string GetSystemName() {
...
 }
 
-int GetNumCPUs() {
+int GetNumCPUsImpl() {
 #ifdef BENCHMARK_HAS_SYSCTL
   int num_cpu = -1;
-  if (GetSysctl("hw.ncpu", &num_cpu)) return num_cpu;
-  fprintf(stderr, "Err: %s\n", strerror(errno));
-  std::exit(EXIT_FAILURE);
+  if (GetSysctl("hw.ncpu", &num_cpu)) return CheckNumCPUs(num_cpu);
```
LebedevRI, line 480: Stale?

thesamesam: Sorry!

LebedevRI (approved):
Cool, thank you!

thesamesam:
Many thanks!

Issue #272, opened 2016-08-09
complexity_test.cc spuriously fails 


EricWF:
The benchmark registered by `BENCHMARK(BM_Complexity_O1) -> Range(1, 1<<18) -> Complexity();` sometimes reports it's complexity as `lgN` as opposed to `([0-9]+)`. This happens when one of the later repetitions happens to run faster due to CPU load.

For example:
```
BM_Complexity_O1/1                 2106 ns       3125 ns      10000
BM_Complexity_O1/8                 2132 ns       1563 ns      10000
BM_Complexity_O1/64                2050 ns       3125 ns      10000
BM_Complexity_O1/512               2032 ns       1563 ns      10000
BM_Complexity_O1/4k                2041 ns       1563 ns      10000
BM_Complexity_O1/32k               2264 ns      15625 ns       1000
BM_Complexity_O1/256k              2128 ns       1563 ns      10000
BM_Complexity_O1_BigO            163.31 lgN     389.19 lgN 
```

I'm not quite sure how to fix this test, or even if this is a bug in the complexity implementation. I would like some guidance on how to tackle this.

This is making our Appveyor build frequently fail so I would like to fix it. For now I'm going to checkin a temporary fix which simply accepts 'lg(N)' as valid test output.

dominichamon:
i think this is what we've been seeing recently on the appveyor windows release tests. it does seem though that the github workflows are more stable.

PR #1757, opened 2024-02-14
Rewrite complexity_test to use (hardcoded) manual time

LebedevRI:
This test is fundamentally flaky, because it tried to read tea leafs, and is inherently misbehaving in CI environments,
since there are unmitigated sources of noise.

Fixes https://github.com/google/benchmark/issues/272
Fixes https://github.com/google/benchmark/issues/1758

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -96,6 +96,7 @@ BenchmarkReporter::Run CreateRunReport(
     } else {
       report.real_accumulated_time = results.real_time_used;
     }
+    report.use_real_time_for_initial_big_o = b.use_manual_time();
     report.cpu_accumulated_time = results.cpu_time_used;
     report.complexity_n = results.complexity_n;
     report.complexity = b.complexity();
--- a/src/complexity.cc
+++ b/src/complexity.cc
@@ -186,8 +186,19 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_cpu = MinimalLeastSq(n, cpu_time, reports[0].complexity_lambda);
     result_real = MinimalLeastSq(n, real_time, reports[0].complexity_lambda);
   } else {
-    result_cpu = MinimalLeastSq(n, cpu_time, reports[0].complexity);
-    result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
+    const BigO* InitialBigO = &reports[0].complexity;
+    const bool use_real_time_for_initial_big_o =
+        reports[0].use_real_time_for_initial_big_o;
+    if (use_real_time_for_initial_big_o) {
+      result_real = MinimalLeastSq(n, real_time, *InitialBigO);
+      InitialBigO = &result_real.complexity;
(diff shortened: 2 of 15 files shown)


dmah42:
what happened? this seemed like a good idea...

LebedevRI:
I don't understand what the heck is going on with automatic complexity fitting,
all windows CI failed here.

LebedevRI:
OOOOH this is a fun one!

LebedevRI:
@dmah42 ok, this is more like it.

src/benchmark_runner.cc
```diff
@@ -96,6 +96,7 @@ BenchmarkReporter::Run CreateRunReport(
     } else {
       report.real_accumulated_time = results.real_time_used;
     }
+    report.use_real_time_for_initial_big_o = b.use_manual_time();
```
dmah42, line 99: to check this logic:

if we're using manual time, we'll report manual time as "real" but then use "real" time for bigO.
if we're not using manual time, then we'll report real time as "real" but use "cpu" time for bigO.

maybe this could be a comment somewhere just to track the interaction between manual, real, and cpu times for bigO.

src/complexity.cc
```diff
@@ -186,8 +186,19 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
     result_cpu = MinimalLeastSq(n, cpu_time, reports[0].complexity_lambda);
     result_real = MinimalLeastSq(n, real_time, reports[0].complexity_lambda);
   } else {
-    result_cpu = MinimalLeastSq(n, cpu_time, reports[0].complexity);
-    result_real = MinimalLeastSq(n, real_time, result_cpu.complexity);
+    const BigO* InitialBigO = &reports[0].complexity;
```
dmah42, line 189: this section needs some comments.  maybe there's a way to simplify the logic?

test/complexity_test.cc
```diff
@@ -69,35 +69,44 @@ int AddComplexityTest(const std::string &test_name,
...
+    // always 1ns per iteration
+    state.SetIterationTime(42 * 1e-9);
   }
   state.SetComplexityN(state.range(0));
 }
-BENCHMARK(BM_Complexity_O1)->Range(1, 1 << 18)->Complexity(benchmark::o1);
-BENCHMARK(BM_Complexity_O1)->Range(1, 1 << 18)->Complexity();
 BENCHMARK(BM_Complexity_O1)
     ->Range(1, 1 << 18)
+    ->UseManualTime()
```
dmah42, line 89: does this mean complexity only works for manual time?  do we have any tests where we use complexity with real time?

LebedevRI:
@dmah42 we always compute complexity for both cpu and real time,
this only changes which of these two is used to auto-detect the Big-O.

dmah42 (approved):

LebedevRI:
@dmah42 thank you. Looks like there's some residual flakiness in CI left, so i'll see if these nits can be revisited later...

PR #1759, opened 2024-02-25
Update README.md

14jdelap:
Remove the `$` at the start of the bash command to run the tests so that it can be copy-pasted by a user without any errors

This change is consistent with the other 1-line bash snippets in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ On a unix system, the build directory should now look something like this:
 Next, you can run the tests to check the build.
 
 ```bash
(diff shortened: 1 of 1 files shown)


LebedevRI:
I'm not sure what we should be doing, but this is not consistent, there's only one snippet that does not use `$`'s.

dmah42:
i agree with your assessment that we should get rid of them.  they're a holdover from pre-markdown/github notes for the most part where differentiating a "command" was more important.

PR #1761, opened 2024-03-05
Fixed LTO issue on no discard variable

tmiguelf:
Fixed LTO issue on no discard variable.
Tested on MSVC

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -156,7 +156,9 @@ static void const volatile* volatile global_force_escape_pointer;
 
 // FIXME: Verify if LTO still messes this up?
 void UseCharPointer(char const volatile* const v) {
-  //forces compiler to issue instruction by assigning dummy value to a volatile preventing LTO
+  // We want to escape the pointer `v` so that the compiler can not eliminate computations
+  // that produced it. To do that, we escape the pointer by storing it into a volatile variable,
+  // since generally, volatile store, is not something the compiler is allowed to elide.
   global_force_escape_pointer = reinterpret_cast<void const volatile*>(v);
 }
 
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -152,11 +152,12 @@ BENCHMARK_EXPORT std::map<std::string, std::string>*& GetGlobalContext() {
   return global_context;
 }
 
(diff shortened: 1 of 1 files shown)


src/benchmark.cc
```diff
@@ -152,8 +152,12 @@ BENCHMARK_EXPORT std::map<std::string, std::string>*& GetGlobalContext() {
...
 }
 
-// FIXME: wouldn't LTO mess this up?
-void UseCharPointer(char const volatile*) {}
+void const volatile* volatile g_LTO_dump;
+
+// FIXME: Verify if LTO still messes this up?
+void UseCharPointer(char const volatile* const v) {
+  g_LTO_dump = reinterpret_cast<void const volatile*>(v);
+}
```
LebedevRI, line 160: This could use some comments when someone encounters it later on.

tmiguelf: done

src/benchmark.cc
```diff
@@ -152,8 +152,13 @@ BENCHMARK_EXPORT std::map<std::string, std::string>*& GetGlobalContext() {
   return global_context;
 }
 
-// FIXME: wouldn't LTO mess this up?
-void UseCharPointer(char const volatile*) {}
+static void const volatile* volatile global_force_escape_pointer;
+
+// FIXME: Verify if LTO still messes this up?
+void UseCharPointer(char const volatile* const v) {
+  //forces compiler to issue instruction by assigning dummy value to a volatile preventing LTO
```
LebedevRI, line 159:
```
// We want to escape the pointer `v` so that the compiler can not eliminate computations
// that produced it. To do that, we escape the pointer by storing it into a volatile variable,
// since generally, volatile store, is not something the compiler is allowed to elide.
```

LebedevRI (approved):
I *think* this works, at least better than the original: https://godbolt.org/z/v1493hd81

LebedevRI:
I'll merge after #1761 unbreaks the CI...

tmiguelf:
Yes, I was having an issue where my code was optimized away and the benchmark::DoNotOptimize failed to prevent optimization. With this it works for my case, now my statistics are more accurate.

LebedevRI:
@tmiguelf thank you!

PR #1762, opened 2024-03-05
Revert "fix some warnings"

LebedevRI:
This reverts commit 1576991177ba97a4b2ff6c45950f1fa6e9aa678c.


dmah42:
but why?  i had local compiler warnings due to potential size differences.

LebedevRI:
> but why? i had local compiler warnings due to potential size differences.

Because the CI was green before that change, and went red after that change.
I think branch protection rules are wrong, they either prevent everyone from bypassing them, or nobody.

dmah42 (approved):

LebedevRI:
@dmah42 thank you!
I'm sure the whatever diagnostics those changes were trying to address
can be addressed without reintroducing that breakage.

PR #1763, opened 2024-03-06
mitigate clang build warnings -Wconversion

dmah42:


LebedevRI:
Yup, implicit conversions is perhaps one of the worst C/C++ "features".
Let's also enable it in cmake/bazel so it's not reintroduced?

include/benchmark/benchmark.h
```diff
@@ -1304,7 +1304,7 @@ class BENCHMARK_EXPORT Benchmark {
  public:
   const char* GetName() const;
   int ArgsCnt() const;
-  const char* GetArgName(int arg) const;
+  const char* GetArgName(size_t arg) const;
```
LebedevRI, line 1307: I'm not sure about this one, we return signed and then take unsigned.
I think things should generally be signed, but even then, this seems unconsistent

LebedevRI (commented):
a comment

include/benchmark/benchmark.h
dmah42: yeah i didn't feel great about that either.  i'll tweak it.

src/benchmark_register.h
```diff
@@ -24,7 +24,7 @@ typename std::vector<T>::iterator AddPowers(std::vector<T>* dst, T lo, T hi,
   static const T kmax = std::numeric_limits<T>::max();
 
   // Space out the values in multiples of "mult"
```
LebedevRI, line 26: I'm somewhat surprized that this is needed, but okay, i guess?

LebedevRI (approved):
Seems fine to me, thank you!

PR #1765, opened 2024-03-09
Fix implicit conversion changes signess warning in perf_counters.cc

ivafanas:
Newly added -Wconversion command line option make this warning to stop compilation with error, because warnings are treated as errors.

Issue appeared on clang-16.


LebedevRI (approved):
Makes sense to me, thank you!

ivafanas:
Thank you!

PR #1769, opened 2024-03-20
allow BENCHMARK_VERSION to be undefined

### LINKED ISSUE #1768: [FR] BENCHMARK_VERSION is not set by default
Since a recent addition adding usage of "BENCHMARK_VERSION", this project is now coupled to the official build tools and is not compatible with anything else.

Default macro variables so we can ignore if they're not set

Setting BENCHMARK_VERSION during compilation via `-DBENCHMARK_VERSION=version` doesn't work as the macro is not converted to a string, see below.

**Additional context**
In my experience, converting define inputs to strings is better done via a helper macro like

```
#define STRINGIFY(x) #x
#define TOSTRING(x) STRINGIFY(x)
```

such that inputs like -DBENCHMARK_VERSION=version

can be used like

```
std::string s{TOSTRING(BENCHMARK_VERSION)};
```

This IMO is simpler from a build perspective to not need any fancy escaping

see: https://coliru.stacked-crooked.com/a/ae9e4273382fe470

PhilipDeegan:
closes #1768 

motivation is to allow simpler override at compilation time
shouldn't impact anyone using the official build tools as it's all internal in that case

I haven't tested the bazel build output for the function `benchmark::GetBenchmarkVersion()` but I'm guessing it's without the extra quote from "stringification" given the modifications

I have tested the cmake build output for the funciton, and the string is without quotes with the proposed changes

PRed on my fork for testing [here](https://github.com/PhilipDeegan/benchmark/pull/1)
I can't explain the sanitizer failures so I'm figuring they are existing

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -757,7 +757,13 @@ int InitializeStreams() {
 
 }  // end namespace internal
 
-std::string GetBenchmarkVersion() { return {BENCHMARK_VERSION}; }
+std::string GetBenchmarkVersion() {
+#ifdef BENCHMARK_VERSION
+  return {BENCHMARK_VERSION};
+#else
+  return {""};
+#endif
+}
 
 void PrintDefaultHelp() {
   fprintf(stdout,


LebedevRI:
I feel this is a roundabout way to achieve the the end result.
How about we just explicitly do
```
std::string GetBenchmarkVersion() {
#ifdef BENCHMARK_VERSION
return {BENCHMARK_VERSION};
#else
return {""};
#endif
}
```
?

LebedevRI:
Hm, and if `BENCHMARK_VERSION` is not defined, `GetBenchmarkVersion()` now returns `"BENCHMARK_VERSION"`:
https://godbolt.org/z/d9xKEbc7o
Is that the expected outcome?

PhilipDeegan:
> Hm, and if `BENCHMARK_VERSION` is not defined, `GetBenchmarkVersion()` now returns `"BENCHMARK_VERSION"`: https://godbolt.org/z/d9xKEbc7o Is that the expected outcome?

I'm kinda surprised it even compiles if it's not defined, but perhaps `BENCHMARK_VERSION` is more informative than an empty string

PhilipDeegan:
> How about we just explicitly do ...

I can do this, but I still think the escapes can be avoided via stringification as @dmah42 suggested

LebedevRI:
> > Hm, and if `BENCHMARK_VERSION` is not defined, `GetBenchmarkVersion()` now returns `"BENCHMARK_VERSION"`: https://godbolt.org/z/d9xKEbc7o Is that the expected outcome?
> 
> I'm kinda surprised it even compiles if it's not defined, but perhaps `BENCHMARK_VERSION` is more informative than an empty string

That's because it's passed into the stringification macro. Normally it indeed fails: https://godbolt.org/z/j5x4a4sWG

PhilipDeegan:
updated to just 

```
std::string GetBenchmarkVersion() {
#ifdef BENCHMARK_VERSION
  return {BENCHMARK_VERSION};
#else
  return {""};
#endif
}
```

dmah42:
oh, that's a shame.  i quite liked the reduced complexity in the bazel config.  i was thinking we'd land somewhere in between where be both stringify it if it's set but return a safe value if it's not.  anyway, this also looks fine.

dmah42 (approved):

LebedevRI:
@PhilipDeegan thank you!

Issue #1776, opened 2024-04-02
[FR] Remove COMPILER_IBMXL macro for z/OS


fanbo-meng:
**Is your feature request related to a problem? Please describe.**
COMPILER_IBMXL is no longer needed to distinguish compilers on z/OS.

**Describe the solution you'd like**
Remove the macro definition and usage.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

LebedevRI:
> **Is your feature request related to a problem? Please describe.**
> COMPILER_IBMXL is no longer needed to distinguish compilers on z/OS.

Why?

fanbo-meng:
I added COMPILER_IBMXL to identify the Clang based IBM XL compiler (xlclang) on z/OS. This compiler is obsolete and replaced by the Open XL compiler, so the macro is no longer needed and the existing code would lead to incorrect asm syntax for Open XL. Please review https://github.com/google/benchmark/pull/1777

LebedevRI:
That wording needs to be in the commit message of that PR

fanbo-meng:
closed by https://github.com/google/benchmark/pull/1777

PR #1777, opened 2024-04-02
Remove COMPILER_IBMXL macro for z/OS

fanbo-meng:
This patch removes definition and usage of COMPILER_IBMXL macro, resolves issue https://github.com/google/benchmark/issues/1776


LebedevRI:
I'm guessing that ecosystem/hardware is so special we (well, you) get to dictate that said obsolete compiler is simply completely unsupported going forward, and thus such change is fine?

dmah42 (approved):

dmah42:
thanks. and thanks for the commit description :)

fanbo-meng:
> I'm guessing that ecosystem/hardware is so special we (well, you) get to dictate that said obsolete compiler is simply completely unsupported going forward, and thus such change is fine?

Yes I'm from the dev team of z/OS C/C++ compiler. That macro isn't used anywhere else other than the one instance in cycleclock.h, and after this change the code would still compile the same way (for that compiler). Thus IMHO this change is fine.

LebedevRI:
Yup, thank you, i just wanted that to be explicitly stated.
Thank you!

Issue #1781, opened 2024-04-11
[BUG]


dhairyarungta:
Unable to build on Mac M1 OSX Sonoma, using gcc version 13.2.0 (Homebrew GCC 13.2.0).

benchmark/src/sysinfo.cc:739:39: warning: conversion from 'long long unsigned int' to 'double' may change value [-Wconversion] 739 | if (GetSysctl(freqStr, &hz)) return hz;

cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ../
The above command ran successfully. Result of run in file1.txt

cmake --build "build" --config Release --parallel 4
produces this warning which terminates the build
benchmark/src/sysinfo.cc:739:39: warning: conversion from 'long long unsigned int' to 'double' may change value [-Wconversion] 739 | if (GetSysctl(freqStr, &hz)) return hz;
result of run in file2.txt and CMakeCache.txt in CMakeCache.txt

I edited CMakeCache.txt, and set Werror to OFF.
//Build Release candidates with -Werror.
BENCHMARK_ENABLE_WERROR:BOOL=OFF

[file2.txt](https://github.com/google/benchmark/files/14945408/file2.txt)

[CMakeCache.txt](https://github.com/google/benchmark/files/14945396/CMakeCache.txt)
[file1.txt](https://github.com/google/benchmark/files/14945397/file1.txt)

The library got built with the warning and no other warnings, but I cannot figure out what was the real problem. I was able to successfully build the library with clang++, clang without any warnings or errors.

LebedevRI:
Please feel free to submit a patch for https://github.com/google/benchmark/blob/f3ec7b8820ca8136c4e1dad4552608b51b47831a/src/sysinfo.cc#L739
```
-  if (GetSysctl(freqStr, &hz)) return hz;
+  if (GetSysctl(freqStr, &hz)) return static_cast<double>(hz);
```

dhairyarungta:
Pull request raised.

PR #1782, opened 2024-04-12
Refactor: Return frequency as double

dhairyarungta:
Adjusted the GetSysctl call in sysinfo.cc to ensure the frequency value is returned as a double rather than an integer.

Fixes https://github.com/google/benchmark/issues/1781.


LebedevRI:
Looks right, but you'll need to deal with that CLA issue i'm afraid.

LebedevRI (approved):
Seems good to me, thank you!

PR #1783, opened 2024-04-15
Modernize wheel building job config

nicholasjng:
It is now possible to build Mac wheels on native machines in Github Actions, so ARM64 Mac wheels are now built and tested on M1 machines.

Also, the artifact up-/download was migrated to v4, which made it necessary to upload wheels to unique artifact names, and then later stitch them together again in a subsequent job.

The cross-platform Mac build injection in setup.py was removed, since it is no longer necessary.

I relanded a monkey-patching of Bazel build files, this time for MODULE.bazel. This is because `rules_python` does not allow running as the root user, which is the case in cibuildwheel+Linux (happens in a Docker container). Since I did not see a quick way of switching to rootless containers, and did not want to hardcode the config change (it can apparently cause cache misses and build failures), I inject the "ignore_root_user_error" flag into the MODULE.bazel file when running in cibuildwheel on Linux.

-------------

That last paragraph wrecked my day, but that's apparently a design choice by rules_python. I verified that the wheels build, see https://github.com/nicholasjng/benchmark/actions/runs/8691995492 (the failure is only because I screwed up the musllinux exclude flag, which I reverted back to how it was.)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -15,16 +15,16 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Install Python 3.11
+      - name: Install Python 3.12
         uses: actions/setup-python@v5
         with:
-          python-version: 3.11
+          python-version: 3.12
       - run: python -m pip install build
       - name: Build sdist
         run: python -m build --sdist
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: dist-sdist
           path: dist/*.tar.gz
 
   build_wheels:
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels.yml
```diff
@@ -15,24 +15,24 @@ jobs:
...
+          name: dist-sdist
           path: dist/*.tar.gz
 
   build_wheels:
     name: Build Google Benchmark wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-13, macos-14, windows-latest]
```
dmah42, line 35: why both macoses?

nicholasjng: MacOS 13 builds Intel x86, MacOS 14 builds ARM64.

https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

(i.e., both build their host architectures, which is x86_64 on macos-13 and arm64 on macos-14.)

dmah42: ah, nothing like obscuring architecture behind OS versions.

dmah42 (approved):

nicholasjng:
Thanks!

PR #1785, opened 2024-05-14
Improve compatibility with Hexagon hardware

steven-johnson:
The customization done via BENCHMARK_OS_QURT works just fine with the Hexagon simulator, but on at least some Hexagon hardware, both `qurt_timer_get_ticks()` and `std::chrono::now()` are broken and always return 0. This fixes the former by using the better-supported (and essentially identical `qurt_sysclock_get_hw_ticks()` call, and the latter by reading a 19.2MHz hardware counter (per suggestion from Qualcomm). Local testing seems to indicate these changes are just as robust under the simulator as before.

### DIFF
--- a/src/timers.cc
+++ b/src/timers.cc
@@ -127,7 +127,7 @@ double ProcessCPUUsage() {
   DiagnoseAndExit("GetProccessTimes() failed");
 #elif defined(BENCHMARK_OS_QURT)
   return static_cast<double>(
-             qurt_timer_timetick_to_us(qurt_timer_get_ticks())) *
+             qurt_timer_timetick_to_us(qurt_sysclock_get_hw_ticks())) *
          1.0e-6;
 #elif defined(BENCHMARK_OS_EMSCRIPTEN)
   // clock_gettime(CLOCK_PROCESS_CPUTIME_ID, ...) returns 0 on Emscripten.
@@ -161,7 +161,7 @@ double ThreadCPUUsage() {
   return MakeTime(kernel_time, user_time);
 #elif defined(BENCHMARK_OS_QURT)
   return static_cast<double>(
-             qurt_timer_timetick_to_us(qurt_timer_get_ticks())) *
+             qurt_timer_timetick_to_us(qurt_sysclock_get_hw_ticks())) *
          1.0e-6;
 #elif defined(BENCHMARK_OS_MACOSX)
   // FIXME We want to use clock_gettime, but its not available in MacOS 10.11.
--- a/src/timers.cc
+++ b/src/timers.cc
@@ -126,6 +126,10 @@ double ProcessCPUUsage() {
     return MakeTime(kernel_time, user_time);
   DiagnoseAndExit("GetProccessTimes() failed");
 #elif defined(BENCHMARK_OS_QURT)
+  // Note that qurt_timer_get_ticks() is no longer documented as of SDK 5.3.0,
+  // and doesn't appear to work on at least some devices (eg Samsung S22),
+  // so let's use the actually-documented and apparently-equivalent
+  // qurt_sysclock_get_hw_ticks() call instead.
   return static_cast<double>(
              qurt_timer_timetick_to_us(qurt_sysclock_get_hw_ticks())) *
          1.0e-6;
@@ -160,6 +164,10 @@ double ThreadCPUUsage() {
                  &user_time);
   return MakeTime(kernel_time, user_time);
 #elif defined(BENCHMARK_OS_QURT)
+  // Note that qurt_timer_get_ticks() is no longer documented as of SDK 5.3.0,
+  // and doesn't appear to work on at least some devices (eg Samsung S22),
+  // so let's use the actually-documented and apparently-equivalent
+  // qurt_sysclock_get_hw_ticks() call instead.
   return static_cast<double>(
              qurt_timer_timetick_to_us(qurt_sysclock_get_hw_ticks())) *
          1.0e-6;
--- a/src/timers.h
+++ b/src/timers.h
@@ -23,17 +23,17 @@ double ThreadCPUUsage();
 // return an absolute time, but since ChronoClockNow() is only used
 // to compute elapsed time, this shouldn't matter.
 struct QuRTClock {
-    typedef uint64_t                           rep;
-    typedef std::ratio<1,19200000>             period;
-    typedef std::chrono::duration<rep, period> duration;
-    typedef std::chrono::time_point<QuRTClock> time_point;
-    static const bool is_steady =              false;
+  typedef uint64_t rep;
+  typedef std::ratio<1,19200000> period;
+  typedef std::chrono::duration<rep, period> duration;
+  typedef std::chrono::time_point<QuRTClock> time_point;
+  static const bool is_steady = false;
 
-    static time_point now() {
-      unsigned long long count;
-      asm volatile(" %0 = c31:30 " : "=r"(count));
-      return time_point(static_cast<duration>(count));
-    }
+  static time_point now() {
+    unsigned long long count;
+    asm volatile(" %0 = c31:30 " : "=r"(count));
+    return time_point(static_cast<duration>(count));
+  }
 };
 
 #else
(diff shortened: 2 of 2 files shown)


LebedevRI:
Can you please explain (in the commit message too)
why this `ChronoClockNow` workaround is being introduced
instead of fixing the underlying bug?

src/timers.h
```diff
@@ -41,6 +56,8 @@ inline double ChronoClockNow() {
   return FpSeconds(ClockType::now().time_since_epoch()).count();
 }
 
+#endif
```
LebedevRI, line 59: In `ChooseClockType`, do we end up with `std::chrono::high_resolution_clock`,
or `std::chrono::steady_clock`? Does either of them work, or neither works?

steven-johnson: We end up with `high_resolution_clock`, but neither of them works.

src/timers.h
```diff
@@ -15,6 +15,21 @@ double ChildrenCPUUsage();
...
 double ThreadCPUUsage();
 
+#if defined(BENCHMARK_OS_QURT)
+
+// std::chrono::now() can return 0 on some Hexagon devices;
+// this reads the value of a 56-bit, 19.2MHz hardware counter
+// and converts it to seconds. Unlike std::chrono, this doesn't
+// return an absolute time, but since ChronoClockNow() is only used
+// to compute elapsed time, this shouldn't matter.
+inline double ChronoClockNow() {
```
LebedevRI, line 25: Could you please wrap this into a clock-like class,
and change `ChooseClockType` to pick it on `BENCHMARK_OS_QURT`?

LebedevRI:
> > Can you please explain (in the commit message too) why this `ChronoClockNow` workaround is being introduced instead of fixing the underlying bug?
> 
> It is not in my power to fix the underlying bug, which (presumably) lies inside the Qualcomm SDK. I'd much prefer it if the implementation of std::chrono that Qualcomm provided actually works.

Makes sense. Please link the upstream bugreport in commit message.

src/timers.h
steven-johnson: That seems a lot more complex than actually needed, but OK.

steven-johnson:
> > > Can you please explain (in the commit message too) why this `ChronoClockNow` workaround is being introduced instead of fixing the underlying bug?
> > 
> > 
> > It is not in my power to fix the underlying bug, which (presumably) lies inside the Qualcomm SDK. I'd much prefer it if the implementation of std::chrono that Qualcomm provided actually works.
> 
> Makes sense. Please link the upstream bugreport in commit message.

There is no upstream bug report, because I haven't yet gotten Qualcomm to admit that this is in fact a bug -- I'm trying to find the correct contacts at Qualcomm.

src/timers.h
steven-johnson: Actually, I'm gonna push back on this a bit: the only reason that the `ChooseClockType` type exists is so that we can call `now()` on it in `ChronoClockNow()`, it's literally the only place in the code base that we use it. Trying to create a class that mimics the existence of `std::chrono` is pointless overhead. If I'm missing something subtle, please let me know, but this seems like a useless exercise.

LebedevRI: https://godbolt.org/z/T84efY5fG

steven-johnson: I'm not disputing what the object code will look like; I'm disputing that constructing a class would be wrong, because `std::chrono` implies a valid date/time, not just a relative time counter, and we can't mimic that with this approach; it means we have to mimic an API designed to do all sorts of arbitrary things with timestamps, rather than one that simply provides a seconds-counter in the form of a double.

LebedevRI: [`std::chrono::steady_clock`](https://en.cppreference.com/w/cpp/chrono/steady_clock):
> This clock is not related to wall clock time (for example, it can be time since last reboot), and is most suitable for measuring intervals.

steven-johnson:
PTAL

src/timers.h
```diff
@@ -41,6 +68,7 @@ inline double ChronoClockNow() {
   return FpSeconds(ClockType::now().time_since_epoch()).count();
 }
 
+
```
LebedevRI, line 71: Nit: extra newline

src/timers.h
```diff
@@ -15,6 +15,29 @@ double ChildrenCPUUsage();
...
+// to compute elapsed time, this shouldn't matter.
+struct QuRTClock {
+    typedef uint64_t                           rep;
+    typedef std::ratio<1,19200000>             period;
+    typedef std::chrono::duration<rep, period> duration;
+    typedef std::chrono::time_point<QuRTClock> time_point;
+    static const bool is_steady =              false;
+
+    static time_point now() {
+      unsigned long long count;
```
LebedevRI, line 33: (You are sure this shouldn't be `rep`?)

LebedevRI:
Cool, looks good to me. It would be good to get some kind of
an acknowledgement from Qualcomm re the bug,
i'm not sure if we should wait for that or not?

LebedevRI:
And one more thought: in the cases where the normal clocks work, what's their frequency, is it better than of this clock?
I.e., should we be always using this clock, or should we detect the bug, and only then use this "fallback" clock?

steven-johnson:
> Cool, looks good to me. It would be good to get some kind of an acknowledgement from Qualcomm re the bug, i'm not sure if we should wait for that or not?

IMHO we should wait for comment

> And one more thought: in the cases where the normal clocks work, what's their frequency, is it better than of this clock? I.e., should we be always using this clock, or should we detect the bug, and only then use this "fallback" clock?

The normal clock frequencies are implementation-defined. As to whether we should detect and fallback, that would add a bit of extra overhead where we probably want to minimize overhead, but I'll defer to QC folk on that.

LebedevRI:
> > Cool, looks good to me. It would be good to get some kind of an acknowledgement from Qualcomm re the bug, i'm not sure if we should wait for that or not?
> 
> IMHO we should wait for comment

Let's wait.

> > And one more thought: in the cases where the normal clocks work, what's their frequency, is it better than of this clock? I.e., should we be always using this clock, or should we detect the bug, and only then use this "fallback" clock?
> 
> The normal clock frequencies are implementation-defined. As to whether we should detect and fallback, that would add a bit of extra overhead where we probably want to minimize overhead, but I'll defer to QC folk on that.

That would need to happen at CMake time, true, but then i'm not sure if it can truly be detected then,
because i suspect it might not happen when the compiled code is run not on the actual hardware,
but emulated on the machine on which it is being compiled.

LebedevRI (dismissed):
LGTM, but let's wait.

steven-johnson:
Update: It's taking much longer than I expected to get any kind of feedback from Qualcomm on this, and my team is getting blocked by this. Can we go ahead and land this, with the understanding that it may be rolled back and/or fixed forward when we eventually do get feedback from QC?

src/timers.cc
```diff
@@ -127,7 +127,7 @@ double ProcessCPUUsage() {
   DiagnoseAndExit("GetProccessTimes() failed");
 #elif defined(BENCHMARK_OS_QURT)
   return static_cast<double>(
-             qurt_timer_timetick_to_us(qurt_timer_get_ticks())) *
+             qurt_timer_timetick_to_us(qurt_sysclock_get_hw_ticks())) *
```
LebedevRI, line 130: Can you please add in-code comment why this uses `qurt_sysclock_get_hw_ticks` instead of `qurt_timer_get_ticks`?

steven-johnson: Done

LebedevRI (approved):

aankit-quic:
@steven-johnson @LebedevRI  I'm really sorry. I still haven't gotten any response from the team who could answer these concerns. 

The PR looks good to me. The qurt APIs essentially do the same thing as what Steven is doing here. I feel the std::chrono should have worked in the first place and is probably a bug, but the responsible team still have not confirmed anything.

LebedevRI:
> Update: It's taking much longer than I expected to get any kind of feedback from Qualcomm on this, and my team is getting blocked by this.

FWIW, i've very much expected that it would take a month+ before anything happens here.
Sure, let's merge this.

LebedevRI:
@steven-johnson @aankit-quic thank you!

PR #1799, opened 2024-06-07
bump standard to C++14

dmah42:
per the dependencies doc, we base our supported standards of the defaults available in compilers or a 10 year expiry from standard release.  we just hit 10 years since C++14 release so that's our new base supported standard.

also updated the docs to align on this.


LebedevRI:
We need to be actually enforcing them via a cmake-time checks though.

dmah42:
> We need to be actually enforcing them via a cmake-time checks though.

given we now officially support C++14 for building as a minimum, maybe that's viable.  it's hard to enumerate every version of every compiler, which is why the linked support doc focuses on standards and build systems.

LebedevRI (approved):
This brings us closer to reality, so looks nice to me.
Thank you!

PR #1802, opened 2024-06-11
cycleclock: Fix type conversion to match function return type on riscv64

kraj:
Fixes builds with clang

src/cycleclock.h:213:10: error: implicit conversion changes signedness: 'uint64_t' (aka 'unsigned long') to 'int64_t' (aka 'long') [-Werror,-Wsign-conversion]
     213 |   return cycles;
         |   ~~~~~~ ^~~~~~
   1 error generated.

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -210,7 +210,7 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
 #else
   uint64_t cycles;
   asm volatile("rdtime %0" : "=r"(cycles));
-  return cycles;
+  return static_cast<int64_t>(cycles);
 #endif
 #elif defined(__e2k__) || defined(__elbrus__)
   struct timeval tv;


dmah42 (approved):

src/cycleclock.h
```diff
@@ -210,7 +210,7 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
 #else
   uint64_t cycles;
   asm volatile("rdtime %0" : "=r"(cycles));
-  return cycles;
+  return static_cast<int64_t>(cycles);
```
dmah42, line 213: maybe we should check the range before the cast?

LebedevRI: We don't do that anywhere else though. I'm not sure what we would even do in that situation?

dmah42: assert i guess?  the C++ style guide talks about this a bit in discussing avoidance of unsigned types ("In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this.")

but i agree we don't currently do it anywhere else.  happy to punt this to another PR.

PR #1803, opened 2024-06-14
[bazel] Use `includes` instead of `strip_include_prefix`

alexkaratarakis:
When using `includes`, consumers will apply the headers using `-isystem`, instead of `-I`. This will allow diagnostics of consumers to not apply to `benchmark`.

More info:

https://bazel.build/reference/be/c-cpp#cc_library.includes

https://bazel.build/reference/be/c-cpp#cc_library.strip_include_prefix

gtest uses `includes` as well:
https://github.com/google/googletest/blob/1d17ea141d2c11b8917d2c7d029f1c4e2b9769b2/BUILD.bazel#L120

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -73,6 +73,7 @@ cc_library(
         ":perfcounters": ["HAVE_LIBPFM"],
         "//conditions:default": [],
     }),
+    includes = ["include"],
     linkopts = select({
         ":windows": ["-DEFAULTLIB:shlwapi.lib"],
         "//conditions:default": ["-pthread"],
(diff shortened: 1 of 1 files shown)


dmah42:
i had originally decided to not use `includes` for exactly this reason: that it is not a system include and would normally be included with a manual installation as `-I` rather than `-isystem`.  however, i recognise the diagnostic issues that can be exposed through this, and that this PR simplifies integration for clients, so i'm happy to accept the change.

thanks so much :)

alexkaratarakis:
> looks like precommit has an opinion about ordering of fields.

Fixed and verified locally:
```
$ pre-commit run --all-files
buildifier...............................................................Passed
buildifier-lint..........................................................Passed
mypy.....................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed

```

dmah42:
unrelated failure.  thank you for the fix.

Issue #1807, opened 2024-06-28
[FR] Add support for adding custom profilers


xdje42:
It is useful to get a trace of a benchmark, e.g., with DynamoRio (https://dynamorio.org/).

What's needed is a public API akin to the MemoryManager API that provides for wrapping the running of the benchmark.
Suitable binaries can register the profiler (akin to how a memory manager is specified), and the framework
can invoke the profiler during the run (again, akin to how memory manager is handled).

PR #1809, opened 2024-07-01
[FR] Add API to provide custom profilers #1807

xdje42:
This API is akin to the MemoryManager API and lets tools provide their own profiler which is wrapped in the same way MemoryManager is wrapped. Namely, the profiler provides Start/Stop methods that are called at the start/end of running the benchmark in a separate pass.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -62,6 +62,8 @@ namespace internal {
 
 MemoryManager* memory_manager = nullptr;
 
+ProfilerManager* profiler_manager = nullptr;
+
 namespace {
 
 static constexpr IterationCount kMaxIterations = 1000000000000;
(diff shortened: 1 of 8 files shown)


src/benchmark_runner.cc
```diff
@@ -482,6 +505,10 @@ void BenchmarkRunner::DoOneRepetition() {
   run_results.non_aggregates.push_back(report);
 
   ++num_repetitions_done;
+
+  if (profiler_manager != nullptr) {
```
dmah42, line 509: nit: can we run this right after the memory manager and before the reporting?

src/benchmark_runner.cc
```diff
@@ -401,6 +403,44 @@ void BenchmarkRunner::RunWarmUp() {
...
+  return memory_result;
+}
+
+void BenchmarkRunner::RunProfilerManager() {
+  // TODO: Provide a way to specify the number of iterations.
+  IterationCount profile_iterations = 1;
+  std::unique_ptr<internal::ThreadManager> manager;
+  manager.reset(new internal::ThreadManager(1));
+  b.Setup();
+  profiler_manager->AfterSetupStart();
```
dmah42, line 435: why doesn't this report alongside the other results?

xdje42: Interesting question. What results make sense to report here?

src/benchmark_runner.cc
xdje42: Sure

src/benchmark_runner.cc
dmah42: i'm not sure what the trace manager would return.  what do you expect to be reported?  for memory manager, no matter what the implementation, we expect number of allocations, max bytes used, total allocated, etc to be common to all memory management.

is there anything the profile manager might have that's common?

dmah42:
looks like clang-format (style Google) needs to be run.

src/benchmark_runner.cc
xdje42: Tracing will generate some output files, so it will want to report the names of the generated files.
The question is how to provide a generic way to communicate random bits of info back to the framework (if the framework wants to be the one to report the results to the user).
In the absence of a way to report back results, I was thinking trace manager would just use printf.

`CreateRunReport` takes `memory_iterations` and `memory_results` as arguments.
We could add `profile_iterations` and `profile_results`, we just need to figure out the type of `profile_results`.
There is `UserCounters` in `State`, so `ProfilerManager` can use that (I think, right?).
It doesn't support string text though. Maybe we could add a simple API that ProfilerManager can use to add some string text to the report? Eg, change `Stop` to return a string to report back to the user?

dmah42: if the tracing creates the output directly then there's no need to report them here as well.  i don't know if that's generally true, but we can certainly start with that as a baseline.

xdje42: SGTM

src/benchmark_runner.cc
```diff
@@ -445,28 +485,15 @@ void BenchmarkRunner::DoOneRepetition() {
            "then we should have accepted the current iteration run.");
   }
 
-  // Oh, one last thing, we need to also produce the 'memory measurements'..
+  // Produce memory measurements if requested.
   MemoryManager::Result* memory_result = nullptr;
   IterationCount memory_iterations = 0;
```
dmah42, line 490: this breaks reporting.  you need to _return_ the number of memory iterations for l501 to make sense.  ie, outside the call to RunMemoryManager you should be setting the memory_iterations to min(16, iters) so that the reporting lower down can report the right thing.  or make IterationCount a pointer/reference but don't do that :)

xdje42: Done

dmah42 (approved):

dmah42 (approved):

PR #1814, opened 2024-07-11
add PERF_FORMAT_TOTAL_TIME_{ENABLED,RUNNING} to support multiplexing

### LINKED ISSUE #1813: [BUG] Perf counters not multiplexed
**Describe the bug**

We do not seem to fully support multiplexing in benchmark perf_counters, which can result in undercounting by a factor of 2..3.

There is an attempt to have multiple 'leaders' and we do specify `PERF_FORMAT_GROUP`, but not `PERF_FORMAT_TOTAL_TIME_ENABLED` nor `PERF_FORMAT_TOTAL_TIME_RUNNING`. The latter two are required to extrapolate the counter values according to how long they are running.

**System**
  - OS: Linux
  - Compiler and version: clang near trunk

**To reproduce**
Request many (>6) counters.

**Expected behavior**
Values are scaled according to the relative time during which the counter was enabled/bound to HW.

**Additional context**

From the [manual](https://man7.org/linux/man-pages/man2/perf_event_open.2.html):

If PERF_FORMAT_GROUP was specified to allow reading all events
          in a group at once:

              struct read_format {
                  u64 nr;            /* The number of events */
                  u64 time_enabled;  /* if PERF_FORMAT_TOTAL_TIME_ENABLED */
                  u64 time_running;  /* if PERF_FORMAT_TOTAL_TIME_RUNNING */
                  struct {
                      u64 value;     /* The value of the event */

dmah42:
Fixes #1813

### DIFF
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -157,7 +157,8 @@ PerfCounters PerfCounters::Create(
     attr.exclude_hv = true;
 
     // Read all counters in a group in one read.
-    attr.read_format = PERF_FORMAT_GROUP;
+    attr.read_format = PERF_FORMAT_GROUP | PERF_FORMAT_TOTAL_TIME_ENABLED |
+                       PERF_FORMAT_TOTAL_TIME_RUNNING;
 
     int id = -1;
     while (id < 0) {


jan-wassenberg:
Nice, thanks for adding :) I think this does not fully fix the issue. Adding the flags changes the in-memory layout of the counters. It remains to add those to the struct, replacing the 'padding' field, and then to extrapolate the counters based on the fraction of enabled/running.

dmah42:
do you have a pointer? i have no idea how this is meant to work.

jan-wassenberg:
Check out cl/638153709 :)

jan-wassenberg:
SG. In the meantime, probably best to revert this commit.

PR #1815, opened 2024-07-15
Clarify the difference between `BENCHMARK_TEMPLATE_F` and `BENCHMARK_TEMPLATE_DEFINE_F` + `BENCHMARK_REGISTER_F`

jiawen:
Add comments highlighting the difference between `BENCHMARK_TEMPLATE_F` and `BENCHMARK_TEMPLATE_DEFINE_F`, mirroring those of `BENCHMARK_F ` and `BENCHMARK_DEFINE_F`.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -665,7 +665,9 @@ BENCHMARK_TEMPLATE_DEFINE_F(MyFixture, DoubleTest, double)(benchmark::State& st)
   }
 }
 
+/* DoubleTest is NOT registered */
 BENCHMARK_REGISTER_F(MyFixture, DoubleTest)->Threads(2);
+/* DoubleTest is now registered */
 ```
 
 <a name="custom-counters" />
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -631,7 +631,7 @@ BENCHMARK_F(MyFixture, FooTest)(benchmark::State& st) {
   }
 }
 
-// Only defines `BarTest` using `MyFixture`.
+// Only defines `BarTest` using the class `MyFixture`.
 BENCHMARK_DEFINE_F(MyFixture, BarTest)(benchmark::State& st) {
    for (auto _ : st) {
      ...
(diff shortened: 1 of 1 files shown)


docs/user_guide.md
```diff
@@ -665,7 +665,9 @@ BENCHMARK_TEMPLATE_DEFINE_F(MyFixture, DoubleTest, double)(benchmark::State& st)
   }
 }
 
+/* DoubleTest is NOT registered */
```
LebedevRI, line 668: I *think* `//`-style comments are used elsewhere, so let's be consistent.
I think it would be good to also mention `IntTest`,
is `BENCHMARK_TEMPLATE_F` enough, and only the `BENCHMARK_TEMPLATE_DEFINE_F`
needs to be followed up with `BENCHMARK_REGISTER_F` or not?

jiawen: 👍, I'm happy to do that (and will write some tests to confirm!).

In the user guide, [Fixtures](https://github.com/google/benchmark/blob/main/docs/user_guide.md#fixtures) is similar and uses `/* ... */` comments - I think it's within scope to update that one.

Shall I update [the one other instance in Preventing Optimization](https://github.com/google/benchmark/blob/main/docs/user_guide.md#preventing-optimization) as well?

LebedevRI: > 👍, I'm happy to do that (and will write some tests to confirm!).
 
> In the user guide, [Fixtures](https://github.com/google/benchmark/blob/main/docs/user_guide.md#fixtures) is similar and uses `/* ... */` comments - I think it's within scope to update that one.
> 
> Shall I update [the one other instance in Preventing Optimization](https://github.com/google/benchmark/blob/main/docs/user_guide.md#preventing-optimization) as well?

Eh, i guess why not?

jiawen: Done. PTAL.

LebedevRI:
Seems nice to me. I did not verify that the behavior of `BENCHMARK_F`/`BENCHMARK_TEMPLATE_F`
is as stated, but it does seem reasonable.

dmah42 (approved):

dmah42:
great. thank you!

PR #1795, opened 2024-05-29
Upgrade google/benchmark to use nanobind-bazel v2.0.0

nicholasjng:
This enables binding extension builds with nanobind@v2.

Due to some backwards-incompatible changes in the treatment of enums, `Counter::Flags` (a flag-based enum) needs to be explicitly marked as arithmetic; setting flags is done as before with the logical "or" operator (now bound like any other class method as `__or__()`, the Python logical or operator slot).

The bindings changes were suggested by @hawkinsp in the corresponding issue on the repo, see https://github.com/google/benchmark/issues/1790.

----------------------------

Fixes #1790.

PR #1817, opened 2024-07-18
Update benchmark Python bindings for nanobind 2.0, and update to nano…

hawkinsp:
…bind 2.0.

Incorporates the nanobind_bazel change from https://github.com/google/benchmark/pull/1795.

nanobind 2.0 reworked the nanobind::enum_ class so it uses a real Python enum or intenum rather than its previous hand-rolled implementation. https://nanobind.readthedocs.io/en/latest/changelog.html#version-2-0-0-may-23-2024

As a consequence of that change, nanobind now checks when casting an integer to a enum value that the integer corresponds to a valid enum. Counter::Flags is a bitmask, and many combinations are not valid enum members.

This change:
a) sets nb::is_arithmetic(), which means Counter::Flags becomes an IntEnum that can be freely cast to an integer. b) defines the | operator for flags to return an integer, not an enum, avoiding the error. c) changes Counter's constructor to accept an int, not a Counter::Flags enum. Since Counter::Flags is an IntEnum now, it can be freely coerced to an int.

If https://github.com/wjakob/nanobind/pull/599 is merged into nanobind, then we can perhaps use a flag enum here instead.

### DIFF
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -118,7 +118,7 @@ NB_MODULE(_benchmark, m) {
   using benchmark::Counter;
   nb::class_<Counter> py_counter(m, "Counter");
 
-  nb::enum_<Counter::Flags>(py_counter, "Flags")
+  nb::enum_<Counter::Flags>(py_counter, "Flags", nb::is_arithmetic())
       .value("kDefaults", Counter::Flags::kDefaults)
       .value("kIsRate", Counter::Flags::kIsRate)
       .value("kAvgThreads", Counter::Flags::kAvgThreads)
@@ -130,7 +130,9 @@ NB_MODULE(_benchmark, m) {
       .value("kAvgIterationsRate", Counter::Flags::kAvgIterationsRate)
       .value("kInvert", Counter::Flags::kInvert)
       .export_values()
-      .def(nb::self | nb::self);
+      .def("__or__", [](Counter::Flags a, Counter::Flags b) {
+        return static_cast<int>(a) | static_cast<int>(b);
+      });
 
   nb::enum_<Counter::OneK>(py_counter, "OneK")
       .value("kIs1000", Counter::OneK::kIs1000)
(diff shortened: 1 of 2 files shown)


dmah42 (approved):
i trust your judgement but i don't really understand the change :)

bindings/python/google_benchmark/benchmark.cc
```diff
@@ -130,18 +130,25 @@ NB_MODULE(_benchmark, m) {
...
+        return static_cast<int>(a) | static_cast<int>(b);
+      });
 
   nb::enum_<Counter::OneK>(py_counter, "OneK")
       .value("kIs1000", Counter::OneK::kIs1000)
       .value("kIs1024", Counter::OneK::kIs1024)
       .export_values();
 
   py_counter
-      .def(nb::init<double, Counter::Flags, Counter::OneK>(),
```
nicholasjng, line 141: In your experience, is there any situation left at this point where you would prefer `nb::init<>` over `.def("__init__", ...)`?

bindings/python/google_benchmark/benchmark.cc
```diff
@@ -130,18 +130,25 @@ NB_MODULE(_benchmark, m) {
...
-           nb::arg("value") = 0., nb::arg("flags") = Counter::kDefaults,
-           nb::arg("k") = Counter::kIs1000)
-      .def("__init__", ([](Counter *c, double value) { new (c) Counter(value); }))
+      .def(
+          "__init__",
+          [](Counter* c, double value, int flags, Counter::OneK oneK) {
+            new (c) Counter(value, static_cast<Counter::Flags>(flags), oneK);
```
nicholasjng, line 146: Won't this throw if `flags` is not a member of `Counter::Flags`?

hawkinsp: No. C++ doesn't care about this. enum flags are common in C++. It's only nanobind that checks membership.

bindings/python/google_benchmark/benchmark.cc
hawkinsp: Well, I'd prefer it wherever it works. But... that's not the case here.

dmah42 (approved):

PR #1819, opened 2024-07-23
Revert to token authentication for PyPI wheel uploads

nicholasjng:
Until the PyPI account is recovered, it should be possible to upload wheels with the GitHub secrets that were previously used.

Changes the PyPI upload action sourcing to point to the v1 stable release branch, which receives rolling updates and is the canonical way of including the wheel publishing action.

Uploading will probably need another release, because setuptools_scm needs to produce a clean tag that the PyPI API allows as an upload.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -81,10 +81,11 @@ jobs:
     name: Publish google-benchmark wheels to PyPI
     needs: [merge_wheels]
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
     steps:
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

nicholasjng:
As a progress update: I am almost done with the new nanobind-bazel release, only the Windows stubgen implementation is giving me some trouble.

After that's done, the following google-benchmark PyPI releases will have stub files packaged with them, which allows for IDE completions, static type-checking, and an altogether better dev workflow for Python developers of GBM..

PR #1824, opened 2024-07-29
Bump nanobind-bazel to v2.1.0, add stubgen target

nicholasjng:
Adds a stub file for the `google_benchmark._benchmark` submodule, generated with the new `nanobind_stubgen` rule released in nanobind_bazel v2.1.0.

Tweaks the setup.py logic a little bit to package stub files with the rest of the build artifacts. Also explicitly adds the generated stub and marker files to the list of package data artifacts.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -38,4 +38,4 @@ use_repo(pip, "tools_pip_deps")
 
 # -- bazel_dep definitions -- #
 
-bazel_dep(name = "nanobind_bazel", version = "2.0.0", dev_dependency = True)
+bazel_dep(name = "nanobind_bazel", version = "2.1.0", dev_dependency = True)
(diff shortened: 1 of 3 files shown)


nicholasjng:
Side note: The generated stub causes the following mypy errors:

1. Unreachable overloads on `Counter.Flags.__init__()`
2. Too narrow typing of `arg` in `Counter.Flags.__or__()` (expects `int`, got `IntEnum`)
3.  Too narrow typing of `arg` in `UserCounters.__eq__()` and `UserCounters.__ne__()` (expects `object`, got `UserCounters`).

These all seem to be problems in `nanobind.stubgen` itself, so there is not a lot to do here except maybe contribute fixes or use a pattern file to rewrite the stub directly at build time.

PR #1837, opened 2024-08-08
Verify RegisterProfilerManager doesn't overwrite an existing registration

xdje42:
Tested:
Configure with:
cmake -GNinja -DCMAKE_BUILD_TYPE=Debug -DBENCHMARK_DOWNLOAD_DEPENDENCIES=on
Then run:
ctest -R profiler_manager_gtest
Before change test fails (expected), after change test passes (expected)

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -663,6 +663,10 @@ void RegisterMemoryManager(MemoryManager* manager) {
 }
 
 void RegisterProfilerManager(ProfilerManager* manager) {
+  // Don't allow overwriting an existing manager.
+  if (manager != nullptr) {
+    BM_CHECK_EQ(internal::profiler_manager, nullptr);
+  }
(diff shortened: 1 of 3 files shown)


dmah42:
i don't see the test in the PR

xdje42:
Yeah, I wasn't sure how to write a crash test. Is there an existing example?

dmah42:
it would be a standard googletest death test (https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests) .. the test file must end with _gtest.cc so we run it against googletest correctly.  but bear in mind we only assert-fatal in debug, not release.

dmah42:
did you manage to figure out the death test?

xdje42:
Oops. I was away during August, and this got dropped to the bottom of my queue.
Death test added.

dmah42 (approved):

PR #1841, opened 2024-08-19
Supply MacOS deployment target to delocate, use build+uv frontend

nicholasjng:
This shaves off multiple minutes from the wheel builds alone.

Also revert to trusted publishing for wheel uploads as it is now set up.

----------------------------

This limits Google Benchmark PyPI distributions to Python 3.10+.

Successful wheel build run on this branch: https://github.com/nicholasjng/benchmark/actions/runs/10452277981 , slightly under 20 minutes elapsed time including aarch64.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install Python 3.12
         uses: actions/setup-python@v5
         with:
-          python-version: 3.12
+          python-version: "3.12"
       - run: python -m pip install build
       - name: Build sdist
         run: python -m build --sdist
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

nicholasjng:
Hm, I have no idea what's going wrong here - builds perfectly fine on my local laptop (MacOS ARM64). I will investigate.

nicholasjng:
Ah, it's related to the use of Python 3.11. The problem disappears when building with Python 3.12. I wonder if that's because of the removal of distutils?

nicholasjng:
Closing in favor of the follow-up PR containing the setuptools-scm revert, which is green.

PR #1842, opened 2024-08-19
Bump oldest supported Python to 3.10, eliminate setuptools-scm

nicholasjng:
The version is now a string again, under the same attribute as it was before. This is a pragmatic decision in order to be able to upload wheels again, possibly directly from main.

We could in the future also set the Python version to a development version if we want to avoid accidental uploads of `main`.

Built on #1841.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools<73"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,25 +1,21 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm[toml]", "wheel"]
+requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
 name = "google_benchmark"
 description = "A library to benchmark code snippets."
-requires-python = ">=3.8"
-license = {file = "LICENSE"}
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -1,25 +1,21 @@
...
-    {name = "Google", email = "benchmark-discuss@googlegroups.com"},
-]
+authors = [{ name = "Google", email = "benchmark-discuss@googlegroups.com" }]
 
 classifiers = [
     "Development Status :: 4 - Beta",
     "Intended Audience :: Developers",
     "Intended Audience :: Science/Research",
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.8",
```
dmah42, line 21: why are we dropping 3.8 and 3.9?

nicholasjng: I was mirroring NumPy's support policy of "every Python released in the previous 42 months", which is also JAX's policy (https://jax.readthedocs.io/en/latest/deprecation.html#python-and-numpy-version-support-policy). If we go for "everything that is not officially EOL" instead, I'll have to add it back.

dmah42: i think it's fine as long as we document it: https://google.github.io/benchmark/dependencies.html

nicholasjng: Sweet, I'll do that right quick.

nicholasjng: Done.

dmah42 (approved):

nicholasjng:
This is ready from my end.

dmah42 (approved):

PR #1844, opened 2024-08-23
Dry Run Argument

Shaan-Mistry:
Added the functionality for a dry run benchmark called through the cli argument --benchmark_dry_run.

**Changes**
- Added boolean flag for --benchmark_dry_run
- Added overriding logic and early stoping
- append /dry_run  to the benchmark name to indicate  there is a dry run
    - ex: BM_StringCreation/dry_run/100/min_time:0.300/repeats:4/threads:4

Please let me know if I need to make any change I will be happy to implement them!

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -80,6 +80,10 @@ BenchmarkReporter::Run CreateRunReport(
   BenchmarkReporter::Run report;
 
   report.run_name = b.name();
+
+  if (FLAGS_benchmark_dry_run) {
+    report.run_name.function_name.append("/dry_run");
+  }
   report.family_index = b.family_index();
   report.per_family_instance_index = b.per_family_instance_index();
   report.skipped = results.skipped_;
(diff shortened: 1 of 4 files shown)


src/benchmark_runner.cc
```diff
@@ -80,6 +80,10 @@ BenchmarkReporter::Run CreateRunReport(
   BenchmarkReporter::Run report;
 
   report.run_name = b.name();
+
+  if (FLAGS_benchmark_dry_run) {
```
dmah42, line 84: i don't think we need to add this to every benchmark.  i think we can add it to the context.

src/benchmark_runner.cc
```diff
@@ -228,20 +232,21 @@ BenchmarkRunner::BenchmarkRunner(
     : b(b_),
       reports_for_family(reports_for_family_),
       parsed_benchtime_flag(ParseBenchMinTime(FLAGS_benchmark_min_time)),
-      min_time(ComputeMinTime(b_, parsed_benchtime_flag)),
-      min_warmup_time((!IsZero(b.min_time()) && b.min_warmup_time() > 0.0)
-                          ? b.min_warmup_time()
-                          : FLAGS_benchmark_min_warmup_time),
+      min_time(FLAGS_benchmark_dry_run ? 0 : ComputeMinTime(b_, parsed_benchtime_flag)),
```
dmah42, line 235: this is starting to get complex.  i wonder if it's worth moving these into the body of the constructor and having 

```
if (FLAGS_benchmark_dry_run) {
  min_time = 0;
  min_warmup_time = 0;
  ...
} else {
  min_time = ComputeMinTime(b_, parsed_benchtime_flag);
  ...
}
```

wdyt?

src/benchmark_runner.cc
```diff
@@ -295,6 +300,10 @@ BenchmarkRunner::IterationResults BenchmarkRunner::DoNIterations() {
   BM_VLOG(2) << "Ran in " << i.results.cpu_time_used << "/"
              << i.results.real_time_used << "\n";
 
+  // In dry run mode, we report 1 iteration even if each thread runs a singular iteration.
+  if (FLAGS_benchmark_dry_run) {
```
dmah42, line 304: i don't understand this change.

src/benchmark_runner.cc
Shaan-Mistry: Ok, ill change that!

src/benchmark_runner.cc
Shaan-Mistry: I agree, I think for the long run it will help reduce complexity. i'll revise that.

src/benchmark_runner.cc
Shaan-Mistry: We chatted on discord about this. I might have misunderstood your answer. When we have a benchmark that has a specified number of threads (n), a dry run would run it for (1 x n) iterations total. Would we report that as 1 or (1 x n) iterations?

src/benchmark_runner.cc
Shaan-Mistry: We could also revert to the original logic and then add a simple if statement for the dry run case.

```
   ...
  iters(has_explicit_iteration_count
                ? ComputeIters(b_, parsed_benchtime_flag)
                : 1),
      perf_counters_measurement_ptr(pcm_) {
  if (FLAGS_benchmark_dry_run) {
    min_time = 0;
    min_warmup_time = 0;
    warmup_done = true;
    repeats = 1;
    iters = 1;
  }
  ...
  ```

dmah42: yes perhaps.  i'm not a fan of this being "hidden" but it does ensure the fields are initialized correctly on construction.

src/benchmark_runner.cc
dmah42: we should report whatever it does.

dmah42:
i tried to make some changes and commit here but it didn't work so they're in https://github.com/google/benchmark/pull/1851.  @Shaan-Mistry would you like to make the same changes in your PR and merge them here or are you ok with me merging #1851 (you still get commits in your name .. i don't want to take away from your work).

Shaan-Mistry:
merging #1851 is good with me!

Issue #1857, opened 2024-09-23
[BUG] pkg-config file has bad version string


xbreak:
**Describe the bug**

The pkg-config file version field contains a leading "v" , possibly introduced in 17bc235a, which breaks plain version constraints in pkg-config.

What worked in 1.8.3:

```console
$ pkg-config benchmark  'benchmark >= 1.8.0' --modversion
1.8.3
$ pkg-config benchmark  'benchmark' --atleast-version=1.8.0
# note: exit 0
```
Doesn't work in 1.9.0, 1.8.5 (and possibly 1.8.4):

```console
$ pkg-config benchmark  'benchmark >= 1.8.0' --modversion
Requested 'benchmark >= 1.8.0' but version of benchmark is v1.8.5
$ pkg-config benchmark  'benchmark' --atleast-version=1.8.0
# note: exit 1
```

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Linux
  - Compiler and version: n/a

**To reproduce**

Given release tar ball `v1.9.0.tar.gz`:

```console
$ tar xzf v1.9.0.tar.gz
$ cd benchmark-1.9.0
$ cmake -DBENCHMARK_ENABLE_TESTING=OFF
...
-- Google Benchmark version: v1.9.0, normalized to 1.9.0
...
$  pkg-config benchmark  'benchmark >= 1.8.0' --modversion
Requested 'benchmark >= 1.8.0' but version of benchmark is v1.8.5
$ PKG_CONFIG_PATH=. pkg-config --modversion 'benchmark > 1.8.0'
Package dependency requirement 'benchmark > 1.8.0' could not be satisfied.
Package 'benchmark' has version 'v1.9.0', required version is '> 1.8.0'
```

**Expected behavior**

```console
$ PKG_CONFIG_PATH=. pkg-config --modversion 'benchmark > 1.8.0'
1.9.0
```

**Possible fix**

```diff
--- cmake/benchmark.pc.in.orig	2024-09-23 09:20:18.558853817 +0000
+++ cmake/benchmark.pc.in	2024-09-23 09:18:31.412725720 +0000
@@ -5,7 +5,7 @@
 
 Name: @PROJECT_NAME@
 Description: Google microbenchmark framework
-Version: @VERSION@
+Version: @NORMALIZED_VERSION@
 
 Libs: -L${libdir} -lbenchmark
 Libs.private: -lpthread @BENCHMARK_PRIVATE_LINK_LIBRARIES@
```
[...]

PR #1858, opened 2024-09-26
Version string correction in pkg-config files

aespondac:
According to issue #1857, the version in the pkg-config files is changed so that they have the nomatized version instead of the version.


dmah42 (approved):

dmah42:
thank you!

LebedevRI:
@aespondac thank you!

PR #1859, opened 2024-09-27
Add enum value from newest Windows SDK

DKLoehr:
Windows SDK version 10.0.26100.0 adds a cache type value, `CacheUnknown`. This adds a case for that type to `sysinfo.cc`, which will otherwise complain about the switch statement being non-exhaustive when building with the new SDK.

Since the value doesn't exist in prior SDK versions, we only add the case conditionally. The condition can be removed if we ever decide to bump up the required SDK version.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -353,6 +353,10 @@ std::vector<CPUInfo::CacheInfo> GetCacheSizesWindows() {
     C.size = static_cast<int>(cache.Size);
     C.type = "Unknown";
     switch (cache.Type) {
+      #ifdef NTDDI_WIN10_NI // Windows SDK version >= 10.0.26100.0
+      case CacheUnknown:
+        break;
+      #endif
       case CacheUnified:
         C.type = "Unified";
         break;


DKLoehr:
We could, but `C.type` is already set to Unknown so it would be a no-op. Happy to do it that way if you think it's clearer.

dmah42 (dismissed):

dmah42 (approved):

dmah42:
looks like the define guard isn't quite right

DKLoehr:
Yes, I can't figure out why, unless I'm blind and missing something obvious. As far as I can tell that define was only added in the most recent SDK version, at the same time as CacheUnknown. Do you know what SDK version is being used by the builders which are failing?

dmah42:
> Yes, I can't figure out why, unless I'm blind and missing something obvious. As far as I can tell that define was only added in the most recent SDK version, at the same time as CacheUnknown. Do you know what SDK version is being used by the builders which are failing?

Current runner version: '2.319.1'
Operating System
  Microsoft Windows Server [2](https://github.com/google/benchmark/actions/runs/11108965351/job/30863405634?pr=1859#step:1:2)022
  10.0.20348
  Datacenter
Runner Image
  Image: windows-2022
  Version: 202[4](https://github.com/google/benchmark/actions/runs/11108965351/job/30863405634?pr=1859#step:1:4)0922.1.0
  Included Software: https://github.com/actions/runner-images/blob/win22/20240[9](https://github.com/google/benchmark/actions/runs/11108965351/job/30863405634?pr=1859#step:1:10)22.1/images/windows/Windows2022-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/win22%2F20240922.1

is any of that useful?

DKLoehr:
Found [this issue](https://github.com/actions/runner-images/issues/10349), it seems like windows-latest has been partially updated but not completely? Not 100% sure if that's the problem, but it seems highly suspicious given that the error we're seeing here looks like we've got only part of the new SDK.

dmah42 (approved):

dmah42:
latest is now passing, 2019/2022 are now failing :)

dmah42 (approved):

dmah42:
thank you!

DKLoehr:
Looking into it.

Issue #1860, opened 2024-10-10
[BUG] Warning 'misc-use-anonymous-namespace' should be disabled within `BENCHMARK` macro


codeinred:
**Describe the bug**

Use of the `BENCHMARK` macro results in a clang-tidy warning, because the macro expands to a global variable declared with `static`.

```cpp
bench.cpp:88:1: warning: variable 'benchmark_uniq_5_benchmark_' declared 'static', move to anonymous namespace instead [misc-use-anonymous-namespace]
   88 | BENCHMARK( myFunc );
```

**To reproduce**
Run clang-tidy on any file containing `BENCHMARK` macro, with the `misc-use-anonymous-namespace` check enabled

**Expected behavior**
No warning is emitted (warning should be suppressed)

**Screenshots**
![Image](https://github.com/user-attachments/assets/b6b843a4-9384-4e87-b659-9f36d3af7fa1)

**Additional context:**

This can be fixed by adding `NOLINTNEXTLINE(misc-use-anonymous-namespace)` within the macro declaration. Adding this silences the warning, only for `BENCHMARK` statement itself.

```cpp
#define BENCHMARK_PRIVATE_DECLARE(n)                                 \
  /* NOLINTNEXTLINE(misc-use-anonymous-namespace) */                 \
  static ::benchmark::internal::Benchmark* BENCHMARK_PRIVATE_NAME(n) \
      BENCHMARK_UNUSED
```

LebedevRI:
This was fixed by https://github.com/google/benchmark/pull/1861 but that PR/commit lacked "Fixes:".

PR #1861, opened 2024-10-10
Fix spurious warning 'misc-use-anonymous-namespace' (#1860)

codeinred:
Disables 'misc-use-anonymous-namespace' for usage of the BENCHMARK macro. This warning is spurious, and the variable declared by the BENCHMARK macro can't be moved into an annonymous namespace.

We don't want to disable it globally, but it can be disabled locally, for the `BENCHMARK` statement, as this warning appears downstream for users.

See:
https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

Before: (warning shown)

image

After: (no warnings)

image

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1554,6 +1554,7 @@ class Fixture : public internal::Benchmark {
   BaseClass##_##Method##_Benchmark
 
 #define BENCHMARK_PRIVATE_DECLARE(n)                                 \
+  /* NOLINTNEXTLINE(misc-use-anonymous-namespace) */                 \
   static ::benchmark::internal::Benchmark* BENCHMARK_PRIVATE_NAME(n) \
       BENCHMARK_UNUSED
 


include/benchmark/benchmark.h
```diff
@@ -1554,6 +1554,7 @@ class Fixture : public internal::Benchmark {
   BaseClass##_##Method##_Benchmark
 
 #define BENCHMARK_PRIVATE_DECLARE(n)                                 \
+  /* NOLINTNEXTLINE(misc-use-anonymous-namespace) */                 \
```
LebedevRI, line 1557: Can this use `//` comment style, like elsewhere in the project?

codeinred: That was what I originally tried to use, but the answer is no. The preprocessor interprets the `\` at the end of the line as extending the comment to the next line. If you try to compile it, it fails with `error: multi-line comment [-Werror,-Wcomment]`.

The comment has to be scoped by using the `/* */` syntax, since the comment is embedded into the macro.

I also tried putting the comment before the macro, but that doesn't work, since the error occurs at the point of expansion.

LebedevRI (approved):
Looks reasonable to me

LebedevRI:
@codeinred thank you!

PR #1864, opened 2024-10-22
update version of gtest to v1.15.2 (latest) and also the cmake config

imrichardcole:
This update removes the following warning:

```
CMake Deprecation Warning at build/third_party/googletest/src/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

Which was coming from cmake/GoogleTest.cmake.in.

It also updates googletest to the latest release version which removes subsequent cmake warnings.


dmah42:
"error: use of the 'maybe_unused' attribute is a C++17 extension [clang-diagnostic-c++17-attribute-extensions]"  looks like gtest moved to C++17 and we're using clang-tidy-14.

so we can bump clang-tidy but i wonder if we also need to make sure when we build for tests we're building on c++17?

imrichardcole:
From the previously shared page - https://opensource.google/documentation/policies/cplusplus-support#support_criteria_4

What would the earliest C++ standard you'd support?

dmah42:
> From the previously shared page - https://opensource.google/documentation/policies/cplusplus-support#support_criteria_4
> 
> What would the earliest C++ standard you'd support? I guess technically it's 10 years since C++14.

https://github.com/google/benchmark/blob/main/CMakeLists.txt#L141  is what we _currently_ support.  though if gtest has moved to require c++ 17 we can do that now based on the 10 year rule.  i have a calendar reminder for the first week of January to flip to 17 then :)

dmah42:
in #1867 i've updated to an interim version.  maybe that version is enough to make the cmake change work?

dmah42:
#1868 bumps partially and updates cmake.  are you ok with me merging that or do you want to change this to match, or keep iterating on this to support c++17?

LebedevRI (approved):
Probably fine?

LebedevRI (approved):
This newer version seems to still work, so why not.

One thing that we might do, some macros were renamed,
but we aren't getting warnings about that because of `-isystem`.

PR #1868, opened 2024-10-23
bump googletest version to match bazel

dmah42:


dmah42:
looks like this requires us to bump to cmake 3.13.  need to figure out if that's viable.

LebedevRI (approved):

PR #1871, opened 2024-10-25
Add a Python matrix to ensure the bindings build on all supported versions

nicholasjng:
Also contains a run of `pre-commit autoupdate`, and a bump of cibuildwheel to its latest tag for CPython 3.13 support.

But, since we build for 3.10+ with SABI from 3.12 onwards, we don't even need a dedicated Python 3.13 build job - the wheels from 3.12 can be reused.

### DIFF
--- a/.github/workflows/test_bindings.yml
+++ b/.github/workflows/test_bindings.yml
@@ -8,23 +8,23 @@ on:
 
 jobs:
   python_bindings:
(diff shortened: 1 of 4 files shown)


dmah42:
3.13 is unhappy.

nicholasjng:
Ah yeah, I didn't commit the 3.13 toolchain def in MODULE.bazel. Just a second

dmah42 (approved):

PR #1875, opened 2024-11-03
update standard to C++17 per C++ build support

dmah42:


dmah42:
this may not be necessary immediately as we're hitting https://github.com/google/googletest/issues/4605.  i might just disable the warning.

dmah42:
> this may not be necessary immediately as we're hitting [google/googletest#4605](https://github.com/google/googletest/issues/4605). i might just disable the warning.

actually, given we're like a month away from dropping support from C++14, maybe we can just pull up the schedule a bit and go all in on 17.

Kataglyphis:
I can reproduce the problem.

dmah42:
@LebedevRI any concerns with pushing this a bit earlier than planned?

dmah42:
@nicholasjng just flagging the 3.13 bindings on windows seem to be failing.

nicholasjng:
Thanks for the info, I think it's related to the interaction of Bazel's Python windows toolchains and the stable ABI.

The easiest fix is to just declare a 3.13 toolchain and revert the setup.py change.

nicholasjng:
Nice! Maybe with the library itself moving to C++17 we could create a project-wide bazelrc hosting the most common build options - that could also shorten the Python bindings instructions.

LebedevRI:
> trying to update some packages for various platforms, googletest requires C++17 which means we now need it too.

Why do we need to? Tests can be not built, they are optional.

dmah42:
> > trying to update some packages for various platforms, googletest requires C++17 which means we now need it too.
> 
> Why do we need to? Tests can be not built, they are optional.

that's true, but per our dependency policy it is time to default to C++17 (and simplify a few things).

PR #1880, opened 2024-11-21
README.md: fix build instructions

0dminnimda:


dmah42:
i wonder if we should do the opposite.

if a user has the googletest libraries installed on their machine then they do not need to download the dependencies.  if it was the case, then we would just make this the default.  as such, the instructions should probably elide the downloading of dependencies for both cases (old cmake and new).

further, as we now require cmake 3.13, the older form can be removed from the documentation entirely.

bosshaas13131313 (approved):

dmah42 (approved):

dmah42:
thank you!

PR #1886, opened 2024-12-10
Remove wheel merge job, merge artifacts on download

nicholasjng:
This is supported by `actions/download-artifact@v4`, and endorsed by cibuildwheel in their documentation (see https://cibuildwheel.pypa.io/en/stable/deliver-to-pypi/#github-actions).

PR #1887, opened 2024-12-10
Change SDK version check

DKLoehr:
Followup to #1859. When that was merged, github's windows builders were only partially updated to the new windows build tools (see [this issue](https://github.com/actions/runner-images/issues/10349) for more information). We worked around this by checking the installed SDK version in a funky way, which seems to have stopped working (#1886) now that they've [fully updated](https://github.com/actions/runner-images/pull/10862).

If we ever decide that the new SDK is a requirement to build, we can remove the check entirely and just include the code unconditionally.


dmah42:
fab. thank you.

PR #1892, opened 2024-12-20
[bazel] Remove selects on CPU

keith:
In a future version of bazel this produces a warning. In this case using
only the platform being windows is enough. Fixes:

```
WARNING: /.../benchmark/BUILD.bazel:29:15: in config_setting rule //:windows: select() on cpu is deprecated. Use platform constraints instead: https://bazel.build/docs/configurable-attributes#platforms.
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -18,27 +18,9 @@ COPTS = [
 ]
 
 config_setting(
-    name = "qnx",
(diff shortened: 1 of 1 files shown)


dmah42:
all CI failures are unrelated.  forcing merge.  thanks :)

dmah42 (approved):

Issue #1869, opened 2024-10-24
[BUG] Does not build with clang-cl


R-Goc:
**Describe the bug**
Due to adding the /MP option to MSVC like compilers, benchmark doesn't build on clang-cl or icx-cl as they don't handle it in any way, and will throw an unused command line option warning. Due to -Werror this will cause the compilation to fail. 

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Windows 11
  - Compiler and version: Clang-cl 19

**To reproduce**
1. sync to commit ... master
2.  cmake -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_LINKER=lld-link -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl
3. cmake --build build

**Expected behavior**
Builds with no modification to cmakelists (removing MP solves it). 
The solution would be checking if it is clang, and if so omitting /MP. Not sure how to check for icx. 

**Screenshots**
![Image](https://github.com/user-attachments/assets/b67840a4-2724-4b19-beed-22d0ec5d7dbf)

**Additional context**
Also should affect icx-cl, as it is based on clang.

PR #1895, opened 2024-12-28
fix: remove clang-cl compilation warning

theComputeKid:
The `/MP` flag only applies to `cl`, not `cl` frontends to other compilers (e.g. `clang-cl`, `icx-cl` etc), whereas the `MSVC` cmake variable applies to every compiler with a `cl` frontend. Therefore, we need to reduce the scope of `/MP` to only MSVC's `cl`, so that the other `cl` frontends can compile without warnings.

Fixes https://github.com/google/benchmark/issues/1869

Reproduction steps to test solution on Windows with the LLVM toolchain installed:
```
cmake -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_LINKER=lld-link -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON
cmake --build build
ctest --test-dir build
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,12 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 if (MSVC)
   # Turn compiler warnings up to 11
   string(REGEX REPLACE "[-/]W[1-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

dmah42:
i'm looking at the windows clang32 issues too ... looks like the msys2 environment doesn't include cc.  either way this PR looks good.

PR #1898, opened 2025-01-08
changes to support newer clang-tidy errors

dmah42:


dmah42:
the segfault on cxx03_test is a bit of a surprise.  i'm debating whether, given we now require c++17 to build the library, it's time to drop the cxx03 stuff and clean everything up anyway.

cc @LebedevRI

dmah42:
so the issue here is that the check for `BENCHMARK_HAS_CXX11` is checking if the compiler supports c++11 and not if the current `--std` is set to C++11 (or greater).  so we end up using c++11 style enum definitions which somehow compile even though the standard is set to not support them.  this then causes issues with where various pointers are pointing, etc, etc, we get a jump to an invalid address.

src/log.h
```diff
@@ -11,6 +11,14 @@
 #define BENCHMARK_HAS_CXX11
 #endif
 
+// This _MSC_VER check should detect VS 2017 v15.3 and newer.
+#if __cplusplus >= 201703L || \
```
theComputeKid, line 15: This is only defined if the [Zc:__cplusplus](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170) flag is enabled.

dmah42: on msvc, yes, which is why we have the rest of the check.

dmah42:
running a straw poll on the discord

dmah42:
see #1903 and #1906 for followons.

PR #1899, opened 2025-01-08
cycleclock: Use cock_gettime() as fallback for any Linux architecture

hdeller:
The Linux kernel provides the clock_gettime() functions since a long time already, so it's possible to use it as a generic fallback option for any architecture if no other (better) possibility has been provided instead.

I noticed the benchmark package failed to build on debian on the SH-4 architecture, so with this change SH-4 is now the first user of this fallback option.


hdeller:
As can be seen here:
https://buildd.debian.org/status/package.php?p=benchmark&suite=sid
sh-4 was the last architecture on which building this package failed.

dmah42 (approved):

PR #1900, opened 2025-01-08
Fix runtime crash when parsing /proc/cpuinfo fails

hdeller:
The testcase fails on sparc64, because the parsing of /proc/cpuinfo fails and thus currently returns "0" CPUs which finally leads to division-by-zero faults in the tests.

Fix the issue by returning at least "1" CPU which allows the tests to run. A error message will be printed in any case.

Long-term the code should be fixed to parse the cpuinfo output on sparch which looks like this:
...
type            : sun4v
ncpus probed    : 48
ncpus active    : 48

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -561,10 +561,12 @@ int GetNumCPUsImpl() {
 }
 
 int GetNumCPUs() {
-  const int num_cpus = GetNumCPUsImpl();
+  int num_cpus = GetNumCPUsImpl();
   if (num_cpus < 1) {
     std::cerr << "Unable to extract number of CPUs.  If your platform uses "
                  "/proc/cpuinfo, custom support may need to be added.\n";
+    /* There is at least one CPU which we run on. */
+    num_cpus = 1;
   }
   return num_cpus;
 }


hdeller:
A failing log on a sparc64 machine can be found here:
https://buildd.debian.org/status/fetch.php?pkg=benchmark&arch=sparc64&ver=1.8.4-1%7Eexp1&stamp=1719401179&raw=0

Currently I suggest to apply this trivial patch as-is to get rid of the problem for now.

In the long run I see two options:
a) Either add code to parse /proc/cpuinfo on sparc (as written in the commit message), or
b) drop all code which parses /proc/cpuinfo, and instead switch to calling "sysconf(_SC_NPROCESSORS_ONLN)" (see man-page).

I suggest to implement b).

src/sysinfo.cc
```diff
@@ -561,10 +561,12 @@ int GetNumCPUsImpl() {
 }
 
 int GetNumCPUs() {
-  const int num_cpus = GetNumCPUsImpl();
+  int num_cpus = GetNumCPUsImpl();
   if (num_cpus < 1) {
     std::cerr << "Unable to extract number of CPUs.  If your platform uses "
```
dmah42, line 566: should this instead be an assert that would require platform support to be added?

hdeller: I generally don't like an asserts, unless there is a really *CRITICAL* problem.
Not being able to correctly detect the number of CPUs is not at all a critical problem.
Currently it prints that the number of CPUs could not be determined at that someone may take a look at it.
With my addition it will then will return 1 CPU (which is of course somewhat true since we wouldn't be here if there was no CPU, although maybe not the perfect answer).

dmah42: that's fair :)

dmah42 (approved):

PR #1901, opened 2025-01-09
Get number of CPUs with sysconf() on Linux

hdeller:
Avoid parsing the /proc/cpuinfo just to get number of CPUs. Instead use the portable function provided by glibc.


hdeller:
This is a follow-up patch to https://github.com/google/benchmark/pull/1900
Let's see if it survives the workflow checks.

dmah42:
clang-format would like some love please.  the other failures are at HEAD.

hdeller:
I dropped the leading space, as mentioned by clang test.
Is it possible to merge both changes into one when you merge?

dmah42 (approved):

PR #1903, opened 2025-01-13
remove cxx03 test, fully unblocking c++1X development

dmah42:
removing cxx03_test allows us to use c++1X features in headers and thus clean up a bunch of clang-tidy warnings too.

after this i'll tackle the sanitizer and windows/clang32 CI bots.


dmah42:
@LebedevRI i'd appreciate your thoughts on taking this step.  i think it's time, but i'm not entirely comfortable making the decision unilaterally.

LebedevRI:
> @LebedevRI i'd appreciate your thoughts on taking this step. i think it's time, but i'm not entirely comfortable making the decision unilaterally.

FWIW, i'm only now starting to actually look at github notifications, or read incoming mail.
Previous two/three years of mail, or most of github notifications over last 3+ months went into trash.

LebedevRI (commented):
Could you please drop the `enum` changes (outside of `test/`) from this PR, i.e., do explicitly specify the underlying type, which i guess is `int`, and do the type narrowing in a follow-up PR?
Latter may or may not have ABI implications and should be more explicit.
https://godbolt.org/z/d46eGcsoj

dmah42:
i'm happy to, but how would we test the ABI implications in that other PR?  i'm not seeing how splitting it is going to make it any safer, unless you just want something to more easily roll back?

dmah42:
done

LebedevRI:
> i'm happy to, but how would we test the ABI implications in that other PR? i'm not seeing how splitting it is going to make it any safer, unless you just want something to more easily roll back?

Without that change, this PR is rather no-changes-intended and is easy to review.
When that change is on it's own, it will at least be more explicit that it's what is going on,
maybe we'll do the enum narrowing maybe we won't.

dmah42:
well, we will do it, because the whole point is to get clang-tidy to stop complaining :)

src/sysinfo.cc
```diff
@@ -84,7 +84,7 @@
 namespace benchmark {
 namespace {
 
-void PrintImp(std::ostream& out) { out << std::endl; }
+void PrintImp(std::ostream& out) { out << '\n'; }
```
LebedevRI, line 87: Should we flush stream[s] before calling `std::exit()`, since it does not do that automatically?

include/benchmark/benchmark.h
```diff
@@ -544,8 +506,7 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp&& value) {
   asm volatile("" : "+m,r"(value) : : "memory");
 #endif
 }
-#endif
-#elif defined(BENCHMARK_HAS_CXX11) && (__GNUC__ >= 5)
+#elif (__GNUC__ >= 5)
```
LebedevRI, line 509: This needs a comment. It took me a while to get that it is for `#if !defined(__GNUC__) || defined(__llvm__) || defined(__INTEL_COMPILER)`

include/benchmark/benchmark.h
```diff
@@ -764,12 +678,7 @@ class ThreadTimer;
 class ThreadManager;
 class PerfCountersMeasurement;
 
-enum AggregationReportMode
-#if defined(BENCHMARK_HAS_CXX11)
-    : unsigned
```
LebedevRI, line 769: Let's keep it if it was already present before this patch

include/benchmark/benchmark.h
```diff
@@ -784,15 +693,7 @@ enum AggregationReportMode
       ARM_FileReportAggregatesOnly | ARM_DisplayReportAggregatesOnly
 };
 
-enum Skipped
-#if defined(BENCHMARK_HAS_CXX11)
-    : unsigned
```
LebedevRI, line 789: Let's keep it if it was already present before this patch

src/sysinfo.cc
dmah42: adding to everywhere we explicitly `exit`

include/benchmark/benchmark.h
dmah42: it took me ages to navigate these ifdefs too.  added comments.

test/output_test_helper.cc
```diff
@@ -490,7 +490,7 @@ static std::string GetTempFileName() {
     std::string name = GetRandomFileName();
     if (!FileExists(name)) return name;
   }
-  std::cerr << "Failed to create unique temporary file name" << std::endl;
+  std::cerr << "Failed to create unique temporary file name\n";
   std::abort();
```
LebedevRI, line 494: (Abort is same as exit)

src/sysinfo.cc
LebedevRI: There's probably other places where we might flush intentionally, but this is better than nothing.

src/check.h
```diff
@@ -57,7 +57,7 @@ class CheckHandler {
 #pragma warning(disable : 4722)
 #endif
   BENCHMARK_NORETURN ~CheckHandler() BENCHMARK_NOEXCEPT_OP(false) {
```
LebedevRI, line 59: Should this/`CallAbortHandler` `flush`?

dmah42: nice.. we have too many exit points :P

LebedevRI (approved):
And github ate my comment.

I really don't like this `flush` mess.
I think we should just never use `std::endl`,
always use `\n`, and explicitly `flush` the nessesary streams when needed.

Otherwise this seems fine to me, thank you for doing this!

dmah42:
sorry @LebedevRI i wiped your approval by merging and adding more flushes.

LebedevRI (approved):

PR #1906, opened 2025-01-13
fix sanitizer cmake builds

dmah42:
1. upgrade to newer LLVM
2. use libc++ includes for Isystem and I command line flags
3. update to C++17 in cmake feature checks


dmah42:
sanitizer failing due to missing <condition_variable> header, but that should be fine with -stdlib=libc++.  🤔

dmah42:
the remaining msan failures are real ones, not build setup ones.

PR #1909, opened 2025-01-22
Lint Python: Add more ruff rules

cclauss:
https://docs.astral.sh/ruff/rules

PR #1910, opened 2025-01-22
ruff rule E501: Fix long lines in Python code

cclauss:
% [`ruff rule E501 `](https://docs.astral.sh/ruff/rules/line-too-long) # Requested at https://github.com/google/benchmark/pull/1909#pullrequestreview-2566713563
# line-too-long (E501)

Derived from the **pycodestyle** linter.

## What it does
Checks for lines that exceed the specified maximum character length.

## Why is this bad?
Overlong lines can hurt readability. [PEP 8], for example, recommends
limiting lines to 79 characters. By default, this rule enforces a limit
of 88 characters for compatibility with Black and the Ruff formatter,
though that limit is configurable via the [`line-length`] setting.

In the interest of pragmatism, this rule makes a few exceptions when
determining whether a line is overlong. Namely, it:

1. Ignores lines that consist of a single "word" (i.e., without any
   whitespace between its characters).
2. Ignores lines that end with a URL, as long as the URL starts before
   the line-length threshold.
3. Ignores line that end with a pragma comment (e.g., `# type: ignore`
   or `# noqa`), as long as the pragma comment starts before the
   line-length threshold. That is, a line will not be flagged as
   overlong if a pragma comment _causes_ it to exceed the line length.
   (This behavior aligns with that of the Ruff formatter.)
4. Ignores SPDX license identifiers and copyright notices
   (e.g., `# SPDX-License-Identifier: MIT`), which are machine-readable
   and should _not_ wrap over multiple lines.

If [`lint.pycodestyle.ignore-overlong-task-comments`] is `true`, this rule will
also ignore comments that start with any of the specified [`lint.task-tags`]
(e.g., `# TODO:`).

## Example
```python
my_function(param1, param2, param3, param4, param5, param6, param7, param8, param9, param10)
```

Use instead:
```python
my_function(
    param1, param2, param3, param4, param5,
    param6, param7, param8, param9, param10
)
```

## Error suppression
Hint: when suppressing `E501` errors within multi-line strings (like
docstrings), the `noqa` directive should come at the end of the string
(after the closing triple quote), and will apply to the entire string, like
so:

```python
"""Lorem ipsum dolor sit amet.

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor.
"""  # noqa: E501
```

## Options
- `line-length`
- `lint.task-tags`
- `lint.pycodestyle.ignore-overlong-task-comments`
- `lint.pycodestyle.max-line-length`

[PEP 8]: https://peps.python.org/pep-0008/#maximum-line-length

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, ubuntu-20.04, macos-latest]
+        os: [ubuntu-22.04, ubuntu-20.04, ubuntu-22.04-arm, macos-latest]
         build_type: ['Release', 'Debug']
         compiler: ['g++', 'clang++']
         lib: ['shared', 'static']


.ycm_extra_conf.py
```diff
@@ -83,10 +83,10 @@ def IsHeaderFile(filename):
 
 
 def GetCompilationInfoForFile(filename):
-    # The compilation_commands.json file generated by CMake does not have entries
-    # for header files. So we do our best by asking the db for flags for a
-    # corresponding source file, if any. If one exists, the flags for that file
-    # should be good enough.
+    # The compilation_commands.json file generated by CMake does not have
+    # entriesfor header files. So we do our best by asking the db for flags for
```
dmah42, line 87: missing a space

dmah42 (dismissed):

dmah42 (approved):

Issue #1913, opened 2025-01-28
[BUG] ProfilerManager should run the benchmark the same number of iterations as the regular run


xdje42:
**Describe the bug**

If a `ProfilerManager` is installed, it will run the benchmark an additional time after the regular run and after the `MemoryManager` run.
Currently the benchmark is hardwired to only run one iteration:

```
IterationCount profile_iterations = 1;
```

This prevents doing things like comparing the results of the `ProfilerManager` run with the regular run as they're run a different number of iterations.
It's more useful to have the `ProfilerManager` run use the same number of iterations.
- the same options can be used to specify how many iterations
- if the user wants a different number of iterations they can run the benchmark twice

xdje42:
Currently working on a patch ...

PR #1914, opened 2025-01-29
[BUG] Fix ProfilerManager number of iterations #1913

xdje42:
Run the external profiler the same number of iterations as the benchmark was run normally.
This makes, for example, a trace collected via ProfilerManager consistent with collected PMU data.


dmah42 (approved):

dmah42:
thank you!

PR #1916, opened 2025-02-04
Retrieve the number of online CPUs on OpenBSD and NetBSD

brad0:


src/sysinfo.cc
```diff
@@ -153,11 +153,11 @@ ValueUnion GetSysctlImp(std::string const& name) {
   int mib[2];
 
   mib[0] = CTL_HW;
-  if ((name == "hw.ncpu") || (name == "hw.cpuspeed")) {
+  if ((name == "hw.ncpuonline") || (name == "hw.cpuspeed")) {
```
dmah42, line 156: not ncpufound?

brad0: The sysctl name is hw.ncpuonline.

dmah42: https://man.openbsd.org/sysctl.2 includes ncpufound which seems to be more predictable in some cases (from brief reading i did).  i'm not an expert but i want to make sure that we're picking the right one.

brad0: ncpufound is the same as ncpu. On OpenBSD X86 systems have SMT disabled by default so the number of CPUs would be double what is actually online and available. This is functiohnality equivalent to what is used on other OS's with ```sysconf(_SC_NPROCESSORS_ONLN)``` instead of ```_SC_NPROCESSORS_CONF```.

dmah42 (approved):

dmah42:
the clang-format is the only failure we need to fix please.

PR #1921, opened 2025-02-07
Remove /proc/cpuinfo fallback path

brad0:
AIX, WASM (fork of musl for libc) and a few others should now use the sysconf
path. /proc is not portable and cpuinfo is Linux specific. It does not work
anywhere else.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -497,6 +497,7 @@ int GetNumCPUsImpl() {
   }
   return hardware_threads.max_hthreads;
 #elif defined(BENCHMARK_HAS_SYSCTL)
+  // *BSD, macOS
   int num_cpu = -1;
   constexpr auto* hwncpu =
 #if defined BENCHMARK_OS_MACOSX
@@ -509,6 +510,7 @@ int GetNumCPUsImpl() {
   if (GetSysctl(hwncpu, &num_cpu)) return num_cpu;
   PrintErrorAndDie("Err: ", strerror(errno));
 #elif defined(_SC_NPROCESSORS_ONLN)
+  // Linux, Solaris, AIX, Haiku, WASM, etc.
   // Returns -1 in case of a failure.
   int num_cpu = static_cast<int>(sysconf(_SC_NPROCESSORS_ONLN));
   if (num_cpu < 0) {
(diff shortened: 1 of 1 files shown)


src/sysinfo.cc
```diff
@@ -517,54 +517,8 @@ int GetNumCPUsImpl() {
...
-  f.close();
-
-  if ((max_id + 1) != num_cpus) {
-    fprintf(stderr,
-            "CPU ID assignments in /proc/cpuinfo seem messed up."
-            " This is usually caused by a bad BIOS.\n");
-  }
-  return num_cpus;
+  // Fallback, no other API exists.
+  return -1;
```
LebedevRI, line 521: Should this be
```
#error Don't know how to detect CPU count for this platform?
```
?

brad0: Well the original code would build even if it would not return an appropriate value.

Depends what the intent is.

LebedevRI: That's what i'm asking, yes. It does seem like 
Actually, for that, we wouldn't need the return:
```
#else
#error Don't know how to detect CPU count for this platform
#endif
  BENCHMARK_UNREACHABLE();
}
```

dmah42: we had this as a "soft error" only because there are so many platforms out there, and this at least returns 1 CPU in the case that we can't identify it (and logs a thing).  i'm not sure we have good enough coverage yet to make it an error but if you all are confident, go for it.

LebedevRI: It would be a loud compile-time error that can not be missed,
not a run-time error that you might not notice when packaging,
so i think it would be more safe than the current more-than-silent error...

I'd do it. Especially if we are confident that this fallback is no longer needed.

dmah42: what it means is that any clients that need platform support will now find out.  that may be good or bad, but basic platform support will be "return 1" within an appropriate platform define so that's easy enough.

LebedevRI: Anyways, i've made my point. Deferring to @dmah42.

dmah42: i'm happy to keep the soft error.  i usually prefer hard errors but i don't really want an influx of random platforms needing support for their way to get CPU.

if we think the new version handles all cases (or should handle all cases) then we can switch it.

dmah42 (approved):

PR #1923, opened 2025-02-10
run clang-tidy using a common config and reduced set of tests

dmah42:


.github/workflows/clang-tidy.yml
```diff
@@ -35,4 +35,4 @@ jobs:
     - name: run
       shell: bash
       working-directory: ${{ runner.workspace }}/_build
-      run: run-clang-tidy -checks=*,-clang-analyzer-deadcode*,-clang-analyzer-optin*
+      run: run-clang-tidy -checks=*,-clang-analyzer-deadcode*,-clang-analyzer-optin*,-fuchsia-*,-llvmlibc-*,-modernize-use-trailing-return-type
```
LebedevRI, line 38: You might want to place this in a `.clang-tidy` file in the root source directory

dmah42: yeah i'm also going to reduce the number of checks we're running.

LebedevRI:
Just a general comment: it might be best to approach this as one check per PR,
otherwise it may turn into too-large-to-meaningfully review.

dmah42:
> Just a general comment: it might be best to approach this as one check per PR, otherwise it may turn into too-large-to-meaningfully review.

yeah, it's a draft PR so i can split it up later.

.github/workflows/clang-tidy.yml
```diff
@@ -35,4 +35,4 @@ jobs:
     - name: run
       shell: bash
       working-directory: ${{ runner.workspace }}/_build
-      run: run-clang-tidy -checks=*,-clang-analyzer-deadcode*,-clang-analyzer-optin*
+      run: run-clang-tidy -config-file=${{ runner.workspace }}/.clang-tidy
```
LebedevRI, line 38: It will be picked up automatically

.clang-tidy
```diff
@@ -1,6 +1,27 @@
 ---
-Checks:          'clang-analyzer-*,readability-redundant-*,performance-*'
-WarningsAsErrors: 'clang-analyzer-*,readability-redundant-*,performance-*'
-HeaderFilterRegex: '.*'
+Checks:          'abseil-*,bugprone-*,clang-analyzer-*,cppcoreguidelines-*,google-*,misc-*,performance-*,readability-*,-clang-analyzer-deadcode*,-clang-analyzer-optin*,-readability-identifier-length'
```
LebedevRI, line 2: You might want the long syntax:
```
---
Checks: >
  jfgioa,
  -gjafpq,

  fgjgwjop,
  -hrta

WarningsAsErrors:
```

.github/workflows/clang-tidy.yml
dmah42: it's not finding it, which suggests the same as when i passed it explicitly: that it's not running with the root as a parent directory, somehow.

dmah42: it finds it locally, fwiw

LebedevRI (approved):
Awesome!

dmah42:
ok fixed it.  clang-tidy is now running properly with the common config.

LebedevRI (approved):

PR #1924, opened 2025-02-11
benchmark declarations can and should be const (clang-tidy)

dmah42:


LebedevRI:
Those preexisting msan failures are weird.

LebedevRI (approved):
CI seems happy with this, so this is probably fine.

dmah42:
> Those preexisting msan failures are weird.

on my list to investigate further

Issue #1925, opened 2025-02-11
[BUG] clang-tidy warnings and errors


dmah42:
**Describe the bug**
ran clang-tidy across the codebase and collated the issues

let's work through them (i wanted to keep them somewhere handy)

[clang-tidy-issues.txt](https://github.com/user-attachments/files/18753892/clang-tidy-issues.txt)

dmah42:
updated: [clang-tidy-issues.txt](https://github.com/user-attachments/files/18773272/clang-tidy-issues.txt)

devtejasx:
> maybe we should start by running clang-tidy on CI (or even on pre-commit) before we add more fixes

It already runs on CI — `.github/workflows/clang-tidy-lint.yml`, on every push and pull request, `run-clang-tidy` over the whole compile database. What it cannot do is fail: `WarningsAsErrors` in `.clang-tidy` is empty, so clang-tidy exits 0 and the findings scroll past a green check mark.

Here is what the last green run on `main` ([30818016611](https://github.com/google/benchmark/actions/runs/30818016611)) actually reported — 1256 findings, top ten checks:

| check | count |
| --- | --- |
| `misc-include-cleaner` | 316 |
| `cppcoreguidelines-avoid-magic-numbers` / `readability-magic-numbers` | 289 |
| `cppcoreguidelines-pro-bounds-array-to-pointer-decay` | 132 |
| `misc-const-correctness` | 129 |
| `cppcoreguidelines-avoid-c-arrays` | 59 |
| `cppcoreguidelines-pro-bounds-pointer-arithmetic` | 42 |
| `misc-non-private-member-variables-in-classes` | 38 |
| `cppcoreguidelines-pro-type-vararg` | 35 |
| `cppcoreguidelines-avoid-non-const-global-variables` | 31 |
| `readability-function-cognitive-complexity` | 12 |

The long tail is small: below those, every remaining check is in single digits, and 13 of them fire fewer than three times.

That suggests a way to make the job mean something without a 1256-finding cleanup first: enforce per check. Put a check in `WarningsAsErrors` once its findings are at zero, and it can never come back; the rest keep being reported as they are now. The cheap ones to start with are at the bottom of the list rather than the top.

[...]

PR #1926, opened 2025-02-11
[clang-tidy] fix warning about decaying array to pointer

dmah42:
Part of #1925

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -49,7 +49,7 @@ cc_library(
         "include/benchmark/export.h",
     ],
     copts = select({
-        ":windows": [MSVC_COPTS],
+        ":windows": MSVC_COPTS,
         "//conditions:default": COPTS,
     }),
     defines = [
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -17,6 +17,10 @@ COPTS = [
     "-Werror=old-style-cast",
 ]
 
+MSVC_COPTS = [
+    "/std:c++17",
+]
+
 config_setting(
     name = "windows",
     constraint_values = ["@platforms//os:windows"],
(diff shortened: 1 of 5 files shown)


BUILD.bazel
```diff
@@ -17,6 +17,10 @@ COPTS = [
     "-Werror=old-style-cast",
 ]
 
+MSVC_COPTS = [
+    "/std:c++17",
+]
```
nicholasjng, line 22: fyi: If C++17 is the default now, it might make sense to add a .bazelrc file, since I'm also injecting the C++17 standard option by hand in the bindings build.

LebedevRI (dismissed):
Seems fine, but is that really all of these?
It would be good to actually enable the check that issued the warning as part of the PR.

BUILD.bazel
dmah42: why wouldn't it being in the bazel BUILD files be sufficient?

dmah42:
> Seems fine, but is that really all of these? It would be good to actually enable the check that issued the warning as part of the PR.

it's not all.  i'm trying to keep the PRs somewhat small.

the check is enabled, and is issuing a warning (not an error).  i want to get clang-tidy clean so then i'll make it a required bot.

LebedevRI:
> > Seems fine, but is that really all of these? It would be good to actually enable the check that issued the warning as part of the PR.
> 
> it's not all. i'm trying to keep the PRs somewhat small.

Sure, but i'm talking about the warnings from this specific check.
 
> the check is enabled, and is issuing a warning (not an error). i want to get clang-tidy clean so then i'll make it a required bot.

dmah42:
> > > Seems fine, but is that really all of these? It would be good to actually enable the check that issued the warning as part of the PR.
> > 
> > 
> > it's not all. i'm trying to keep the PRs somewhat small.
> 
> Sure, but i'm talking about the warnings from this specific check.

yes, i understood that.  it's not all of the fixes for this one check.  that would make the PR quite a bit bigger.

> 
> > the check is enabled, and is issuing a warning (not an error). i want to get clang-tidy clean so then i'll make it a required bot.

LebedevRI (approved):

PR #1927, opened 2025-02-11
[clang-tidy] use unique_ptr for benchmark registration

dmah42:
Part of #1925

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -631,7 +631,7 @@ class Counter {
   Counter(double v = 0., Flags f = kDefaults, OneK k = kIs1000)
       : value(v), flags(f), oneK(k) {}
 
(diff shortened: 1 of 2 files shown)


include/benchmark/benchmark.h
```diff
@@ -1349,36 +1351,28 @@ class LambdaBenchmark : public Benchmark {
  public:
   void Run(State& st) override { lambda_(st); }
 
- private:
   template <class OLambda>
```
LebedevRI, line 1354: Is this needed to make `std::make_unique<>()` work?

dmah42: yes.  there are clever workarounds (https://seanmiddleditch.github.io/enabling-make-unique-with-private-constructors/) but i didn't think it was necessary to go to such lengths.

LebedevRI (approved):

PR #1933, opened 2025-02-13
[clang-tidy] resolve some global clang-tidy issues

dmah42:
part of #1925

### DIFF
--- a/src/commandlineflags.h
+++ b/src/commandlineflags.h
@@ -11,28 +11,33 @@
 #define FLAG(name) FLAGS_##name
 
 // Macros for declaring flags.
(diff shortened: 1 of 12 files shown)


src/commandlineflags.h
```diff
@@ -11,28 +11,33 @@
 #define FLAG(name) FLAGS_##name
 
 // Macros for declaring flags.
+// NOLINTBEGIN
```
LebedevRI, line 14: IMHO it's best to specify the specific check in parenthesis, but maybe that doesn't work for in scoped case?

LebedevRI (dismissed):
Seems fine.

PR #1934, opened 2025-02-14
change setup and teardown callback type

### LINKED ISSUE #1872: [FR] Let SetUp and TearDown be std::function
My benchmarks are parameterized (not by numbers) and I register them dynamically using RegisterBenchmark for various combinations of parameters. Setup and Teardown need those parameters. Custom setup and teardown take function pointers so I can't pass functors or lambdas with the parameters curried in. 

Setup and teardown can be std::function

manually creating all my benchmarks, setup, and teardown functions for each set of parameters. but that is not sustainable

EfesX:
Custom Benchmark::Setup and Benchmark::Teardown take std::function. So it is possible to pass any functor there

Fixes #1872


include/benchmark/benchmark.h
```diff
@@ -1076,6 +1077,9 @@ typedef void(Function)(State&);
 // Each method returns "this" so that multiple method calls can
 // chained into one expression.
 class BENCHMARK_EXPORT Benchmark {
+  // Define alias of Setup/Teardown callback function type
+  using callback_function = std::function<void(const benchmark::State&)>;
```
LebedevRI, line 1081: shouldn't this be public?

test/benchmark_setup_teardown_test.cc
```diff
@@ -36,7 +39,7 @@ BENCHMARK(BM_with_setup)
     ->Arg(5)
     ->Arg(7)
     ->Iterations(100)
-    ->Setup(DoSetup1)
+    ->Setup(DoSetup1{})
```
LebedevRI, line 42: Is this nessesary?

test/benchmark_setup_teardown_test.cc
```diff
@@ -14,17 +14,20 @@ namespace singlethreaded {
 static int setup_call = 0;
 static int teardown_call = 0;
 }  // namespace singlethreaded
-static void DoSetup1(const benchmark::State& state) {
-  ++singlethreaded::setup_call;
 
-  // Setup/Teardown should never be called with any thread_idx != 0.
-  assert(state.thread_index() == 0);
-}
+struct DoSetup1 {
```
LebedevRI, line 18: Err, please just add a new test.

LebedevRI:
I do believe we could theoretically do this now that C++03 support is gone.
Also, #1609 comes to mind.

src/benchmark_register.cc
```diff
@@ -337,13 +337,13 @@ Benchmark* Benchmark::Apply(void (*custom_arguments)(Benchmark* benchmark)) {
   return this;
 }
 
-Benchmark* Benchmark::Setup(void (*setup)(const benchmark::State&)) {
+Benchmark* Benchmark::Setup(callback_function setup) {
```
LebedevRI, line 340: (I don't know)

I don't remember what the best practice is, should it be taken by `&&` and perhaps moved?

EfesX: It looks like you're right. && - a more appropriate way

EfesX:
> Also, #1609 comes to mind.

I could try to work on this in a separate PR

test/benchmark_setup_teardown_test.cc
EfesX: Actually, I do not know... Is there a need for a new test at all, or are the existing ones enough?

LebedevRI:
Where did the tests go?

EfesX:
> Where did the tests go?

The tests is back

src/benchmark_api_internal.h
```diff
@@ -68,7 +68,6 @@ class BenchmarkInstance {
   IterationCount iterations_;
   int threads_;  // Number of concurrent threads to us
 
-  typedef void (*callback_function)(const benchmark::State&);
   callback_function setup_ = nullptr;
   callback_function teardown_ = nullptr;
```
LebedevRI, line 72: `= nullptr` is probably no longer needed

include/benchmark/benchmark.h
```diff
@@ -1159,8 +1164,10 @@ class BENCHMARK_EXPORT Benchmark {
   // of threads, thread-index, benchmark arguments, etc.
   //
   // The callback must not be NULL or self-deleting.
```
LebedevRI, line 1166: outdated comment

include/benchmark/benchmark.h
```diff
@@ -1157,10 +1162,10 @@ class BENCHMARK_EXPORT Benchmark {
   //
   // The callback will be passed a State object, which includes the number
   // of threads, thread-index, benchmark arguments, etc.
-  //
-  // The callback must not be NULL or self-deleting.
-  Benchmark* Setup(void (*setup)(const benchmark::State&));
-  Benchmark* Teardown(void (*teardown)(const benchmark::State&));
+  Benchmark* Setup(callback_function&&);
+  Benchmark* Setup(const callback_function&);
```
dmah42, line 1166: why do we need two overloads?

test/benchmark_setup_teardown_cb_types_test.cc
```diff
@@ -0,0 +1,83 @@
...
+};
+
+void BM_DoSomething(benchmark::State& state) {
+  for (auto _ : state) {
+  }
+}
+
+void function(const benchmark::State& /*unused*/) {}
+
+int main(int argc, char** argv) {
```
dmah42, line 21: this test doesn't appear to be, well, testing anything.  can you please take a look at how the other tests are structured and see about making this more useful?

include/benchmark/benchmark.h
EfesX: `Benchmark* Setup(const callback_function&)` takes an object of type `callback_function` as input by value. 

For example, following code cannot be compiled without this overload:

```c++
callback_function func = [](const benchmark::State&){};
bm->Setup(func); // Compile Time Error

bm->Setup(std::move(func)); // OK
```

As an alternative, it is possible to use a template parameter. In that case, the need for two overloads disappears, but it seems that `BM_CHECK(setup != nullptr)` would no longer work.

```c++
template<class Callback>
Benchmark* Setup(Callback&& setup) {
//  BM_CHECK(setup != nullptr); // Compile Time Error
  setup_ = std::forward<Callback>(setup);
  return this;
}
```

EfesX: Should I use the overloads or is it better to use the template parameter?

dmah42: the overloads are fine.

dmah42 (approved):

LebedevRI (approved):
Seems reasonable-ish.

LebedevRI:
@EfesX thank you!

PR #1941, opened 2025-02-26
fix memory manager result bug

### LINKED ISSUE #1939: [BUG] Invalid Pointer Dereference due to std::vector auto-resizing in Memory Manager Result push_back
**Describe the bug**
When a `Memory Manager` class is registered, the results of each iteration are appended to an std::vector, and a pointer to the newly added element is used as a reference thereafter.
https://github.com/google/benchmark/blob/8d4fdd6e6e003867045e0bb3473b5b423818e4b7/src/benchmark_runner.cc#L434-L435

However, when the vector automatically resizes, these pointers become invalid, leading to corrupted statistical measurements.
The problem only occurs if the number of repetitions is greater than one.

**To reproduce**
1. Register a `Memory Manager` class 
2. Run any benchmark with at least 3 repetitions
3. User the Json reporter to see the corrupted memory related stats.

**Expected behavior**
Pointers to stored values should remain valid, or the implementation should ensure they are updated correctly when the vector reallocates.

**Additional context**
Possible fixes:
- Use `std::list` or instead of `std::vector` to avoid reallocation issues.
- Store indices instead of pointers if possible.
- Reserve `GetNumRepeats()` values to the `std::vector` to ensure no automatic growth is performed.

EfesX:
Fixes #1939

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -394,7 +394,8 @@ class MemoryManager {
         : num_allocs(0),
           max_bytes_used(0),
           total_allocated_bytes(TombstoneValue),
-          net_heap_growth(TombstoneValue) {}
+          net_heap_growth(TombstoneValue),
+          is_valid(false) {}
 
     // The number of allocations made in total between Start and Stop.
     int64_t num_allocs;
@@ -410,6 +411,8 @@ class MemoryManager {
     // ie., total_allocated_bytes - total_deallocated_bytes.
     // Init'ed to TombstoneValue if metric not available.
     int64_t net_heap_growth;
+
+    bool is_valid;
   };
 
   virtual ~MemoryManager() {}
@@ -1721,7 +1724,6 @@ class BENCHMARK_EXPORT BenchmarkReporter {
           complexity_n(0),
           report_big_o(false),
           report_rms(false),
-          memory_result(NULL),
           allocs_per_iter(0.0) {}
 
     std::string benchmark_name() const;
@@ -1777,7 +1779,7 @@ class BENCHMARK_EXPORT BenchmarkReporter {
     UserCounters counters;
 
     // Memory metrics.
-    const MemoryManager::Result* memory_result;
+    MemoryManager::Result memory_result;
     double allocs_per_iter;
   };
 
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -389,7 +389,7 @@ BENCHMARK_EXPORT void SetDefaultTimeUnit(TimeUnit unit);
 // benchmark.
 class MemoryManager {
  public:
-  static const int64_t TombstoneValue;
+  static constexpr int64_t TombstoneValue = std::numeric_limits<int64_t>::max();
 
   struct Result {
     Result()
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -306,7 +306,7 @@ namespace benchmark {
 class BenchmarkReporter;
 class State;
 
-typedef int64_t IterationCount;
+using IterationCount = int64_t;
 
 // Define alias of Setup/Teardown callback function type
 using callback_function = std::function<void(const benchmark::State&)>;
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -306,6 +306,8 @@ namespace benchmark {
 class BenchmarkReporter;
 class State;
 
+typedef int64_t IterationCount;
+
 // Define alias of Setup/Teardown callback function type
 using callback_function = std::function<void(const benchmark::State&)>;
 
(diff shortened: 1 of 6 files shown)


src/benchmark_runner.cc
```diff
@@ -508,13 +504,14 @@ void BenchmarkRunner::DoOneRepetition() {
   }
 
   // Produce memory measurements if requested.
-  MemoryManager::Result* memory_result = nullptr;
+  MemoryManager::Result memory_result;
   IterationCount memory_iterations = 0;
   if (memory_manager != nullptr) {
     // Only run a few iterations to reduce the impact of one-time
     // allocations in benchmarks that are not properly managed.
     memory_iterations = std::min<IterationCount>(16, iters);
```
LebedevRI, line 512: I would propose to change `(bool)memory_result.is_valid` with `memory_iterations > 0`.

EfesX: I put `memory_iteration` in `MemoryManager::Result` to use it in `JSONReporter::PrintRunData`

include/benchmark/benchmark.h
```diff
@@ -306,6 +306,8 @@ namespace benchmark {
 class BenchmarkReporter;
 class State;
 
+typedef int64_t IterationCount;
```
LebedevRI, line 309: `using IterationCount = int64_t`

src/json_reporter.cc
```diff
@@ -306,8 +306,8 @@ void JSONReporter::PrintRunData(Run const& run) {
     out << ",\n" << indent << FormatKV(c.first, c.second);
   }
 
-  if (run.memory_result != nullptr) {
-    const MemoryManager::Result memory_result = *run.memory_result;
+  if (run.memory_result.memory_iterations > 0) {
+    const MemoryManager::Result& memory_result{run.memory_result};
```
LebedevRI, line 310: `    const MemoryManager::Result& memory_result = run.memory_result;`

LebedevRI:
Yep, something like this should work i suppose.

include/benchmark/benchmark.h
```diff
@@ -410,6 +413,8 @@ class MemoryManager {
     // ie., total_allocated_bytes - total_deallocated_bytes.
     // Init'ed to TombstoneValue if metric not available.
     int64_t net_heap_growth;
+
+    IterationCount memory_iterations;
```
LebedevRI, line 417: Ah, i thought we already had `memory_iterations` in `Run`, but apparently not.

test/memory_results_gtest.cc
```diff
@@ -0,0 +1,101 @@
+#include <vector>
+
```
LebedevRI, line 2: Just to check: you have verified that this test fails on its own, without the fix, correct?

EfesX: Of course. I was able to confirm and reproduce the issue in this test initially.

After that, when I fixed the issue, the test had to be changed slightly.

LebedevRI:
Looks like more casts are needed  :)

test/memory_results_gtest.cc
```diff
@@ -0,0 +1,101 @@
...
+TEST_F(MemoryResultsTest, ResultsTest) {
+  counts::reset();
+  MemoryManager* mm = new TestMemoryManager;
+  RegisterMemoryManager(mm);
+
+  RunSpecifiedBenchmarks(&reporter);
+  EXPECT_EQ(reporter.store.size(), N_REPETITIONS);
+
+  for (int64_t i = 0; i < static_cast<int64_t>(reporter.store.size()); i++) {
```
dmah42, line 93: i suspect having `i` be size_t might be the way to go here.

LebedevRI:
Hmm, those msvc failure look weird. I don't think this PR can cause them.

EfesX:
> Hmm, those msvc failure look weird. I don't think this PR can cause them.

I reproduced the failure on my Windows computer. I don’t understand why, but it seems like the failure is caused by my PR. 

e1eaa52 might resolve the issue.

test/memory_results_gtest.cc
```diff
@@ -0,0 +1,102 @@
...
+
+using benchmark::ClearRegisteredBenchmarks;
+using benchmark::ConsoleReporter;
+using benchmark::MemoryManager;
+using benchmark::RegisterBenchmark;
+using benchmark::RunSpecifiedBenchmarks;
+using benchmark::State;
+using benchmark::internal::Benchmark;
+
+namespace counts {
```
dmah42, line 18: this would also work as an unnamed namespace, fyi.

test/memory_results_gtest.cc
```diff
@@ -0,0 +1,102 @@
...
+};
+
+TEST_F(MemoryResultsTest, NoMMTest) {
+  RunSpecifiedBenchmarks(&reporter);
+  EXPECT_EQ(reporter.store.size(), 0);
+}
+
+TEST_F(MemoryResultsTest, ResultsTest) {
+  counts::reset();
+  MemoryManager* mm = new TestMemoryManager;
```
dmah42, line 87: auto mm = std::make_unique<TestMemoryManager>();

at the very least you want () after `TestMemoryManager`.

LebedevRI (approved):

LebedevRI:
@EfesX thank you!

Issue #1943, opened 2025-03-08
C++14 is required for the example


jun-shibata:
It's a small thing, but the compile option `-std=c++11` in this example should be updated.

https://github.com/google/benchmark?tab=readme-ov-file#basic-usage

The error occurs because `make_unique` is used.

```script
benchmark/include/benchmark/benchmark.h:1371:12: error: 'make_unique' is not a member of 'std'
 1371 |       std::make_unique<internal::FunctionBenchmark>(name, fn));
```

Please fix the compilation command.

```diff
# Example on linux after running the build steps above. Assumes the
# `benchmark` and `build` directories are under the current directory.
- $ g++ mybenchmark.cc -std=c++11 -isystem benchmark/include \
+ $ g++ mybenchmark.cc -std=c++14 -isystem benchmark/include \
  -Lbenchmark/build/src -lbenchmark -lpthread -o mybenchmark
```

LebedevRI:
Fixed by https://github.com/google/benchmark/pull/1945.

jun-shibata:
Thank you!

PR #1945, opened 2025-03-08
Compilation example was wrong. Fixed standard

maxvdec:
The README.md file put a compile example with the flag `-std=c++11`, and because of the use of new standard library functions, the flag in the compile example had to be changed to `-std=c++14`, this solving issue: #1943

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ be under the build directory you created.
 ```bash
 # Example on linux after running the build steps above. Assumes the
 # `benchmark` and `build` directories are under the current directory.
-$ g++ mybenchmark.cc -std=c++11 -isystem benchmark/include \
+$ g++ mybenchmark.cc -std=c++14 -isystem benchmark/include \
   -Lbenchmark/build/src -lbenchmark -lpthread -o mybenchmark
 ```
 


README.md
```diff
@@ -190,7 +190,7 @@ be under the build directory you created.
 ```bash
 # Example on linux after running the build steps above. Assumes the
 # `benchmark` and `build` directories are under the current directory.
-$ g++ mybenchmark.cc -std=c++11 -isystem benchmark/include \
+$ g++ mybenchmark.cc -std=c++14 -isystem benchmark/include \
```
dmah42, line 193: this should be c++17 to be up to date with the library.

LebedevRI:
CLA check still seems to be having issues.

dmah42:
forced a rescan

README.md
LebedevRI: Are there C++17 features in the header? If so, docs are outdated now too.
By the way, with cxx03 test gone, i guess we want to add a cxx14 test,
to ensure that the header can actually be used with the ${version} standard.

LebedevRI: Aha. Then it certainly might be great to "guard" against that with a test.

LebedevRI (approved):
This is an improvement over the current incorrect status-quo.

PR #1946, opened 2025-03-12
CI: unbreak MSan job

LebedevRI:


.github/workflows/sanitizer.yml
```diff
@@ -50,14 +50,17 @@ jobs:
...
+      run: |
+        sudo apt update
+        sudo apt install clang-19
 
     - name: configure clang
       run: |
+        sudo ln -s /usr/bin/clang-19 /usr/local/bin/clang
+        sudo ln -s /usr/bin/clang++-19 /usr/local/bin/clang++
+        sudo ln -s /usr/bin/clang /usr/local/bin/cc
+        sudo ln -s /usr/bin/clang++ /usr/local/bin/c++
```
dmah42, line 63: you should be able to just `echo "CC=/usr/bin/clang-19" >> $GITHUB_ENV` instead

LebedevRI: SomeI guess it doesn't matter for us, but in general, doing just that might not be enough in some cases.

dmah42:
fantastic. thank you!

PR #1955, opened 2025-03-24
Threading api v3

krzikalla:
As requested, here comes a new pull request to enable the user to perform multi-threading benchmarks using other threading APIs.

This one takes another approach, which requires only a few changes in the existing benchmark code. 
Personally I like this approach the most, as it decouples the actual benchmark code and the user threading code. In addition, there is no special `thread_state` anymore.

Remark: line 305 of `benchmark_runner.cc`:  ` manager->WaitForAllThreads();` has been removed. I don't think, that this causes any issues, as all threads join directly afterwards (all tests run fine). In fact I think, that some more things regarding threads in `benchmark_runner.cc` could be refactored. E.g. I don't understand this code:
```
  // Acquire the measurements/counters from the manager, UNDER THE LOCK!
  {
    i.results = manager->results;
  }
```
IMHO the mutex is not needed there, as all spawned threads have already been joined.


LebedevRI:
Ok, now that's completely different from the original PR.
This is very straight-forward and reasonable. I like this.

What about the second part of that PR, `ManualThreading()`?
I think that feature is useful too, but like i said,
let's have separate PR's for these.

test/manual_threading_test.cc
```diff
@@ -0,0 +1,166 @@
...
+    })
+    ->Threads(2)
+    ->MeasureProcessCPUTime()
+    ->UseManualTime();
+
+// ========================================================================= //
+// ---------------------------- TEST CASES END ----------------------------- //
+// ========================================================================= //
+
+int main(int argc, char* argv[]) { RunOutputTests(argc, argv); }
```
LebedevRI, line 166: Do we want to actually test the output of this test?

test/manual_threading_test.cc
```diff
@@ -0,0 +1,166 @@
...
+    fn(0);
+
+    for (std::thread& thread : pool) {
+      thread.join();
+    }
+  }
+
+ private:
+  std::vector<std::thread> pool;
+};
```
LebedevRI, line 63: You'd want to actually test that the library is actually using your custom runner,
and not the default one. (i.e. add some file-level static variables, increment them in `RunThreads`,
and check them afterwards)

test/manual_threading_test.cc
krzikalla: Actually I had copied this from internal_threading_test.cc. It's now gone.

dmah42 (approved):

krzikalla:
Does anyone know, what these failed windows tests are all about? 
Some windows tests run successfully, others exit with exit code 0xc0000135 (which is related to the .NET framework(???)). 
Then again, when a run fails, it fails at my newly introduced test (`manual_threading_test`), so it somehow seems to be related to my changes.

src/benchmark_api_internal.h
```diff
@@ -41,6 +41,10 @@ class BenchmarkInstance {
   int threads() const { return threads_; }
   void Setup() const;
   void Teardown() const;
+  std::unique_ptr<ThreadRunnerBase> GetUserThreadRunner() const {
+    return benchmark_.threadrunner_ ? benchmark_.threadrunner_(threads_)
+                                    : nullptr;
+  }
```
LebedevRI, line 47: I think it would be better if `GetUserThreadRunner` would return the factory,
and we'd never to have to second-guess it's output later on,
and explcitly call it with the thread count, once.
```
  ??? GetThreadRunner() const {
    return benchmark_.threadrunner_ ? benchmark_.threadrunner_
                                    : [](int n){ return std::make_unique<ThreadRunnerDefault>(n); };
  }
```

test/manual_threading_test.cc
```diff
@@ -0,0 +1,174 @@
...
+
+void BM_ManualThreading(benchmark::State& state) {
+  for (auto _ : state) {
+    MyBusySpinwait();
+    state.SetIterationTime(time_frame_in_sec);
+  }
+  state.counters["invtime"] =
+      benchmark::Counter{1, benchmark::Counter::kIsRate};
+}
```
LebedevRI, line 75: This can also be in `namespace {`

src/benchmark_runner.cc
```diff
@@ -289,23 +318,10 @@ BenchmarkRunner::IterationResults BenchmarkRunner::DoNIterations() {
...
-                           /*profiler_manager=*/nullptr);
-  }
-  // And run one thread here directly.
-  // (If we were asked to run just one thread, we don't create new threads.)
-  // Yes, we need to do this here *after* we start the separate threads.
-  RunInThread(&b, iters, 0, manager.get(), perf_counters_measurement_ptr,
-              /*profiler_manager=*/nullptr);
-
-  // The main thread has finished. Now let's wait for the other threads.
-  manager->WaitForAllThreads();
```
LebedevRI, line 305: I'm mildly uncomfortable that we are introducing non-obvious threading-related changes
in an otherwise straight-forward refactoring PR.
Would you be open to dropping this call in a preparatory PR?

krzikalla: I can do this. Should I do a wider refactoring? E.g. I still don't understand this `// Acquire the measurements/counters from the manager, UNDER THE LOCK!` comment some lines below (and the lock itself). I have searched the blame for the root of this comment, but didn't find something until 2018. There was some mentioning of benchmarks running interleaved, but even then - all variables in question are owned by exactly one BenchmarkRunner.

LebedevRI: I think what we can "safely" do is:
1. Drop `ThreadManager::WaitForAllThreads`
2. Drop `ThreadManager::end_condition_`
3. Drop `ThreadManager::end_cond_mutex_`

I wouldn't touch anything else there, it does not seem nessesary for this change...

LebedevRI:
Rebase pls

test/manual_threading_test.cc
```diff
@@ -0,0 +1,174 @@
...
+// Creation of threads is done before the start of the measurement,
+// joining after the finish of the measurement.
+void BM_ManualThreading(benchmark::State& state) {
+  for (auto _ : state) {
+    MyBusySpinwait();
+    state.SetIterationTime(time_frame_in_sec);
+  }
+  state.counters["invtime"] =
+      benchmark::Counter{1, benchmark::Counter::kIsRate};
+}
```
LebedevRI, line 75: This can be in the namespace

docs/user_guide.md
```diff
@@ -863,6 +863,42 @@ BENCHMARK(BM_test)->Range(8, 8<<10)->UseRealTime();
 
 Without `UseRealTime`, CPU time is used by default.
 
+### Manual Multithreaded Benchmarks
```
LebedevRI, line 866: (nit) Please linewrap like the rest of this file, clang-format the snippets

src/benchmark_runner.cc
```diff
@@ -182,6 +183,39 @@ IterationCount ComputeIters(const benchmark::internal::BenchmarkInstance& b,
...
+    for (std::thread& thread : pool) {
+      thread.join();
+    }
+  }
+
+ private:
+  std::vector<std::thread> pool;
+};
+
+template <class T>
```
LebedevRI, line 211: `T` here is always `threadrunner_factory` is it not?

include/benchmark/benchmark.h
```diff
@@ -1093,6 +1093,16 @@ inline BENCHMARK_ALWAYS_INLINE State::StateIterator State::end() {
   return StateIterator();
 }
 
+// Base class for user-defined multi-threading
+struct ThreadRunnerBase {
```
LebedevRI, line 1097: Let's try
```
struct BENCHMARK_EXPORT ThreadRunnerBase {
```

src/benchmark_runner.cc
```diff
@@ -182,6 +183,38 @@ IterationCount ComputeIters(const benchmark::internal::BenchmarkInstance& b,
   return iters_or_time.iters;
 }
 
+class BENCHMARK_EXPORT ThreadRunnerDefault : public ThreadRunnerBase {
```
LebedevRI, line 186: No, i really did mean only on `ThreadRunnerBase`.

test/CMakeLists.txt
```diff
@@ -189,6 +189,9 @@ benchmark_add_test(NAME perf_counters_test COMMAND perf_counters_test --benchmar
 compile_output_test(internal_threading_test)
 benchmark_add_test(NAME internal_threading_test COMMAND internal_threading_test --benchmark_min_time=0.01s)
 
+compile_output_test(manual_threading_test)
+add_test(NAME manual_threading_test COMMAND manual_threading_test --benchmark_min_time=0.01s)
```
LebedevRI, line 193: Ugh. Please follow the pattern of the neighbouring code :)
Use `benchmark_add_test()`, not `add_test()`, and drop `BENCHMARK_EXPORT`.

LebedevRI:
Awesome. I will need to sleep on this before doing further review passes, but this does look reasonable.

include/benchmark/benchmark.h
```diff
@@ -1093,6 +1093,16 @@ inline BENCHMARK_ALWAYS_INLINE State::StateIterator State::end() {
...
+  virtual ~ThreadRunnerBase() {}
+  virtual void RunThreads(const std::function<void(int)>& fn) = 0;
+};
+
+// Define alias of ThreadRunner factory function type
+using threadrunner_factory =
```
dmah42, line 1103: do you intend this to be a type that is part of the public API?  if so, it should be `ThreadRunnerFactory` but i suspect it's just here to save typing, in which case we shouldn't have the alias here.

test/manual_threading_test.cc
```diff
@@ -0,0 +1,174 @@
...
+
+}  // end namespace
+
+BENCHMARK(BM_ManualThreading)
+    ->Iterations(1)
+    ->ThreadRunner([](int num_threads) {
+      return std::make_unique<ManualThreadRunner>(num_threads);
```
dmah42, line 80: is this ever going to be anything but

```
[](int nt) { return std::make_unique<T>(nt); }
```

?

can we maybe simplify the boilerplate and templatise it?  something like:

```
->ThreadRunner<ManualThreadRunner>();
```
?

include/benchmark/benchmark.h
krzikalla: `callback_function` is the same kind of alias. I am unsure, if `callback_function` is part of the public API.

test/manual_threading_test.cc
krzikalla: I thought about it, but decided against it for two reasons:
1. API consistency: gbenchmark doesn't use this kind of templates anywhere else.
2. Custom arguments: A template would restrict the `ManualThreadRunner` constructor to one argument (`nt`). A factory function makes it easier to inject more arguments to such a class (even if those arguments will be globals).

include/benchmark/benchmark.h
dmah42: that looks like a style oversight: https://google.github.io/styleguide/cppguide.html#Aliases

i don't think we expect clients to type this type anywhere (?) so it shouldn't be in the public API.

test/manual_threading_test.cc
dmah42: makes sense. thanks!

include/benchmark/benchmark.h
krzikalla: Moved it to the internal namespace, which was easier than expected :-)

dmah42:
i'm happy with this but i'll let @LebedevRI make the final decision and merge :)

thank you so much @krzikalla

LebedevRI (approved):
This is probably fine.

Test is rather too bare-bones,
perhaps we should be checking the output?

LebedevRI:
@krzikalla thank you!

Any thoughts on the second part of the original PR?

PR #1960, opened 2025-03-26
Use the top-level ::benchmark namespace to resolve `make_unique`

### LINKED ISSUE #1959: [BUG] BENCHMARK macros do not reference full namespace for make_unique
**Describe the bug**
In https://github.com/google/benchmark/pull/1948 the various `BENCHMARK` macros were changed to something like:
```
#define BENCHMARK(...)                                                \
  BENCHMARK_PRIVATE_DECLARE(_benchmark_) =                            \
      (::benchmark::internal::RegisterBenchmarkInternal(              \
          benchmark::internal::make_unique<                           \
              ::benchmark::internal::FunctionBenchmark>(#__VA_ARGS__, \
                                                        __VA_ARGS__)))
```

This makes a relatively namespace lookup for `benchmark::internal::make_unique` even though the macros can be invoked from any namespace.

If the user code defined their own `benchmark` namespace, then the compiler fails finding `make_unique`. This produces errors in the Apache Arrow continuous benchmarking, for example [here](https://buildkite.com/apache-arrow/arrow-bci-benchmark-on-amd64-c6a-4xlarge-linux/builds/937#0195d209-e346-4c84-8895-b82b61d1a899/95-2687).

dmah42:
Fixes #1959

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1391,7 +1391,8 @@ class LambdaBenchmark : public Benchmark {
 inline internal::Benchmark* RegisterBenchmark(const std::string& name,
                                               internal::Function* fn) {
   return internal::RegisterBenchmarkInternal(
-      benchmark::internal::make_unique<internal::FunctionBenchmark>(name, fn));
+      ::benchmark::internal::make_unique<internal::FunctionBenchmark>(name,
+                                                                      fn));
 }
 
 template <class Lambda>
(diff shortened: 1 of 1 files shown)


LebedevRI:
Yeah, i was wondering about that when writing that patch. Makes sense.

LebedevRI (approved):

LebedevRI:
@dmah42 @pitrou thank you!

pitrou:
Thanks for the quick fix :)

Issue #1767, opened 2024-03-19
[FR] Reuse benchmark functions with arbitrary fixture templates


PhilipDeegan:
**Is your feature request related to a problem? Please describe.**

Currently, looking at the docs, it does not appear like we can resuse benchmark functions per fixture template, which I find somewhat strange.

**Describe the solution you'd like**

something like this

```
template<typename T>
class MyFixture : public benchmark::Fixture {};

BENCHMARK_TEMPLATE_F(MyFixture, Test)(benchmark::State& st) {
   for (auto _ : st) {
     ...
  }
}

BENCHMARK_REGISTER_F(MyFixture, Test, double);
BENCHMARK_REGISTER_F(MyFixture, Test, int);
```

**Describe alternatives you've considered**
the current function per template type is an alternative, but seems redundant.

**Additional context**
n/a

PR #1967, opened 2025-04-11
Templated fixture methods

krzikalla:
With this PR it becomes possible to define a method once for a fixture template and then reuse it for several template arguments.
This also fixes https://github.com/google/benchmark/issues/1767#issue-2194848517.


dmah42 (approved):

dmah42:
thank you!

PR #1968, opened 2025-04-11
Macro fix

krzikalla:
There is a bug in the previous PR. If you use the macro across several translation units, then only one benchmark might get registered, since BENCHMARK_PRIVATE_NAME generates unique names only per translation unit. This PR should fix this.

I am unsure if it would be a good idea to declare the instantiation class (`class UniqueName ...`) in an unnamed namespace.


dmah42 (approved):

dmah42:
oh well spotted.

Issue #1962, opened 2025-03-28
[BUG] Empty benchmarks cause endless loops


krzikalla:
**Describe the bug**
```
void MyBenchmark(benchmark::State& state)
{
  if (!global_condition)
  {
    return;
  }
  for (auto _ : state)
  {
    ...
  }
}
```
The above benchmark causes an endless loop in  `BenchmarkRunner::DoOneRepetition` in the current version, if `global_condition == false`.
I can only tell, that in version 1.5.x such benchmarks were simply ignored (not even output was generated for those).

**Expected behavior**
I'd like to have the behavior of version 1.5.x: in case of a premature return (i.e. `i.iters==0`) just ignore the benchmark.

krzikalla:
I guess I could track down the bug to the following change:

https://github.com/google/benchmark/blob/8039b4030795b1c9b8cedb78e3a2a6fb89574b6e/src/benchmark_runner.cc#L240C15-L240C20

vs.

https://github.com/google/benchmark/blob/afa23b7699c17f1e26c88cbf95257b20d78d6247/src/benchmark_runner.cc#L325

Back in 1.5.1 the pre-computed iters was used. Nowadays the actual number of iterations are used.

LebedevRI:
I would say that
https://github.com/google/benchmark/blob/2918a094b0068f28e4a6723406252855d57527b3/src/benchmark_runner.cc#L144-L145
```diff
-  BM_CHECK(st.skipped() || st.iterations() >= st.max_iterations)
-    << "Benchmark returned before State::KeepRunning() returned false!";
+ if(!(st.skipped() || st.iterations() >= st.max_iterations)) {
+   st.SkipWithError("FOUND BUG IN YOUR BENCHMARK")
+ }
```

dmah42:
but it's not a user error, it's intentional.  ie a global flag to not run the benchmark.  the user should be using `st.SkipWithMessage` to skip the benchmark, i think?

LebedevRI:
> but it's not a user error, it's intentional. ie a global flag to not run the benchmark.

But it is a user error, because it violates the defined contract:
"either run the loop, or explicitly skip the benchmark."

> the user should be using `st.SkipWithMessage` to skip the benchmark, i think?

Yes.

krzikalla:
> But it is a user error, because it violates the defined contract:
> "either run the loop, or explicitly skip the benchmark."

Is this contract documented? I haven't found anything about `SkipWithMessage` at all.

[...]

PR #1976, opened 2025-05-01
Improve user UX on empty benchmarks

LebedevRI:
Fixes https://github.com/google/benchmark/issues/1962

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -142,8 +142,11 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
 
   State st = b->Run(iters, thread_id, &timer, manager,
                     perf_counters_measurement, profiler_manager_);
-  BM_CHECK(st.skipped() || st.iterations() >= st.max_iterations)
-      << "Benchmark returned before State::KeepRunning() returned false!";
+  if (!(st.skipped() || st.iterations() >= st.max_iterations)) {
+    st.SkipWithError(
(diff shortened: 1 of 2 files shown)


src/benchmark_runner.cc
```diff
@@ -142,8 +142,9 @@ void RunInThread(const BenchmarkInstance* b, IterationCount iters,
 
   State st = b->Run(iters, thread_id, &timer, manager,
                     perf_counters_measurement, profiler_manager_);
-  BM_CHECK(st.skipped() || st.iterations() >= st.max_iterations)
-      << "Benchmark returned before State::KeepRunning() returned false!";
+  if (!(st.skipped() || st.iterations() >= st.max_iterations)) {
+    st.SkipWithError("FOUND BUG IN YOUR BENCHMARK");
```
dmah42, line 146: it might not be a bug... i mean, i think the user should call "SkipWithMessage" which is a bug if they're not, but maybe that's the message we need to put here.

LebedevRI: Yes, the message is very much a placeholder, please feel free to suggest better verbiage.

dmah42: "The benchmark didn't run, nor was it explicitly skipped.  Please call 'SkipWithXXX` in your benchmark as appropriate."

or we could just output "WARNING: Benchmark did not run and it's unclear if this is intentional."

dmah42 (approved):

LebedevRI:
@dmah42 thank you!

Issue #461, opened 2017-10-17
Different results, depending on wether certain benchmarks are enabled.


DenisYaroshevskiy:
Hi!

If I enable set_union_unbalanced benchmark, I get:

`
set_union_benchmark<std_set_union>              9301 ns       9238 ns      74616
set_union_benchmark<set_union_unbalanced>       9586 ns       9550 ns      76047
`
However, if it's disabled (commented out), I get:

`
set_union_benchmark<std_set_union>       3376 ns       3364 ns     207403
`

Results are consistent and depend a lot on partition_point_biased code.
Seems like a bug.
code: https://gist.github.com/DenisYaroshevskiy/604f1e331b384579b2a4fb5eac3ac8ef

System info:
<img width="504" alt="2017-10-17 18 31 07" src="https://user-images.githubusercontent.com/11256077/31674026-650a70ba-b369-11e7-8580-a186b193406e.png">

clang --version
Apple LLVM version 9.0.0 (clang-900.0.38)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

DenisYaroshevskiy:
Upd: Seems like the issue goes away if I use BENCHMARK instead of BENCHMARK_TEMPLATE.
Is this even possible?

dominichamon:
```
#define BENCHMARK(n)                                     \
  BENCHMARK_PRIVATE_DECLARE(n) =                         \
      (::benchmark::internal::RegisterBenchmarkInternal( \
          new ::benchmark::internal::FunctionBenchmark(#n, n)))
```

```
#define BENCHMARK_TEMPLATE1(n, a)                        \
  BENCHMARK_PRIVATE_DECLARE(n) =                         \
      (::benchmark::internal::RegisterBenchmarkInternal( \
          new ::benchmark::internal::FunctionBenchmark(#n "<" #a ">", n<a>)))
```

I don't think that replacing one with the other would make any difference.

can you put the files into a gist so it's easier to take a look at them?

DenisYaroshevskiy:
That makes sence. Didn't reproduce since though.

Here is gist, I didn't know about it, don't use github that often: https://gist.github.com/DenisYaroshevskiy/c1c28e3fd95f8a7eb53ddab40dc8464f

dominichamon:
```
[...]

PR #1977, opened 2025-05-01
Detect and report ASLR status

LebedevRI:
Theoretically, we could just disable ASLR in `main()`, but that seems to be disallowed by default
by some other security features.

Refs. https://github.com/google/benchmark/issues/461

### DIFF
--- a/docs/reducing_variance.md
+++ b/docs/reducing_variance.md
@@ -50,15 +50,18 @@ If you see this error:
 you might want to disable the ASLR security hardening feature while running the
 benchmark.
 
-To globally disable ASLR, run
+To globally disable ASLR on Linux, run
 ```
 echo 0 > /proc/sys/kernel/randomize_va_space
 ```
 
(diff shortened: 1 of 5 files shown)


docs/reducing_variance.md
```diff
@@ -39,6 +39,27 @@ The benchmarks you subsequently run will have less variance.
...
+If you see this error:
+
+```
+***WARNING*** ASLR is enabled, the results may have unreproducible noise in them.
+```
+
+you might want to disable the ASLR security hardening feature while running the
+benchmark.
+
+To globally disable ASLR, run
```
dmah42, line 53: these instructions only apply to linux i think?  windows also has ASLR available.

dmah42: (i assume you don't care given the change in the code, but it might be worth mentioning in the documentation)

dmah42 (approved):

dmah42 (approved):

PR #1978, opened 2025-05-01
Automatic ASLR disablement

LebedevRI:
While ASLR is a useful security hardening feature, it introduces unreproducible noise into benchmarks, and we really really really don't want any noise,
especially easily avoidable one.

Unless prevented by some other security hardening, we can disable ASLR for the current process,
and restart it, thus eliminating this noise.

Fixes https://github.com/google/benchmark/issues/461

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -26,6 +26,10 @@
 #include <unistd.h>
 #endif
 
+#ifdef BENCHMARK_OS_LINUX
+#include <sys/personality.h>
+#endif
+
 #include <algorithm>
 #include <atomic>
 #include <condition_variable>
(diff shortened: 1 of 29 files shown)


src/benchmark.cc
```diff
@@ -813,6 +817,29 @@ int InitializeStreams() {
 
 }  // end namespace internal
 
+bool MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
+  // On e.g. Hexagon simulator, argv may be NULL.
+  if (!argv) return false;
+
+#ifdef BENCHMARK_OS_LINUX
+  const auto curr_personality = personality(0xffffffff);
+  BM_CHECK(curr_personality != -1)
```
dmah42, line 826: this will only fail in DEBUG. should it be a stronger check?

LebedevRI: `man 2 personality` says:
```
       If persona is not 0xffffffff, then personality() sets the caller's execution domain to the value specified by persona.  Specifying persona as 0xffffffff provides a way of retrieving the current persona without changing it.
```
```
RETURN VALUE
       On success, the previous persona is returned.  On error, -1 is returned, and errno is set to indicate the error.

ERRORS
       EINVAL The kernel was unable to change the personality.
```
So generally it shouldn't just fail. But sure, let's do that.

dmah42 (dismissed):

dmah42 (approved):

dmah42:
that's great.  thank you!

PR #1981, opened 2025-05-11
Move cmake export under install condition

MBkkt:
Otherwise build with libc++ from source fails, at least in our case. 
Maybe it makes some sense for others


LebedevRI (changes requested):
This is invalid and will not be merged.

LebedevRI:
Let me rephrase. That `export` has no relation to the install step,
it is intentionally not guarded by `BENCHMARK_ENABLE_INSTALL`.
It outputs into build directory, so that the library can be consumed
by another CMake without the need for the install step.
This is very standard and is working correctly.

So it would be helpful to re-analyze whatever issue you are seeing,
this isn't the underlying problem for it.

Issue #1982, opened 2025-05-11
[BUG] BENCHMARK_ENABLE_WERROR: implicit conversion loses integer precision


SunBlack:
**Describe the bug**

Tried to push the latest release to vcpkg (see https://github.com/microsoft/vcpkg/pull/45467), but the android builds are failing

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Android
  - Compiler and version: arm64/x64 (Clang 18.0.3 on x64 as far as I can see)

**To reproduce**
See vcpkg portfile

**Expected behavior**
Compiles, but it failes due to:
```
/mnt/vcpkg-ci/b/benchmark/src/v1.9.3-f02919e468.clean/src/benchmark.cc:836:52: error: implicit conversion loses integer precision: 'unsigned long' to 'unsigned int' [-Werror,-Wshorten-64-to-32]
  835 |   const auto prev_personality = personality(
      |                                 ~~~~~~~~~~~
  836 |       static_cast<unsigned long>(curr_personality) | ADDR_NO_RANDOMIZE);
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
```

The line was added last week within #1978. As `BENCHMARK_ENABLE_WERROR` is enabled by default, this should be fixed.

LebedevRI:
`personality(2)` says:
```
SYNOPSIS
       #include <sys/personality.h>

       int personality(unsigned long persona);
```
but on that os, the argument is narrower?

LebedevRI:
Does https://github.com/google/benchmark/pull/1983 help?

PR #1983, opened 2025-05-11
`MaybeReenterWithoutASLR()`: be more cautious about argument types

LebedevRI:
It seems, on android the argument is narrower than on linux. Let's try to support that while not introducing any explicit lossy casts.

Fixes https://github.com/google/benchmark/issues/1982

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -815,6 +815,12 @@ int InitializeStreams() {
   return 0;
 }
 
+template <typename T>
+std::make_unsigned_t<T> get_as_unsigned(T v) {
+  using UnsignedT = std::make_unsigned_t<T>;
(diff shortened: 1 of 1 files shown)


SunBlack:
I've experimented with it a bit now, but haven't really been successful.

The bugfix itself works (see vcpkg). I then tried to find out somehow where the issue really comes from, but debugging against the vcpkg-CI just takes a long time and locally CMake ignores the request to cross-compile with Android. But it's not worth investing any more time for me now.

In case someone has more experience and wants to play with it: The Android Dockerfile is available [here](https://github.com/microsoft/vcpkg/tree/master/scripts/azure-pipelines/android) (the image `vcpkgandroidwus.azurecr.io/vcpkg-android:2025-04-08` can unfortunately not be used without authentication).

Issue #1984, opened 2025-05-16
[BUG] `MaybeReenterWithoutASLR()` causes infinite loop


chfast:
**Describe the bug**

The invocation of `MaybeReenterWithoutASLR()` at the beginning of `main()` cause infinite loop. I'm guessing the process is constantly executing `execv()`.

**System**

This actually happens in a very specific environment: in a x86-64 docker for the 32-bit build. I use similar docker images for 64-bits and they all work fine.

**To reproduce**

**Screenshots**

When running with gdb, these logs show up constantly:
```
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 2428 is executing new program: /home/builder/build/bin/evmone-bench
[Thread debugging using libthread_db enabled]
```

**Additional context**

LebedevRI:
So for that to happen,
1. `ADDR_NO_RANDOMIZE` personality needs to be *NOT* set
2. We detect that
3. We set `ADDR_NO_RANDOMIZE` personality
4. `personality()` does not return `-1`, as it's supposed to:
```
[...]

PR #1985, opened 2025-05-16
Fix `MaybeReenterWithoutASLR()` in docker

chfast:
In some docker configurations the `personality()` function may return inconsistent results. Double check if the persona has been updated before reentering, otherwise we risk infinite loop.

Fixes https://github.com/google/benchmark/issues/1984.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -845,6 +845,13 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
   // Have we failed to change the personality? That may happen.
   if (prev_personality == -1) return;
 
+  // Make sure the parsona has been updated with the no-ASLR flag,
+  // otherwise we will try to reenter infinitely.
+  // This seems impossible, but can happen in some docker configurations.
+  const auto new_personality = personality(0xffffffff);
+  if ((internal::get_as_unsigned(new_personality) & ADDR_NO_RANDOMIZE) == 0)
+    return;
+
   execv(argv[0], argv);
   // The exec() functions return only if an error has occurred,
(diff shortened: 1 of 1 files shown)


src/benchmark.cc
```diff
@@ -845,6 +845,13 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
   // Have we failed to change the personality? That may happen.
   if (prev_personality == -1) return;
 
+  // Make sure the parsona has been updated,
+  // otherwise we will try to reenter infinitely.
+  // This seems impossible, but can happen in some docker configurations.
+  const auto new_personality = personality(0xffffffff);
+  if (internal::get_as_unsigned(new_personality) != proposed_personality)
```
LebedevRI, line 852: Hm, i would think we should just check that `ADDR_NO_RANDOMIZE` flag is now set.
Mainly, i don't think it is _guaranteed_ that setting one flag wouldn't set/unset something else.

chfast: Changed.

src/benchmark.cc
```diff
@@ -845,6 +845,13 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
   // Have we failed to change the personality? That may happen.
   if (prev_personality == -1) return;
 
+  // Make sure the parsona has been updated with the no-ADLR flag,
```
LebedevRI, line 848: s/ADLR/ASLR/

LebedevRI (approved):
LGTM, thank you!

dmah42:
nice catch

PR #1987, opened 2025-05-19
fix: resolve editable installation issue for python package

senyai:
Before this `pip install -e .` produced ` error: [Errno 2] No such file or directory: '/tmp/tmpi02jtn4q.build-lib/google_benchmark/_benchmark.pyi'`

closes #1986

PR #1989, opened 2025-05-20
fix: resolve editable installation issue for python package

senyai:
Before this `pip install -e .` produced ` error: [Errno 2] No such file or directory: '/tmp/tmpi02jtn4q.build-lib/google_benchmark/_benchmark.pyi'`

Second try of #1987. I hope code makes more sense now. Yes, `_benchmark.pyi`, `_benchmark.so` and `py.typed` go to the source tree, but this is how it is supposed to work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,7 +131,11 @@ class BuildBazelExtension(build_ext.build_ext):
         pkgname = "google_benchmark"
         pythonroot = Path("bindings") / "python" / "google_benchmark"
         srcdir = temp_path / "bazel-bin" / pythonroot
-        libdir = Path(self.build_lib) / pkgname
+        if not self.inplace:
+            libdir = Path(self.build_lib) / pkgname
+        else:
+            build_py = self.get_finalized_command("build_py")
+            libdir = build_py.get_package_dir(pkgname)
         for root, dirs, files in os.walk(srcdir, topdown=True):
             # exclude runfiles directories and children.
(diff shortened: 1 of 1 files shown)


dmah42:
looks like a linter issue that needs resolving @senyai

dmah42 (approved):

setup.py
```diff
@@ -131,7 +131,11 @@ def bazel_build(self, ext: BazelExtension) -> None:  # noqa: C901
         pkgname = "google_benchmark"
         pythonroot = Path("bindings") / "python" / "google_benchmark"
         srcdir = temp_path / "bazel-bin" / pythonroot
-        libdir = Path(self.build_lib) / pkgname
+        if not self.inplace:
+            libdir = Path(self.build_lib) / pkgname
+        else:
+            build_py = self.get_finalized_command("build_py")
+            libdir = build_py.get_package_dir(pkgname)
```
nicholasjng, line 138: Have you tested with `self.inplace`? This looks wrong, since `get_package_dir` returns a `str`, which would make the path concatenation in `shutil.copyfile` below fail, see https://github.com/pypa/setuptools/blob/main/setuptools/command/build_py.py#L275.

At minimum, I think this should be a `Path(build_py.get_package_dir(pkgname))`.

senyai: Yes, I tested it and it works `fp` in `libdir / fp` is `Path`, and 'str / Path' results in `Path`

PR #1990, opened 2025-05-20
python binding: add range check

### LINKED ISSUE #1965: [BUG] segmentation fault in python module
Using default python on Ubuntu 22.04
* `pip install google-benchmark`
* run

```python
import google_benchmark as benchmark


@benchmark.register
def pass_test(state):
    assert state.range(0) is not None
    while state:
        pass


for i in range(3):
    for j in range(4):
        benchmark.option.args((i, j))(pass_test)

if __name__ == "__main__":
    benchmark.main()
```

senyai:
fixes segmentation fault in
```python
import google_benchmark as benchmark

@benchmark.register
def av_test(state):
    state.range(9)

benchmark.main()
```

closes #1965

@nicholasjng, what do you think about performance impact?

### DIFF
--- a/bindings/python/google_benchmark/benchmark.cc
+++ b/bindings/python/google_benchmark/benchmark.cc
@@ -167,7 +167,15 @@ NB_MODULE(_benchmark, m) {
       .def_prop_rw("items_processed", &State::items_processed,
                    &State::SetItemsProcessed)
       .def("set_label", &State::SetLabel)
-      .def("range", &State::range, nb::arg("pos") = 0)
+      .def(
+          "range",
+          [](const State& state, std::size_t pos = 0) -> int64_t {
+            if (pos < state.range_size()) {
+              return state.range(pos);
+            }
+            throw nb::index_error("pos is out of range");
+          },
+          nb::arg("pos") = 0)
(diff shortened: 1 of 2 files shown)


LebedevRI:
Given that the current status quo is a weird crash, i think this is pretty obviously okay.

An argument could be made that this should be sunk into the C++ interface, and `range()` should throw.
Maybe. I guess the general expectation is that people do test with debug-built libbenchmark too.

nicholasjng:
Looks good. 

Do we want to allow Pythonic indexing as e.g. `state.range(-1)` for `state.range(state.range_size() - 1)`? Or is that too slow in a benchmark loop?

dmah42:
> Given that the current status quo is a weird crash, i think this is pretty obviously okay.
> 
> An argument could be made that this should be sunk into the C++ interface, and `range()` should throw. Maybe. I guess the general expectation is that people do test with debug-built libbenchmark too.

C++ already handles this case and asserts.  we could change this assert to do something else in NDEBUG but i suspect performance would take a hit (per @LebedevRI )

dmah42 (approved):

LebedevRI:
@senyai thank you!

PR #1994, opened 2025-05-23
[doc] Added nice to reducing_variance

### LINKED ISSUE #1991: [DOC] Using nice to reduce variance
Since context switches and scheduling competition was mentioned as a source of variance in results, running benchmarks with `nice` set to `-20` should minimize context switches by prioritizing the benchmark process. I ran a few tests and saw that the reduction in variance was evident in benchmarks that used more memory. I saw more than a 30% drop in coefficient of variance and also about 8% drop in mean runtime when I ran the below benchmark with and without setting niceness to -20.

```
#define N 100000

static void BM_StringCopy(benchmark::State& state) {
  std::string x = "hello";
  std::string s;
  for (auto _ : state)  for(int i = 0; i < N ; i++) s += x;
}

BENCHMARK(BM_StringCopy);
BENCHMARK_MAIN();
```
Terminal Commands to execute (may take a few minutes to run) : 
```
taskset -c 0 ./mybenchmark --benchmark_display_aggregates_only=true --benchmark_repetitions=200 

sudo nice -n  -20 taskset -c 0 ./mybenchmark --benchmark_display_aggregates_only=true --benchmark_repetitions=200
```
Varying `N` should show that the drop in coefficient of variance is proportional to `N`.

I was wondering if this could be added to the [reducing variance](https://github.com/Proxihox/benchmark/blob/main/docs/reducing_variance.md) docs?

Proxihox:
Added a suggestion to reduce variance in benchmarks using `nice`/`chrt`.
Closes #1991

### DIFF
--- a/docs/reducing_variance.md
+++ b/docs/reducing_variance.md
@@ -105,23 +105,28 @@ Linux workstation are:
 
 1. Use the performance governor as [discussed
 above](user_guide#disabling-cpu-frequency-scaling).
-1. Disable processor boosting by:
+2. Disable processor boosting by:
    ```sh
    echo 0 | sudo tee /sys/devices/system/cpu/cpufreq/boost
    ```
    See the Linux kernel's
    [boost.txt](https://www.kernel.org/doc/Documentation/cpu-freq/boost.txt)
    for more information.
-2. Set the benchmark program's task affinity to a fixed cpu.  For example:
+3. Set the benchmark program's task affinity to a fixed cpu.  For example:
    ```sh
    taskset -c 0 ./mybenchmark
    ```
-3. Disabling Hyperthreading/SMT.  This can be done in the Bios or using the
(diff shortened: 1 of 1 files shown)


LebedevRI (approved):
LG, thank you!
@dmah42 ?

dmah42:
thank you!

docs/reducing_variance.md
```diff
@@ -105,23 +105,28 @@ Linux workstation are:
 
 1. Use the performance governor as [discussed
 above](user_guide#disabling-cpu-frequency-scaling).
```
LebedevRI, line 107: This is(was?) kinda broken currently on AMD by the way.
I had to pass `amd_pstate=disable` to truly disable it.

AO696: thanks!

Issue #1232, opened 2021-09-24
undefined reference to `__imp_SHGetValueA'


C0C41N:
I am trying to run this command

`g++ src/benchmark.cpp -std=c++11 -lbenchmark -lpthread -o x.exe`

but when I do I get this error

_C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../lib/libbenchmark.a(sysinfo.cc.obj):sysinfo.cc:(.text+0x637): undefined reference to `__imp_SHGetValueA'
collect2.exe: error: ld returned 1 exit status_

Here's the benchmark.cpp file

```
#include <benchmark/benchmark.h>

static void BM_StringCreation(benchmark::State& state) {
	for (auto _ : state) std::string empty_string;
}
// Register the function as a benchmark
BENCHMARK(BM_StringCreation);

// Define another benchmark
static void BM_StringCopy(benchmark::State& state) {
	std::string x = "hello";
	for (auto _ : state) std::string copy(x);
}
BENCHMARK(BM_StringCopy);

BENCHMARK_MAIN();
```

**System**
Windows 10
g++ (Rev2, Built by MSYS2 project) 10.3.0
gtest & benchmark libraries are included. (gtest works fine btw)

![image](https://user-images.githubusercontent.com/51693606/134703995-dfaf71da-7a04-4575-b494-9e89f1635dc8.png)

dominichamon:
`SHGetValueA` is part of the shlwapi.h and per the docs you may need to link against Shlwapi.lib on some configurations: https://google.github.io/benchmark/platform_specific_build_instructions.html

C0C41N:
Thanks for the tip @dominichamon

I just added -lShlwapi and now it's working fine

`g++ src/benchmark.cpp -std=c++11 -lbenchmark -lpthread -lShlwapi -o x.exe`

dominichamon:
#imstillworthy

PR #1996, opened 2025-06-06
Fix bug: link Shlwapi in Libs.private for Windows

### LINKED ISSUE #1940: [BUG] pkgconfig file missing required libs for linking
**Describe the bug**
Issue #1232 was closed by mentioning adding `-lShlwapi` to the linker's command line would fix issue.
However, [pkgconfig file](https://github.com/google/benchmark/blob/main/cmake/benchmark.pc.in#L11) doesn't seem to reflect that. 

**System**
Win11 64 bits / msys2, gcc version 14.2.0 (Rev2, Built by MSYS2 project)

**To reproduce**
1. Commit ff5c94d860dcffd4f4159edf79d805bcfaca3cb8 (but probably any)
2. cmake -DBENCHMARK_ENABLE_GTEST_TESTS=0  -DCMAKE_BUILD_TYPE=Release <source>
3. cmake --build . && cmake --install .
4. In lib/pkgconfig/benchmark.pc, neither `Libs.private` nor `Libs` contain `-lShlwapi`

**Expected behavior**
The `Libs` line should contain all info needed for linking.

xshthkr:
Fix missing -lShlwapi in pkg-config file on Windows

The static benchmark library depends on Shlwapi for `SHGetValueA`, used in `src/sysinfo.cc` (fixed by manually linking, #1232). The dependency was never reflected in the generated `benchmark.pc` file.

Users relying on `pkg-config` to obtain linker flags would encounter undefined reference errors on Windows unless manually adding `-lShlwapi` themselves.

This fix ensures `-lShlwapi` is added to `Libs.private` in the pkg-config file when building on Windows, so static linkers pick it up automatically.

Tested on Windows 11 x86_64, msys2

Fixes #1940 
Closes https://github.com/google/benchmark/issues/639

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -63,6 +63,7 @@ Rainer Orth <ro@cebitec.uni-bielefeld.de>
 Roman Lebedev <lebedev.ri@gmail.com>
 Sayan Bhattacharjee <aero.sayan@gmail.com>
 Shapr3D <google-contributors@shapr3d.com>
+Shashank Thakur <shashankt2004@gmail.com>
 Shuo Chen <chenshuo@chenshuo.com>
 Staffan Tjernstrom <staffantj@gmail.com>
 Steinar H. Gunderson <sgunderson@bigfoot.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -89,6 +89,7 @@ Ray Glover <ray.glover@uk.ibm.com>
 Robert Guo <robert.guo@mongodb.com>
 Roman Lebedev <lebedev.ri@gmail.com>
 Sayan Bhattacharjee <aero.sayan@gmail.com>
+Shashank Thakur <shashankt2004@gmail.com>
 Shuo Chen <chenshuo@chenshuo.com>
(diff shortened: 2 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ endif(HAVE_LIB_RT)
 # We need extra libraries on Windows
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   target_link_libraries(benchmark PRIVATE shlwapi)
+  set(BENCHMARK_PRIVATE_LINK_LIBRARIES -lShlwapi)
```
LebedevRI, line 65: I suppose you are just copying from the next block, but this is a rather roundabout fix,
the previous line should be sufficient, anything else should be derived from it.

LebedevRI (approved):
Looks fine as a fix.

src/CMakeLists.txt
xshthkr: I have attempted to fix that in #1998.

PR #1998, opened 2025-06-10
Improve handling of private link libraries in pkg-config generation

xshthkr:
This PR improves how `Shlwapi` and other platform specific libraries are linked, building on #1996. This addresses the repetitive logic in the current implementation for Windows and Solaris.

The platform specific libraries were originally hardcoded twice. In this PR, `BENCHMARK_PRIVATE_LINK_LIBRARIES` is derived automatically from the target's linked libraries, filtered to exclude CMake targets (like `Threads::Threads`) and formatted with `-l` prefixes.

Tested on Windows 11 (MSYS2 MinGW64)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -62,13 +62,11 @@ endif(HAVE_LIB_RT)
 # We need extra libraries on Windows
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   target_link_libraries(benchmark PRIVATE shlwapi)
-  set(BENCHMARK_PRIVATE_LINK_LIBRARIES -lShlwapi)
 endif()
 
 # We need extra libraries on Solaris
 if(${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
   target_link_libraries(benchmark PRIVATE kstat)
-  set(BENCHMARK_PRIVATE_LINK_LIBRARIES -lkstat)
 endif()
 
 if (NOT BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


LebedevRI:
> I cant seem to find a Cmake native way of doing this. It seems that CMake doesnt natively automatically populate `.pc` files with `PRIVATE` libs.

Actually, that's a good point: shouldn't that be guarded with `NOT BUILD_SHARED_LIBS` check?

xshthkr:
> > I cant seem to find a Cmake native way of doing this. It seems that CMake doesnt natively automatically populate `.pc` files with `PRIVATE` libs.
> 
> Actually, that's a good point: shouldn't that be guarded with `NOT BUILD_SHARED_LIBS` check?

You're right. Guarded with `NOT BUILD_SHARED_LIBS` check.

LebedevRI (commented):
Just to check, you have tried placing generator expressions into `.pc` and seeing if maybe `configure_file()` does the right thing for them?
I haven't tried, but maybe `"-l$<JOIN:$<TARGET_PROPERTY:benchmark,LINK_LIBRARIES>, -l>"`

LebedevRI (approved):
Thank you, i think this is an improvement, although this is still totally not a cmake way to do it. Although i don't know what is.

xshthkr:
> For the record, the following kinda works, but doesn't resolve cmake target names to the actual library names. But then the current code doesn't do that either. IMHO the current handling is broken.

Yeah. Raw cmake target names arent handled in either approaches. Although the generator expression approach tries to be more native.

PR #2003, opened 2025-06-15
python: fix segfault on empty `argv` in `main()`

senyai:
test:
`python -c 'import google_benchmark as b;b.register(print);b.main([]);'`


bindings/python/google_benchmark/benchmark.cc
```diff
@@ -17,14 +17,13 @@ std::vector<std::string> Initialize(const std::vector<std::string>& argv) {
...
   // so it persists, and replace the pointer below.
-  static std::string executable_name(argv[0]);
-  std::vector<char*> ptrs;
-  ptrs.reserve(argv.size());
-  for (auto& arg : argv) {
-    ptrs.push_back(const_cast<char*>(arg.c_str()));
+  static std::string executable_name(argv.empty() ? "unknown" : argv[0]);
+  int argc = static_cast<int>(std::max(std::size_t{1}, argv.size()));
+  std::vector<char*> ptrs(argc);
+  for (size_t i = 0; i < argv.size(); ++i) {
```
LebedevRI: Might as well start from 1 then

senyai:
I think there's more explicit and expressive way to update the logic:

```
if (!argv.empty()) {
  static std::string executable_name(argv[0]);
  ptrs[0] = const_cast<char*>(executable_name.c_str());
}
```

should I change it, @LebedevRI?

senyai:
Now I'm glad with this simple fix @LebedevRI

senyai:
btw, my initial command for reproducing the issue is no longer giving me segmentation fault, it just says `ValueError: argv cannot be an empty list, and must contain the program name as the first element.` Not sure why, I didn't change anything. But this command `python -c 'import google_benchmark as b;b.register(print);b.initialize([]);'` should still give you a segfault in main.

bindings/python/google_benchmark/benchmark.cc
```diff
@@ -14,16 +14,18 @@ namespace {
...
     ptrs.push_back(const_cast<char*>(arg.c_str()));
   }
-  ptrs[0] = const_cast<char*>(executable_name.c_str());
+  if (!ptrs.empty()) {
+    // The `argv` pointers here become invalid when this function returns, but
+    // benchmark holds the pointer to `argv[0]`. We create a static copy of it
+    // so it persists, and replace the pointer below.
+    static std::string executable_name(argv[0]);
+    ptrs[0] = const_cast<char*>(executable_name.c_str());
+  }
```
LebedevRI, line 28: Doesn't this just exchange one bug for another?
The comment suggests that something uses the binary name (`argv[0]`) later on,
but now instead of "crashing" here, we'll crash the code that tries to take `argv[0]`.

bindings/python/google_benchmark/benchmark.cc
```diff
@@ -14,16 +14,18 @@ namespace {
 namespace nb = nanobind;
 
 std::vector<std::string> Initialize(const std::vector<std::string>& argv) {
-  // The `argv` pointers here become invalid when this function returns, but
```
LebedevRI, line 17:
```diff
-std::vector<std::string> Initialize(const std::vector<std::string>& argv) {
-  // The `argv` pointers here become invalid when this function returns, but
-  // benchmark holds the pointer to `argv[0]`. We create a static copy of it
-  // so it persists, and replace the pointer below.
-  static std::string executable_name(argv[0]);
-  std::vector<char*> ptrs;
-  ptrs.reserve(argv.size());
-  for (auto& arg : argv) {
-    ptrs.push_back(const_cast<char*>(arg.c_str()));
-  }
-  ptrs[0] = const_cast<char*>(executable_name.c_str());
-  int argc = static_cast<int>(argv.size());
-  benchmark::Initialize(&argc, ptrs.data());
-  std::vector<std::string> remaining_argv;
-  remaining_argv.reserve(argc);
-  for (int i = 0; i < argc; ++i) {
-    remaining_argv.emplace_back(ptrs[i]);
-  }
-  return remaining_argv;
-}

+std::vector<std::string> Initialize(const std::vector<std::string>& argv) {
+  // The `argv` pointers here become invalid when this function returns, but
+  // benchmark holds the pointer to `argv[0]`. We create a static copy of it
+  // so it persists, and replace the pointer below.
+  std::vector<char*> ptrs;
+  if(argv.empty()) {
+    static std::string executable_name("unknown");
+    ptrs.emplace_back(executable_name.c_str());
+  } else {
+    ptrs.reserve(argv.size());
+    for (auto& arg : argv) {
+      ptrs.push_back(const_cast<char*>(arg.c_str()));
+    }
+    static std::string executable_name(argv[0]);
+    ptrs[0] = const_cast<char*>(executable_name.c_str());
+  }
+  int argc = static_cast<int>(ptrs.size());
+  benchmark::Initialize(&argc, ptrs.data());
+  std::vector<std::string> remaining_argv;
+  remaining_argv.reserve(argc);
+  for (int i = 0; i < argc; ++i) {
+    remaining_argv.emplace_back(ptrs[i]);
+  }
+  return remaining_argv;
+}
```

bindings/python/google_benchmark/benchmark.cc
senyai: The code that tries to take `argv[0]` is in `benchmark::Initialize`, and there's already a check that ensures that it does not happen when `argc` is 0. That's why this little change is enough. I'm not sure how to make the comment better.

bindings/python/google_benchmark/benchmark.cc
senyai: It is okay to pass empty array to `benchmark::Initialize`
https://github.com/google/benchmark/blob/b20cea674170b2ba45da0dfaf03953cdea473d0d/src/benchmark.cc#L753-L754

LebedevRI:
I don't really touch python stuff here.
I've suggested how i think this would be best done, but i'm perfectly happy to just leave it for other reviewers.

bindings/python/google_benchmark/benchmark.cc
dmah42: this seems to be replacing something C++ like (`vector<string>`) with something C like (`vector<char*>`) which is always going to be unsafe.    should `benchmark::Initialize` just take a `vector<string>` instead?  and then update `BENCHMARK_MAIN` to do the translation from C-strings to C++ strings (which should always be safer).

only issue would be custom mains that _don't_ currently do that translation but presumably they'll continue to work if we keep the old (current) `Initialize` method around?

senyai: The `std::vector<char*>` isn't part of my changes. Could we defer discussions about interface changes to a separate conversation?

dmah42: sure, but i'm not comfortable with the creation of a static string to solve the problem. i think the appropriate way to solve this would be a static vector pointer for ptrs that gets allocated on Initialize (with a check that it isn't already initialized or at least non-nullptr).

alternatively, changing benchmark to do the deep copy of argv[0] internally instead of handling it here in the binding.  that might be even better, and doesn't require API changes.

LebedevRI: > sure, but i'm not comfortable with the creation of a static string to solve the problem. i think the appropriate way to solve this would be a static vector pointer for ptrs that gets allocated on Initialize (with a check that it isn't already initialized or at least non-nullptr).
>
> alternatively, changing benchmark to do the deep copy of argv[0] internally instead of handling it here in the binding. that might be even better, and doesn't require API changes.

To reword this, `benchmark::Initialize()` would still take `char**`,
but instead of retaining the `argv[0]` as a pointer,
it would create a copy of it, so here, we wouldn't need a `static` string,
we could safely deallocate our `argv[0]` that we created here.

To be honest, i'm not sure why the current `static string` approach
is worse than that? It is as safe, and results in less unnessesary allocations anc copies.

Also, seen the alternative diff in my other comment? https://github.com/google/benchmark/pull/2003/files#r2172575805

dmah42: yeah.. and i think there's probably a couple of changes we could make either way to make this safer, but i also think the approach in this PR is safe given that we check for argc 0 already, and it does keep the change localised to the python binding where the problem arises.

LebedevRI:
LG

PR #2004, opened 2025-06-18
Guard adding intel compiler flag behind check for intel compiler.

EricWF:
For reasons I don't yet fully understand, failing to guard the check for flag -wd654 will cause it to be added to certain clang configurations, and then cause compilation to fail.

While exactly why this happens needs investigation, the fix proposed in this patch seems direct and simple enough to apply regardless.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,9 @@ else()
     # See #631 for rationale.
     add_cxx_compiler_flag(-wd1786)
     add_cxx_compiler_flag(-fno-finite-math-only)
+    # ICC17u2: overloaded virtual function "benchmark::Fixture::SetUp" is only partially
+    # overridden (because of deprecated overload)
+    add_cxx_compiler_flag(-wd654)
   endif()
   # Disable deprecation warnings for release builds (when -Werror is enabled).
   if(BENCHMARK_ENABLE_WERROR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -230,9 +230,13 @@ else()
...
-  # ICC17u2: overloaded virtual function "benchmark::Fixture::SetUp" is only partially overridden
-  # (because of deprecated overload)
-  add_cxx_compiler_flag(-wd654)
+  if  (CMAKE_CXX_COMPILER_ID STREQUAL "Intel" OR
+       CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
+      # ICC17u2: overloaded virtual function "benchmark::Fixture::SetUp" is only partially overridden
+      # (because of deprecated overload)
+      add_cxx_compiler_flag(-wd654)
+  endif()
+
```
LebedevRI, line 236: While there, can you move this line into the previous such guard block ^ ?

EricWF: Indeed I can.

I moved it up into the previous matching block.

LebedevRI (approved):
LG, thank you!

LebedevRI:
Looks like you need to rebase this yourself, since it seems you've unclicked the checkbox.

EricWF:
> Looks like you need to rebase this yourself, since it seems you've unclicked the checkbox.

I certainly didn't mean to. 

Thanks for the review.

LebedevRI (approved):

EricWF:
Thanks again.

PR #2005, opened 2025-06-26
core(perf): use string move constructor for AddCustomContext

romintomasetti:
Avoid unnecessary string copies.


romintomasetti:
I've just signed the `CLA`.

PR #2012, opened 2025-07-22
Fix build when targeting Arm64EC using Clang

dpaoliello:
These fix building benchmark targeting Arm64EC Windows when using Clang (i.e., with Clang's preprocessor defines rather than the MSVC-compatibility defines).

Copied from <https://github.com/llvm/llvm-project/pull/150068>

### DIFF
--- a/src/cycleclock.h
+++ b/src/cycleclock.h
@@ -36,6 +36,9 @@
 // declarations of some other intrinsics, breaking compilation.
 // Therefore, we simply declare __rdtsc ourselves. See also
 // http://connect.microsoft.com/VisualStudio/feedback/details/262047
+//
+// Note that MSVC defines the x64 preprocessor macros when building
+// for Arm64EC, despite it using Arm64 assembly instructions.
 #if defined(COMPILER_MSVC) && !defined(_M_IX86) && !defined(_M_ARM64) && \
     !defined(_M_ARM64EC)
 extern "C" uint64_t __rdtsc();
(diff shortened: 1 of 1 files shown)


src/cycleclock.h
```diff
@@ -79,7 +79,7 @@ inline BENCHMARK_ALWAYS_INLINE int64_t Now() {
   int64_t ret;
   __asm__ volatile("rdtsc" : "=A"(ret));
   return ret;
-#elif defined(__x86_64__) || defined(__amd64__)
+#elif (defined(__x86_64__) || defined(__amd64__)) && !defined(__arm64ec__)
```
LebedevRI, line 88: Is `__x86_64__` or `__amd64__` really defined when `__arm64ec__` is defined?

dpaoliello: Yes:
https://github.com/llvm/llvm-project/blob/b13bca7387a7aad6eaf3fa1c55bd06fe091f65ed/clang/lib/Basic/Targets/AArch64.cpp#L404-L414

You should think of Arm64EC as "x64 code running on an Arm64 machine with a weird instruction set". So, by default, it needs to use the x64 layout for objects to be compatible with any real x64 code that is linked into the same binary or called across binaries.

LebedevRI: Yes, but ARM is ARM, it is not Intel arch, which x86_64 and AMD64 are.

LebedevRI: Please add a code comment to explicitly state this, err, _**very**_ interesting design decision...

LebedevRI (approved):
Thanks!

PR #2016, opened 2025-08-06
2011: Installing contents from benchmark's tools/ subdirectory.

### LINKED ISSUE #2011: Packaging tools subdirectory on install
I am interested in using the utils in the tools subdirectory in our benchmark packages. We are using conan to package benchmark and the maintainers of the contain recipe stated: "It looks like the CMakeLists scripts for this project don't package or do anything with tools". 

I was wondering if there is a reason for not packaing the tools/ subdirectory and if it could be included on install? Perhaps there is an option to include on install that we are unaware of? 

Thanks!

BuildMonkey:
This Pull Request (PR) relates to the following issue: 
https://github.com/google/benchmark/issues/2011#issuecomment-3107688780

I modeled my change against how benchmark docs are installed, except tools do not have a comparable CMAKE_INSTALL_DOCDIR variable recognised by build systems. Instead, I used CMAKE_INSTALL_PREFIX.

Additional, like docs, I have tools subdirectory installed by default, but we can change if this isn't the optimal behaviour.

Fixes #2011.


dmah42:
let me know when you've signed the CLA and i will take a look.  thank you!

src/CMakeLists.txt
```diff
@@ -194,3 +194,9 @@ else()
       DESTINATION ${CMAKE_INSTALL_DOCDIR})
   endif()
 endif()
+
+if (BENCHMARK_ENABLE_INSTALL AND BENCHMARK_INSTALL_TOOLS)
+  install(
+    DIRECTORY "${PROJECT_SOURCE_DIR}/tools/"
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/tools)
```
LebedevRI: This destination is wrong. It _probaly_ should be
```
if(NOT DEFINED CMAKE_INSTALL_PYTOOLSDIR)
  define(CMAKE_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/googlebenchmark/tools" CACHE PATH "")
endif()

<...>
  install(
    DIRECTORY "${PROJECT_SOURCE_DIR}/tools/"
    DESTINATION ${CMAKE_INSTALL_PYTOOLSDIR})
```

https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

BuildMonkey: Thanks for the feedback. I noticed set() was used for variables so I used that since I had problems with the recommended define(). 

It looks like tools is the first component to go under googlebenchmark/ so just want to make sure that's the right place.

Thanks again!

src/CMakeLists.txt
```diff
@@ -194,3 +194,13 @@ else()
...
 endif()
+
+if (NOT DEFINED CMAKE_INSTALL_PYTOOLSDIR)
+  set(CMAKE_INSTALL_PYTOOLSDIR "${CMAKE_INSTALL_DATADIR}/googlebenchmark/tools" CACHE PATH "")
+endif()
+
+if (BENCHMARK_ENABLE_INSTALL AND BENCHMARK_INSTALL_TOOLS)
+  install(
+    DIRECTORY "${PROJECT_SOURCE_DIR}/tools/"
+    DESTINATION ${CMAKE_INSTALL_PYTOOLSDIR})
```
dmah42, line 203: i'm not a python package management expert, but doesn't python expect things to be in some central location?

i see options like

```
FindPython(..)
execute_process(COMMAND ${Python_EXECUTABLE} -m site --user-site
    OUTPUT_VARIABLE PYTHON_USER_SITE
    OUTPUT_STRIP_TRAILING_WHITESPACE)
message(STATUS "installing <pybind module.so> to ${PYTHON_USER_SITE}")
install(TARGETS <module> LIBRARY DESTINATION ${PYTHON_USER_SITE})
```

offered elsewhere for this sort of thing.

i don't have any strong preference either way so if this works for your needs then its fine by me.

as for the googlebenchmark prefix: i suspect this is so the Python import paths are self-describing?

nicholasjng: Would it make sense to move the tools into the `google-benchmark` bindings package? That would give users a single entrypoint to first run the benchmarks, and directly thereafter process the results.

LebedevRI: > Would it make sense to move the tools into the `google-benchmark` bindings package?

If you drop abseil dep for bindings - yes, please.

nicholasjng: Sure. I will create a separate issue for that, if you could just assign me, that would be great.

src/CMakeLists.txt
LebedevRI: Yeah no idea why i've written `define()`, that's not a thing. I meant `set()`.

src/CMakeLists.txt
LebedevRI: > i'm not a python package management expert, but doesn't python expect things to be in some central location?
> 
> i see options like
> 
> ```
> FindPython(..)
> execute_process(COMMAND ${Python_EXECUTABLE} -m site --user-site
>     OUTPUT_VARIABLE PYTHON_USER_SITE
>     OUTPUT_STRIP_TRAILING_WHITESPACE)
> message(STATUS "installing <pybind module.so> to ${PYTHON_USER_SITE}")
> install(TARGETS <module> LIBRARY DESTINATION ${PYTHON_USER_SITE})
> ```
> 
> offered elsewhere for this sort of thing.
> 
> i don't have any strong preference either way so if this works for your needs then its fine by me.
> 
> as for the googlebenchmark prefix: i suspect this is so the Python import paths are self-describing?

Yes, that is the alternative. My reasoning is that the script does not have a globally-unique descriptive name,
and isn't meant to be in $PATH, but meant to be invoked with it's full path,
and it can't be used as a library in a script. `find /usr/share/ -iname \*.py ` shows quite a bit of existing scripts,
so i think this would not be inventing something new.

BuildMonkey: Hello! 

It seems to me this has raised discussion on where to place. Have we reached consensus? I am more than happy to change the work to match what people are in agreement with.

Thanks!

dmah42: i'll defer to @LebedevRI for this.

LebedevRI: I'd defer to whoever owns the python bindings.
Other than that, i think regardless of what happens with #2017, 
the current diff seems like a reasonable improvement.

LebedevRI (approved):

LebedevRI:
@BuildMonkey you need to deal with the CLA check.

LebedevRI (dismissed):

src/CMakeLists.txt
```diff
@@ -194,3 +194,13 @@ else()
       DESTINATION ${CMAKE_INSTALL_DOCDIR})
   endif()
 endif()
+
+if (NOT DEFINED CMAKE_INSTALL_PYTOOLSDIR)
+  set(CMAKE_INSTALL_PYTOOLSDIR "${CMAKE_INSTALL_DATADIR}/googlebenchmark/tools" CACHE PATH "")
+endif()
```
LebedevRI: Thinking about it more, i don't believe those guards should be there,
let's just always `set()`...

LebedevRI (approved):

Issue #2039, opened 2025-09-12
[BUG] Multithreaded benchmark results in invalid JSON


ldionne:
**Describe the bug**
In some cases (but strangely, not all cases), a multi-threaded benchmark results in invalid JSON output when using `--benchmark_format=json`.

**System**
I was able to reproduce this on macOS and on Compiler Explorer, so I think this clearly is independent of the platform.

**To reproduce**
A benchmark like this reproduces the issue (unreliably):

```c++
#include "benchmark/benchmark.h"

static void BM_Ostream_number(benchmark::State& state) {
  while (state.KeepRunning()) {

  }
}
BENCHMARK(BM_Ostream_number)->Arg(3)->Threads(2);

BENCHMARK_MAIN();
```

To really reproduce the issue reliably, it's best to perform something in the loop. For example: https://godbolt.org/z/Y97e6veaE

That example results in the following JSON:

```
{
  "context": {
    "date": "2025-09-12T13:38:31+00:00",
    "host_name": "ce",
    "executable": "./output.s",
    "num_cpus": 2,
    "mhz_per_cpu": 3598,
    "cpu_scaling_enabled": false,
    "aslr_enabled": false,
    "caches": [
    ],
    "load_avg": [0.679688,0.585449,0.533691],
    "library_version": "v1.9.4-39-g16144154",
    "library_build_type": "debug",
    "json_schema_version": 1
  },
  "benchmarks": [
    {
      "name": "BM_Ostream_number/3/threads:2",
      "family_index": 0,
      "per_family_instance_index": 0,
      "run_name": "BM_Ostream_number/3/threads:2",
      "run_type": "iteration",
      "repetitions": 1,
      "repetition_index": 0,
      "threads": 2,
      "iterations": 1,172,451,248,
      "real_time": 5.9873469468256479e-01,
      "cpu_time": 5.9655409228580569e-01,
      "time_unit": "ns"
    }
  ]
}
```

Notice that the `iterations` key of `BM_Ostream_number/3/threads:2` contains multiple comma-separated values, which is invalid JSON.

**Expected behavior**
The JSON should be valid, otherwise tooling can't do anything with the benchmark results.

PR #2040, opened 2025-09-12
JSON reporter: make int dumping always happen in C locale

LebedevRI:
The test fails without the fix.

Fixes https://github.com/google/benchmark/issues/2039


LebedevRI:
@dmah42 thank you!

PR #2042, opened 2025-09-16
Adding explicit cast to aid in template argument deduction.

dmah42:
The `BENCHMARK(my_func)` macro needs to register the function my_func so the library can run it. It uses `std::make_unique` to create an object that stores metadata about the benchmark, including its name ("my_func") and a pointer to the function itself (my_func).

The specific constructor being called via `std::make_unique` is for `benchmark::internal::FunctionBenchmark`.

When a benchmark function name is overloaded, the compiler can't determine which function is being referred to until the call is made with specific arguments, or enough context is provided.

The addition of a `static_cast` to the expected `Function*` type resolves any ambiguity as the compiler is forced to pick the one with the matching signature.

### DIFF
--- a/test/overload_gtest.cc
+++ b/test/overload_gtest.cc
@@ -31,4 +31,3 @@ void MyTemplatedOverloadedBenchmark(benchmark::State& state) {
 BENCHMARK_TEMPLATE(MyTemplatedOverloadedBenchmark, 1);
 
 BENCHMARK_MAIN();
-


LebedevRI:
The test does not seem to fail without the fix. Does the test need explicit `-std=c++20` or something?

dmah42:
> The test does not seem to fail without the fix. Does the test need explicit `-std=c++20` or something?

it was failing under an Android NDK build which i'd rather not start replicating here, if possible.  do you have concerns about the change otherwise? i agree ideally we'd have a test that fails/passes (and i do but.. it's in https://github.com/google/pthreadpool).

i saw this trying to import the latest version and this project under Android NDK failed due to overloaded functions and poor argument deduction.

LebedevRI (approved):

Issue #2017, opened 2025-08-07
python bindings: Add tools, drop hard abseil dependency


nicholasjng:
Follow-up of https://github.com/google/benchmark/pull/2016.

tl,dr: It would be to add the analysis tools of benchmark results to the `google-benchmark` Python package instead of keeping them in a separate directory. Benefits of this are easier installation for the user, being able to use them outside of CMake, and better organization of Python sources for GBM.

Bonus: The `absl-py` dependency should be dropped. Since that is used for flag parsing, bindings for the flag values should be added as well as an `argparse` solution.

LebedevRI:
I **_personally_** think dropping abseil dep would be good,
but i'm not quite sure why/how specifically it would be better
if the compare.py script would be installed from bindinds.

It has no particular relationship with bindings,
it does not have a globally-unique descriptive name,
it isn't meant to be in $PATH, but meant to be invoked with it's full path,
it can't be used as a library in a script.

`find /usr/share/ -iname \*.py` shows quite a bit of existing scripts,
so i think #2016 would not be inventing something new.

PR #2052, opened 2025-10-13
bindings(python): Remove hard `absl-py` dependency

nicholasjng:
This commit refactors the public `google_benchmark` APIs to work without `absl-py` installed in the current virtual environment. It also removes `absl-py` from the list of PyPI package dependencies.

To restore the previous behavior, users can implement a custom `main()` function using `absl.app` as command-line parser + runner, for example as the previous default `main()` function that is removed by this very commit.

-----------------

Happy to hear your thoughts, this is part of the analysis tools refactoring (#2017).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -50,7 +50,7 @@ jobs:
         uses: astral-sh/setup-uv@eb1897b8dc4b5d5bfe39a428a8f2304605e0983c # v7.0.0
 
       - name: Build wheels on ${{ matrix.os }} using cibuildwheel
-        uses: pypa/cibuildwheel@7c619efba910c04005a835b110b057fc28fd6e93 # v3.2.0
+        uses: pypa/cibuildwheel@9c00cb4f6b517705a3794b22395aedc36257242c # v3.2.1
         env:
           CIBW_BUILD: "cp310-* cp311-* cp312-*"
           CIBW_BUILD_FRONTEND: "build[uv]"


bindings/python/google_benchmark/__init__.py
```diff
@@ -53,6 +52,17 @@ def my_benchmark(state):
...
 
+HAS_ABSL = False
+with contextlib.suppress(ImportError):
+    # check if absl-py is available.
+    import absl
+
+    HAS_ABSL = True
+    del absl
+
+del contextlib
```
nicholasjng: Not sure about this bit yet - 1), it still requires a `from absl import app` statement in all APIs below (it will be cached in `sys.modules` in all but the first call, but still), and 2) it could break users importing `contextlib` before `google-benchmark`, as this call marks _all_ previously loaded contextlib modules for GC.

bindings/python/google_benchmark/__init__.py
```diff
@@ -122,19 +120,19 @@ def register(undefined=None, *, name=None):
     return options.func
 
 
-def _flags_parser(argv):
-    argv = _benchmark.Initialize(argv)
-    return app.parse_flags_with_usage(argv)
+def _flags_parser(argv: list[str] | None = None) -> list[str]:
+    import sys
+
+    return _benchmark.Initialize(argv or sys.argv)
```
nicholasjng: I could probably save the `sys` import here by changing the `_benchmark.Initialize` interface to take a `std::optional<std::vector<std::string>>& argv` in `benchmark.cc` in a follow-up.

bindings/python/google_benchmark/__init__.py
```diff
@@ -122,19 +120,19 @@ def register(undefined=None, *, name=None):
...
 
 
-def _run_benchmarks(argv):
-    if len(argv) > 1:
-        raise app.UsageError("Too many command-line arguments.")
+def _run_benchmarks(argv: list[str] | None = None) -> None:
```
dmah42: do we still need to pass argv in?

bindings/python/google_benchmark/__init__.py
```diff
@@ -122,19 +120,19 @@ def register(undefined=None, *, name=None):
     return options.func
 
 
-def _flags_parser(argv):
-    argv = _benchmark.Initialize(argv)
-    return app.parse_flags_with_usage(argv)
+def _flags_parser(argv: list[str] | None = None) -> list[str]:
```
dmah42: maybe rename this to `initialize`?

nicholasjng: I'd argue this can go entirely, since  `initialize` is defined directly below the diff in the same file.

bindings/python/google_benchmark/__init__.py
nicholasjng: This API is probably of no more use, since it just calls `RunSpecifiedBenchmarks()` directly. I left it in since removal was not in scope of the original PR (making absl optional), but I can also remove it entirely.

dmah42: yeah i think we can hugely simplify this.

PR #2062, opened 2025-10-22
Fix warnings from Clang

dingxiangfei2009:
It appears that the library hits some lints when it is compiled with Clang.

This patch addresses the following issues.
- `-Wformat-nonliteral` from colour printing functions
- `-Wmissing-prototypes` from a few internal functions

### DIFF
--- a/test/basic_test.cc
+++ b/test/basic_test.cc
@@ -3,6 +3,7 @@
 
 #define BASIC_BENCHMARK_TEST(x) BENCHMARK(x)->Arg(8)->Arg(512)->Arg(8192)
 
+namespace {
 void BM_empty(benchmark::State& state) {
   for (auto _ : state) {
     auto iterations = static_cast<double>(state.iterations()) *
@@ -174,5 +175,6 @@ static_assert(
                      benchmark::State::StateIterator>::value_type,
                  typename benchmark::State::StateIterator::value_type>::value,
     "");
+}  // end namespace
 
 BENCHMARK_MAIN();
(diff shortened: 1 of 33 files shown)


src/benchmark.cc
```diff
@@ -723,14 +723,14 @@ namespace internal {
 
 void (*HelperPrintf)();
 
-void PrintUsageAndExit() {
+static void PrintUsageAndExit() {
```
LebedevRI: IIRC it's "better" to just wrap them into anonymous namespace instead.

dingxiangfei2009: Applied.

I also updated the tests similarly.

src/colorprint.h
```diff
@@ -17,12 +17,15 @@ enum LogColor {
...
+__attribute__((__format__(__printf__, 1, 2))) std::string FormatString(
+    const char* msg, ...);
 
-void ColorPrintf(std::ostream& out, LogColor color, const char* fmt,
-                 va_list args);
-void ColorPrintf(std::ostream& out, LogColor color, const char* fmt, ...);
+__attribute__((__format__(__printf__, 3, 0))) void ColorPrintf(
+    std::ostream& out, LogColor color, const char* fmt, va_list args);
+__attribute__((__format__(__printf__, 3, 4))) void ColorPrintf(
+    std::ostream& out, LogColor color, const char* fmt, ...);
```
LebedevRI: These `__attribute__` need to be hidden by preprocessor macro,
and only enabled for non-MSVC.

dingxiangfei2009: Applied

PR #2073, opened 2025-11-20
Take a closure for benchmark configuration

dingxiangfei2009:
It turns out it is useful when a benchmark is generated on the fly before the enumeration on benchmark registration.

There is a use case in which benchmarks shall be built from configuration flags, before the benchmark init. Without captures, it is actually hard to pass additional data.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -1229,7 +1229,7 @@ class BENCHMARK_EXPORT Benchmark {
   // Pass this benchmark object to *func, which can customize
   // the benchmark by calling various methods like Arg, Args,
   // Threads, etc.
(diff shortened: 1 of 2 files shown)


include/benchmark/benchmark.h
```diff
@@ -1229,7 +1229,7 @@ class BENCHMARK_EXPORT Benchmark {
   // Pass this benchmark object to *func, which can customize
   // the benchmark by calling various methods like Arg, Args,
   // Threads, etc.
-  Benchmark* Apply(void (*custom_arguments)(Benchmark* benchmark));
+  Benchmark* Apply(const std::function<void(Benchmark* benchmark)>&);
```
dmah42, line 1232: should this be an overload rather than replacing the fn pointer version?  i suspect we have some integrations that might be relying on the void* version...

Issue #2080, opened 2025-11-26
[BUG] Rate counters are per-thread in multi-threaded benchmarks, kAvgThreadsRate does not make sense


FRosner:
**Describe the bug**

When using counters that represent a rate (`benchmark::Counter::kIsRate`), the rate appears to be computed per thread. Subsequently, when using `kAvgThreadsRate`, the rate is computed per thread and then divided by the number of threads (again). This is misleading.

**System**

OS

```
ProductName:		macOS
ProductVersion:		15.6.1
BuildVersion:		24G90
```

Compiler

```
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

Library

```
FetchContent_Declare(
        benchmark
        GIT_REPOSITORY https://github.com/google/benchmark.git
        GIT_TAG v1.9.4
)
```
 
**To reproduce**

Write a benchmark that simply sleeps for 1 second, then increments a counter by 1. With 1 thread, we expect a counter of 1, and a rate of 1/s. With 10 threads, we expect a counter of 10 and a rate of 10/s. However, the rate is reported as 1/s and the per-thread rate as 0.1/s. This seems wrong?

```c++
static void BM_IntegerAddition(benchmark::State& state) {
    for (auto _ : state) {
        std::this_thread::sleep_for(std::chrono::seconds(1));
        benchmark::DoNotOptimize(1 + 2);
    }
    state.counters["counter"] = benchmark::Counter(1);
    state.counters["counter_rate"] = benchmark::Counter(1, benchmark::Counter::kIsRate);
    state.counters["counter_thread_rate"] = benchmark::Counter(1, benchmark::Counter::kAvgThreadsRate);
}

BENCHMARK(BM_IntegerAddition)
    ->Threads(1)
    ->Threads(10)
    ->Iterations(1)
    ->UseRealTime();
```

```
---------------------------------------------------------------------------------------------------------------
Benchmark                                                     Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------------------------------
BM_IntegerAddition/iterations:1/real_time/threads:1  1005048541 ns        29000 ns            1 counter=1 counter_rate=0.994977/s counter_thread_rate=0.994977/s
[...]

PR #2081, opened 2025-11-27
#2080: Fix rate and thread rate counter aggregates

FRosner:
## Changes

When using counters that represent a global rate (`benchmark::Counter::kIsRate`), before this PR, the rate was effectively computed per thread because we pass the sum of all seconds (wall or CPU time) passed across all threads. This breaks the definition of the global rate and subsequently, when using `kAvgThreadsRate`, the rate is divided by the number of threads (again), yielding non-sense results.

This is a regression introduced by https://github.com/google/benchmark/pull/1836. This PR fixes it by dividing the total seconds count by the number of threads before passing it to the counter finalization, which then computes the rates etc.

We're also fixing the test expectations.

## References

- Fixes https://github.com/google/benchmark/issues/2080.
- Requires https://github.com/google/benchmark/pull/2089

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
-    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
     - name: mount bazel cache
       uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
--- a/.github/workflows/build-and-test-min-cmake.yml
+++ b/.github/workflows/build-and-test-min-cmake.yml
@@ -19,9 +19,9 @@ jobs:
         os: [ubuntu-latest, macos-latest]
 
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
-      - uses: lukka/get-cmake@bb2faa721a800324b726fec00f7c1ff7641964d1 # latest
+      - uses: lukka/get-cmake@9e07ecdcee1b12e5037e42f410b67f03e2f626e1 # latest
         with:
           cmakeVersion: 3.13.0
 
--- a/.github/workflows/build-and-test-perfcounters.yml
+++ b/.github/workflows/build-and-test-perfcounters.yml
@@ -23,7 +23,7 @@ jobs:
         os: [ubuntu-latest]
         build_type: ['Release', 'Debug']
     steps:
-    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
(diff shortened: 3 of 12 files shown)


src/counter.cc
```diff
@@ -23,7 +23,12 @@ double Finish(Counter const& c, IterationCount iterations, double cpu_time,
...
-    v /= cpu_time;
+    // The CPU time is the total time taken by all threads (https://github.com/google/benchmark/pull/1836)
+    // If we used that as the denominator, we'd be calculating the rate per thread here. This is why
+    // we have to divide the total cpu_time by the number of threads for global counters to get a global rate.
+    //
+    // See https://github.com/google/benchmark/issues/2080
+    v /= (cpu_time / num_threads);
   }
   if ((c.flags & Counter::kAvgThreads) != 0) {
     v /= num_threads;
```
dmah42, line 29: i should know this but i've lost track: can flags be both `IsRate` and `AvgThreads`?  if so, are we then dividing twice incorrectly?

FRosner: > i should know this but i've lost track: can flags be both IsRate and AvgThreads

IIUC, yes, that's what `kAvgThreadsRate` will do:

https://github.com/google/benchmark/blob/2279f2acc8f2ca1bb51195a12966411bed31b26b/include/benchmark/benchmark.h#L655

>  if so, are we then dividing twice incorrectly?

I think it is correct. For `IsRate` we are multiplying by the number of threads (note the brackets, `a / (b / c)` = `a / b * c`). Then for `kAvgThreads` we are dividing by the number of threads again `a / b * c / c` = `a / b`, so we get what we'd expect for the per-thread average?

But this should just be tested in some unit tests. I need to check where the existing tests are.

src/counter.cc
```diff
@@ -23,7 +23,12 @@ double Finish(Counter const& c, IterationCount iterations, double cpu_time,
               double num_threads) {
   double v = c.value;
   if ((c.flags & Counter::kIsRate) != 0) {
-    v /= cpu_time;
+    // The CPU time is the total time taken by all threads (https://github.com/google/benchmark/pull/1836)
+    // If we used that as the denominator, we'd be calculating the rate per thread here. This is why
+    // we have to divide the total cpu_time by the number of threads for global counters to get a global rate.
+    //
+    // See https://github.com/google/benchmark/issues/2080
+    v /= (cpu_time / num_threads);
```
LebedevRI: Please hoist the `cpu_time / num_threads` into the outermost caller of `Finish()`.

FRosner: You mean here? 

https://github.com/google/benchmark/blob/6a8dee95ae1151142fe4be94482cdfd11b7111bd/src/benchmark_runner.cc#L126

LebedevRI: I _think_ so, yes.
Something like
```
// NOTE: <...>
double thread_seconds = seconds / num_threads;
internal::Finish(&report.counters, results.iterations, thread_seconds, 
```

FRosner: Alright! I'll get that done, and see what doc changes we might need, and also check the tests. If you have any pointers on where the tests are for this, let me know, otherwise I'll go search for it.

LebedevRI: Looking at https://github.com/google/benchmark/pull/1836,
i kinda expect that no tests will be affected by this,
which is why this regression wasn't caught in the first place.

FRosner: Ok I'll see where I can add some tests. I got sick last week so I didn't make any progress. I'll try to get to it this week.

FRosner:
I updated the docs in https://github.com/google/benchmark/pull/2081/commits/894fa391171bf910fdc08041e19da6bc31d47dd6. Let me know if you'd like me to add an example or if that's enough :)

LebedevRI (approved):

Issue #1896, opened 2024-12-31
[BUG] Certain performance counters are not measured and are always zero


double-fault:
**Describe the bug**

Certain counters such as `frontend_retired.l2_miss` are always reported as zero using the [performance counters integration with libpfm](https://github.com/google/benchmark/blob/main/docs/perf_counters.md). 

To confirm that this is indeed a bug I have made a small example where I'm flushing a function from all caches using the `clflush` instruction and then calling it, which should lead to an L2 miss. perf stat and record confirm that `frontend_retired.l2_miss` should be non-zero and the miss is happening within the function, however the counters reported by google benchmark are always zero. I have also noticed this issue with a few other events such as `frontend_retired.latency_ge_1`. 

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Ubuntu 22.04, Kernel 6.8.0-49-generic
  - Compiler and version: GNU assembler version 2.38 (x86_64-linux-gnu) using BFD version (GNU Binutils for Ubuntu) 2.38, g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
  - Processor: 11th Gen Intel(R) Core(TM) i5-11400H

**To reproduce**

Make a file `fun.S` with the following contents.

```
.section .data

.section .text

.global fun
.global not_fun

.align 512
not_fun:
        nop
        ret

.align 512
fun:
        clflush not_fun
        mfence
        call not_fun
        ret
```
The function `fun` flushes `not_fun` from all the caches and then calls it, which will cause a cache miss as it tries to fetch the first instruction of the function. 

Make two more files, `prog1.S` and `prog2.cpp` respectively with the following contents.

```
.section .data

.section .text

.global _start
_start:
        mov $10000000, %rcx

loop:
        call fun
        dec %rcx
        jnz loop       

        xor %rdi,%rdi
        mov $60, %rax
        syscall
```
```
#include <benchmark/benchmark.h>

extern "C" void fun();

void BM(benchmark::State &state) {
        for (auto _ : state) {
                fun();
        }
}

BENCHMARK(BM);

BENCHMARK_MAIN();
```
and a Makefile:
```
all: fun.o prog1.o prog2.o
	g++ prog1.o fun.o -o prog1.out -std=c++17 -nostdlib -no-pie -O3 -Wall -Werror -fno-omit-frame-pointer -isystem benchmark/include -Lbenchmark/build/src -lbenchmark -lpthread -lpfm 
[...]

PR #2083, opened 2025-11-28
Add performance counter multiplexing

### LINKED ISSUE #1813: [BUG] Perf counters not multiplexed

double-fault:
Fixes #1813, #1896. 

Lot of the code already existed but it was incomplete, like tracking of multiple leaders already existed but there was no extrapolation based off time running.

There were multiple groups in #1896 and the later groups always returned zero as the first group was pinned.


dmah42 (approved):

dmah42:
i think this looks good. i've approved but maybe @LebedevRI wants to also weigh in.

double-fault:
@LebedevRI ping

LebedevRI:
> @LebedevRI ping

I've already provided all the feedback that i have.

src/perf_counters.cc
```diff
@@ -29,29 +28,37 @@ namespace internal {
...
-      // Actual data bytes are all bytes minus initial padding
-      std::size_t data_bytes =
-          static_cast<std::size_t>(read_bytes) - sizeof(uint64_t);
-      // This should be very cheap since it's in hot cache
-      std::memmove(ptr, ptr + sizeof(uint64_t), data_bytes);
-      // Increment our counters
-      ptr += data_bytes;
-      size -= data_bytes;
-    } else {
+    ssize_t read_bytes = ::read(lead, &buffer, sizeof(buffer));
```
LebedevRI, line 47: This `struct` thing is nice, but i'm like 80% sure there is UB here.
The most safe way would be to read into `array<uint64_t, 3+kMaxCounters>`,
and then produce that `buffer` struct from it.

LebedevRI: Also, can that be sunk into `for (int lead : leaders) {`?

src/perf_counters.h
```diff
@@ -39,46 +39,77 @@
...
+   public:
+    void Set(uint64_t time_enabled, uint64_t time_running, uint64_t value) {
+      time_enabled_ = time_enabled;
+      time_running_ = time_running;
+      value_ = value;
+    }
+
+    double GetEstimate() const {
+      return static_cast<double>(value_ * time_enabled_) /
+             static_cast<double>(time_running_);
```
LebedevRI, line 88: This looks iffy. What are the typical perf counter value ranges?
And what about the `time_enabled`/`time_running`,
what units are they? Cycles?

We are either going to be losing precision here,
and/or having a multiplication overflow.

LebedevRI (changes requested):

dmah42:
@double-fault will you be able to resolve these comments?

double-fault:
Yeah I think I had looked at the comments I should be able to make the fixes, was a bit busy and then forgot about it lol. I'll do it this/next week, thanks for the reminder

LebedevRI:
> Yeah I think I had looked at the comments I should be able to make the fixes, was a bit busy and then forgot about it lol. I'll do it this/next week, thanks for the reminder

Thank you! IIRC other than those nits, this was fine otherwise. But i didn't test this.

double-fault:
Cool, IIRC I could not figure out how to test the changes objectively outside of obviously just running the code on some benchmarks and seeing if the values make intuitive sense. I'll have a look at the existing tests to see if I can extend those to include this case after making the fixes

rael346:
@double-fault hi, any updates on this, I'm seeing similar issue with L2 related info

Issue #2078, opened 2025-11-22
[BUG] Unneccesary re-run of cxx_feature_check on cmake config re-run


theComputeKid:
It is not necessary to run `cxx_feature_check` for `HAVE_GNU_POSIX_REGEX`, `HAVE_POSIX_REGEX`, `HAVE_STEADY_CLOCK` or `HAVE_PTHREAD_AFFINITY` on every cmake re-config.

E.g. log on cmake config:

```
-- Google Benchmark version: v1.9.4, normalized to 1.9.4
-- Performing Test HAVE_GNU_POSIX_REGEX -- failed to compile
-- Performing Test HAVE_POSIX_REGEX -- failed to compile
-- Performing Test HAVE_STEADY_CLOCK -- success
-- Performing Test HAVE_PTHREAD_AFFINITY -- failed to compile
```
No compiler tests should be performed on a re-config as it is tied to the C/CXX compiler.

The issue is because of the following code:
```
cxx_feature_check(STD_REGEX)
cxx_feature_check(GNU_POSIX_REGEX)
cxx_feature_check(POSIX_REGEX)
if(NOT HAVE_STD_REGEX AND NOT HAVE_GNU_POSIX_REGEX AND NOT HAVE_POSIX_REGEX)
  message(FATAL_ERROR "Failed to determine the source files for the regular expression backend")
endif()
```

This is not necessary. A suggestion to solve this is to change `cxx_feature_check` such that it sets the necessary feature variables in the cache instead of a normal variable. On config re-run, CMake will see that the variable is already set and not run the tests again.

To reproduce, run this twice and inspect the log:
```
cmake -B build -G Ninja -DBENCHMARK_DOWNLOAD_DEPENDENCIES=1
```

Expected behaviour: the second call to cmake should be a no-op and run no compiler tests, and print nothing to the console.

LebedevRI:
Please feel free to submit a patch.

PR #2084, opened 2025-11-28
Remove redundant feature checks on re-run of CMake config step

theComputeKid:
- On a config re-run, CMake will not check features again as it will read previously defined cache variables
- Logic was difficult to follow, so refactored `cxx_feature_check` for simplicity

On CMake config first run:

Before (slow): checks all REGEX engines
After (fast): short circuits on first REGEX engine found.

On CMake config re-run:

Before (slow due to checks):
```
-- Google Benchmark version: v1.9.4, normalized to 1.9.4
-- Performing Test HAVE_PTHREAD_AFFINITY -- failed to compile
```

After (fast as checks skipped): No output (as checks not performed)

Fixes https://github.com/google/benchmark/issues/2078

### DIFF
--- a/cmake/CXXFeatureCheck.cmake
+++ b/cmake/CXXFeatureCheck.cmake
@@ -62,6 +62,8 @@ function(cxx_feature_check FEATURE)
     )
     if(COMPILE_STATUS)
       set(RUN_STATUS 0)
+      message(WARNING
+              "If you see build failures due to cross compilation, try setting ${VAR} to 0")
     endif()
   else()
     cxx_feature_check_print("Compiling and running to test ${FEATURE}")


cmake/CXXFeatureCheck.cmake
```diff
@@ -10,22 +10,30 @@
 #
 # include(CXXFeatureCheck)
 # cxx_feature_check(STD_REGEX)
-# Requires CMake 2.8.12+
+# Requires CMake 3.13+
 
-if(__cxx_feature_check)
-  return()
-endif()
```
LebedevRI, line 17: Why is double-include guard is being removed?

HamzaAB0: Fixed. That was an oversight. Thanks.

cmake/CXXFeatureCheck.cmake
```diff
@@ -35,48 +48,51 @@ function(cxx_feature_check FILE)
...
-      try_compile(COMPILE_${FEATURE}
-              ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/cmake/${FILE}.cpp
-              CXX_STANDARD 17
-              CXX_STANDARD_REQUIRED ON
-              CMAKE_FLAGS ${FEATURE_CHECK_CMAKE_FLAGS}
-              LINK_LIBRARIES ${BENCHMARK_CXX_LIBRARIES}
-              OUTPUT_VARIABLE COMPILE_OUTPUT_VAR)
-      if(COMPILE_${FEATURE})
-        message(WARNING
-              "If you see build failures due to cross compilation, try setting HAVE_${VAR} to 0")
```
LebedevRI, line 50: Why?

cmake/CXXFeatureCheck.cmake
```diff
@@ -35,48 +48,51 @@ function(cxx_feature_check FILE)
...
+      COMPILE_OUTPUT_VARIABLE COMPILE_OUTPUT
+      RUN_OUTPUT_VARIABLE RUN_OUTPUT
+    )
   endif()
 
-  if(COMPILE_${FEATURE})
-    if(DEFINED RUN_${FEATURE} AND RUN_${FEATURE} EQUAL 0)
-      message(STATUS "Performing Test ${FEATURE} -- success")
-      set(HAVE_${VAR} 1 PARENT_SCOPE)
-      add_definitions(-DHAVE_${VAR})
```
LebedevRI, line 71: Why?

theComputeKid:
> Is this AI-made?

No.

cmake/CXXFeatureCheck.cmake
HamzaAB0: Why what?

LebedevRI: Since the line is only on the LHS of the diff, why this line and the other snippet being removed.
This literally renders this whole check pointless, since we no longer communicate it's results.

cmake/CXXFeatureCheck.cmake
HamzaAB0: Why did I remove this? Because it is redundant, setting HAVE_${VAR} overrides detection regardless of whether cross-compilation is happening or not. If we want that message, we can put it in. Or perhaps you can explain more and I can put it back in.

cmake/CXXFeatureCheck.cmake
HamzaAB0: I am not sure what you mean as the `CACHE` token here is communicating results at the parent scope:
```
set(${VAR} TRUE CACHE BOOL "" FORCE)
```
No other variables need to escape the function scope.

cmake/CXXFeatureCheck.cmake
LebedevRI: I guess my main point here is that, in addition to fixing the issue it fixes,
it also arbitrairly removes some externally-observable side-effects,
and i don't get why it does that.

cmake/CXXFeatureCheck.cmake
LebedevRI: https://github.com/google/benchmark/blob/2279f2acc8f2ca1bb51195a12966411bed31b26b/src/re.h#L22C14-L49

HamzaAB0: Yes, `HAVE_STD_REGEX`, `HAVE_GNU_POSIX_REGEX` or `HAVE_POSIX_REGEX` are being properly propagated in my patch, depending on which REGEX engine is detected. If it wasn't, that file wouldn't compile with my changes and CI would fail.

As an example, in my local machine, it detects HAVE_STD_REGEX and does the right thing. 

I am not sure I see the issue.

cmake/CXXFeatureCheck.cmake
HamzaAB0: Fixed. Re-added warning to regain previous behaviour.

cmake/CXXFeatureCheck.cmake
LebedevRI: > Yes, `HAVE_STD_REGEX`, `HAVE_GNU_POSIX_REGEX` or `HAVE_POSIX_REGEX` are being properly propagated in my patch, depending on which REGEX engine is detected.

cmake cache variables - sure.

Preprocessor defines - no, they are not.
Could you please point me at where they are being set.

The answer i was looking for was: "are you illiterate or something, just on the line above, there's `add_compile_definitions()`".

> If it wasn't, that file wouldn't compile with my changes and CI would fail.

This is not quite true, as you can see from that snippet, it has fallback defaults.

PR #2087, opened 2025-12-05
Correct identification macro for cuda compiler diagnostic control

m00978:
* According to NVIDIA CUDA compiler driver NVCC specification


LebedevRI:
Could you please demonstrate the issue without this patch,
and that it is gone with this patch applied? (Also, which version of the compiler?)

PR #2101, opened 2026-01-12
Make the Benchmark class public

jmr:
Move ::benchmark::internal::Benchmark to ::benchmark::Benchmark.

It's a bit odd that the documented way to pass arguments to a benchmark is with `benchmark::internal::Benchmark`.  Make this public instead.

https://github.com/google/benchmark/blob/v1.9.4/docs/user_guide.md#passing-arguments

https://raw.githubusercontent.com/google/benchmark/refs/tags/v1.9.4/docs/user_guide.md#:~:text=void%20CustomArguments(benchmark%3A%3A-,internal%3A%3ABenchmark,-*%20b)%20%7B%0A%20%20for%20(int

Keep ::benchmark::internal::Benchmark as a deprecated forwarding alias. Uses of Benchmark in the internal namespace need to explicitly use ::benchmark::Benchmark to avoid the deprecation warning.

### DIFF
--- a/include/benchmark/benchmark.h
+++ b/include/benchmark/benchmark.h
@@ -103,7 +103,7 @@ BENCHMARK(BM_SetInsert)->Ranges({{1<<10, 8<<10}, {128, 512}});
 // arbitrary set of arguments to run the microbenchmark on.
 // The following example enumerates a dense range on
 // one parameter, and a sparse range on the second.
-static void CustomArguments(benchmark::internal::Benchmark* b) {
+static void CustomArguments(benchmark::Benchmark* b) {
   for (int i = 0; i <= 10; ++i)
     for (int j = 32; j <= 1024*1024; j *= 8)
       b->Args({i, j});
@@ -488,8 +488,9 @@ void RegisterProfilerManager(ProfilerManager* profiler_manager);
 BENCHMARK_EXPORT
 void AddCustomContext(std::string key, std::string value);
 
-namespace internal {
 class Benchmark;
+
+namespace internal {
 class BenchmarkImp;
 class BenchmarkFamilies;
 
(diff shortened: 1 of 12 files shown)


dmah42:
I like this change a lot, it's always bothered me that we expose "internal".  The window failures look like legitimate issues @jmr .

dmah42:
this looks good, but will users of the library see broken builds if they're using `internal`? or will it only be an issue if they have `-Werror`?

jmr:
> it's an API change

It's an ABI change, but the old API continues to work.

> i think we'll need a release after this too.

Why? Isn't `master` just whatever is being worked on now?

dmah42:
only because there's been a few changes since the last release and this is relatively significant.  also it's nicer to import internally point releases rather than commit SHAs.

include/benchmark/benchmark.h
```diff
@@ -1125,20 +1126,17 @@ struct ThreadRunnerBase {
   virtual void RunThreads(const std::function<void(int)>& fn) = 0;
 };
 
-namespace internal {
-
 // Define alias of ThreadRunner factory function type
 using threadrunner_factory =
     std::function<std::unique_ptr<ThreadRunnerBase>(int)>;
 
 typedef void(Function)(State&);
```
LebedevRI: Missing typedefs for these

src/benchmark_register.cc
```diff
@@ -206,13 +208,15 @@ bool FindBenchmarksInternal(const std::string& re,
...
+}  // end namespace internal
+
 //=============================================================================//
 //                               Benchmark
 //=============================================================================//
 
 Benchmark::Benchmark(const std::string& name)
     : name_(name),
-      aggregation_report_mode_(ARM_Unspecified),
+      aggregation_report_mode_(internal::ARM_Unspecified),
```
LebedevRI, line 219: Should the enum be made non-`internal::` too?

jmr: Doesn't look user-facing: https://github.com/search?q=repo%3Agoogle%2Fbenchmark+aggregationreportmode&type=code

include/benchmark/benchmark.h
jmr: Are these intended to be user-facing?  Maybe they should go back to internal?

LebedevRI: I'm not sure about `Function`, `FunctionBenchmark` and such,
but that whole `threadrunner_factory` thing
is clearly meant to be externally-avaliable.

jmr: Done.

nunezmati185-arch: Si

PR #2105, opened 2026-01-19
Optimization to Skip this entire family 

urmez:


src/benchmark_register.cc
```diff
@@ -166,6 +166,26 @@ bool BenchmarkFamilies::FindBenchmarks(
...
+    // Optimization: Quick check if family name could match the filter
+    // before iterating through all args combinations
+    bool family_name_disabled = family->name_.rfind(kDisabledPrefix, 0) == 0;
+    bool family_could_match = false;
+    
+    if (!family_name_disabled) {
+      // Check if the filter spec starts with the family name
+      // or if the family name appears as a substring in the spec
+      family_could_match = spec.find(family->name_) != std::string::npos || 
+                           spec == "." ||
```
LebedevRI: The `.` special case should probably be checked first.

src/benchmark_register.cc
```diff
@@ -166,6 +166,26 @@ bool BenchmarkFamilies::FindBenchmarks(
       benchmarks->reserve(benchmarks->size() + family_size);
     }
 
+    // Optimization: Quick check if family name could match the filter
+    // before iterating through all args combinations
+    bool family_name_disabled = family->name_.rfind(kDisabledPrefix, 0) == 0;
```
LebedevRI: Why `rfind()` though, instead of `find()`?
That being said i think it might make sense to define a helper function `string_has_prefix()` for this...

urmez: No rfind now. let me know what you think of latest code.

test/filter_optimization_test.cc
```diff
@@ -0,0 +1,115 @@
...
+    // Verify optimization provides at least 5x speedup
+    if (speedup < 5.0) {
+      std::cerr << "ERROR: Expected at least 5x speedup, got " << speedup << "x\n";
+      std::cerr << "Optimization may not be working correctly!\n";
+      return -1;
+    }
+  } else {
+    std::cerr << "ERROR: Literal filter completed too fast to measure accurately\n";
+    return -1;
+  }
```
LebedevRI, line 100: It is not a good idea to make tests depend on host machie performance.

src/benchmark_register.cc
```diff
@@ -166,6 +171,23 @@ bool BenchmarkFamilies::FindBenchmarks(
...
+    // we can skip entire families if the family name doesn't contain the literal.
+    // This is safe because all instances will have names starting with the family name.
+    // For positive filters: skip if literal not found in family name
+    // For negative filters: skip if literal IS found (all instances will match the negative filter)
+    if (is_literal && !family->name_.empty()) {
+      bool family_contains_literal = family->name_.find(spec) != std::string::npos;
+      if (!is_negative_filter && !family_contains_literal) {
+        // Positive filter: family name doesn't contain literal, skip family
+        continue;
+      }
```
LebedevRI, line 184: Won't this errneously skip `BM_Family/arg1/arg2` given filter `BM_Family/arg` ?

LebedevRI (changes requested):

PR #2123, opened 2026-02-06
Replicate generic hardware events on all CPU PMUs

pcc:
On systems with more than one PMU for the CPUs (e.g. Apple M series SOCs), generic hardware events are only created for an arbitrary PMU. Usually this is the big cluster's PMU, which can cause inaccuracies when the process is scheduled onto a little core. To fix this, teach PerfCounters to register generic hardware events on all CPU PMUs.

CPU PMUs are identified using the same method as perf.

### DIFF
--- a/src/perf_counters.cc
+++ b/src/perf_counters.cc
@@ -16,9 +16,15 @@
 
 #include <cstring>
 #include <memory>
+#include <optional>
 #include <vector>
 
 #if defined HAVE_LIBPFM
+#include <dirent.h>
+#include <fcntl.h>
+#include <linux/perf_event.h>
+#include <sys/stat.h>
+
 #include "perfmon/pfmlib.h"
 #include "perfmon/pfmlib_perf_event.h"
 #endif
@@ -68,7 +74,7 @@ bool PerfCounters::Initialize() {
 
 bool PerfCounters::IsCounterSupported(const std::string& name) {
   Initialize();
-  perf_event_attr_t attr;
+  perf_event_attr attr;
   std::memset(&attr, 0, sizeof(attr));
   pfm_perf_encode_arg_t arg;
   std::memset(&arg, 0, sizeof(arg));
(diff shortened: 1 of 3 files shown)


src/perf_counters.cc
```diff
@@ -14,6 +14,10 @@
 
 #include "perf_counters.h"
 
+#include <dirent.h>
```
LebedevRI: This header seems to not exist on windows.
(i hadn't realized we were building with pfm on that platform, ha.)

src/perf_counters.cc
```diff
@@ -158,50 +205,54 @@ PerfCounters PerfCounters::Create(
...
-      static constexpr size_t kNrOfSyscallRetries = 5;
-      // Retry syscall as it was interrupted often (b/64774091).
-      for (size_t num_retries = 0; num_retries < kNrOfSyscallRetries;
-           ++num_retries) {
-        id = perf_event_open(&attr, 0, -1, group_id, 0);
-        if (id >= 0 || errno != EINTR) {
-          break;
+    uint64_t base_config = attr.config;
+    for (uint64_t pmu : GetPMUTypesForEvent(attr)) {
+      attr.config = (pmu << 32) | base_config;
```
captain5050: Fwiw, the 32 constant is PERF_PMU_TYPE_SHIFT in perf_event.h:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/perf_event.h#n55

src/perf_counters.cc
pcc: Looks like this file is built unconditionally while most of the code is guarded by `HAS_PFM`, which I imagine would only be defined on Linux due to the many Linux dependencies in libpfm. So I moved my new header inclusions to the `HAS_PFM` block.

src/perf_counters.cc
pcc: Done

src/perf_counters.cc
```diff
@@ -79,6 +84,49 @@ bool PerfCounters::IsCounterSupported(const std::string& name) {
   return (ret == PFM_SUCCESS);
 }
 
+static std::vector<uint64_t> GetPMUTypesForEvent(const perf_event_attr& attr) {
+  // Replicate generic hardware events on all CPU PMUs.
+  if (attr.type == PERF_TYPE_HARDWARE && attr.config < PERF_COUNT_HW_MAX) {
+    std::vector<uint64_t> types;
+    DIR* dir = opendir("/sys/bus/event_source/devices");
+    if (!dir) {
+      return {0};
```
LebedevRI: Nitpick: i think it would be cleaner to have `QueryPMUTypesForEvent()`
that would return `None` when failing to read from the `/sys`,
and have `GetPMUTypesForEvent()` return the fall-back `{0}`.

LebedevRI:
Does this need any kind of guarding or is this always the right thing to do?

src/perf_counters.cc
pcc: I guess you mean `std::nullopt`? Done.

pcc:
> > can we add google test unit tests for this please? i don't know how tricky it is to test it but i'm a little concerned having no tests at all.
> 
> We can test it by verifying that counters are non-zero when the process is pinned to each CPU in the system. We can adapt the test ` PerfCountersTest.Read1Counter` for this purpose. Actually, on my Mac Studio that test fails at head (after patching BCR to fix the build failures with `--define pfm=1`: [bazelbuild/bazel-central-registry#7502](https://github.com/bazelbuild/bazel-central-registry/pull/7502) [bazelbuild/bazel-central-registry#7503](https://github.com/bazelbuild/bazel-central-registry/pull/7503)) and it also fails after this change (because it didn't expect more than one counter with the same name). I'll take care of that as well.

Done

dmah42 (dismissed):

Issue #1897, opened 2025-01-06
[FR]


double-fault:
**Is your feature request related to a problem? Please describe.**

I have a multi-threaded benchmark where I'd like the performance counters to only measure the main thread. 

**Describe the solution you'd like**

Setting the `inherit` attribute to `false` [instead of `true`](https://github.com/google/benchmark/blob/main/src/perf_counters.cc#L151) before passing the struct to `perf_event_open` would solve my issue. Some sort of configuration option which allows me to set this would be ideal. Maybe this can be extended to other attributes as well instead of hardcoding them. 

**Describe alternatives you've considered**

Currently, I have to set it to `false` and recompile google benchmark.

double-fault:
@dmah42 do you have any suggestions on what the interface to set the config should look like - as in, should it be a cmd line argument, an argument in the benchmark macros, etc. If you can suggest what the interface should be like maybe I can implement it when I get some time.

dmah42:
i think a command line argument is probably the right way to go as it affects the entire run.

shreejaykurhade:
Implemented and tested

dmah42:
the PR brought up some concerns (from me) about this request. @double-fault can you be more concrete about the use-case this would solve for you?

LebedevRI:
Perhaps the decision "all threads vs main thread" could be based on `->MeasureProcessCPUTime()` thing?
I'm not sure what should happen for custom user timers then though.

dmah42:
blending it in to an existing configuration may be even more confusing.  i'm just eager to understand the use-case before adding Yet Another Config Option.

double-fault:
Went through some old code, it seems like one of the things I was trying to benchmark is a futex wakeup - so the same process would spawn two threads, but you only want to measure the main thread. Seems like a good purpose to add some sort of config option. 

Eg. code of mine I found: https://github.com/double-fault/microbenchmarks/blob/main/futex/main.cpp I think I had set `inherit` to false and recompiled google benchmark locally to make these benchmarks meaningful, but can't really remember.

LebedevRI:
As per disscussion in https://github.com/google/benchmark/pull/2179, the desired behaviour has not been convincing,
so closing for now.

PR #2129, opened 2026-02-18
Allow Mod of Perf Attributes

IshanPhadte776:
- First Contribution I have made for open source, Some formatting issues might be present 

- Fixing this issue (https://github.com/google/benchmark/issues/1897) 

I added an optional Command Line Argument 
-- benchmark_perf_counters_inherit (Either True or False) and True by Default 

True = Benchmark Parent and Child Processes 
False = Benchmark just the parent process 

True if unspecified

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -141,6 +141,9 @@ BM_DEFINE_bool(benchmark_counters_tabular, false);
 // information about libpfm: https://man7.org/linux/man-pages/man3/libpfm.3.html
 BM_DEFINE_string(benchmark_perf_counters, "");
 
+// Setting this to false allows measuring only the main benchmark thread.
+BM_DEFINE_bool(benchmark_perf_counters_inherit, true);
+
 // Extra context to include in the output formatted as comma-separated key-value
 // pairs. Kept internal as it's only used for parsing from env/command line.
 BM_DEFINE_kvpairs(benchmark_context, {});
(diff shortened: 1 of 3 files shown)


LebedevRI (commented):
I think `inherit` is too undecipherable,
it should probably be `benchmark_perf_counters_for_all_threads` / `benchmark_perf_counters_all_threads` or something..

src/benchmark.cc
```diff
@@ -891,6 +898,7 @@ void PrintDefaultHelp() {
           "          [--benchmark_counters_tabular={true|false}]\n"
 #if defined HAVE_LIBPFM
           "          [--benchmark_perf_counters=<counter>,...]\n"
+          "          [--benchmark_perf_counters_inherit={true|false}]\n" // <--- Add this
```
dmah42, line 901: remove the AI generated comment please

src/perf_counters.cc
```diff
@@ -135,7 +135,7 @@ static std::vector<uint64_t> GetPMUTypesForEvent(const perf_event_attr& attr) {
 }
 
 PerfCounters PerfCounters::Create(
-    const std::vector<std::string>& counter_names) {
+    const std::vector<std::string>& counter_names, bool inherit) {
```
dmah42, line 138: boolean parameters are notoriously hard to track as the callsite becomes opaque : https://abseil.io/tips/94

src/perf_counters.cc
```diff
@@ -311,20 +312,21 @@ bool PerfCounters::Initialize() { return false; }
 bool PerfCounters::IsCounterSupported(const std::string&) { return false; }
 
 PerfCounters PerfCounters::Create(
```
dmah42, line 314: i think this would be clearer as two methods: Create and CreateInherited (or whatever naming you land on)

LebedevRI (changes requested):

5kng:
Is this still being worked on? Happy to give it a stab. @IshanPhadte776

PR #2135, opened 2026-02-27
Add BENCHMARK_NAMED macro for named benchmarks without lambda

### LINKED ISSUE #2128: Large number of benchmarks exposes compiler/linker scalability issues
We generate a very large number of benchmarks in our project, many thousands in some cases. We mostly use `BENCHMARK_CAPTURE`, but not to actually capture arguments, we just need to give the benchmark a unique name.

The problem is `BENCHMARK_CAPTURE` has an implied lambda, and this lambda contributes heavily to the compiler/linker issues. This change: https://github.com/google/XNNPACK/commit/4a547fef38f9f6b8d50c16b6aac53d1ba35ca54b#diff-e5c2e748172f56235fc68e43d4cfe0e530a3874cab87dccb3d5216bdda5f2263R89 cut compile/link times by 2x on good linkers, and exponentially in bad linkers.

Of course, the above change uses a bunch of "private" benchmark implementation details, which we'd rather not do.

This request is to provide a way of doing the same thing without using private APIs.

A more general solution to this problem might be a way to customize the name of a benchmark specified with `Args`, which would be more generally useful.

5kng:
Closes #2128. BENCHMARK_CAPTURE creates a lambda even when no arguments are captured, causing compiler/linker scalability issues with thousands of benchmarks. BENCHMARK_NAMED provides the same func/name format but passes the function pointer directly (no lambda), consistent with the existing BENCHMARK macro.

### DIFF
--- a/test/benchmark_test.cc
+++ b/test/benchmark_test.cc
@@ -234,7 +234,6 @@ void BM_non_template_args(benchmark::State& state, int, double) {
 }
 BENCHMARK_CAPTURE(BM_non_template_args, basic_test, 0, 0);
 
-
 template <class T, class U, class... ExtraArgs>
 void BM_template2_capture(benchmark::State& state, ExtraArgs&&... extra_args) {
   static_assert(std::is_same<T, void>::value, "");


test/benchmark_test.cc
```diff
@@ -234,6 +234,15 @@ void BM_non_template_args(benchmark::State& state, int, double) {
 }
 BENCHMARK_CAPTURE(BM_non_template_args, basic_test, 0, 0);
 
+// Test BENCHMARK_NAMED: same function, named variant, no lambda
```
LebedevRI: This doesn't actually test anything.

5kng: Oops. I moved the test to register_benchmark_test.cc and added ADD_CASES assertions to verify that the registered benchmark names match the expected func/test_case_name format, inclduing chaining w/ ->Threads().

LebedevRI:
> Please fix clang-format nit

Almost there...

PR #2127, opened 2026-02-17
Revert "Remove redundant feature checks on re-run of CMake config ste…

dsharlet:
…p (#2084)"

This reverts commit 6a8dee95ae1151142fe4be94482cdfd11b7111bd.

This commit has broken our build, e.g.: 
```
-- Compiling and running to test GNU_POSIX_REGEX
CMake Error at build/linux/armhf/googlebenchmark-source/CMakeLists.txt:330 (message):
  Failed to determine the source files for the regular expression backend


-- Performing Test GNU_POSIX_REGEX -- failed
-- Compile Output: Change Dir: '/__w/XNNPACK/XNNPACK/build/linux/armhf/CMakeFiles/CMakeTmp'

Run Build Command(s): /usr/bin/ninja -v cmTC_0f13b
[1/2] ccache /usr/bin/arm-linux-gnueabihf-g++   -Wall  -Wextra  -Wshadow  -Wfloat-equal  -Wold-style-cast  -Wconversion  -Wformat=2  -Werror  -Wsuggest-override  -pedantic  -pedantic-errors  -fstrict-aliasing  -Wno-deprecated-declarations  -Wno-deprecated  -Wstrict-aliasing  -std=c++17 -o CMakeFiles/cmTC_0f13b.dir/gnu_posix_regex.cpp.o -c /__w/XNNPACK/XNNPACK/build/linux/armhf/googlebenchmark-source/cmake/gnu_posix_regex.cpp
FAILED: CMakeFiles/cmTC_0f13b.dir/gnu_posix_regex.cpp.o 
ccache /usr/bin/arm-linux-gnueabihf-g++   -Wall  -Wextra  -Wshadow  -Wfloat-equal  -Wold-style-cast  -Wconversion  -Wformat=2  -Werror  -Wsuggest-override  -pedantic  -pedantic-errors  -fstrict-aliasing  -Wno-deprecated-declarations  -Wno-deprecated  -Wstrict-aliasing  -std=c++17 -o CMakeFiles/cmTC_0f13b.dir/gnu_posix_regex.cpp.o -c /__w/XNNPACK/XNNPACK/build/linux/armhf/googlebenchmark-source/cmake/gnu_posix_regex.cpp
/__w/XNNPACK/XNNPACK/build/linux/armhf/googlebenchmark-source/cmake/gnu_posix_regex.cpp:1:10: fatal error: gnuregex.h: No such file or directory
    1 | #include <gnuregex.h>
      |          ^~~~~~~~~~~~
[...]

PR #2137, opened 2026-02-27
cmake: cache HAVE_* results to skip feature checks on re-runs

### LINKED ISSUE #2078: [BUG] Unneccesary re-run of cxx_feature_check on cmake config re-run

5kng:
Without caching, HAVE_${VAR} was lost between cmake runs, so the early-exit guard never fired and all feature test messages re-printed on every reconfigure.

Now HAVE_${VAR} is written to the CMake cache (INTERNAL) on both success and failure. Also fix the guard to not promote a cached 0 to 1.

Fixes #2078. A previous attempt (#2084) was reverted in #2127 due to a cross-compilation bug; @dsharlet noted the CXXFeatureCheck.cmake change had a bug - this fixes it.


cmake/CXXFeatureCheck.cmake
```diff
@@ -67,16 +69,19 @@ function(cxx_feature_check FILE)
   if(COMPILE_${FEATURE})
     if(DEFINED RUN_${FEATURE} AND RUN_${FEATURE} EQUAL 0)
       message(STATUS "Performing Test ${FEATURE} -- success")
+      set(HAVE_${VAR} 1 CACHE INTERNAL "")
       set(HAVE_${VAR} 1 PARENT_SCOPE)
```
LebedevRI: Is there a point in also setting it in the parent scope when it's already set in cache?

5kng: Yeah fair point, the cache variable is already visible in the parent scope, so the PARENT_SCOPE set is redundant. I'll remove it now.

cmake/CXXFeatureCheck.cmake
```diff
@@ -24,8 +24,10 @@ function(cxx_feature_check FILE)
   string(TOUPPER ${FILE} VAR)
   string(TOUPPER "HAVE_${VAR}" FEATURE)
   if (DEFINED HAVE_${VAR})
-    set(HAVE_${VAR} 1 PARENT_SCOPE)
-    add_definitions(-DHAVE_${VAR})
+    if(HAVE_${VAR})
+      set(HAVE_${VAR} 1 PARENT_SCOPE)
```
LebedevRI: Hm, and perh here too?

5kng: Removed too, ty.

PR #2138, opened 2026-02-27
refactor ToExponentAndMantissa to use StrFormat instead of std::stringstream

dmah42:


src/string_util.cc
```diff
@@ -33,11 +33,11 @@ const int64_t kUnitsSize = arraysize(kBigSIUnits);
 
 void ToExponentAndMantissa(double val, int precision, double one_k,
                            std::string* mantissa, int64_t* exponent) {
```
LebedevRI: IMHO this should be returning `std::pair<std::string, int64_t>`

dmah42: i don't hate output parameters, but they should at least be non-const references i think.  but i'm also with returning a pair which does make sense for this API.

src/string_util.cc
```diff
@@ -70,20 +69,18 @@ void ToExponentAndMantissa(double val, int precision, double one_k,
...
-          mantissa_stream << scaled;
+          *mantissa += StrFormat("%g", scaled);
           *exponent = -static_cast<int64_t>(i + 1);
-          *mantissa = mantissa_stream.str();
           return;
         }
       }
     }
-    mantissa_stream << val;
+    *mantissa += StrFormat("%g", val);
```
LebedevRI: I'd suggest extracting `StrFormat("%g", x)` into a lambda

src/string_util.cc
LebedevRI: (No strong opinion, consider it a suggestion)

src/string_util.cc
dmah42: that might make it less explicit/less readable. i'd rather not.

LebedevRI: My point is that all these calls are meant to have the same format specifier,
but now each call site provides it's own format string.
If not a lambda, i'd suggest to at least define a macro for `"%d"`

dmah42: done

src/string_util.cc
```diff
@@ -49,41 +51,41 @@ void ToExponentAndMantissa(double val, int precision, double one_k,
...
+    format_mantissa(val);
+    exponent = 0;
   } else {
-    mantissa_stream << val;
-    *exponent = 0;
+    format_mantissa(val);
+    exponent = 0;
   }
-  *mantissa = mantissa_stream.str();
+  return std::make_pair(mantissa, exponent);
```
LebedevRI, line 92: Nit: it might be better to sink return into each of the branches.

src/string_util.cc
```diff
@@ -104,12 +106,11 @@ std::string ExponentToPrefix(int64_t exponent, bool iec) {
 
 std::string ToBinaryStringFullySpecified(double value, int precision,
                                          Counter::OneK one_k) {
-  std::string mantissa;
-  int64_t exponent = 0;
-  ToExponentAndMantissa(value, precision,
-                        one_k == Counter::kIs1024 ? 1024.0 : 1000.0, &mantissa,
-                        &exponent);
-  return mantissa + ExponentToPrefix(exponent, one_k == Counter::kIs1024);
+  auto mantissa_and_exponent = ToExponentAndMantissa(
```
LebedevRI: I'm not sure if structured bindings can be used nowadays?
This could be something like `auto [mantissa, exponent] = ToExponentAndMantissa(<...>`

LebedevRI (approved):

PR #2139, opened 2026-02-28
multiple fixes to avoid Windows x64 crashes

dmah42:
* fix: cast size_t widths to int for variadic printer to avoid Windows x64 issues with bitwidth differences
* extend minimum time as Windows can have a coarse timer (15ms)
* vsnprintf can consume va_list so we need to copy it to avoid UB

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -184,7 +184,7 @@ compile_output_test(templated_fixture_method_test)
 benchmark_add_test(NAME templated_fixture_method_test COMMAND templated_fixture_method_test --benchmark_min_time=0.01s)
 
 compile_output_test(user_counters_test)
-benchmark_add_test(NAME user_counters_test COMMAND user_counters_test --benchmark_min_time=0.01s)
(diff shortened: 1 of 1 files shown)


src/console_reporter.cc
```diff
@@ -196,8 +196,8 @@ void ConsoleReporter::PrintRunData(const Run& result) {
       }
     }
     if ((output_options_ & OO_Tabular) != 0) {
-      printer(Out, COLOR_DEFAULT, " %*s%s", cNameLen - strlen(unit), s.c_str(),
-              unit);
+      printer(Out, COLOR_DEFAULT, " %*s%s",
+              static_cast<int>(cNameLen - strlen(unit)), s.c_str(), unit);
```
LebedevRI, line 200: I'm not really sure why this would be needed.
Other changes make sense.

dmah42: same as the other issues.. `cNameLen - strlen(unit)` is a size_t, so 64-bit on windows-64, and we're putting it into a `%*s` which assumes int type aiui. (see also line 140).

dmah42:
i'd like to land this one first as it actually fixes the windows tests (finally!) and then we can start having clead CI/CD on every PR.

Issue #2173, opened 2026-04-19
[BUG] Flaky failure of PerfCountersTest.MultiThreaded


musicinmybrain:
**Describe the bug**

While looking at [enabling `libpfm` support](https://src.fedoraproject.org/rpms/google-benchmark/pull-request/5) in Fedora’s `google-benchmark` package, we’ve found that `PerfCountersTest.MultiThreaded` fails flakily on the `ppc64le` architecture relatively frequently (perhaps 1/4 to 1/2 of the time), and rarely on some other architectures, at least `aarch64`.

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Fedora (Rawhide)
  - Compiler and version: `gcc (GCC) 16.0.1 20260416 (Red Hat 16.0.1-0)`

**To reproduce**
We see this while building the [`google-benchmark` package](https://src.fedoraproject.org/rpms/google-benchmark/), but there should not be anything unique here: I expect it will be possible to reproduce it in any `ppc64le` environment with `-DBENCHMARK_ENABLE_LIBPFM:BOOL=ON`.

I can do test-builds of Fedora packages on `ppc64le` in Fedora infrastructure, but I do not have *interactive* access to real `ppc64le` hardware.

**Expected behavior**
All tests pass.

**Screenshots**

```
74/84 Test #79: perf_counters_gtest ........................***Failed    0.20 sec
Running main() from gmock_main.cc
[==========] Running 9 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 9 tests from PerfCountersTest
[ RUN      ] PerfCountersTest.Init
[       OK ] PerfCountersTest.Init (0 ms)
[ RUN      ] PerfCountersTest.OneCounter
[       OK ] PerfCountersTest.OneCounter (0 ms)
[ RUN      ] PerfCountersTest.NegativeTest
A performance counter name was the empty string
Unknown performance counter name: not a counter name
A performance counter name was the empty string
Unknown performance counter name: not a counter name
Unknown performance counter name: bad event name
[       OK ] PerfCountersTest.NegativeTest (0 ms)
[ RUN      ] PerfCountersTest.Read1Counter
[       OK ] PerfCountersTest.Read1Counter (0 ms)
[ RUN      ] PerfCountersTest.Read2Counters
[       OK ] PerfCountersTest.Read2Counters (0 ms)
[ RUN      ] PerfCountersTest.ReopenExistingCounters
[...]

PR #2175, opened 2026-04-20
In PerfCountersTest.MultiThreaded, serialize worker threads

musicinmybrain:
This is based on the suggestion by @LebedevRI in https://github.com/google/benchmark/issues/2173#issuecomment-4278264213. See also the discussion in https://github.com/google/benchmark/issues/2173#issuecomment-4279129698. Before merging this, a reviewer should verify the claim that this change actually preserves the original intent of the test.

In ten tests in Fedora infrastructure on `i686`, `x86_64`, `aarch64`, and `ppc64le` architectures, this seems to consistently avoid the flaky failure reported in https://github.com/google/benchmark/issues/2173.

### DIFF
--- a/test/perf_counters_gtest.cc
+++ b/test/perf_counters_gtest.cc
@@ -1,3 +1,4 @@
+#include <mutex>
 #include <random>
 #include <thread>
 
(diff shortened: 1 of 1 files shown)


LebedevRI:
I'm pretty sure the test is checking that PFM (still) tracks perf counters of child threads,
by postulating that changing the number of said child threads should be changing
the perf counter values observed. IOW i'm confident in this fix.

LebedevRI (approved):
LG

LebedevRI:
Thank you!

PR #2177, opened 2026-04-23
Replace deprecated _ReadWriteBarrier with std::atomic_signal_fence in MSVC path

### LINKED ISSUE #747: _ReadWriteFence is deprecated in MSVC 2015 (possibly earlier)
With the library being rebased on C++11, we should be using std::atomic_thread_fence instead per https://docs.microsoft.com/en-us/cpp/intrinsics/readwritebarrier?view=vs-2015

On the assumption that the fence is only being used to prevent read/write re-ordering, it should be sufficient / equivalent to use std::memory_order_acq_rel.

qorexdevs:
_ReadWriteBarrier() has been deprecated since MSVC 2015. This replaces
the three calls in the MSVC DoNotOptimize overloads with
std::atomic_signal_fence(std::memory_order_acq_rel), which is the
standard C++11 equivalent (compiler-only barrier, no hardware fence).

Same function is already used for ClobberMemory() in this file, and
<atomic> is already included.

Fixes #747

### DIFF
--- a/include/benchmark/utils.h
+++ b/include/benchmark/utils.h
@@ -127,19 +127,19 @@ BENCHMARK_DEPRECATED_MSG(
     "undesired compiler optimizations in benchmarks")
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
   internal::UseCharPointer(&reinterpret_cast<char const volatile&>(value));
-  _ReadWriteBarrier();
+  std::atomic_signal_fence(std::memory_order_acq_rel);
 }
 
 template <class Tp>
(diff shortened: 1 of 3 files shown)


include/benchmark/utils.h
LebedevRI, line 37: Why not call `ClobberMemory()` instead of inlining it?

qorexdevs: yeah good call, missed that ClobberMemory() already wraps atomic_signal_fence. will update

qorexdevs:
@googlebot I signed it!

LebedevRI (approved):
LG, thank you!

Issue #720, opened 2018-11-19
Document JSON format


EricWF:
Currently we don't provide any documentation as to the format of the JSON we produce. Instead of requiring users to read the code, we should document and then stick to a format for our JSON.

PR #2181, opened 2026-04-26
Docs: JSON output format

shreejaykurhade:
## Summary #720 
- Document the JSON output structure in the user guide
- List context fields, benchmark result fields, and optional fields
- Update the example JSON output to reflect the current reporter fields

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -88,74 +88,26 @@ BM_SetInsert/1024/10                       33157      33648      21431  1.13369M
 ```
 
 The JSON format outputs human readable JSON split into two top level
-attributes: `context` and `benchmarks`.
+attributes: `context` and `benchmarks`. This format is useful for tools that
+need to consume benchmark results without parsing console output.
 
 The `context` object contains information about the run in general, including
-information about the machine, library, and date:
+the date, host, CPU, caches, load average, benchmark library version, and
+`json_schema_version`. Extra context added with `benchmark::AddCustomContext` or
+`--benchmark_context` is emitted as additional string fields in `context`.
 
-| Field | Type | Description |
-| ----- | ---- | ----------- |
-| `date` | string | Local date and time when the benchmark ran. |
-| `host_name` | string | Name of the host machine. |
-| `executable` | string | Benchmark executable path, when known. |
-| `num_cpus` | integer | Number of logical CPUs. |
-| `mhz_per_cpu` | integer | Approximate CPU frequency in MHz. |
-| `cpu_scaling_enabled` | boolean | Present when CPU scaling status is known. |
-| `aslr_enabled` | boolean | Present when ASLR status is known. |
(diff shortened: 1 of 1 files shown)


shreejaykurhade:
@dmah42 please review it

dmah42:
I'm on the fence about this. I appreciate document updates in general but documenting the JSON output so rigidly means we would need to keep it updated, while the JSON output itself can be extended without breaking any users. what value do you see this bringing to users?

shreejaykurhade:
That concern makes sense. My intent here is not to freeze the JSON format or imply that every listed field is permanently guaranteed. The value I see is giving users a readable reference for the current schema so they do not need to reverse-engineer `json_reporter.cc` or infer behavior from test fixtures.

A few concrete cases where I think this helps:
- Tool authors can understand the difference between always-present fields, optional fields, aggregate rows, skipped/error rows, and user counters.
- Users consuming benchmark JSON can discover `json_schema_version` and know where it lives.
- The docs can clarify that additional fields may appear, especially custom counters and context fields, so consumers should remain tolerant.

If the table feels too rigid, I’m happy to soften the wording further, for example by explicitly saying this documents the current schema version and that compatible additions may be made over time.

LebedevRI:
I'm on the fence about this too.

shreejaykurhade:
Ya have a point.
@dmah42 @LebedevRI That concern makes sense. I reworked the docs to avoid presenting the JSON output as a rigid field-by-field contract.

The updated version now focuses on consumer guidance: it explains the top-level structure, why JSON output is useful for tools, which fields are commonly present, and explicitly says consumers should ignore unknown fields and tolerate optional fields being absent because the output may gain new fields over time.

I also shortened the example so it is illustrative rather than exhaustive.

docs/user_guide.md
```diff
@@ -87,45 +87,60 @@ BM_SetInsert/1024/8                        32065      32913      21375  949.487k
...
       "name": "BM_SetInsert/1024/1",
+      "run_name": "BM_SetInsert/1024/1",
+      "run_type": "iteration",
       "iterations": 94877,
       "real_time": 29275,
       "cpu_time": 29836,
+      "time_unit": "ns",
       "bytes_per_second": 134066,
       "items_per_second": 33516
-    },
```
dmah42, line 113: not sure why you removed this bit..

shreejaykurhade:
@dmah42 Good catch. I restored that bit: the example now includes the additional benchmark entries again, and I also brought back the note that `bytes_per_second`, `items_per_second`, and other user counters are emitted as additional numeric fields.

The docs still avoid treating the JSON output as a strict field-by-field contract, but this keeps the useful counter/output guidance in place.

dmah42:
this is fine. i'm not sure it adds a lot of useful information, but it doesn't hurt.

dmah42 (approved):

Issue #1902, opened 2025-01-09
[FR] Report cross-thread statistics for multithreaded runs


serhiihuralniksc:
**Is your feature request related to a problem? Please describe.**
When running multithreaded benchmarks - in certain cases there is a need to understand a "fairness" of the measured code, i.e. if work is equally distributed among all threads and/or if all threads were able to acquire some resource in an equal (or intentionally unequal) manner.
Currently counters for all threads are summed and normalized which in fact leads to a loss of information about how work of each thread compares to its peers.

**Describe the solution you'd like**
Is it possible to output the same set of statistics for multithreaded runs on a cross-thread basis as it exists today for repeated runs? I.e. to see median and stddev between N threads of the same multithreaded run?

**Describe alternatives you've considered**
Tried to leverage existing API, like custom counters but it appears that it also undergoes the same averaging for all threads so basically info about individual thread results is lost already. So there is no alternative as I understand it.

dmah42:
you're right that this isn't offered.  the context here is that multi-threading something to benchmark it is anticipated to be to see what the impact of the threading (and thread counts) has on the overall performance (and throughput, etc).  separating out the thread reports would run counter to that expectation.

that being said, i understand the desire and i don't think the library currently supports it, so if someone wants to come up with a design for how we can offer this without complicating the reporting or main loop too much, i'm open to hearing about it.

jlecavelier:
Just for the record I raised that issue in a previous ticket :
https://github.com/google/benchmark/issues/1849
`
Wouldn't you agree that it is reasonable for someone who registers a benchmark to run in a multithreaded fashion, explicitly overriding the memory manager, to expect gathering of memory statistics happening as they set it up, in a multi threaded fashion, and not in serial?`

I will do some digging and check out how I got around that issue.
I do think that this is a reasonable issue to tackle

Frank20021:
## Design: cross-thread statistics for multithreaded runs

I am interested in implementing this. Below is a concrete design that fits the current runner/reporter architecture. I used automated assistance to locate the relevant code paths; I have read those paths myself and this design is mine to defend.

### Problem

In a multithreaded run, each `State` has its own counters and `ThreadTimer`. At the end of `RunInThread` (`src/benchmark_runner.cc`), those values are summed into a single `ThreadManager::Result` via `Increment`, then the per-thread `State` and timer are destroyed. `CreateRunReport` then calls `Finish` on the **already summed** counters.

[...]

PR #2182, opened 2026-04-27
feat: Add opt-in cross-thread statistics for multithreaded counters

shreejaykurhade:
## Summary  

This PR addresses #1902 by enabling visibility into cross-thread imbalance in multithreaded counters.

Currently, only aggregated (summed) counter values are reported, which makes it difficult to assess fairness across benchmark-created threads. This change introduces an opt-in mechanism to compute and report distribution statistics across threads.

When enabled, benchmark-created thread counter values are retained long enough to compute the benchmark’s configured statistics across threads. The existing summed counter row remains unchanged, and additional aggregate rows:

- `_thread_mean`
- `_thread_median`
- `_thread_stddev`
- `_thread_cv`

For example, if three benchmark threads report `work` values `1`, `2`, and `3`, the normal row still reports `work=6`, while the opt-in thread-stat rows report mean `2`, median `2`, stddev `1`, and cv `50%`.

## Testing

- `cmake --build build`
- `cmake --build build --config Debug`
- `ctest --test-dir build --output-on-failure`

All 74 CTest tests pass.


shreejaykurhade:
@dmah42 please review

dmah42:
is anyone asking for this? i'm beginning to think you've pointed an agent at the codebase and asked it to come up with "improvements".

shreejaykurhade:
Yes, this is intended to address #1902, which asks for cross-thread statistics to inspect fairness/imbalance in multithreaded counters.

I do use Codex to understand parts of the codebase (e.g., structure and JSON fields), but the design and implementation here are deliberate.

The change is opt-in and doesn’t affect existing behavior.

Happy to adjust if needed.

shreejaykurhade:
@dmah42 check it please

PR #2183, opened 2026-04-27
BENCHMARK_ENABLE_WERROR=Off should also disable -pedantic-errors

ednolan:
Otherwise Clang 22 builds fail with:

```
/build/_deps/benchmark-src/include/benchmark/benchmark.h:1442:30: error: '__COUNTER__' is a C2y extension [-Werror,-Wc2y-extensions]
 1442 | #if defined(__COUNTER__) && (__COUNTER__ + 1 == __COUNTER__ + 0)
      |                              ^
```


LebedevRI:
The change itself makes sense, but did you check that it solves the issue you were seeing?
Because that warning should already be silenced regardless:
https://github.com/google/benchmark/blob/f43ed80fa2b6ca9411081892bf961513f56489df/include/benchmark/registration.h#L21-L38

ednolan:
You're right: the warning was only not silenced for me because I was testing with version 1.9.2 instead of the latest. (So updating to 1.9.5 solved the problem without needing the CMake change).

I still think the proposed change is useful, though.

LebedevRI (approved):
LG, thank you!

Issue #1217, opened 2021-08-27
Memory reporting capabilities are unclear and undocumented.


haberman:
I am currently using google/benchmark to benchmark https://github.com/protocolbuffers/upb. I would very much like for those benchmarks to also include memory usage.

Does this project support memory benchmarking? The [user guide](https://github.com/google/benchmark/blob/main/docs/user_guide.md) does not mention memory usage benchmarking at all. However I notice that the header has [a `RegisterMemoryManager()` function](https://github.com/google/benchmark/blob/dc1a97174d5b84b1ce6720f1c98f85e6031b4723/include/benchmark/benchmark.h#L302-L304), and there is [a unit test](https://github.com/google/benchmark/blob/main/test/memory_manager_test.cc) verifying this functionality.

From running the test, it appears that the memory usage data is surfaced to JSON, but not CSV or console reports. Is this an oversight or intentional? Ideally this data would be shown on CSV and console reports also.

oontvoo:
@dominichamon Any thought on this? 
(I've similar question to @haberman  as I'm thinking about what to do with the internal BenchmakMemoryUsage() )

dominichamon:
i'll leave this open to cover the addition of the memory reporting to CSV/Console reporters.

q-ycong-p:
There is a section in [User Guide](https://github.com/google/benchmark/blob/main/docs/user_guide.md#memory-usage) that briefly mentions the supported functionality of benchmarking memory usage. But the documentation gives no concrete instruction of how to use it. Please provide details and examples.

Ominsk:
The following is an example that I created to try the MemoryManager functionality.
I'm interested to know how many bytes are allocated during a single benchmark execution.

I encountered a problem and I'm not sure if what I'm doing is right as there is no documentation.

Steps:
- First create a subclass of `benchmark::MemoryManager` and override both `benchmark::MemoryManager::Start` and `benchmark::MemoryManager::Stop`.

- Second you need to wrap your memory operator or function such that you can store the size inside the MemoryManager object.

```cpp
#include <memory>
#include <benchmark/benchmark.h>

class CustomMemoryManager: public benchmark::MemoryManager {
public:

    int64_t num_allocs;
    int64_t max_bytes_used;


    void Start() BENCHMARK_OVERRIDE {
        num_allocs = 0;
        max_bytes_used = 0;
    }

    void Stop(Result* result) BENCHMARK_OVERRIDE {
        result->num_allocs = num_allocs;
[...]

PR #2185, opened 2026-05-04
Docs: clarify MemoryManager usage

qorexdevs:
Adds a small MemoryManager registration example to the user guide.\n\nRefs #1217

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1370,6 +1370,14 @@ a report on the number of allocations, bytes used, etc.
 This data will then be reported alongside other performance data, currently
 only when using JSON output.
 
+The memory manager does not collect allocations by itself. It is meant to
+bridge data from an allocator hook or profiler that already tracks memory
+activity. Reset that collector in `Start()`, then copy its final counters into
+`MemoryManager::Result` in `Stop()`.
+
+Use [Custom Counters](#custom-counters) instead when memory stats need to be
+reported in console or CSV output, or when you need allocator-specific columns.
+
 <a name="profiling" />
 
 ## Profiling


docs/user_guide.md
```diff
@@ -1370,6 +1370,38 @@ a report on the number of allocations, bytes used, etc.
...
+}
+BENCHMARK(BM_work);
+
+int main(int argc, char** argv) {
+  benchmark::MaybeReenterWithoutASLR(argc, argv);
+  MyMemoryManager mm;
+  benchmark::RegisterMemoryManager(&mm);
+  benchmark::Initialize(&argc, argv);
+  benchmark::RunSpecifiedBenchmarks();
+  benchmark::RegisterMemoryManager(nullptr);
```
dmah42: is this necessary?

docs/user_guide.md
```diff
@@ -1370,6 +1370,38 @@ a report on the number of allocations, bytes used, etc.
...
 
+For example, a custom manager can report allocation data collected from
+allocator hooks or another profiler:
+
+```c++
+class MyMemoryManager : public benchmark::MemoryManager {
+ public:
+  void Start() override {}
+
+  void Stop(Result& result) override {
```
dmah42: is this a helpful example? it's a bit of a toy, but i think the complexity of using the memory manager is not in this interface, but in how to use it alongside something like tcmalloc.

maybe this should be an expanded tutorial on how to use it in its own documentation instead of this (what i consider fairly obvious) interface documentation.

qorexdevs:
Makes sense. I removed the toy example and replaced it with a shorter note that MemoryManager only bridges allocator or profiler data into JSON, with a pointer to custom counters for console/CSV or allocator-specific columns.

docs/user_guide.md
```diff
@@ -1370,6 +1370,14 @@ a report on the number of allocations, bytes used, etc.
 This data will then be reported alongside other performance data, currently
 only when using JSON output.
 
+The memory manager does not collect allocations by itself. It is meant to
+bridge data from an allocator hook or profiler that already tracks memory
+activity. Reset that collector in `Start()`, then copy its final counters into
+`MemoryManager::Result` in `Stop()`.
+
+Use [Custom Counters](#custom-counters) instead when memory stats need to be
```
dmah42, line 1378: no don't do this. we have the memory manager for exactly this reason. this would encourage users to run memory profiling in their performance sensitive micro benchmarks.

i don't think this PR is very useful.

qorexdevs:
Fair point. I'll close this then. Thanks for taking a look.

Issue #2184, opened 2026-04-29
[BUG] MaybeReenterWithoutASLR() infinite loop under AppArmor (not covered by #1984 fix)


sjanel:
## Description

`MaybeReenterWithoutASLR()` causes an infinite `execv()` loop when running
google-benchmark binaries inside VSCode's integrated terminal on a Linux
machine with AppArmor enabled. This is a different root cause from #1984
(which was fixed in #1985/v1.9.4).

## Environment

- google-benchmark **v1.9.5**
- Linux (Ubuntu), kernel with AppArmor enabled
  (`lockdown,capability,landlock,yama,apparmor,ima,evm`)
- Triggered specifically inside VSCode integrated terminal (Electron child
  process tree). Works fine in gnome-terminal on the same machine.

## Root Cause Analysis

The fix from #1985 checks the personality *before* the `execv()`:

```cpp
const auto new_personality = personality(0xffffffff);
if ((internal::get_as_unsigned(new_personality) & ADDR_NO_RANDOMIZE) == 0)
    return;  // guard: flag didn't stick
execv(argv[0], argv);
```

In my case, this guard does NOT trigger because personality() does
report ADDR_NO_RANDOMIZE as set within the current process. However,
execv() causes the kernel/AppArmor to reset the personality in the
new process image. The newly exec'd process then sees ASLR is still active,
tries to disable it again, succeeds "locally", exec's again — infinite loop.

Evidence from GDB (addresses change on each exec = ASLR still active):

```bash
process 452521 is executing new program: /tmp/mini_bench_debug
Breakpoint 1, main (argc=2, argv=0x7ffc55608698)
process 452521 is executing new program: /tmp/mini_bench_debug
Breakpoint 1, main (argc=2, argv=0x7fff874b84f8)
process 452521 is executing new program: /tmp/mini_bench_debug
Breakpoint 1, main (argc=2, argv=0x7fff7f778408)
... (loops forever at 100% CPU)
```

Additional evidence:

```bash
VIRT=12KB in top (binary never fully loads due to constant re-exec)
cat /proc/$PID/maps shows only [stack] and [vsyscall]
cat /proc/$PID/syscall reports running (no syscall, pure userspace loop)
cat /sys/kernel/security/lsm → lockdown,capability,landlock,yama,apparmor,ima,evm
No LD_PRELOAD, no kernel module interference, no seccomp
```

## Difference from #1984

| Aspect | #1984 (Docker) | This bug (AppArmor) |
[...]

PR #2193, opened 2026-05-20
Avoid repeated ASLR re-exec attempts

kiwigitops:
## Summary
- add an environment sentinel before re-execing after ADDR_NO_RANDOMIZE is set
- skip the re-exec if the process already attempted the ASLR-disable path once
- avoid re-execing if the sentinel cannot be recorded

## Context
This addresses the AppArmor case from #2184 where `personality()` reports `ADDR_NO_RANDOMIZE` as set before `execv()`, but the new process image loses that flag and re-enters the same code path forever.

## Testing
- `git diff --check`

I did not add a direct unit test because the behavior depends on Linux exec/personality interaction under an AppArmor-constrained process tree.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -864,6 +864,11 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
   if ((internal::get_as_unsigned(new_personality) & ADDR_NO_RANDOMIZE) == 0)
     return;
 
+  // Some security profiles clear ADDR_NO_RANDOMIZE across exec even though the
+  // personality appears updated here. Avoid re-entering forever in that case.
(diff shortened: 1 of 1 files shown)


src/benchmark.cc
```diff
@@ -860,6 +860,11 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
   if ((internal::get_as_unsigned(new_personality) & ADDR_NO_RANDOMIZE) == 0)
     return;
 
+  // Some security profiles clear ADDR_NO_RANDOMIZE across exec even though the
+  // personality appears updated here. Avoid re-entering forever in that case.
+  static const char kAslrReexecEnv[] = "BENCHMARK_ASLR_NO_REEXEC";
```
LebedevRI: Why is this an array of strings if we only need a single string?

PR #2194, opened 2026-05-22
Add -lpfm to pkg-config Libs.private when libpfm is enabled

### LINKED ISSUE #1908: [BUG]  pkgconfig benchmark.pc not modified with addition of -lpfm when  BENCHMARK_ENABLE_LIBPFM=On
When the option BENCHMARK_ENABLE_LIBPFM=On is used, the installed CMAKE files provide the necessary test to find libpfm when other software using cmake is being built with benchmark as a dependency.

However no requirement to include libpfm as a library is added to the pkgconfig benchmark.pc file is added.

So unless one manually appends -lpfm to  the benchmark.pc line for Libs

    --- benchmark.pc.ORIG      2025-01-17 21:05:44.622488319 +0100
    +++ benchmark.pc        2025-01-17 21:07:35.562542402 +0100
    @@ -7,6 +7,6 @@
    Description: Google microbenchmark framework
    Version: 1.9.1
 
    -Libs: -L${libdir} -lbenchmark
    +Libs: -L${libdir} -lbenchmark -lpfm
    Libs.private: -lpthread 
    Cflags: -I${includedir}

when building software which uses benchmark via pkg-config eg movit (high-performance, high-quality video filters)  the following error for undefined references to pfm_get_os_event and pfm_initialize  occurs when linking

            /usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /var/tmp/root/BUILD/BUILD_movit_20250117_195233/ccAqTAKg.ltrans3.ltrans.o: in function `benchmark::internal::PerfCounters::Create(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)':

           <artificial>:(.text+0x8445): undefined reference to `pfm_get_os_event_encoding' 
           /usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: <artificial>:(.text+0x894c): undefined reference to `pfm_initialize'

At first I tried appending -lpfm on the Libs.private line but that did not work but next test with it on the Libs line did work.

kiwigitops:
## Summary

When `BENCHMARK_ENABLE_LIBPFM=ON`, libpfm is linked into the benchmark library via the `PFM::libpfm` imported target. The pkg-config generation derives `Libs.private` from the target's `LINK_LIBRARIES`, but the loop only emits `-l` flags for entries that are *not* CMake targets. Imported targets like `PFM::libpfm` are silently skipped, so `-lpfm` never makes it into `benchmark.pc`.

As a result, projects that statically link benchmark through pkg-config fail to resolve `pfm_initialize` / `pfm_get_os_event_encoding` at link time.

This adds `-lpfm` to the private link libraries for the `PFM::libpfm` target, alongside the existing private dependencies (`-lpthread`, etc.).

Closes #1908.

## Testing
- `git diff --check`


LebedevRI (commented):
This should probably be using `get_target_property()` instead of
guessing the lib name, oh, and, this seems to completely not handle `-L`?

src/CMakeLists.txt
```diff
@@ -113,6 +113,18 @@ if(NOT BUILD_SHARED_LIBS)
...
         list(APPEND BENCHMARK_PRIVATE_LINK_LIBRARIES "-l${LIB}")
+      elseif(TARGET "${LIB}")
+        get_target_property(_target_type "${LIB}" TYPE)
+        if(_target_type MATCHES "^(UNKNOWN|STATIC|SHARED|MODULE)_LIBRARY$")
+          get_target_property(_imported_loc "${LIB}" IMPORTED_LOCATION)
+          if(_imported_loc)
+            get_filename_component(_imported_dir "${_imported_loc}" DIRECTORY)
+            get_filename_component(_imported_name "${_imported_loc}" NAME)
+            string(REGEX REPLACE "^lib" "" _imported_name "${_imported_name}")
+            string(REGEX REPLACE "\\.(so|a|dylib)([0-9.]*)$" "" _imported_name "${_imported_name}")
```
LebedevRI, line 124: Should we care about `.dll` ?

LebedevRI (approved):
Looks good to me, thank you!

PR #2199, opened 2026-05-24
Skip errored runs when computing repetition statistics

### LINKED ISSUE #1406: [BUG] Uninitialized pointer access when using --benchmark_repetitions in some scenarios
**Describe the bug**
An uninitialized pointer is dereferenced when calculating statistics for a repeated benchmark that fails a certain number of times, which causes a segv/access violation/other mayhem.

This doesn't _seem_ to happen for benchmarks that always fail, nor if you run without `--benchmark_repetitions=N`.

**System**
  - OS: Windows 10 Pro 21H2 (x64)
  - Compiler and version: MSVC 19.29.30141.0

In my experience this happens on various compiler (MSVC, GCC) and execution environments (Windows x64, Linux arm64), as well as in both Debug and Release builds.

**To reproduce**
```cpp

static int count = 0;

static void BM_StringCreation(benchmark::State &state)
{
    if (++count == 3)
    {
        state.SkipWithError("Simulate a flaky benchmark");
        return;
    }

    for (auto _ : state)
        std::string empty_string;
}
BENCHMARK(BM_StringCreation);

int main(int argc, char *argv[])
{
    benchmark::RunSpecifiedBenchmarks();
    benchmark::Shutdown();
    return 0;
}
```

Compile the given code above on MSVC 19 with benchmark 1.6.1, and run using:
```
./example.exe --benchmark_repetitions=5
```

**Expected behavior**
Benchmark does not crash. Statistics are calculated for only the passing runs.
(If this is incorrect usage of the library, it should be documented somewhere.)

**Screenshots**
image

**Additional context**
Looks like the uninitialized pointer is `reports[0].statistics`:
https://github.com/google/benchmark/blob/6d50251d8e6ed6b7f6eb1e12c558d2808fb4ddaa/src/statistics.cc#L159

It seems to not be initialized here (on line 98) when a run has an error:
https://github.com/google/benchmark/blob/6d50251d8e6ed6b7f6eb1e12c558d2808fb4ddaa/src/benchmark_runner.cc#L88-L99

Haihan-Jiang:
Fixes #1406.

## Summary
- compute repetition aggregates from non-errored runs only
- use the first successful run as the metadata/statistics reference instead of `reports[0]`
- ignore skipped runs when building aggregate counter inputs and labels
- add a regression test for a skipped first run followed by successful repetitions

## Validation
- cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBENCHMARK_ENABLE_TESTING=ON -DBENCHMARK_DOWNLOAD_DEPENDENCIES=ON -DCMAKE_CXX_FLAGS="-isystem /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1"
- cmake --build build --target statistics_gtest skip_with_error_test -- -j4
- ctest --test-dir build -R `statistics_gtest|skip_with_error_test` --output-on-failure

Note: the explicit CMAKE_CXX_FLAGS include path is only needed on my local macOS CommandLineTools install because its default C++ header search path does not find SDK libc++ headers.

### DIFF
--- a/src/statistics.cc
+++ b/src/statistics.cc
@@ -117,8 +117,8 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
   };
   auto successful_run =
       std::find_if(reports.begin(), reports.end(), is_successful);
-  const auto successful_count =
-      std::count_if(reports.begin(), reports.end(), is_successful);
+  const auto successful_count = static_cast<std::vector<Run>::size_type>(
+      std::count_if(reports.begin(), reports.end(), is_successful));
 
   if (successful_count < 2) {
     // We don't report aggregated data if there was a single run.
@@ -129,8 +129,8 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
   std::vector<double> real_accumulated_time_stat;
   std::vector<double> cpu_accumulated_time_stat;
 
-  real_accumulated_time_stat.reserve(reports.size());
-  cpu_accumulated_time_stat.reserve(reports.size());
+  real_accumulated_time_stat.reserve(successful_count);
+  cpu_accumulated_time_stat.reserve(successful_count);
 
   // All repetitions should be run with the same number of iterations so we
   // can take this information from the first benchmark.
@@ -152,7 +152,7 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
                  .emplace(cnt.first,
                           CounterStat{cnt.second, std::vector<double>{}})
                  .first;
-        it->second.s.reserve(reports.size());
+        it->second.s.reserve(successful_count);
       } else {
         BM_CHECK_EQ(it->second.c.flags, cnt.second.flags);
       }
--- a/src/statistics.cc
+++ b/src/statistics.cc
@@ -117,7 +117,7 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
   };
   auto successful_run =
       std::find_if(reports.begin(), reports.end(), is_successful);
-  const auto successful_count = static_cast<std::vector<Run>::size_type>(
+  const auto successful_count = static_cast<size_t>(
       std::count_if(reports.begin(), reports.end(), is_successful));
 
   if (successful_count < 2) {
--- a/src/statistics.cc
+++ b/src/statistics.cc
@@ -112,13 +112,13 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
   typedef BenchmarkReporter::Run Run;
   std::vector<Run> results;
 
-  auto successful_run = std::find_if(reports.begin(), reports.end(),
-                                     [](Run const& run) {
-                                       return run.skipped == 0u;
-                                     });
-  const auto successful_count = std::count_if(
-      reports.begin(), reports.end(),
-      [](Run const& run) { return run.skipped == 0u; });
+  const auto is_successful = [](Run const& run) {
+    return run.skipped == internal::NotSkipped;
+  };
+  auto successful_run =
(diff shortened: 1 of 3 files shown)


src/statistics.cc
```diff
@@ -169,26 +177,30 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
   }
 
   // Only add label if it is same for all runs
-  std::string report_label = reports[0].report_label;
-  for (std::size_t i = 1; i < reports.size(); i++) {
-    if (reports[i].report_label != report_label) {
+  std::string report_label = successful_run->report_label;
+  for (const Run& run : reports) {
+    if (run.skipped != 0u) {
```
LebedevRI: Here and elsewhere, this is a very weird way to check a boolean...

test/statistics_gtest.cc
```diff
@@ -32,4 +51,26 @@ TEST(StatisticsTest, CV) {
               0.32888184094918121, 1e-15);
 }
 
+TEST(StatisticsTest, ComputeStatsSkipsErroredFirstRun) {
```
LebedevRI: This is a wrong place for this test, since it is not about the actual statistic's computation.
`test/repetitions_test.cc` might be a better place,
and the test should follow the pattern of those tests.

LebedevRI (changes requested):
Please clang-format the diff

Haihan-Jiang:
Applied clang-format 18 to the touched test file and pushed c0a4bda. Local validation:

- `/Users/haihan/Library/Python/3.9/bin/clang-format --dry-run --Werror src/statistics.cc test/repetitions_test.cc`
- `cmake --build build --target statistics_gtest skip_with_error_test -- -j4`
- `ctest --test-dir build -R 'statistics_gtest|skip_with_error_test' --output-on-failure`

src/statistics.cc
LebedevRI: These should probably also use `successful_count`?

LebedevRI: Same

LebedevRI (changes requested):

Haihan-Jiang:
I do not see any active threads left now, and the clang-format check is passing after c0a4bda. Happy to make another pass if I missed something.

LebedevRI:
> I do not see any active threads left now, and the clang-format check is passing after [c0a4bda](https://github.com/google/benchmark/commit/c0a4bda543b394db4aeb68ae90f4e9d6793be0c3). Happy to make another pass if I missed something.

There are last two inline comments.

Haihan-Jiang:
Thanks, I missed those two. Updated in e5409c3: the time accumulators and per-counter buffers now reserve the successful run count as well. I also added the explicit size_type cast so the warning flags stay clean.

Locally checked with clang-format --dry-run, the statistics/skip-with-error build targets, and the matching ctest filter.

src/statistics.cc
```diff
@@ -117,8 +117,8 @@ std::vector<BenchmarkReporter::Run> ComputeStats(
   };
   auto successful_run =
       std::find_if(reports.begin(), reports.end(), is_successful);
-  const auto successful_count =
-      std::count_if(reports.begin(), reports.end(), is_successful);
+  const auto successful_count = static_cast<std::vector<Run>::size_type>(
```
LebedevRI: 1. Why, C++, why?!
2. Let's just hardcode `size_t` here

Haihan-Jiang: Done in 7caecf6. I switched the cast to size_t and reran the same local checks: clang-format dry run, the statistics/skip-with-error build targets, and the matching ctest filter.

LebedevRI (approved):
Awesome!

Issue #1498, opened 2022-10-04
[BUG] Invoking executable without args gives regex error


TylerSeanRau:
**Describe the bug**
When I run the following benchmark setup I always see the message `Failed to match any benchmarks against regex: .` when I invoke the executable (giving `--help` causes a segfault

**System**
Which OS, compiler, and compiler version are you using:
  - OS: Debian 10 for WSL2
  - Compiler and version: gcc 8.3

**To reproduce**
1. Create a new directory
2. Place this CMakeLists.txt in that directory:
```cmake
cmake_minimum_required(VERSION 3.14)
project(my_project)

set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED YES)

include(FetchContent)

set(BENCHMARK_ENABLE_TESTING off)

FetchContent_Declare(
  googletest
    GIT_REPOSITORY https://github.com/google/googletest.git
    GIT_TAG release-1.12.0)

FetchContent_Declare(
  googlebenchmark
    GIT_REPOSITORY https://github.com/google/benchmark.git
    GIT_TAG v1.7.0)

FetchContent_MakeAvailable(
  googletest
  googlebenchmark)

find_package(OpenMP)

add_library(
  libbench
  bench.cpp)

target_link_libraries(
  libbench
  PRIVATE
    benchmark::benchmark
    pthread
    OpenMP::OpenMP_CXX)

add_executable(bench /dev/null)
target_link_libraries(
  bench
    PRIVATE
      libbench
      benchmark::benchmark_main
      OpenMP::OpenMP_CXX)
```
3. Use this `.cpp` file
    * or write your own as far as I can tell it doesn't matter
    * unsure if all these headers matter but I've left them in case they do
```cpp
#include <iostream>
#include <string>
#include <unordered_map>
#include <vector>
#include <fcntl.h>
#include <unistd.h>
#include <sstream>
#include <string>
#include <benchmark/benchmark.h>
#include <vector>
#include <random>
#include <iostream>
#include <sys/mman.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <mutex>
#include <omp.h>
#include <array>
#include <memory>

static void BM_vector_push_back(benchmark::State& state) {
  for (auto _ : state) {
    std::vector<int> v;
    v.reserve(1);
    benchmark::DoNotOptimize(v.data()); // Allow v.data() to be clobbered.
[...]

PR #2200, opened 2026-05-24
Document static library benchmark registration

Haihan-Jiang:
Addresses #1498.

## Summary
- document why benchmarks registered from a separate static library may not appear at runtime
- explain that BENCHMARK uses static initialization and otherwise-unused object files may not be linked
- show an OBJECT library CMake pattern that forces benchmark registration objects into the final executable

## Validation
- git diff --check -- README.md

This is a documentation-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,16 +220,16 @@ target_link_libraries(MyTarget benchmark::benchmark)
 ```
 
 When benchmark sources are shared through an intermediate CMake target, prefer
-an object library:
+an object library instead of a static library:
 
 ```cmake
-add_library(my_benchmarks OBJECT bench.cc)
-target_link_libraries(my_benchmarks benchmark::benchmark_main)
-add_executable(my_benchmark)
-target_link_libraries(my_benchmark my_benchmarks)
+add_library(shared_benchmarks OBJECT bench.cc)
+target_link_libraries(shared_benchmarks benchmark::benchmark_main)
+add_executable(runnable_benchmarks)
+target_link_libraries(runnable_benchmarks shared_benchmarks)
 ```
 
 This links the object file that contains `BENCHMARK` registrations into the
 final executable. If those registrations are placed only in an intermediate
-static library, the linker may omit the otherwise-unused object file and the
+`STATIC` library, the linker may omit the otherwise-unused object file and the
 benchmarks will not be registered at runtime.
--- a/README.md
+++ b/README.md
@@ -218,3 +218,16 @@ Either way, link to the library as follows.
 ```cmake
 target_link_libraries(MyTarget benchmark::benchmark)
 ```
+
+If benchmark registrations live outside the final executable, make sure the
+benchmark object files are linked into it. The `BENCHMARK` macro registers
+benchmarks through static initialization, and many linkers do not pull
+otherwise-unused object files out of a static library. Prefer an object library
+so CMake links the registration objects directly:
+
+```cmake
+add_library(my_benchmarks OBJECT bench.cc)
+target_link_libraries(my_benchmarks benchmark::benchmark_main)
+add_executable(my_benchmark)
+target_link_libraries(my_benchmark my_benchmarks)
+```
--- a/README.md
+++ b/README.md
@@ -219,11 +219,8 @@ Either way, link to the library as follows.
 target_link_libraries(MyTarget benchmark::benchmark)
 ```
 
-If benchmark registrations live outside the final executable, make sure the
-benchmark object files are linked into it. The `BENCHMARK` macro registers
-benchmarks through static initialization, and many linkers do not pull
-otherwise-unused object files out of a static library. Prefer an object library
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -218,3 +218,17 @@ Either way, link to the library as follows.
...
+
+If benchmark registrations live in a separate static library, make sure the
+benchmark object files are linked into the final executable. The `BENCHMARK`
+macro registers benchmarks through static initialization, and many linkers do
+not pull otherwise-unused object files out of a static library. In that case the
+executable can start successfully but report that no benchmarks matched the
+filter. Prefer an object library, or use your linker equivalent of
+`--whole-archive`, so the registration objects are included:
+
+```cmake
```
LebedevRI, line 230: I do not understand.
As per https://cmake.org/cmake/help/latest/command/target_link_libraries.html#linking-object-libraries,
one should simply
```
add_library(my_benchmarks OBJECT bench.cc)
target_link_libraries(my_benchmarks benchmark::benchmark_main)
add_executable(my_benchmark)
target_link_libraries(my_benchmark my_benchmarks)
```

README.md
```diff
@@ -218,3 +218,16 @@ Either way, link to the library as follows.
 ```cmake
 target_link_libraries(MyTarget benchmark::benchmark)
 ```
+
+If benchmark registrations live outside the final executable, make sure the
```
LebedevRI: All of the wording reads very weirdly here.
```
Also, please be aware of how static library linking work - when linking a static library
into some other target (e.g. the final executable), the symbols, namely,
static variable initializers, which are produced by `BENCHMARK` macro,
that ended up in a static library, will not be copied over from the static library
into the destination object, and therefore the benchmarks that were registered
in said `STATIC` library, will not be present in the linked result.

Therefore, if you are using an intermediate `STATIC` library
you may need to use `OBJECT` intermediate library instead.
```

README.md
```diff
@@ -218,3 +218,18 @@ Either way, link to the library as follows.
 ```cmake
 target_link_libraries(MyTarget benchmark::benchmark)
 ```
+
+When benchmark sources are shared through an intermediate CMake target, prefer
+an object library:
```
LebedevRI: But the problem isn't that a non-object library is used,
the problem is that a `STATIC` library is used, is it not?
It seems like this distinction should be explained.

README.md
```diff
@@ -218,3 +218,18 @@ Either way, link to the library as follows.
...
+
+```cmake
+add_library(my_benchmarks OBJECT bench.cc)
+target_link_libraries(my_benchmarks benchmark::benchmark_main)
+add_executable(my_benchmark)
+target_link_libraries(my_benchmark my_benchmarks)
```
LebedevRI: When i've first looked at this snippet, it was really confusing because the name differ by a single char suffix added, it would be more readable to rename to `shared_bench` `runnable_benchmarks` or something.

Haihan-Jiang:
Thanks, updated the README in the latest push to make both points clearer:

- the problematic case is specifically an intermediate `STATIC` library;
- the example target names now distinguish the shared object library from the runnable benchmark executable.

README.md
```diff
@@ -218,3 +218,18 @@ Either way, link to the library as follows.
 ```cmake
 target_link_libraries(MyTarget benchmark::benchmark)
 ```
+
+When benchmark sources are shared through an intermediate CMake target, prefer
```
LebedevRI: s/prefer/choose/, since it's a correctness problem

README.md
```diff
@@ -218,3 +218,18 @@ Either way, link to the library as follows.
...
+add_library(shared_benchmarks OBJECT bench.cc)
+target_link_libraries(shared_benchmarks benchmark::benchmark_main)
+add_executable(runnable_benchmarks)
+target_link_libraries(runnable_benchmarks shared_benchmarks)
+```
+
+This links the object file that contains `BENCHMARK` registrations into the
+final executable. If those registrations are placed only in an intermediate
+`STATIC` library, the linker may omit the otherwise-unused object file and the
+benchmarks will not be registered at runtime.
```
LebedevRI:
```suggestion
-`STATIC` library, the linker may omit the otherwise-unused object file and the
-benchmarks will not be registered at runtime.
+`STATIC` library, the linker may not copy static registration symbols,
+and thus benchmarks will not be part of the final executable.
```
or something

LebedevRI (changes requested):

Haihan-Jiang:
Updated the README wording for the remaining review comments and pushed the documentation-only follow-up. Local validation: `git diff --check`.

LebedevRI (approved):
@dmah42 sounds reasonable?

README.md
Haihan-Jiang: Thanks, that example was the direction I ended up using. The README now shows the OBJECT library linked directly into the benchmark executable, so the object file with the BENCHMARK registration is pulled in.

PR #2201, opened 2026-05-24
Clarify DoNotOptimize const-ref warning

### LINKED ISSUE #1580: [BUG] No indication in error of how to fix "DoNotOptimize<long long>' is deprecated: The const-ref version ..."
**Describe the bug**
I recently updated the version of this repository used by my bazel(isk) WORKSPACE and I started encountering the deprecation warning about the const-ref form of DoNotOptimize. While I appreciate the warning and want to do better, it's unclear to me what I need to do differently to make it go away.

**System**
  - OS: OSX (uname -v: Darwin Kernel Version 19.6.0)
  - Compiler and version: llvm (probably. whatever bazelisk picks)

**To reproduce**
1. sync to commit d29044d5da4b32e449073a365ad6524fb9bdf884
2. bazelisk test :empty_test
3. See error: "problems/open/empty_test.cc:19:16: warning: 'DoNotOptimize<long long>' is deprecated: The const-ref version of this method can permit undesired compiler optimizations in benchmarks [-Wdeprecated-declarations]"

**Expected behavior**
I would like a description (or link to documentation) about what I need to change in that program to be getting ahead of the deprecation.

**Additional context**
BUILD
```
load("@rules_cc//cc:defs.bzl", "cc_test")
load("//euler/computer_science:data_file.bzl", "data_cc_file")

cc_test(
    name = "empty_test",
    timeout = "short",
    srcs = ["empty_test.cc"],
    tags = ["benchmark"],
    deps = [
        "//euler/computer_science:vlog",
        "@com_google_absl//absl/log",
        "@com_google_absl//absl/log:check",
        "@com_google_googletest//:gtest",
        "@com_monkeynova_gunit_main//:test_main",
    ],
)

```

empty_test.cc
```
#include "absl/log/check.h"
#include "absl/log/log.h"
#include "euler/computer_science/vlog.h"
#include "gtest/gtest.h"

int64_t RunProblem(int64_t range) {
  int64_t ret = -1;
  VLOG(1) << "RunProblem(" << range << ") = " << ret;
  return ret;
}

TEST(Euler, Given) { EXPECT_EQ(RunProblem(-1), -1); }

TEST(Euler, Problem) { EXPECT_EQ(RunProblem(-1), -1); }

void BM_ProblemXXX(benchmark::State& state) {
  for (auto _ : state) {
    benchmark::DoNotOptimize( RunProblem(state.range(0)));
  }
}

BENCHMARK(BM_ProblemXXX)->Arg(-1);
```

Haihan-Jiang:
## Summary
- make the deprecated `DoNotOptimize(T const&)` diagnostic tell users what to do next
- point users to storing expression results in a local non-const variable and passing that lvalue
- add the same before/after guidance to the user guide

Fixes #1580.

## Testing
- `cmake --build build --target donotoptimize_test`
- `./build/test/donotoptimize_test`
- `cmake --build build --target benchmark_gtest`
- `./build/test/benchmark_gtest`
- `git diff --check`
- compiled a local snippet with `-Wdeprecated-declarations` to verify the new warning text

`bazel test //test:donotoptimize_test` could not run in this local environment because the Bazel C++ toolchain cannot find the standard library header `<cmath>` while compiling `src/check.cc`.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1462,10 +1462,10 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
-Since that is not the behavior the user intended, such problematic cases
-produce a compile-time diagnostic. Pass a non-const lvalue instead. For an
-expression result, store the result in a local variable before calling
-`DoNotOptimize`:
+Since that is not the behaviour the user intended,
+such problematic cases will result in a diagnostic
+at compile time. The correct approach, for an expression result,
+is to use an intermediate local variable:
 
 ```c++
   // Avoid: may call the deprecated const-reference overload.
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1462,12 +1462,12 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
-Passing a temporary or `const` value may call the deprecated const-reference
-overload. Store the value in a local non-const variable and pass that lvalue to
-make the result observable:
+If `DoNotOptimize` warns about the deprecated const-reference overload, pass it
+a non-const lvalue instead. For an expression result, store the result in a
+local variable before calling `DoNotOptimize`:
 
 ```c++
-  // Avoid: calls the deprecated const-reference overload for the temporary.
+  // Avoid: may call the deprecated const-reference overload.
   while (...) DoNotOptimize(foo(0));
 
   // Prefer: materialize the result, then pass the local lvalue.
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1462,9 +1462,10 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
-If `DoNotOptimize` warns about the deprecated const-reference overload, pass it
-a non-const lvalue instead. For an expression result, store the result in a
-local variable before calling `DoNotOptimize`:
+Since that is not the behavior the user intended, such problematic cases
+produce a compile-time diagnostic. Pass a non-const lvalue instead. For an
+expression result, store the result in a local variable before calling
+`DoNotOptimize`:
 
 ```c++
   // Avoid: may call the deprecated const-reference overload.
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1462,6 +1462,21 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
+Passing a temporary or `const` value may call the deprecated const-reference
+overload. Store the value in a local non-const variable and pass that lvalue to
+make the result observable:
+
+```c++
+  // Avoid: calls the deprecated const-reference overload for the temporary.
+  while (...) DoNotOptimize(foo(0));
+
+  // Prefer: materialize the result, then pass the local lvalue.
+  while (...) {
+    auto result = foo(0);
+    DoNotOptimize(result);
+  }
+```
+
 The second tool for preventing optimizations is `ClobberMemory()`. In essence
 `ClobberMemory()` forces the compiler to perform all pending writes to global
 memory. Memory managed by block scope objects must be "escaped" using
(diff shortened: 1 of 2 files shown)


docs/user_guide.md
```diff
@@ -1462,6 +1462,21 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
+Passing a temporary or `const` value may call the deprecated const-reference
```
LebedevRI: All of the wording here is backwards.

docs/user_guide.md
```diff
@@ -1462,6 +1462,21 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
+If `DoNotOptimize` warns about the deprecated const-reference overload, pass it
+a non-const lvalue instead. For an expression result, store the result in a
+local variable before calling `DoNotOptimize`:
```
LebedevRI: This new line is really a continuation of the previous line,
but it still reads completely incongruent from it...
Perhaps something closer to:
```
Since that is not the behaviour the user intended,
such problematic cases will result in a diagnostic
at compile time. The correct approach,
for an expression result, store the result in a
local variable before calling `DoNotOptimize`:
```

LebedevRI (changes requested):

Haihan-Jiang:
Thanks, updated the wording in the latest push so the warning guidance now says to pass a non-const lvalue, and for expression results to store the value in a local variable before calling DoNotOptimize.

LebedevRI:
Did you push the update?

> Thanks, updated the wording in the latest push so the warning guidance now says to pass a non-const lvalue, and for expression results to store the value in a local variable before calling DoNotOptimize.

Please post a whale emoji if you are a human.

Haihan-Jiang:
Yes, the update is in 77be0e7. To be transparent: I used AI assistance while drafting the wording, then reviewed and pushed the change myself.

LebedevRI:
> Yes, the update is in [77be0e7](https://github.com/google/benchmark/commit/77be0e7cb963d3b22f4e28ce0ccbfe83dcd27ebc).

It is not. The inline comment was posted after that commit was pushed.

> To be transparent: I used AI assistance while drafting the wording, then reviewed and pushed the change myself.

LebedevRI (changes requested):

Haihan-Jiang:
You're right, that note was not addressed by the commit I pointed to. I pushed another update in 439646c that ties the warning back to the preceding example and keeps the local-variable guidance.

docs/user_guide.md
```diff
@@ -1462,6 +1462,22 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
+Since that is not the behavior the user intended, such problematic cases
+produce a compile-time diagnostic. Pass a non-const lvalue instead. For an
+expression result, store the result in a local variable before calling
+`DoNotOptimize`:
```
LebedevRI:
```suggestion
Since that is not the behaviour the user intended,
such problematic cases will result in a diagnostic
at compile time. The correct approach, for an expression result,
is to use an intermediate local variable:
```

include/benchmark/utils.h
```diff
@@ -37,12 +37,15 @@ inline BENCHMARK_ALWAYS_INLINE void ClobberMemory() {
   std::atomic_signal_fence(std::memory_order_acq_rel);
 }
 
+#define BENCHMARK_DONOTOPTIMIZE_CONST_REF_DEPRECATED_MSG                     \
+  "DoNotOptimize(T const&) can permit undesired compiler optimizations. "    \
```
LebedevRI:
```suggestion
#define BENCHMARK_DONOTOPTIMIZE_CONST_REF_DEPRECATED_MSG                       \
  "DoNotOptimize(T const&) can permit undesired compiler optimizations. "      \
```

LebedevRI (approved):

PR #2205, opened 2026-05-26
Clarify benchmark_main CMake target usage

### LINKED ISSUE #1856: CMake: Difference between benchmark::benchmark and benchmark::benchmark_main
Hi, I saw that on https://github.com/google/benchmark?tab=readme-ov-file#usage-with-cmake the Readme indicates that I should link against both `benchmark::benchmark` and `benchmark::benchmark_main` when using CMake. What's the difference between the two? Wouldn't it be enough to just link against `benchmark::benchmark_main`?

Thanks.

Haihan-Jiang:
Fixes #1856

The CMake usage text said to link both `benchmark::benchmark` and `benchmark::benchmark_main`, while the example only linked `benchmark::benchmark`. This clarifies when to use each target.

`benchmark::benchmark_main` already links `benchmark::benchmark` transitively, so CMake users only need to link one of the two targets depending on whether they provide their own `main`.

Tested:
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,8 +202,11 @@ flag for option information or see the [User Guide](docs/user_guide.md).
 ### Usage with CMake
 
 If using CMake, it is recommended to link against the project-provided
-`benchmark::benchmark` and `benchmark::benchmark_main` targets using
-`target_link_libraries`.
+`benchmark::benchmark` or `benchmark::benchmark_main` targets using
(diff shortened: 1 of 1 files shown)


dmah42 (approved):

dmah42:
i don't think git diff --check does anything. running the tests using cmake/bazel is the usual way to test a change. but given this is documentation, it likely doesn't matter.

PR #2207, opened 2026-05-26
Fix ComputeStats crash when first repetition is skipped

### LINKED ISSUE #2206: [BUG] ComputeStats crash when first repetition is skipped
`benchmark::ComputeStats()` crashes when the first repetition of a benchmark is skipped (via `SkipWithError` / `SkipWithMessage`) and `--benchmark_repetitions > 1` is used.     
  * In Release builds, it segfaults due to a null-pointer dereference of `reports[0].statistics`.
  * In Debug builds, it aborts on `BM_CHECK_EQ` because skipped runs have `iterations == 0`, which mismatches `run_iterations`.
  
**System**
  * OS: Linux (Ubuntu 24.04)
  * Compiler and version: GCC 13.3 (C++17)
  
**To reproduce**
  * Sync to `main` (e.g., commit `223269f9`).
  * Register a benchmark with `->Repetitions(3)` that calls `state.SkipWithError("skip")` on the first run but runs normally on subsequent repetitions.
  * Build the benchmark executable.
  * Run the benchmark: `example_benchmark --benchmark_repetitions=3`.
  * Observe crash (SIGSEGV or `BM_CHECK_EQ` abort).
  
**Expected behavior**
The benchmark should skip the first repetition, output the error message, calculate statistics for the remaining successful runs, and exit cleanly without crashing.             
  
**Screenshots**
Not applicable.
  
**Additional context**
  * Discovered during multi-platform benchmarking; runs on macOS and Windows succeeded, but approximately 4% of runs on Linux (Ubuntu 24.04) crashed.
  * Core dump analysis traced the crash to `ComputeStats()` in `src/statistics.cc` and `CreateRunReport()` in `src/benchmark_runner.cc`.
  * We will open a PR with a proposed fix and a regression test shortly.

dforsten:
Fixes #2206

When  --benchmark_repetitions > 1  and the first repetition is skipped (via  SkipWithError / SkipWithMessage ), ComputeStats()  crashes due to:

1. Dereferencing null  reports[0].statistics  (only set for non-skipped runs).                                                                                                   
2. Expecting  reports.front().iterations  to match every non-skipped run's iteration count.

Fixed by:

- Picking the first non-skipped report as the baseline for  run_iterations ,  report_label , and statistics.
- Ensuring  report.statistics  is always initialized in  CreateRunReport() , even for skipped runs.                                                                              
- Skipping skipped runs in aggregation and consistency loops.                                                                                                                    

Also adds a regression test reproducing the crash.

AI-Assisted-by: Gemini (Google DeepMind Antigravity)

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -105,6 +105,10 @@ BenchmarkReporter::Run CreateRunReport(
   report.repetition_index = repetition_index;
   report.repetitions = repeats;
 
+  // Set `statistics` for every report, skipped or not, so the pointer is
+  // always valid in aggregation paths.
+  report.statistics = &b.statistics();
+
   if (report.skipped == 0u) {
     if (b.use_manual_time()) {
       report.real_accumulated_time = results.manual_time_used;
@@ -116,7 +120,6 @@ BenchmarkReporter::Run CreateRunReport(
     report.complexity_n = results.complexity_n;
     report.complexity = b.complexity();
     report.complexity_lambda = b.complexity_lambda();
-    report.statistics = &b.statistics();
     report.counters = results.counters;
 
     if (memory_iterations > 0) {
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -143,6 +143,9 @@ benchmark_add_test(NAME diagnostics_test COMMAND diagnostics_test --benchmark_mi
 compile_benchmark_test(skip_with_error_test)
 benchmark_add_test(NAME skip_with_error_test COMMAND skip_with_error_test --benchmark_min_time=0.01s)
 
+compile_benchmark_test(skipped_repetition_test)
+benchmark_add_test(NAME skipped_repetition_test COMMAND skipped_repetition_test --benchmark_min_time=0.01s --benchmark_repetitions=3)
+
 compile_benchmark_test(donotoptimize_test)
 # Enable errors for deprecated deprecations (DoNotOptimize(Tp const& value)).
 check_cxx_compiler_flag(-Werror=deprecated-declarations BENCHMARK_HAS_DEPRECATED_DECLARATIONS_FLAG)
--- a/test/skipped_repetition_test.cc
+++ b/test/skipped_repetition_test.cc
@@ -1,0 +1,79 @@
+
+#undef NDEBUG
+#include <cassert>
+#include <vector>
+
+#include "../src/check.h"  // NOTE: check.h is for internal use only!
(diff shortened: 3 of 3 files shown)


dforsten:
The issue was encountered on a project using the vcpkg "benchmark" port version 1.9.5.
We worked around it by creating an overlay port and applying patches with the same fixes as in this PR.

That works for now, but it would be great to get it fixed upstream and eventually published to the vcpkg registry so we can eliminate that overlay port.

LebedevRI:
Please rebase now that https://github.com/google/benchmark/pull/2199 has been merged.
Why do we need to always have valid statistics pointer even for skipped runs?

dforsten:
> Why do we need to always have valid statistics pointer even for skipped runs?

That was the initial fix for the crash I encountered, but further investigation led to fixing the issues in statistics.cc which avoided both the crash and produces correct results.

In principle the change in benchmark_runner.cc could be reverted, but on the other hand it makes the code more robust since the pointer is always valid.

LebedevRI (changes requested):
(Awaiting diff update)

dforsten:
Well, it looks like #2199 may actually fixed this crash issue.
I have to test by adapting our vcpkg overlay port to use the latest main branch commit without the patches we applied.

src/benchmark_runner.cc
```diff
@@ -105,6 +105,10 @@ BenchmarkReporter::Run CreateRunReport(
   report.repetition_index = repetition_index;
   report.repetitions = repeats;
 
+  // Set `statistics` for every report, skipped or not, so the pointer is
+  // always valid in aggregation paths.
+  report.statistics = &b.statistics();
```
LebedevRI, line 110: As far as i can tell, the pointer is null-initialized by default,
so this doesn't prevent a dangling pointer.
Why is this nessesary?

dforsten: Prior to the latest change to statistics.cc it prevented a crash (the reported results were still incorrect though).
With the fixes to statistics.cc it is moot.

dforsten: I adapted our port overlay for the benchmark library to use the commit of the #2199 merge, with all custom patches removed, and am running our tests to verify if the crashes indeed are fixed.

dforsten: Looks good on our side as well, #2199 fixed the issue we were encountering.
I am closing this PR and await the arrival of the fix in the vcpkg registry.
Thank you!

LebedevRI:
@dforsten thank you!

Issue #1675, opened 2023-10-09
[BUG] DoNotOptimize build break with GCC


marcalff:
See downstream bug:

* https://github.com/open-telemetry/opentelemetry-cpp/issues/2343

When building the following code:

```
void BM_ParentBasedSamplerConstruction(benchmark::State &state)
{
  while (state.KeepRunning())
  {
    benchmark::DoNotOptimize(ParentBasedSampler(std::make_shared<AlwaysOnSampler>())); // line 49
  }
}
```

and

```
void BM_TraceIdRatioBasedSamplerConstruction(benchmark::State &state)
{
  while (state.KeepRunning())
  {
    benchmark::DoNotOptimize(TraceIdRatioBasedSampler(0.01)); // line 58
  }
}
```

the build fails with GCC 7.4.1, GCC 7.5.0, GCC 12.3.0.

The error reported is:

```
[ 56%] Building CXX object sdk/test/trace/CMakeFiles/sampler_benchmark.dir/sampler_benchmark.cc.o
In file included from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/sdk/test/trace/sampler_benchmark.cc:17:
/opt/benchmark-1.8.2/include/benchmark/benchmark.h: In instantiation of ‘typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type benchmark::DoNotOptimize(Tp&&) [with Tp = opentelemetry::v1::sdk::trace::ParentBasedSampler; typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type = void]’:
/home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/sdk/test/trace/sampler_benchmark.cc:49:29:   required from here
/opt/benchmark-1.8.2/include/benchmark/benchmark.h:540:3: error: read-only reference ‘value’ used as ‘asm’ output
  540 |   asm volatile("" : "+m"(value) : : "memory");
      |   ^~~
[...]

PR #2209, opened 2026-05-27
Clarify DoNotOptimize const-ref diagnostics

Haihan-Jiang:
Summary:
- make the deprecated const-reference DoNotOptimize diagnostic point to the actionable workaround
- add the same expression-result guidance to the user guide

Context: #1675 called out that the current failure mode can be hard to act on. This does not change DoNotOptimize semantics; it just makes the intended usage clearer.

Test:
- cmake --build build --target donotoptimize_test
- ctest --test-dir build -R '^donotoptimize_test$' --output-on-failure
- git diff --check

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1462,6 +1462,21 @@ known. For example:
   // while (...) DoNotOptimize(__result__);
 ```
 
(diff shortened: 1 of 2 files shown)


LebedevRI:
???

You need to fix your bot.

PR #2213, opened 2026-05-28
Add minimal Rust bindings PoC under bindings/rust/

### LINKED ISSUE #2212: [RFC] Add native Rust bindings via top-level `bindings/rust`
## Problem Statement

As Rust continues to mature as a systems programming language, an increasing number of polyglot codebases (C++ and Rust) require uniform performance tracking. Currently, organizations using `google/benchmark` across their C++ infrastructure lack a native way to harness the same execution engine, CLI flags (`--benchmark_filter`), and structured output formats (JSON/CSV) for their Rust services.

While the Rust ecosystem has profiling tools (e.g., `criterion`, `divan`), using a separate framework breaks parity in mixed-language environments and complicates unified CI/CD performance regression tracking. One of our maintainers has also reached out to the maintainers of criterion and has confirmed that it's only supported "best effort".

We want to explore adding **official, native Rust bindings** to `google/benchmark` without disrupting the existing C++ developer experience, build architecture, or performance characteristics.

## Proposed Architecture

To minimize maintenance overhead and guarantee that the bindings do not fall out of sync with core C++ API changes, we propose hosting the Rust bindings **co-located within the main repository** under a new `bindings/rust/` directory.

### Repository Layout

The proposed directory structure isolates the Rust toolchain dependencies entirely to its own subdirectory:

```text
google/benchmark/
├── CMakeLists.txt
├── src/
├── include/
└── bindings/
    ├── python/          # Existing python bindings
    └── rust/            # Proposed Rust bindings
        ├── Cargo.toml   # Rust package manifest
        ├── build.rs     # Cargo build script (drives CMake)
        └── src/
            ├── ffi.rs   # Declarative CXX bridge boundary
            └── lib.rs   # Safe, idiomatic Rust public API

```

### Build Integration (Bi-directional Isolation)

A key requirement is that C++ maintainers should **not** need a Rust toolchain installed to work on the library, and Rust users should experience a standard `cargo build` flow.

* **For Rust Users:** The `bindings/rust/build.rs` script will programmatically invoke `CMake` in the background to build the C++ static library artifact, then link it to the Rust crate automatically.
* **For C++ Users:** The top-level `CMakeLists.txt` will ignore the `bindings/rust` directory by default. We can introduce an optional flag, `-DBENCHMARK_ENABLE_RUST_BINDINGS=ON`, which will look for `cargo` and run the Rust test suite during CI.

### The Interop Layer (`cxx`)

Rather than maintaining a raw, unsafe C-string and pointer-heavy FFI layer, we propose using the **`cxx` crate**. `cxx` enforces type safety and static guardrails at the C++/Rust boundary, generating the required header plumbing automatically during compilation.

A preliminary layout of the bridge module (`ffi.rs`) would target the runtime loop and initialization:

```rust
#[cxx::bridge(namespace = "benchmark")]
mod ffi {
    unsafe extern "C++" {
        include!("benchmark/benchmark.h");

        type State;
        fn KeepRunning(self: Pin<&mut State>) -> bool;
        fn SkipWithError(self: Pin<&mut State>, msg: &str);
        
        fn Initialize(argc: &mut i32, argv: *mut *mut c_char);
        fn RunSpecifiedBenchmarks() -> usize;
    }
}

```

### Public Rust API & Static Initialization Challenge

Because Rust intentionally lacks life-before-main (`__attribute__((constructor))`) macro mechanisms for safety reasons, we cannot easily replicate the C++ `BENCHMARK(BM_StringCopy);` macro style.

Instead, clients of the public Rust bindings will utilize explicit **runtime registration** inside their own `fn main()`, mapping a Rust closure to the underlying C++ runner:

```rust
// Example of how an end-user might write a benchmark binary
fn main() {
    // Initialize arguments passed from the command line
    benchmark::initialize(); 

    // Explicitly register the benchmark with the C++ engine
    benchmark::register_benchmark("BM_RustVectorPush", |state| {
        while state.keep_running() {
            let mut vec = Vec::new();
            vec.push(42);
            benchmark::do_not_optimize(vec);
        }
    });

    // Hand over control to the C++ test runner
    benchmark::run_specified_benchmarks();
}

```

## Critical Trade-offs & Open Questions

Before committing to an implementation, we would love community feedback on the following areas:

1. **FFI Overhead:** Crossing the FFI boundary via `state.KeepRunning()` on every single iteration loop adds a small execution cost. For ultra-fast microbenchmarks (sub-nanosecond), could this skew timing accuracy compared to pure C++ loops? Could we hide this by implementing `state::keep_running()` using `KeepRunningBatch` under the hood?
2. **Community Maintenance:** Is there sufficient long-term interest from contributors with both C++ and Rust expertise to help review and maintain the `bindings/rust/` directory as the core engine evolves?

## Next Steps

If the community are supportive of this direction, we will start by submitting a minimal Proof of Concept (PoC) PR demonstrating a working `build.rs` loop that successfully binds and runs a simple benchmark.

rsd-darshan:
Closes #2212

## Summary

This is the minimal Proof of Concept proposed in the RFC. It demonstrates a working build-and-run loop that bridges Rust into the C++ benchmark runner without touching the C++ build for users who never invoke `cargo`.

- **`build.rs`** drives CMake in the background to build `libbenchmark.a`, then compiles a thin C++ bridge via `cxx-build`. C++ maintainers are completely unaffected.
- **`src/bridge.h` / `src/bridge.cc`** provide non-inline wrappers around `State::KeepRunning()`, which is `BENCHMARK_ALWAYS_INLINE` and has no linkable symbol. This mirrors how the existing Python bindings re-wrap `Initialize` for the same reason.
- **`src/lib.rs`** contains the `cxx` bridge declarations and a safe `BenchState` wrapper exposing `keep_running()` and `skip_with_error()`.
- **`examples/basic.rs`** registers two benchmarks (`BM_VectorPush`, `BM_StringCopy`) and runs them end-to-end.
- **`CMakeLists.txt`** gains an opt-in `-DBENCHMARK_ENABLE_RUST_BINDINGS=ON` flag that locates `cargo` and wires a custom target so the Rust suite runs alongside the C++ suite in CI.

## Test plan

- [x] `cd bindings/rust && cargo build` succeeds
- [x] `cargo run --example basic` produces benchmark output
- [x] `cmake -DBENCHMARK_ENABLE_RUST_BINDINGS=ON ..` finds cargo and registers the custom target
- [x] C++ build with no Rust flags is unaffected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ CMakeSettings.json
 dist/
 *.egg-info*
 uv.lock
+
+# Rust build artifacts
+bindings/rust/target/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ option(BENCHMARK_ENABLE_GTEST_TESTS "Enable building the unit tests which depend
 option(BENCHMARK_USE_BUNDLED_GTEST "Use bundled GoogleTest. If disabled, the find_package(GTest) will be used." ON)
 
 option(BENCHMARK_ENABLE_LIBPFM "Enable performance counters provided by libpfm" OFF)
+option(BENCHMARK_ENABLE_RUST_BINDINGS "Enable building and testing the Rust bindings (requires cargo)" OFF)
 
 # Export only public symbols
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
@@ -365,4 +366,21 @@ if (BENCHMARK_ENABLE_TESTING)
     endif()
   endif()
   add_subdirectory(test)
+endif()
+
+if(BENCHMARK_ENABLE_RUST_BINDINGS)
+  find_program(CARGO_EXE cargo)
+  if(NOT CARGO_EXE)
+    message(FATAL_ERROR
+      "BENCHMARK_ENABLE_RUST_BINDINGS is ON but cargo was not found. "
+      "Install the Rust toolchain (https://rustup.rs) and re-run CMake.")
+  endif()
+  message(STATUS "Rust bindings enabled. cargo: ${CARGO_EXE}")
+  add_custom_target(benchmark_rust_bindings ALL
+    COMMAND ${CARGO_EXE} test --manifest-path
+            ${CMAKE_CURRENT_SOURCE_DIR}/bindings/rust/Cargo.toml
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/bindings/rust
+    COMMENT "Building and testing Rust bindings"
+    VERBATIM
+  )
 endif()
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "google-benchmark"
+version = "0.1.0"
+edition = "2021"
+description = "Rust bindings for the google/benchmark C++ microbenchmarking library"
+license = "Apache-2.0"
+repository = "https://github.com/google/benchmark"
+
+[dependencies]
+cxx = "1"
+
+[build-dependencies]
+cxx-build = "1"
+cmake = "0.1"
+
+[[example]]
+name = "basic"
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -1,0 +1,63 @@
+use std::path::PathBuf;
+
+fn main() {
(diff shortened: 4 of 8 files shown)


dmah42:
RFC means request for comments. we are not ready for any implementation and i don't want to distract people with an implementation at this point. so thanks, but a bit premature.

PR #2214, opened 2026-05-28
Disable allocation zeroing in benchmarks on Android

charlesmunger:
This is particularly problematic because large allocations partially used can get eagerly memset even if they hit the backing allocator, faulting in a ton of pages, so the behavior can diverge substantially from real in-app performance.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -30,6 +30,10 @@
 #include <unistd.h>
 #endif
 
+#ifdef __ANDROID_NDK__
+#include <malloc.h>
+#endif
+
 #ifdef BENCHMARK_OS_LINUX
 #include <sys/personality.h>
 #endif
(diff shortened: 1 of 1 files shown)


LebedevRI:
I'm a bit confused by the diff.
Is the idea to disable hardening features to measure the best-case performance,
or because benchmarks happen to have hardening enabled, unlike the real production?
(Latter case seems especially weird to me, is hardening not specifically for the production?)

dmah42:
my understanding is that this is intended to mimic production allocation behaviour for various android versions.

LebedevRI:
(I don't have a leg in this either way, i just feel like the actual ansver is something that should be documented at least here as an issue comment.)

charlesmunger:
>Is the idea to disable hardening features to measure the best-case performance,
or because benchmarks happen to have hardening enabled, unlike the real production?

This hardening feature is on by default, but disabled for application processes, so in practice only system services built into the OS have it on. However, the disabling happens at runtime during the initialization of application processes before they load code, and a benchmark binary doesn't have that, so it runs in the "default" misleading configuration.

https://android.googlesource.com/platform/bionic/+/HEAD/docs/defines.md

The idea is that benchmark binaries built with the toolchain used to build apps should perform the way the code would if it was compiled into an app.

dmah42 (approved):

PR #2219, opened 2026-06-10
Handle missing locale in test on more platforms

alyssais:
My musl installation has no en_US.UTF-8 locale.  Rather than trying to hardcode a complete list of platforms where this might happen, just accept the situation where we don't have that locale available to test with, with the same behavior as the test had before on GCC MINGW64.

### DIFF
--- a/test/locale_impermeability_test.cc
+++ b/test/locale_impermeability_test.cc
@@ -10,9 +10,10 @@
 
 namespace {
 void BM_ostream(benchmark::State& state) {
-#if !defined(__MINGW64__) || defined(__clang__)
-  // GCC-based versions of MINGW64 do not support locale manipulations,
-  // don't run the test under them.
+#if !defined(__GLIBCXX__) || defined(__GLIBC__) || defined(__DragonFly__)
+  // libstdc++ only supports locale manipulations on GNU and
+  // DragonflyBSD platforms at the time of writing, don't run the test
+  // on other platforms when using libstdc++.
   std::locale::global(std::locale("en_US.UTF-8"));
 #endif
   while (state.KeepRunning()) {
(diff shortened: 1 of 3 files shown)


test/locale_impermeability_test.cc
```diff
@@ -10,11 +10,12 @@
...
 namespace {
 void BM_ostream(benchmark::State& state) {
-#if !defined(__MINGW64__) || defined(__clang__)
-  // GCC-based versions of MINGW64 do not support locale manipulations,
-  // don't run the test under them.
-  std::locale::global(std::locale("en_US.UTF-8"));
-#endif
+  try {
+    std::locale::global(std::locale("en_US.UTF-8"));
+  } catch (std::runtime_error&) {
```
dmah42: https://google.github.io/styleguide/cppguide.html#Exceptions

we don't use exceptions in this project.

alyssais: How do I tell if std::locale failed, or otherwise check whether a locale exists, then?  I can't help that the standard library throws an exception here.

dmah42: well that's why we have it wrapped in the platform macros. i agree, the natural approach would be what you have done, but it is not something we can budge on.

alyssais: Maybe we can at least use a more accurate platform macro check then.  Please see my new attempt.

dmah42:
i appreciate your flexibility, thank you. i agree this is a much better check and fixes other cases than your musl issue (Windows with mingw64+clang+libstdc++ should also be skipped i think, and now it should be).

dmah42:
hm. looks like maybe github actions have changed something that means the windows 2025 builds are failing. that's unrelated to your change i think.

dmah42 (approved):

PR #2229, opened 2026-06-23
Fix tsan/asan/hwasan/etc failures on Android

charlesmunger:
Some people run their benchmarks with sanitizers, and they should not spuriously fail.

This unblocks some test failures in google3

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -913,7 +913,9 @@ void Initialize(int* argc, char** argv, void (*HelperPrintf)()) {
   // Android 12 (API level 31) introduced zeroing of allocated memory in bionic
   // as a hardening feature; however, this is not enabled for apps.
   if (__builtin_available(android 31, *)) {
-    BM_CHECK_EQ(mallopt(M_BIONIC_ZERO_INIT, 0), 1);
+    // Not asserting on the return value, as sanitizers interpose the allocator
+    // and always return 0
+    (void)mallopt(M_BIONIC_ZERO_INIT, 0);
   }
 
   // The default configuration of bionic is to return pages to the OS as soon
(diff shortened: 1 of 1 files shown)


LebedevRI:
Could you please adjust the commit message to contain the previous output of each of those lines?

LebedevRI:
@charlesmunger poke

LebedevRI:
> @charlesmunger poke

charlesmunger:
I'm not sure what you mean? The asan allocator returns 0 for all mallopt calls regardless of the params. The output is just that the BM_CHECK triggers.

LebedevRI:
> I'm not sure what you mean? The asan allocator returns 0 for all mallopt calls regardless of the params. The output is just that the BM_CHECK triggers.

That's what the commit message should say. I'm just asking you to put the reasoning in the commit message.

dmah42:
i don't suppose there's any way to check if we're under sanitizer at runtime? it would be great to keep this check when _not_ running under sanitizer.

LebedevRI:
> i don't suppose there's any way to check if we're under sanitizer at runtime? it would be great to keep this check when _not_ running under sanitizer.

Sanitizers are compile-time opt-in, they can be detected the usual C way.
I'm not really sure what is the correct list of sanitizers that override mallopt, but i'd guess something like
```
#if __has_feature(address_sanitizer) || __has_feature(memory_sanitizer) || __has_feature(thread_sanitizer) || __has_feature(hwaddress_sanitizer)
// no BM_CHECK
#else
// can check return value
#end
```

charlesmunger:
I did try that at first, but new sanitizers are being added all the time, like dfsan and typesanitizer.

LebedevRI (approved):

LebedevRI:
@charlesmunger thank you!

PR #2233, opened 2026-06-25
Run an autofix pass on github workflows using zizmor

dmah42:
Zizmor is going to be automatically installed in this repo shortly. To avoid any churn, this is a pre-run to fix up any potential vulnerabilities identified by the tool.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -31,6 +31,8 @@ jobs:
         run: brew install ninja
 
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
 
       - name: build
         uses: threeal/cmake-action@725d1314ccf9ea922805d7e3f9d9bcbca892b406 # v2.1.0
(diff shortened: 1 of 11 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -78,21 +80,29 @@ jobs:
...
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
 
       - uses: lukka/get-cmake@f5b8fbb4d77cec1acc5a5f9f0df4beffaf5d98d9 # latest
 
       - name: configure cmake
         run: >
-          cmake -S . -B ${{ runner.workspace }}/_build/
+          cmake -S . -B $env:RUNNER_WORKSPACE/_build/
```
LebedevRI, line 90: This is literally the first time i'm seeing this syntax.

LebedevRI: I don't suppose there is a way to make this be `${RUNNER_WORKSPACE}`?

dmah42: apparently this is exactly the risk. the way the action works is that it does a literal text replacement, opening the door to potential malicious commands (i think it's unlikely but hey).

dmah42: oh and this is using the powershell syntax, basically. if it was linux it would be $RUNNER_WORKSPACE

LebedevRI:
I think this can be merged?

dmah42:
> I think this can be merged?

seems to be missing your review. probably because i had to merge.

LebedevRI (approved):

PR #2239, opened 2026-07-02
fix(linux): prevent infinite re-exec loop under AppArmor

### LINKED ISSUE #2184: [BUG] MaybeReenterWithoutASLR() infinite loop under AppArmor (not covered by #1984 fix)

anishesg:
The `MaybeReenterWithoutASLR()` function in `src/benchmark.cc` caused infinite execv() loops when running benchmarks under AppArmor-enabled LSMs. The existing fix from #1985 only checked whether personality(ADDR_NO_RANDOMIZE) succeeded in the current process before calling execv(). However, some LSMs like AppArmor can silently reset personality flags during the execve() system call transition, even though the flag was successfully set in the parent process.

This fix implements a fork-based verification: before committing to re-exec the parent process, we fork a test child, exec it with a special argument, and have it report back whether ADDR_NO_RANDOMIZE actually survived the exec boundary. Only if the child confirms the flag persists do we proceed with re-executing the parent. This prevents infinite loops in environments where the LSM strips the personality flag during exec.

The test child is invoked with `--benchmark_aslr_test_child=<fd>`, checks its personality after exec, writes the result back through the provided file descriptor, and exits. This approach maintains hermeticity by using process arguments rather than environment variables.

Fixes #2184


src/benchmark.cc
```diff
@@ -835,11 +837,26 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
 
 }  // end namespace internal
 
-void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
+void MaybeReenterWithoutASLR(int argc, char** argv) {
   // On e.g. Hexagon simulator, argv may be NULL.
   if (!argv) return;
```
LebedevRI:
```suggestion
void MaybeReenterWithoutASLR(int argc, char** argv) {
  (void)argc;
  // On e.g. Hexagon simulator, argv may be NULL.
  if (!argv) return;
```

src/benchmark.cc
```diff
@@ -835,11 +837,26 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
...
 
 #ifdef BENCHMARK_OS_LINUX
+  // Check if we are a test child process that should report personality and exit
+  for (int i = 1; i < argc; ++i) {
+    if (std::strncmp(argv[i], "--benchmark_aslr_test_child=", 28) == 0) {
```
LebedevRI: Just put the string into a static constexpr variable, and `strlen` it

src/benchmark.cc
```diff
@@ -864,9 +881,68 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
...
+
+    // Add test argument with pipe write fd (allocate on heap)
+    char* test_arg = static_cast<char*>(malloc(64));
+    if (!test_arg) _exit(1);
+    std::snprintf(test_arg, 64,
+                  "--benchmark_aslr_test_child=%d", pipefd[1]);
+    new_argv[arg_count] = test_arg;
+    new_argv[arg_count + 1] = nullptr;
+
+    execv(argv[0], new_argv);
```
LebedevRI: I haven't seen if ASAN splats on this,
but i really don't think this is allowed.
Can you even touch heap in a child?
Certainly you can't alloc before calling execv?

LebedevRI:
Thank you for taking a look!
Can you repro the original issue?

This does roughly look how i imagined it would look.
But i'm not yet confident the fix is safe.

src/benchmark.cc
```diff
@@ -864,9 +881,68 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
...
+    while (argv[arg_count] != nullptr) ++arg_count;
+
+    // Build new argv with test argument (allocate on heap to survive exec)
+    char** new_argv = static_cast<char**>(
+        malloc(sizeof(char*) * (arg_count + 2)));
+    if (!new_argv) _exit(1);
+
+    for (int i = 0; i < arg_count; ++i) {
+      new_argv[i] = argv[i];
+    }
```
LebedevRI: I'm not sure there is a need to preserve existing arguments,
just the `{argv[0], <magic>, nullptr}` is enough.

src/benchmark.cc
anishesg: done, switched to static constexpr with sizeof

src/benchmark.cc
anishesg: good catch, switched to stack allocation which is safe before exec

src/benchmark.cc
anishesg: simplified to just {argv[0], magic, nullptr}

anishesg:
I haven't personally reproduced the original issue reported in #2184, but the reporter confirmed the infinite loop occurs on Ubuntu 24.04 with AppArmor enabled. The fork-based verification approach should be safe - if the personality doesn't survive exec, we just skip re-exec and continue normally.

src/benchmark.cc
LebedevRI: Err, i didn't actually mean a stack-based array, just that it only needs to have 3 elements.
(in principle, i'm not sure we need to pass argv[0], either)

src/benchmark.cc
LebedevRI: I'm not sure any (either stack or heap) is safe before calling `execv`.
It would only be safe if it copies the actual underlying strings, does it do that?
I'm hoping ASAN will answer this. If this isn't safe, we need to produce `new_argv` before `fork()`.

src/benchmark.cc
```diff
@@ -835,11 +837,30 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
...
+  // Check if we are a test child process that should report personality and exit
+  for (int i = 1; i < argc; ++i) {
+    if (std::strncmp(argv[i], kTestChildArg, kTestChildArgLen) == 0) {
+      const int write_fd = std::atoi(argv[i] + kTestChildArgLen);
+      const auto test_personality = personality(0xffffffff);
+      // Write 1 if ADDR_NO_RANDOMIZE is set, 0 otherwise
+      char result = ((test_personality != -1) &&
+                     (internal::get_as_unsigned(test_personality) & ADDR_NO_RANDOMIZE))
+                        ? 1 : 0;
+      (void)write(write_fd, &result, 1);
```
LebedevRI: Looks like GCC wants some better check than cast-to-void here.

LebedevRI (commented):
Could you please make the CI happy?

src/benchmark.cc
```diff
@@ -864,9 +885,55 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
...
+  }
+
+  if (pid == 0) {
+    // Child: prepare to exec with test argument
+    close(pipefd[0]);
+
+    // Build test argument on stack (safe before exec)
+    char test_arg[64];
+    std::snprintf(test_arg, sizeof(test_arg),
+                  "--benchmark_aslr_test_child=%d", pipefd[1]);
```
LebedevRI:
```suggestion
                  "%s%d", kTestChildArg, pipefd[1]);
```

src/benchmark.cc
```diff
@@ -835,11 +837,30 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
 
 }  // end namespace internal
```
LebedevRI, line 839:
```suggestion

#ifdef BENCHMARK_OS_LINUX
static constexpr const char kTestChildArg[] = "--benchmark_aslr_test_child=";
bool ValidateNoASLRPersonalitySticks(char* argv0) {
  // Verify that the personality change survives exec() by testing in a child.
  // Some LSMs (e.g., AppArmor) may reset personality flags during execve(),
  // even though they were successfully set in the parent process.
  // This prevents infinite re-exec loops when the kernel silently resets
  // the personality after each exec.
  int pipefd[2];
  if (pipe(pipefd) == -1) return false;

  pid_t pid = fork();
  if (pid == -1) {
    close(pipefd[0]);
    close(pipefd[1]);
    return false;
  }

  if (pid == 0) {
    // Child: prepare to exec with test argument
    close(pipefd[0]);

    // Build test argument on stack (safe before exec)
    char test_arg[64];
    int test_arg_len = std::snprintf(test_arg, sizeof(test_arg), "%s%d",
                                     kTestChildArg, pipefd[1]);
    BM_CHECK_LT(test_arg_len, sizeof(test_arg));
    (void)test_arg_len;

    // Simple argv with just the executable and test argument
    char* child_argv[] = {argv0, test_arg, nullptr};

    execv(argv0, child_argv);
    // If exec fails, exit
    _exit(1);
  }

  // Parent: wait for child to report back
  close(pipefd[1]);

  char result = 0;
  ssize_t nread = read(pipefd[0], &result, 1);
  close(pipefd[0]);

  int status;
  waitpid(pid, &status, 0);

  // Did we successfully read the result and it indicates ADDR_NO_RANDOMIZE?
  return status == 0 && nread == 1 && result == 1;
}
#endif

```

src/benchmark.cc
LebedevRI:
```suggestion
void MaybeReenterWithoutASLR(int argc, char** argv) {
  (void)argc;

  // On e.g. Hexagon simulator, argv may be NULL.
  if (!argv) return;

#ifdef BENCHMARK_OS_LINUX
  static constexpr size_t kTestChildArgLen = sizeof(kTestChildArg) - 1;

  // Check if we are a test child process that should report personality and
  // exit
  if (argc == 2 &&
      std::strncmp(argv[1], kTestChildArg, kTestChildArgLen) == 0) {
    const int write_fd = std::atoi(argv[1] + kTestChildArgLen);
    const auto test_personality = personality(0xffffffff);
    // Write 1 if ADDR_NO_RANDOMIZE is set, 0 otherwise
    char result =
        ((test_personality != -1) &&
         (internal::get_as_unsigned(test_personality) & ADDR_NO_RANDOMIZE))
            ? 1
            : 0;
    const auto nbytes = write(write_fd, &result, 1);
    close(write_fd);
    if (test_personality == -1 || nbytes != 1) std::exit(1);
    std::exit(0);
  }

  const auto curr_personality = personality(0xffffffff);

  // We should never fail to read-only query the current personality,
  // but let's be cautious.
  if (curr_personality == -1) return;

  // If ASLR is already disabled, we have nothing more to do.
  if (internal::get_as_unsigned(curr_personality) & ADDR_NO_RANDOMIZE) return;

  // Try to change the personality to disable ASLR.
  const auto proposed_personality =
      internal::get_as_unsigned(curr_personality) | ADDR_NO_RANDOMIZE;
  const auto prev_personality = personality(proposed_personality);

  // Have we failed to change the personality? That may happen.
  if (prev_personality == -1) return;

  // Make sure the parsona has been updated with the no-ASLR flag,
  // otherwise we will try to reenter infinitely.
  // This seems impossible, but can happen in some docker configurations.
  const auto new_personality = personality(0xffffffff);
  if (new_personality == -1) return;
  if ((internal::get_as_unsigned(new_personality) & ADDR_NO_RANDOMIZE) == 0)
    return;

  // Additionally, ensure that the personality change would survive exec().
  if (ValidateNoASLRPersonalitySticks(argv[0])) {
    execv(argv[0], argv);
  }

  // Personality doesn't survive exec() boundary, or execv() failed.
  // We need to try to un-change the personality to "re-enable" ASLR,
  // at least so that there is a warning in the output,
  // and continue as-is.
  const auto restored_personality =
      internal::get_as_unsigned(new_personality) & ~ADDR_NO_RANDOMIZE;
  personality(restored_personality);
  // This may or may not have failed, but we're out of options here.
#else
  return;
#endif
}
```

src/benchmark.cc
```diff
@@ -835,11 +837,185 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
...
+  // If we successfully read the result and it indicates ADDR_NO_RANDOMIZE
+  // survived exec, proceed with re-exec. Otherwise, skip to avoid infinite loop.
+  if (nread == 1 && result == 1) {
+    execv(argv[0], argv);
+  }
+
+  // Personality doesn't survive exec boundary, skip re-exec.
+#else
+  return;
+#endif
```
LebedevRI:
```suggestion

```

src/benchmark.cc
```diff
@@ -835,11 +837,85 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
...
+
+  if (pid == 0) {
+    // Child: prepare to exec with test argument
+    close(pipefd[0]);
+
+    // Build test argument on stack (safe before exec)
+    char test_arg[64];
+    int test_arg_len = std::snprintf(test_arg, sizeof(test_arg), "%s%d",
+                                     kTestChildArg, pipefd[1]);
+    BM_CHECK_LT(test_arg_len, sizeof(test_arg));
```
LebedevRI:
```suggestion
    BM_CHECK_LT(static_cast<size_t>(test_arg_len), sizeof(test_arg));
```

src/benchmark.cc
```diff
@@ -861,12 +937,23 @@ void MaybeReenterWithoutASLR(int /*argc*/, char** argv) {
...
+  if (ValidateNoASLRPersonalitySticks(argv[0])) {
+    execv(argv[0], argv);
+  }
+
+  // Personality doesn't survive exec() boundary, or execv() failed.
+  // We need to try to un-change the personality to "re-enable" ASLR,
+  // at least so that there is a warning in the output,
+  // and continue as-is.
+  const auto restored_personality =
+      internal::get_as_unsigned(new_personality) & ~ADDR_NO_RANDOMIZE;
```
LebedevRI:
```suggestion
  const auto restored_personality =
      internal::get_as_unsigned(new_personality) &
      ~internal::get_as_unsigned(ADDR_NO_RANDOMIZE);
```

src/benchmark.cc
```diff
@@ -835,11 +837,85 @@ std::make_unsigned_t<T> get_as_unsigned(T v) {
...
+#ifdef BENCHMARK_OS_LINUX
+static constexpr const char kTestChildArg[] = "--benchmark_aslr_test_child=";
+bool ValidateNoASLRPersonalitySticks(char* argv0) {
+  // Verify that the personality change survives exec() by testing in a child.
+  // Some LSMs (e.g., AppArmor) may reset personality flags during execve(),
+  // even though they were successfully set in the parent process.
+  // This prevents infinite re-exec loops when the kernel silently resets
+  // the personality after each exec.
+  int pipefd[2];
+  if (pipe(pipefd) == -1) return false;
```
LebedevRI:
```suggestion
  if (pipe(pipefd) != 0) return false;
```

LebedevRI (approved):
@dmah42 thoughts?

dmah42 (approved):

LebedevRI:
@anishesg @dmah42 thank you!

PR #2246, opened 2026-07-12
Fix DoNotOptimize compilation error with partially const structs (Fixes #1997)

### LINKED ISSUE #1997: [BUG] DoNotOptimize on values with const members again
**Describe the bug**

```

struct X
{
  X() : value{} {}

  const int value;
};

template<class T>
void foo(T&& f)
{
    X x;
    f(x);
}

int main() {
}
```

MaciekChoszczyk:
Fixes #1997.

### What's wrong?
Passing structs with `const` members to `DoNotOptimize` causes compilation errors because the `asm` constraints (`+m`, `+r`) request write access to memory that the compiler considers `const`.

### Fix
Cast the input value to a raw byte array (`char(&)[sizeof(Tp)]`) to strip `const` qualifiers for the optimizer. This safely allows write access in the assembly block.

### Notes
- `const_cast` is not removing 'const' from internal struct members.
- `std::bit_cast` was considered but avoided to maintain compatibility with older C++ standards.
- This approach aligns with the existing workaround used for MSVC, which was missing from the GCC/Clang implementations.

### DIFF
--- a/include/benchmark/utils.h
+++ b/include/benchmark/utils.h
@@ -53,18 +53,18 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp const& value) {
 template <class Tp>
 inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
 #if defined(__clang__)
-  asm volatile("" : "+r,m"(value) : : "memory");
+  asm volatile("" : "+r,m"(reinterpret_cast<char(&)[sizeof(Tp)]>(value)) : : "memory");
 #else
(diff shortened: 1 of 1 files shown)


dmah42:
this seems deeply unwise, casting away const. maybe there's a better option? and maybe this isn't an issue as people can just... not use const?

LebedevRI:
I also don't like this. I fear more concrete solution would require a compiler builtin.

dmah42:
i'm going to close this out. it's just not a safe approach to a non-problem.

PR #2247, opened 2026-07-13
Exclude setup/teardown from MemoryManager measurement window (#2149)

devtejasx:
RunMemoryManager called memory_manager->Start() before allocating the internal ThreadManager and running the benchmark's Setup(), and called Stop() after resetting the ThreadManager and running Teardown(). As a result the MemoryManager attributed the library's own bookkeeping allocation and any per-benchmark fixture allocations to the benchmark, so the reported allocation count could never be accurate.

Setup()/Teardown() already run outside the timed region; make memory measurement consistent by bracketing only RunInThread with Start()/ Stop(), and move the ThreadManager creation and Setup()/Teardown() outside that window.

Add memory_manager_ordering_test, which registers a MemoryManager and asserts no Setup/Teardown occurs between Start() and Stop(). It fails on the previous ordering and passes with this change.


test/memory_manager_ordering_test.cc
```diff
@@ -0,0 +1,70 @@
...
+  assert(start < stop && "Start must precede Stop");
+
+  // Nothing between Start and Stop may be a Setup or Teardown: those run
+  // outside the measured window (this is what #2149 fixes).
+  for (auto it = start + 1; it != stop; ++it) {
+    assert(*it != "Setup" &&
+           "Setup ran inside the MemoryManager Start/Stop window");
+    assert(*it != "Teardown" &&
+           "Teardown ran inside the MemoryManager Start/Stop window");
+  }
```
LebedevRI: We can't rely on `assert` for correctness checks, this should be a gtest-based test.

test/memory_manager_ordering_test.cc
```diff
@@ -0,0 +1,70 @@
...
+  void Start() override { events.emplace_back("Start"); }
+  void Stop(Result& result) override {
+    events.emplace_back("Stop");
+    result.num_allocs = 0;
+    result.max_bytes_used = 0;
+  }
+};
+
+void DoSetup(const benchmark::State&) { events.emplace_back("Setup"); }
+void DoTeardown(const benchmark::State&) { events.emplace_back("Teardown"); }
```
LebedevRI: Instead of collecting strings,
can we just have a boolean for MM and a boolean for Fixture,
and perform correctness checking in callbacks?

test/memory_manager_ordering_test.cc
devtejasx: Done — replaced the assert-based test with a proper gtest (`memory_manager_ordering_gtest.cc`, wired up via `add_gtest`). All checks are `EXPECT_*` now.

test/memory_manager_ordering_test.cc
devtejasx: Done — no more string collection. The manager sets an `in_measurement_window` bool in `Start()`/`Stop()`, and the Setup/Teardown callbacks EXPECT it to be false directly, plus `*_ran` bools so the test fails loudly instead of passing vacuously if a path never executes.

src/benchmark_runner.cc
```diff
@@ -462,17 +462,23 @@ void BenchmarkRunner::RunWarmUp() {
 
 MemoryManager::Result BenchmarkRunner::RunMemoryManager(
     IterationCount memory_iterations) {
-  memory_manager->Start();
+  // Set up the thread manager and run the user's Setup() outside the
```
dmah42, line 465: not sure what this comment is adding to the code.

PR #2248, opened 2026-07-13
Fix default reporter use after free

### LINKED ISSUE #2240: [BUG]Heap-use-after-free in benchmark `RunSpecifiedBenchmarks
A heap-use-after-free vulnerability in google/benchmark 1.9.4. When `RunSpecifiedBenchmarks()` is called with a filter that matches no registered benchmark, the static reporter pointer is freed but the cached pointer returned by `CreateDefaultDisplayReporter()` is not invalidated, causing a use-after-free on the next call.

### Impact

- Denial-of-service or potential memory corruption
- Any application using the benchmark library with dynamic filtering is affected

### Reproduction

All materials are available in my research repository:
https://github.com/eglonnnn/opensource-fuzz-vulnerability-research/tree/main/Heap-use-after-free%20in%20benchmark%20RunSpecifiedBenchmarks

devtejasx:
Fixes #2240.

Root cause: CreateDefaultDisplayReporter() cached the reporter it creates in a function-local static, while every caller owns and frees the returned pointer (RunSpecifiedBenchmarks() wraps it in a std::unique_ptr). The first call therefore freed the cached object, and every later call returned a dangling pointer, producing the heap-use-after-free reported in #2240.

Fix: return a fresh, caller-owned reporter on every call, and document the ownership contract at the public declaration in benchmark_api.h.

Testing:
- Added a regression test covering repeated CreateDefaultDisplayReporter() calls.
- The test verifies that each call returns an independent caller-owned reporter.
- The test fails against the old implementation and passes with this fix.
- Release build/tests: 86/86 passed.
- Debug build/tests: 86/86 passed.
- Builds remain -Werror clean.

Note: the branch was rebased onto current main to remove an unrelated commit that was previously included by mistake.

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -55,8 +55,9 @@ BENCHMARK_EXPORT void SetBenchmarkFilter(std::string value);
 BENCHMARK_EXPORT int32_t GetBenchmarkVerbosity();
 
 // Creates the display reporter selected by --benchmark_format. Returns a
-// fresh instance on every call; the caller owns and frees it.
-BENCHMARK_EXPORT BenchmarkReporter* CreateDefaultDisplayReporter();
+// fresh instance on every call, owned by the caller.
+BENCHMARK_EXPORT std::unique_ptr<BenchmarkReporter>
+CreateDefaultDisplayReporter();
 
 BENCHMARK_EXPORT size_t RunSpecifiedBenchmarks();
 BENCHMARK_EXPORT size_t RunSpecifiedBenchmarks(std::string spec);
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -54,6 +54,8 @@ BENCHMARK_EXPORT void SetBenchmarkFilter(std::string value);
 
 BENCHMARK_EXPORT int32_t GetBenchmarkVerbosity();
 
+// Creates the display reporter selected by --benchmark_format. Returns a
+// fresh instance on every call; the caller owns and frees it.
 BENCHMARK_EXPORT BenchmarkReporter* CreateDefaultDisplayReporter();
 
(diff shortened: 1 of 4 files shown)


LebedevRI:
Diff contains spurious commits.
Blame points at https://github.com/google/benchmark/commit/6e51dcbcc3965b3f4b13d4bab5e43895c1a73290 - @dmah42 i do not understand why that `static` was introduced in an otherwise NFC refactoring.

devtejasx:
Rebased — the branch is now a single commit (f1f1485) on top of current main containing only the 4-file fix. On the `static`: whatever the original intent in 6e51dcb, it can't be correct in the current API since both call sites (`RunSpecifiedBenchmarks` overloads and `benchmark_main`) wrap the returned pointer in a `unique_ptr` and free it, so the cached pointer dangles from the second call on — which is exactly what #2240 hit. This PR drops the cache so every caller gets a fresh, owned instance, with a regression test covering it.

include/benchmark/benchmark_api.h
```diff
@@ -54,6 +54,8 @@ BENCHMARK_EXPORT void SetBenchmarkFilter(std::string value);
 
 BENCHMARK_EXPORT int32_t GetBenchmarkVerbosity();
 
+// Creates the display reporter selected by --benchmark_format. Returns a
```
dmah42, line 57: if the caller is meant to own the returned type, then we should be returning a `unique_ptr`.

https://google.github.io/styleguide/cppguide.html#Ownership_and_Smart_Pointers

devtejasx:
Done — `CreateDefaultDisplayReporter()` now returns `std::unique_ptr<BenchmarkReporter>`, making the caller's ownership explicit in the signature rather than in a comment. Both call sites and the regression test updated accordingly. Also rebased onto current main.

devtejasx:
Withdrawing this PR.

PR #2249, opened 2026-07-15
Pass benchmark State to ProfilerManager hooks

### LINKED ISSUE #2234: [FR] Pass State into the ProfilerManager
Hello

Some custom profilers for regional profiling require registering a region with a string name. It's convenient to use the name of the benchmark itself as this name. Therefore, I suggest adding a mechanism to the ProfilerManager that would get the benchmark name

One way to implement this is to pass State as an argument to the AfterSetupStart and BeforeTeardownStop methods. Then, the ProfilerManager will have access to the benchmark name via the name() method

To maintain backward compatibility with the current interface, I propose adding a new function overload and leaving the old one (you can mark it as deprecated if you prefer)

virtual void AfterSetupStart() {};
virtual void AfterSetupStart(const State&) { AfterSetupStart() };

And in benchmark.cc we pass *this into this methods

Currently, this is the only context for using this feature. However, I think CustomProfiler's ability to look into State could be useful for other tasks as well. This would allow more context to be passed to profilers. But if you think that passing the State is too much, we could add a lightweight method to pass only the name of the bench

Thank you

devtejasx:
Custom profilers doing regional profiling need information about the benchmark being profiled, e.g. its name to register a named region. Add AfterSetupStart(const State&) and BeforeTeardownStop(const State&) overloads to ProfilerManager and call them with the State of the profiling run. The default implementations forward to the existing parameterless hooks, which are no longer pure virtual, so existing ProfilerManager implementations keep working unchanged.

Fixes google/benchmark#2234

### DIFF
--- a/include/benchmark/managers.h
+++ b/include/benchmark/managers.h
@@ -24,6 +24,8 @@
 
 namespace benchmark {
 
+class State;
+
 class MemoryManager {
  public:
   static constexpr int64_t TombstoneValue = std::numeric_limits<int64_t>::max();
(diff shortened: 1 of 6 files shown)


include/benchmark/managers.h
```diff
@@ -54,8 +56,20 @@ void RegisterMemoryManager(MemoryManager* memory_manager);
 class ProfilerManager {
  public:
   virtual ~ProfilerManager() {}
-  virtual void AfterSetupStart() = 0;
-  virtual void BeforeTeardownStop() = 0;
+  // Kept for backwards compatibility; new code should override the
```
dmah42, line 59: so this is not backwards compatible. current users now have to add the `using` statement to compile correctly.

devtejasx: you're right adding the overload hides the base method and breaks source compat for existing implementers under Woverloaded virtual. I'll rework it so the running State is reachable without changing the existing hook signatures, keeping current ProfilerManager subclasses compiling unchanged, and push an update.

Then I actually do the rework so that reply is backed by a real change. This one is worth pursuing.

devtejasx:
Reworked to address the review feedback: instead of overloading the hooks (which hid the base methods under -Woverloaded-virtual), the existing pure-virtual hooks are kept unchanged and a protected `ProfilerManager::GetState()` accessor is added, so existing subclasses compile unchanged. This branch was force-pushed, so this PR can't be reopened; the reworked change is in #2253.

PR #2250, opened 2026-07-15
Support per-benchmark custom context in JSON output

### LINKED ISSUE #1890: [FR] Allow to add more information in the custom output for json
I generate benchmarks with different variations.
However, last I checked, the only field I can populate is `benchmark_name`.
So I first populate the name with the things I need:
https://github.com/DenisYaroshevskiy/unsq_eve/blob/a2c914f0c37e57911ba47751db0463a33f48bf22/src/bench/bench.h#L100
And then parse them out in the visualization.

It would be great to have some way to add an object to the outputted json.

devtejasx:
Add Benchmark::AddCustomContext(key, value) to attach string key-value pairs to a benchmark at registration time. The pairs are reported with every result of that benchmark (including aggregates) as a custom_context object in the JSON output, complementing the existing run-wide benchmark::AddCustomContext() which only goes into the top-level context. This allows tagging generated benchmark variants with structured metadata instead of encoding it into the benchmark name and parsing it back out of reports.

Fixes google/benchmark#1890

### DIFF
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -21,6 +21,7 @@
 #endif
 
 #include <functional>
+#include <map>
 #include <memory>
 #include <string>
 #include <utility>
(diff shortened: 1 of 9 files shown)


include/benchmark/benchmark_api.h
```diff
@@ -141,6 +142,12 @@ class BENCHMARK_EXPORT Benchmark {
   Benchmark* DenseThreadRange(int min_threads, int max_threads, int stride = 1);
   Benchmark* ThreadPerCpu();
   Benchmark* ThreadRunner(threadrunner_factory&& factory);
+  // Adds a key-value pair that is reported alongside every result of this
```
dmah42, line 145: 1. is this the right place for this method and 2. do we need the long comment? do other methods have these long comments, or do we document these things elsewhere?

dmah42:
can you explain the use-case more clearly? the benchmark already supports plenty of ways to change the name, or add parameters... why do we need custom context?

devtejasx:
Fair point for the metadata-tagging case I had in mind, a naming/ArgNames convention plus the existing run-wide context is enough, and it's not worth adding new API surface for. I'll close this. Thanks for taking a look.

PR #2251, opened 2026-07-15
Replace offsetof layout check with a conforming runtime check

### LINKED ISSUE #1877: [BUG] fails to compile on windows with clang-18 / clang-19
```cpp
  // Offset tests to ensure commonly accessed data is on the first cache line.
  const int cache_line_size = 64;
                    (cache_line_size - sizeof(error_occurred_)),
                "");
```

error:

```cpp
/src/benchmark.cc:191:17: error: offset of on non-standard-layout type 'State' [-Werror,-Winvalid-offsetof]
  191 |   static_assert(offsetof(State, error_occurred_) <=
      |                 ^               ~~~~~~~~~~~~~~~
E:\llvm_\LLVM\lib\clang\18\include\__stddef_offsetof.h:16:24: note: expanded from macro 'offsetof'
   16 | #define offsetof(t, d) __builtin_offsetof(t, d)
      |                        ^                     ~
1 error generated.
```

devtejasx:
In #1877, using `offsetof` on `State` (not a standard-layout type) started erroring under clang-18/19 with `-Winvalid-offsetof`. PR #1821 unblocked builds by suppressing the warning, but the issue was left open, asking for ideas for how to capture the spirit of the check in another way.

This replaces the conditionally-supported `offsetof` static_assert - and the warning-suppression pragma blocks it required for four compiler vendors (Intel, GCC/Clang, NVCC, NVHPC) - with a fully conforming runtime check in the `State` constructor: a `BM_CHECK_LE` comparing `reinterpret_cast<const char*>(&skipped_) - reinterpret_cast<const char*>(this)` against the cache line size. Pointer arithmetic over the object representation is well-defined, so no vendor pragmas are needed.

Trade-off vs. the static_assert: the check now fires in debug builds at State construction time rather than at compile time. Since every test in the suite constructs `State` objects, any layout regression is still caught by CI debug runs. `BM_CHECK` compiles away entirely under `NDEBUG`, so release builds get no new overhead.

Net: -36/+12 lines, all four pragma stacks gone.

Testing: full CMake test suite passes in both Release and Debug (where the check is active) locally with MinGW-w64 GCC 15.2 on Windows 11.

Fixes #1877

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -42,6 +42,7 @@
 #include <algorithm>
 #include <atomic>
 #include <condition_variable>
+#include <cstddef>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


dmah42:
let's be clear: this is a regression. before we would have the check at compile time under all compile targets, now it's only under debug and only when run and this particular library is almost never run under debug (it is in CI but that's post commit which is quite the shift-right).

is there no better option for a compile time check? that's why the issue was left open.

LebedevRI:
Good thing we are at C++17 then.
https://github.com/google/benchmark/blob/eb15fc80dd8a1d2c922fff3551db80dfc6ef2c52/CMakeLists.txt#L148

PR #2252, opened 2026-07-15
Fix a batch of clang-tidy findings in src/

devtejasx:
Works through a slice of the clang-tidy findings collated in #1925, scoped to `src/` and split into one commit per check for easier review:

- `misc-const-correctness`: declare immutable locals `const` (13 files, includes the `MutexLock` guards and the `State` objects constructed for Setup/Teardown callbacks)
- `readability-uppercase-literal-suffix`: `0u` -> `0U`
- `readability-simplify-boolean-expr` / `readability-isolate-declaration` / `readability-use-anyofallof`: De Morgan simplifications in `benchmark_runner.cc` and `commandlineflags.cc`, split a multi-variable declaration in `sysinfo.cc`, replace a hand-written find loop in `SameNames` with `std::all_of`
- `misc-use-anonymous-namespace`: move the file-static `IgnoreColorPrint`/`FormatTime` helpers in `console_reporter.cc` into an anonymous namespace

Not included on purpose: `readability-convert-member-functions-to-static` on `PerfCountersMeasurement::Read` (the suggestion is only valid for the non-libpfm stub, making it static would break the Linux/libpfm build), and the `cppcoreguidelines-avoid-non-const-global-variables` findings on flag definitions and registration globals, which look intentional.

The remaining findings were re-checked with clang-tidy 20 against a fresh `compile_commands.json`; the checks addressed here now report clean for `src/` in a Windows configuration (Linux-only code paths, e.g. libpfm perf counters, were not analyzed and may retain findings).

Testing: full CMake test suite passes locally in Debug and the library and tests build warning-free in Release (`-Wall -Wextra -Werror`, MinGW-w64 GCC 15.2 on Windows 11).

Refs #1925

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -290,7 +290,7 @@ void State::ResumeTiming() {
 void State::SkipWithMessage(const std::string& msg) {
   skipped_ = internal::SkippedWithMessage;
   {
-    MutexLock l(manager_->GetBenchmarkMutex());
+    MutexLock const l(manager_->GetBenchmarkMutex());
     if (internal::NotSkipped == manager_->results.skipped_) {
       manager_->results.skip_message_ = msg;
       manager_->results.skipped_ = skipped_;
@@ -305,7 +305,7 @@ void State::SkipWithMessage(const std::string& msg) {
 void State::SkipWithError(const std::string& msg) {
   skipped_ = internal::SkippedWithError;
   {
-    MutexLock l(manager_->GetBenchmarkMutex());
+    MutexLock const l(manager_->GetBenchmarkMutex());
     if (internal::NotSkipped == manager_->results.skipped_) {
       manager_->results.skip_message_ = msg;
       manager_->results.skipped_ = skipped_;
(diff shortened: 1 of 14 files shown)


dmah42:
is there any kind of enforcement we can apply? we have a clang-tidy workflow that runs on PRs but it wasn't flagging any of these issues.

LebedevRI:
> is there any kind of enforcement we can apply? we have a clang-tidy workflow that runs on PRs but it wasn't flagging any of these issues.

Looks like that is using clang-tidy-18, which isn't the freshest. Perhaps updating to a newer version would change things.
Additionally, are those lints even enabled? IIRC there is no way to mass-enable all lints in clang-tidy.

dmah42:
> > is there any kind of enforcement we can apply? we have a clang-tidy workflow that runs on PRs but it wasn't flagging any of these issues.
> 
> Looks like that is using clang-tidy-18, which isn't the freshest. Perhaps updating to a newer version would change things. Additionally, are those lints even enabled? IIRC there is no way to mass-enable all lints in clang-tidy.

ah fair. well if this PR could also bump clang-tidy and enable lints we care about that would be great. i believe i did select the set i cared about in https://github.com/google/benchmark/blob/main/.clang-tidy and the ones fixed in this PR don't look all that useful.

devtejasx:
Withdrawing this PR.

PR #2253, opened 2026-07-15
Pass benchmark State to ProfilerManager hooks via GetState() accessor

### LINKED ISSUE #2234: [FR] Pass State into the ProfilerManager

devtejasx:
Supersedes #2249, which could not be reopened after its branch was force-pushed to address review feedback.

Profilers that do regional profiling need information about the benchmark being profiled, e.g. its name to register a named profiling region. This exposes the running benchmark's `State` to the `ProfilerManager` hooks.

Instead of overloading `AfterSetupStart()`/`BeforeTeardownStop()` with `State`-taking variants (which hide the base methods and force existing subclasses to add `using` declarations under `-Woverloaded-virtual`), the existing pure-virtual hooks are left unchanged and a protected `ProfilerManager::GetState()` accessor is added. The library sets the active `State` immediately before invoking each hook; the profiling run is single-threaded, so no synchronization is needed. Existing `ProfilerManager` implementations compile unchanged.

Fixes google/benchmark#2234

Testing:
- Added coverage in profiler_manager_test that reads the benchmark name through GetState() from within AfterSetupStart().
- Release build/tests and Debug build/tests pass (profiler_manager tests plus the full suite).
- Builds remain -Werror clean.

### DIFF
--- a/include/benchmark/managers.h
+++ b/include/benchmark/managers.h
@@ -61,9 +61,10 @@ class ProfilerManager {
 
  protected:
   // The State of the benchmark run being profiled, giving access to e.g. the
-  // benchmark name (for a named profiling region). Only valid to call from
-  // within AfterSetupStart() and BeforeTeardownStop().
-  const State& GetState() const { return *state_; }
+  // benchmark name (for a named profiling region). Non-null only while
+  // AfterSetupStart() or BeforeTeardownStop() is executing; any other caller
+  // observes nullptr.
+  const State* GetState() const { return state_; }
 
  private:
   friend class State;
--- a/include/benchmark/managers.h
+++ b/include/benchmark/managers.h
@@ -24,6 +24,8 @@
 
 namespace benchmark {
 
+class State;
+
 class MemoryManager {
  public:
   static constexpr int64_t TombstoneValue = std::numeric_limits<int64_t>::max();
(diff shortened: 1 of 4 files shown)


include/benchmark/managers.h
```diff
@@ -56,6 +58,16 @@ class ProfilerManager {
   virtual ~ProfilerManager() {}
   virtual void AfterSetupStart() = 0;
   virtual void BeforeTeardownStop() = 0;
+
+ protected:
+  // The State of the benchmark run being profiled, giving access to e.g. the
```
dmah42, line 63: what happens if it's called outside of those methods? how can we enforce that? maybe return a pointer instead and allow it to be null?

devtejasx:
Good point — a reference gave no safe answer for out-of-window calls. Changed `GetState()` to return `const State*`, and the enforcement is now active rather than documented: `State` clears the stored pointer immediately after each hook returns, so any caller outside `AfterSetupStart()`/`BeforeTeardownStop()` observes `nullptr`. Test now asserts both the non-null name inside the hook and the `nullptr` outside; docs updated. Also rebased onto main.

devtejasx:
Withdrawing this PR.

PR #2256, opened 2026-07-18
Respect min_time per thread in multi-threaded benchmarks (#2117)

### LINKED ISSUE #2117: [BUG] Minimum benchmark time with multiple threads
**Describe the bug**
https://github.com/google/benchmark/pull/1836 fixed https://github.com/google/benchmark/issues/1834 and now the reported output makes more sense. However, the minimum benchmark time doesn't seem to be respected.

**System**
Arch Linux, GCC 15.2.1, `benchmark` 1.9.5-1

**To reproduce**
Consider
```c++

static void BM_MultiThreaded(benchmark::State& state) {
  int a = 0;

  for (auto _ : state) {
    for (int i = 0; i < 1000; i++) {
      benchmark::DoNotOptimize(a++);
    }
  }
}
BENCHMARK(BM_MultiThreaded)->ThreadRange(1, 8);

BENCHMARK_MAIN();
```

The output is somewhat expected:
```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
BM_MultiThreaded/threads:1        275 ns          274 ns      2585850
BM_MultiThreaded/threads:2        273 ns          272 ns      2259184
BM_MultiThreaded/threads:4        279 ns          278 ns      2449204
BM_MultiThreaded/threads:8        308 ns          307 ns      2148808
```
However, notice that the number of iterations stays constant - it should increase as more threads are used to run the iterations. For example:
```
 $ /usr/bin/time -v ./benchmark-multithreaded --benchmark_min_time=1s --benchmark_filter=threads:8
[...]
	User time (seconds): 1.72
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.25
```

With previous versions of `benchmark`, this could be fixed with `UseRealTime()` but that doesn't seem to work anymore since (probably) https://github.com/google/benchmark/pull/1836.

**Expected behavior**
Each benchmark should spend the minimum time (per thread).

devtejasx:
Fixes #2117.

### Problem

`real_time` and `manual_time` (and whole-process CPU time) are accumulated across all threads in `RunInThread` (they are summed under the manager lock). #1836 removed the per-thread scaling of those values so that the **reported** `Time` is correct wall-clock time. However, the very same values also feed the **min-time stopping decision** via `IterationResults::seconds`.

The consequence: a benchmark running `N` threads stops as soon as the *summed* time across threads reaches `min_time`, i.e. after only `min_time / N` of wall-clock time per thread. This is exactly what #2117 reports — the iteration count doesn't grow with more threads, `--benchmark_min_time` isn't honored, and `UseRealTime()` no longer fixes it (it did before #1836).

```
BM_MultiThreaded/threads:1   ... 2585850
BM_MultiThreaded/threads:8   ... 2148808   # should be ~8x, wall time ~min_time/8
```

### Fix

Add `BenchmarkRunner::GetTimeForDecision()`, which returns the **per-thread** duration by dividing the accumulated real/manual/process-CPU time by the thread count — but only for the stopping decision (`PredictNumItersNeeded` and `ShouldReportIterationResults`). The summed values stored in `IterationResults::results` are left untouched, so reported times and global counter rates are unaffected. The `real_time_used >= 5 * min_time` runaway guard is scaled the same way.

This restores the pre-#1836 decision behavior without reintroducing #1834.

### Test

`test/min_time_threads_gtest.cc` runs the same benchmark at 1 and 4 threads with `UseRealTime()` and asserts that the per-thread wall-clock time does not collapse as the thread count grows. It fails before this change (4-thread ≈ ¼ of 1-thread) and passes after.


LebedevRI:
Duplicate of https://github.com/google/benchmark/pull/2118.

devtejasx:
Withdrawing this PR.

PR #2257, opened 2026-07-18
Exclude setup/teardown from the MemoryManager measurement window (#2149)

devtejasx:
Fixes #2149.

### Problem

When a `MemoryManager` is registered, `RunMemoryManager` called `memory_manager->Start()` *before* allocating the internal `ThreadManager` and running the user's `Setup()`, and called `Stop()` *after* `Teardown()` and the `ThreadManager` destruction. As a result the library's own bookkeeping allocation and any per-benchmark fixture / `Setup()` / `Teardown()` allocations were attributed to the benchmark's reported memory usage, inflating `num_allocs` etc. for allocations that are not part of the measured work.

### Fix

Move `memory_manager->Start()` to just before `RunInThread` (after the `ThreadManager` is constructed and `Setup()` has run) and `Stop()` to just after `RunInThread` (before `Teardown()` and manager teardown), so the measurement window covers only the benchmark iterations — matching how the timed window already excludes setup/teardown.

### Test

`test/memory_manager_ordering_gtest.cc` registers a `MemoryManager` and a benchmark whose `Setup`/`Teardown` allocate, and asserts that those allocations fall outside the measured window.

(This supersedes #2247, which I closed to convert its ad-hoc test into a gtest.)


devtejasx:
Withdrawing this PR.

PR #2262, opened 2026-07-22
Fix typos, add codespell configuration

mwtoews:
This fixes a few typos identified and corrected using codespell. A configuration is also added, in case this tool is used again in the future. This check can be added to CI and/or pre-commit, but this is not done in this PR.

Note that codespell has previously been used with fe65457e80ef8ab6bbd6da418ab7cb6f6102afda

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,9 @@ Documentation = "https://github.com/google/benchmark/tree/main/docs"
 Repository = "https://github.com/google/benchmark.git"
 Discord = "https://discord.gg/cz7UX7wKC2"
 
+[tool.codespell]
+ignore-words-list = ["bion", "globa", "wron"]
+
 [tool.setuptools]
 package-dir = { "" = "bindings/python" }
 zip-safe = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,9 +34,6 @@ Documentation = "https://github.com/google/benchmark/tree/main/docs"
 Repository = "https://github.com/google/benchmark.git"
 Discord = "https://discord.gg/cz7UX7wKC2"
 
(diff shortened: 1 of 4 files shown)


test/output_test_helper.cc
```diff
@@ -30,7 +30,7 @@ using TestCaseList = std::vector<TestCase>;
 using SubMap = std::vector<std::pair<std::string, std::string>>;
 
 TestCaseList& GetTestCaseList(TestCaseID ID) {
-  // Uses function-local statics to ensure initialization occurs
+  // Uses function-local statistics to ensure initialization occurs
```
LebedevRI: Uh?

mwtoews: Ah, false positive. Perhaps "static structures" instead?

LebedevRI: There is no such thing as 'static structures', as far as i understand.
`static` refers to the lifetime of a variable.

mwtoews: So resolution should just use non-plural "Uses function-local static to ensure ..."

pyproject.toml
```diff
@@ -34,6 +34,9 @@ Documentation = "https://github.com/google/benchmark/tree/main/docs"
 Repository = "https://github.com/google/benchmark.git"
 Discord = "https://discord.gg/cz7UX7wKC2"
 
+[tool.codespell]
+ignore-words-list = ["bion", "globa", "wron"]
+
```
LebedevRI: I think this is not an appropriate place for this.
This is not a python project.
If you'd drop this, i'd merge.

LebedevRI (approved):

LebedevRI:
@mwtoews thank you!

PR #2265, opened 2026-07-23
Fix use-after-free in CreateDefaultDisplayReporter (#2240)

### LINKED ISSUE #2240: [BUG]Heap-use-after-free in benchmark `RunSpecifiedBenchmarks

devtejasx:
Fixes #2240.

`CreateDefaultDisplayReporter()` transfers ownership of the returned reporter to
the caller, but cached that owning pointer in a function-local `static`. After
the first `RunSpecifiedBenchmarks()` call destroyed the reporter, a subsequent
call handed back the same, now-dangling pointer, causing a heap-use-after-free
in `BenchmarkReporter::GetOutputStream()`.

Per @dmah42's review, the ownership transfer is now expressed in the type: the
function returns `std::unique_ptr<BenchmarkReporter>` instead of a raw pointer,
and the `static` is gone, so every call returns a fresh reporter. With that
signature the old bug is unrepresentable, so no regression test is included.

### API change

This changes the signature of an exported function:

```diff
-BENCHMARK_EXPORT BenchmarkReporter* CreateDefaultDisplayReporter();
+BENCHMARK_EXPORT std::unique_ptr<BenchmarkReporter>
+CreateDefaultDisplayReporter();
```

Callers that stored the result in a raw pointer need `.get()`, and callers that
already wrapped it in a `unique_ptr` (which is what the old contract required)
simplify. The only in-tree caller is `RunSpecifiedBenchmarks()`. This is a
source-level break for external users, so it may be worth calling out in the
next release notes.

Release and Debug builds pass; test suite unchanged from `main`.

---

responsibility for it.

### DIFF
--- a/src/benchmark.cc
+++ b/src/benchmark.cc
@@ -585,12 +585,9 @@ ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
 
 }  // end namespace internal
 
-BenchmarkReporter* CreateDefaultDisplayReporter() {
-  static auto* default_display_reporter =
-      internal::CreateReporter(FLAGS_benchmark_format,
-                               internal::GetOutputOptions())
-          .release();
-  return default_display_reporter;
+std::unique_ptr<BenchmarkReporter> CreateDefaultDisplayReporter() {
+  return internal::CreateReporter(FLAGS_benchmark_format,
+                                  internal::GetOutputOptions());
 }
 
 size_t RunSpecifiedBenchmarks() {
@@ -629,7 +626,7 @@ size_t RunSpecifiedBenchmarks(BenchmarkReporter* display_reporter,
   std::unique_ptr<BenchmarkReporter> default_display_reporter;
   std::unique_ptr<BenchmarkReporter> default_file_reporter;
   if (display_reporter == nullptr) {
-    default_display_reporter.reset(CreateDefaultDisplayReporter());
+    default_display_reporter = CreateDefaultDisplayReporter();
     display_reporter = default_display_reporter.get();
   }
   auto& Out = display_reporter->GetOutputStream();
--- a/include/benchmark/benchmark_api.h
+++ b/include/benchmark/benchmark_api.h
@@ -54,7 +54,8 @@ BENCHMARK_EXPORT void SetBenchmarkFilter(std::string value);
 
 BENCHMARK_EXPORT int32_t GetBenchmarkVerbosity();
 
-BENCHMARK_EXPORT BenchmarkReporter* CreateDefaultDisplayReporter();
+BENCHMARK_EXPORT std::unique_ptr<BenchmarkReporter>
+CreateDefaultDisplayReporter();
 
 BENCHMARK_EXPORT size_t RunSpecifiedBenchmarks();
 BENCHMARK_EXPORT size_t RunSpecifiedBenchmarks(std::string spec);


src/benchmark.cc
```diff
@@ -586,11 +586,14 @@ ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
 }  // end namespace internal
 
 BenchmarkReporter* CreateDefaultDisplayReporter() {
-  static auto* default_display_reporter =
-      internal::CreateReporter(FLAGS_benchmark_format,
-                               internal::GetOutputOptions())
-          .release();
-  return default_display_reporter;
+  // Ownership of the returned reporter is transferred to the caller (note the
```
dmah42: if ownership is being transferred, it should be returning a unique_ptr<>.

devtejasx:
Good call — done in 283ee51. `CreateDefaultDisplayReporter()` now returns `std::unique_ptr<BenchmarkReporter>`, so the ownership transfer is in the type rather than implied by a `.release()` at the call site, and the `static` is gone.

```diff
-BENCHMARK_EXPORT BenchmarkReporter* CreateDefaultDisplayReporter();
+BENCHMARK_EXPORT std::unique_ptr<BenchmarkReporter>
+CreateDefaultDisplayReporter();
```

`RunSpecifiedBenchmarks()` was the only in-tree caller and just assigns now instead of `.reset(...)`. It is a source-level break for external users, so I've called that out in the PR body in case it's worth a release note.

Release and Debug both build clean and the test suite is unchanged from `main`. Rebased onto current `main`.

src/benchmark.cc
```diff
@@ -585,12 +585,13 @@ ConsoleReporter::OutputOptions GetOutputOptions(bool force_no_color) {
...
 }  // end namespace internal
 
-BenchmarkReporter* CreateDefaultDisplayReporter() {
-  static auto* default_display_reporter =
-      internal::CreateReporter(FLAGS_benchmark_format,
-                               internal::GetOutputOptions())
-          .release();
-  return default_display_reporter;
+std::unique_ptr<BenchmarkReporter> CreateDefaultDisplayReporter() {
+  // Ownership of the returned reporter is transferred to the caller, so a fresh
```
dmah42: this comment is useless at this point.

test/create_default_display_reporter_gtest.cc
```diff
@@ -0,0 +1,26 @@
...
+
+namespace benchmark {
+namespace {
+
+// Regression test for a heap-use-after-free (issue #2240):
+// CreateDefaultDisplayReporter() transfers ownership of the returned reporter
+// to the caller, so it must return a fresh object on every call. A previous
+// implementation cached the pointer in a function-local `static`, so the second
+// call returned an already-deleted reporter -> use-after-free.
+TEST(CreateDefaultDisplayReporterTest, ReturnsFreshOwnedReporterEachCall) {
```
dmah42: this is testing nothing now.

src/benchmark.cc
devtejasx: Agreed, dropped it in 480bb68 — the return type says it now.

test/create_default_display_reporter_gtest.cc
devtejasx: You're right — with the `unique_ptr` return the old bug is unrepresentable, so the test asserts something the type system already guarantees. Deleted it and its `add_gtest` entry in 480bb68. The PR is now just the signature change and the `static` removal.

dmah42 (approved):

dmah42:
thank you.

devtejasx:
Your welocme

PR #2268, opened 2026-07-25
Fix compile error with clang -Werror -Wunused-template

aokblast:
This sysctl override is unused now so it causes compile error with latest LLVM.

### DIFF
--- a/src/sysinfo.cc
+++ b/src/sysinfo.cc
@@ -149,7 +149,7 @@ struct ValueUnion {
   }
 
   template <class T, int N>
-  std::array<T, N> GetAsArray() {
+  BENCHMARK_MAYBE_UNUSED std::array<T, N> GetAsArray() {
     const int arr_size = sizeof(T) * N;
     BM_CHECK_LE(arr_size, size);
     std::array<T, N> arr;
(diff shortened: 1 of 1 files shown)


LebedevRI:
Are these ever used at all? I may be wrong, but i'm not seeing it. In which case, let's just drop them altogether?

aokblast:
I don't suffer from any compile fail in here. Remove this now.

LebedevRI:
> I don't suffer from any compile fail in here. Remove this now.

I suppose the problem may be that those functions are used in some other codepath that isn't being compiled on your system/isn't covered by CI, in which case they'd need to be guarded with an appropriate preprocessor magic.
And indeed, looks like they are used on macos.

aokblast:
Oh. I get it. I test it on FreeBSD, which doesn't pass through this pah. Do you think my previous patch is good?

LebedevRI:
I'd suggest wrapping them with `#ifdef BENCHMARK_OS_MACOSX`, but maybe `BENCHMARK_MAYBE_UNUSED` is good-enough.

LebedevRI (approved):

LebedevRI:
@aokblast thank you!

aokblast:
Thanks for your review!

Issue #2235, opened 2026-06-26
[BUG] Benchmark registration order affects reported outcomes


ccomendant:
**Describe the bug**
Hello,
I have the following bechmark:

```c++
#include <benchmark/benchmark.h>
#include <gtest/gtest.h>

/**
 * defines / pre-allocates a std::vector<std::wstring> examples (within g_dataset)
 */
#include "Helpers.h"

#include <chrono>
#include <cstdint>
#include <ranges>
#include <vector>
#include <execution>
#include <codecvt>

std::string DoWork(const std::wstring& input)
{
    std::string result(input.size(), '\0');
    std::transform(std::execution::par_unseq, std::begin(input), std::end(input), std::begin(result),
                   [](wchar_t inChar) { return (char)(inChar & 0x0ff); });
    return result;
}

std::wstring DoWork1(const std::string& input)
{
    std::wstring result(input.size(), L'\0');
    std::transform(std::execution::par_unseq, std::begin(input), std::end(input), std::begin(result),
                   [](wchar_t inChar) { return (wchar_t)(inChar & 0x0ff); });
    return result;
}

template <typename InStringT,
    typename Fn,
    typename OutStringT = std::invoke_result_t<Fn,InStringT>>
void RunOnExamples(const std::vector<InStringT>& examples, Fn&& fn,
                  std::vector<OutStringT>& results)
{
    for (size_t iterations = 0; iterations < examples.size(); iterations++) {
        results[iterations] = fn(examples[iterations]);
    }
}

static void BM_DoWork(benchmark::State& state)
{
    state.PauseTiming();

    auto expectedResults = std::vector<std::string>(g_dataset.examples.size());

    state.ResumeTiming();

    while (state.KeepRunningBatch(g_dataset.examples.size())) {
        RunOnExamples(g_dataset.examples, DoWork, expectedResults);
    }

    state.SetItemsProcessed(state.iterations() * g_dataset.examples.size());
}

static void BM_DoWork1(benchmark::State& state)
{
    state.PauseTiming();

    auto expectedResults = std::vector<std::wstring>(g_dataset.examples.size());

    std::vector<std::string> examples(g_dataset.examples.size());
    static std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> myconv;
    for (size_t i = 0; i < g_dataset.examples.size(); i++) {
        examples[i] = myconv.to_bytes(g_dataset.examples[i]);
    }

    state.ResumeTiming();

[...]

PR #2274, opened 2026-08-01
Diagnose invalid PauseTiming()/ResumeTiming() calls (#2235)

devtejasx:
PauseTiming() and ResumeTiming() guarded their preconditions with BM_CHECK, which compiles away under NDEBUG. In a release build, calling PauseTiming() outside the benchmark loop -- e.g. before it, to try to exclude setup work -- therefore reached ThreadTimer::StopTimer() with a timer that was never started, and accumulated `ChronoClockNow() - 0`, an absolute clock reading, into the run's real time. The same happens to the CPU time. Every benchmark in the process then reports a huge constant offset plus whatever the clocks had advanced by, which is why the results in #2235 grow monotonically with registration order and are unrelated to the work being measured.

Turn the preconditions into a reported benchmark error: an invalid call now leaves the timer untouched and skips the run with a message naming the misuse, in every build configuration. The user sees what went wrong instead of plausible-looking numbers, and a run that would have been garbage is no longer reported as a result.

Rewrite diagnostics_test to cover the new contract -- pause/resume before the loop, pause after it, and pausing or resuming twice -- in both debug and release builds; previously it could only run in debug, since the behaviour it checked existed only there.

---

responsibility for it.


dmah42:
what impact does this have on the generated assembly for the core timing loop? did you consider why we keep these methods light weight or did you just point some AI crap at the project and suggest changes without a deeper understanding of the library?

devtejasx:
Fair question. I've reworked the patch around it and measured.

**The core timing loop is not touched.** `KeepRunning()`, `KeepRunningBatch()`, `KeepRunningInternal()` and `StateIterator` all live in `state.h`, which this PR does not modify. Compiling the same benchmark translation unit against the library before and after, gcc 15 at `-O2`, `objdump -d` of the benchmark function is byte-for-byte identical.

**`PauseTiming()`/`ResumeTiming()` themselves.** The condition now keys off the timer rather than the loop flags: the timer runs exactly between the first and the last iteration and is stopped while timing is paused, so `timer_->running()` on its own answers "are we inside the loop with the timer going". What that adds to the fast path is

```
mov  0x98(%rcx),%rbx     ; timer_
cmpb $0x0,0x1(%rbx)      ; timer_->running_
je   <cold path>
```

one load and a predicted-not-taken branch, in a function that already reads two clocks (`ChronoClockNow()` and `ReadCpuTimerOfChoice()`). The message is handed to an out-of-line helper so the `std::string` temporary that `SkipWithError()` takes is built on the cold path rather than here.

AI assistance is disclosed in the description, per AGENTS.md. On the substance: `BM_CHECK` compiles out under `NDEBUG`, so in a release build `PauseTiming()` before the loop reaches `ThreadTimer::StopTimer()` with `start_real_time_ == 0` and accumulates `ChronoClockNow() - 0` — an absolute clock reading rather than a duration — into the run's real time, and likewise for CPU time. That is what produces the fixed ~86 s offset in #2235 and why every subsequent benchmark in the process reports a larger number regardless of registration order.

If the position is that misuse of these two functions is a debug-only concern, that is a reasonable call and I'll close this. The reason I thought it was worth raising is that the release-build symptom is plausible-looking wrong numbers rather than a crash.

src/benchmark.cc
```diff
@@ -260,9 +260,29 @@ State::State(std::string name, IterationCount max_iters,
...
+}
+
+}  // namespace
+
 void State::PauseTiming() {
+  // The timer runs exactly between the first and the last iteration, and is
+  // stopped while timing is paused. Stopping it at any other point would fold
+  // an absolute clock reading into the accumulated time instead of a
+  // duration, so report the misuse rather than produce a meaningless result.
+  if (BENCHMARK_BUILTIN_EXPECT(!timer_->running(), false)) {
```
dmah42: why is this check `!_timer_->running()` rather than the original `started_ && !finished_ && !skipped()`?

src/benchmark.cc
```diff
@@ -280,7 +300,17 @@ void State::PauseTiming() {
 }
 
 void State::ResumeTiming() {
-  BM_CHECK(started_ && !finished_ && !skipped());
+  // Restarting a timer that is already running discards the slice it is in the
+  // middle of, and starting one outside of the loop leaves it running past the
+  // point where its value is read.
+  if (BENCHMARK_BUILTIN_EXPECT(timer_->running() || !started_ || finished_,
```
dmah42: same here. this has changed the check

src/benchmark.cc
```diff
@@ -260,9 +260,29 @@ State::State(std::string name, IterationCount max_iters,
 #endif
 }
 
+namespace {
+
+// Kept out of line so that the timer entry points below stay small: the string
+// temporary that SkipWithError() takes is built here, off their fast path.
```
dmah42: this comment makes no sense. no temporary is "built" as the strings are literals.

test/diagnostics_test.cc
```diff
@@ -1,107 +1,94 @@
...
-  try {
-    state.PauseTiming();
-    std::abort();
-  } catch (std::logic_error const&) {
-  }
-  try {
-    state.ResumeTiming();
-    std::abort();
-  } catch (std::logic_error const&) {
+// Set by each benchmark once it has observed its own misuse being diagnosed,
```
dmah42: having these be global increases the risk of order dependent testing if tests accidentally share the variables. is there any reason why they're global? presumably they can be local to the benchmark and asserted on after the for loop in each case.

test/diagnostics_test.cc
```diff
@@ -1,107 +1,94 @@
...
+// so that main() can tell a diagnosed misuse from a benchmark that never ran.
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
```
dmah42: this is a big red flag right here.

dmah42 (changes requested):

LebedevRI (changes requested):
I don't like this.
Can we instead move `State::PauseTiming()` into a header, use normal `assert()`,
and then call the original `State::PauseTiming()` renamed as `State::PauseTimingImpl()` and kept in a non-header?

devtejasx:
Done — rebuilt the way you suggested, @LebedevRI.

`PauseTiming()` and `ResumeTiming()` are now small inline wrappers in `state.h` that `assert()` and forward to `PauseTimingImpl()`/`ResumeTimingImpl()`, which keep the existing bodies in `benchmark.cc`.

@dmah42 — that also answers both of your questions on the check. The predicate is back to the original `started_ && !finished_ && !skipped()`; the timer-based form is gone, along with the comment about the temporary. What the change amounts to now is *where* the assertion is evaluated: it was compiled into the library, so whether it fires depends on how the library was built, and distributions ship a release build — the check is then absent no matter how the benchmark itself was compiled. As an inline wrapper it follows the `NDEBUG` of whoever writes the benchmark, and costs nothing once they define it.

One consequence worth flagging: `diagnostics_test` worked by installing the library's abort handler and catching what `BM_CHECK` threw. A plain `assert()` does not go through that handler, so the test cannot work as written. It is now `diagnostics_gtest` using `ASSERT_DEATH_IF_SUPPORTED`, the same pattern as `min_time_parse_gtest` and `profiler_manager_gtest` — which also removes the globals you flagged, since each case is a separate benchmark and the death test runs in its own process. There is one non-death case asserting that the documented pause/resume-inside-the-loop usage still runs.

Release and debug suites pass.

test/diagnostics_gtest.cc
```diff
@@ -0,0 +1,68 @@
...
+  ASSERT_DEATH_IF_SUPPORTED(
+      benchmark::RunSpecifiedBenchmarks("BM_pause_after_loop"), "PauseTiming");
+#endif
+}
+
+// The assertions must not fire on the documented usage.
+TEST(Diagnostics, PauseAndResumeInsideLoop) {
+  EXPECT_EQ(benchmark::RunSpecifiedBenchmarks("BM_valid"), 1u);
+}
+
```
LebedevRI, line 99: what about pause of paused, resume of resumed, pause of skipped, resume of skipped, etc?

LebedevRI: And note that skipping can happen before/during/after the loop, and time may be either running or paused, so some more variations to test.
Oh, and we should actually test that the timings reported aren't absurd in the cases that aren't diagnosed.

devtejasx:
Covered, @LebedevRI — and working through the list changed the patch, not just the test.

**pause of paused / resume of resumed.** `started_ && !finished_ && !skipped()` does not catch either, and both corrupt the accumulated time exactly the way #2235 does. The assertion now includes the timer state via a `TimerIsRunning()` accessor that is only ever named from inside the assertions, so it is never called once `NDEBUG` is defined. (`ThreadTimer` is not visible in the public header, hence the accessor rather than touching `timer_` directly.)

**pause of skipped / resume of skipped.** These must *not* be diagnosed, and the test proves it: with an assertion there, `BM_skipped_while_paused` — which is just

```c++
for (auto _ : state) {
  benchmark::ScopedPauseTiming pause(state);
  state.SkipWithMessage("...");
}
```

— aborts on the `ScopedPauseTiming` destructor. `SkipWith*()` has already stopped the timer and the run is being abandoned, so both calls now return early when the state is skipped. I checked that by removing the guard and re-running:

```
Assertion failed: started_ && !finished_ && !TimerIsRunning() && "ResumeTiming() called ..."
```

**The variations.** `diagnostics_gtest` now covers pause and resume before the loop, pause and resume after it, pause of paused, resume of resumed, and skipping before the loop, during it with the timer running, during it with timing paused, and after it — eleven benchmarks, each `Iterations(1)`, whole file runs in about a second.

**Timings that are not diagnosed.** The last test captures the report through a reporter and asserts the real and CPU time of the undiagnosed run are non-negative and under a second; the bug this PR is about produces values in the tens of billions of nanoseconds, so the bound is loose on purpose. One limitation worth stating: the tests are compiled with `-UNDEBUG`, so this cannot exercise what an `NDEBUG` build does with the same misuse — there, the calls are simply unchecked, as they are on main today.

LebedevRI (commented):
So what currently happens when pausing a skipped benchmark?

LebedevRI:
Am i talking to an LLM or did human write last comment?

devtejasx:
I wrote that comment myself. I do use AI as a drafting tool (and disclosed that in the previous PR), but I review, test, and understand every change before submitting it. The previous reply was too long and probably read like an LLM-generated summary rather than a code review discussion—that's on me. I'll keep the responses focused on the specific technical questions from here.

include/benchmark/state.h
```diff
@@ -52,9 +52,26 @@ class BENCHMARK_EXPORT BENCHMARK_INTERNAL_CACHELINE_ALIGNED State {
...
 
-  void PauseTiming();
+  // Pause the timer. Only valid while the benchmark loop is running and
+  // timing is not already paused. Does nothing once the benchmark has been
+  // skipped, so that a pause left open by SkipWith*() unwinds cleanly.
+  void PauseTiming() {
+    if (skipped()) return;
+    assert(started_ && !finished_ && TimerIsRunning() &&
+           "PauseTiming() called outside of the benchmark loop, or while "
+           "timing was already paused");
```
LebedevRI: Does `TimerIsRunning()` return `true` when outside of the measuring loop?

devtejasx:
Thanks, I went through those cases and updated both the implementation and the tests.

I found that calling PauseTiming or ResumeTiming more than once also gives incorrect timing results, so the tests now check for those cases too.

For skipped benchmarks, PauseTiming and ResumeTiming should not fail. If a benchmark is skipped while timing is paused, ScopedPauseTiming will later call ResumeTiming during cleanup. Since the benchmark has already been skipped and timing has already stopped, that call is simply ignored.

I also added tests for

Calling PauseTiming and ResumeTiming before the loop 
Calling PauseTiming and ResumeTiming  after the loop
Calling PauseTiming twice
Calling ResumeTiming twice
Skipping a benchmark before, during, and after the loop, including while timing is paused

I also added a test for the normal case to make sure valid benchmarks still report reasonable timing values.

This description focuses on the expected behavior instead of the internal implementation details.

include/benchmark/state.h
```diff
@@ -52,9 +52,26 @@ class BENCHMARK_EXPORT BENCHMARK_INTERNAL_CACHELINE_ALIGNED State {
 
   inline bool KeepRunningBatch(IterationCount n);
 
-  void PauseTiming();
+  // Pause the timer. Only valid while the benchmark loop is running and
+  // timing is not already paused. Does nothing once the benchmark has been
+  // skipped, so that a pause left open by SkipWith*() unwinds cleanly.
+  void PauseTiming() {
+    if (skipped()) return;
```
LebedevRI: I'm confused, why are we relaxing the check, and now always adding control flow to every callsite that wasn't there before?

devtejasx:
The change is meant to keep scopedpauseTiming wroking correctly .when skipwith is called,the benchmark has already stopped and been marked as skipped.if a benchmark::scopepausetiming object is still active,its destructor will automatically call resumetiming.making pausetiming and resumetiming do nothing after a skip allow that automatic cleanup to finish safely instead of triggering an assertion .This behavior only applies after a benchmark has been skipped .neomal benchmark behaviour and checks for incorrect API usage stay the same

LebedevRI:
But you see the problem, right? The PR is intended to simply move the existing assertion so that it actually triggers not just when the whole library is built with assertions.
And now it also contains subtle behavioural changes.

devtejasx:
Yes, I see the concern. The original goal was only to make the existing misuse checks apply based on the benchmark user's build configuration instead of the library build configuration.
The extra behavior changes came from trying to handle the edge cases uncovered while adding tests. I agree that they make the patch harder to review because they mix two different changes.these should be reply 
I will split this back to the original intent: move the assertions into the caller-visible header path without changing the runtime behavior. If we want to change the skipped benchmark behavior or other edge cases, that can be discussed separately.

devtejasx:
Reverted. The patch is back to only moving the assertion.

```c++
void PauseTiming() {
  assert(started_ && !finished_ && !skipped() &&
         "PauseTiming() called outside of the benchmark loop");
  PauseTimingImpl();
}
```

Same condition as the BM_CHECK it replaces. No `skipped()` early return, no `TimerIsRunning()`, no added control flow at any call site. Under NDEBUG the wrapper is empty and the call goes straight to `PauseTimingImpl()`, exactly as now.

To answer the two questions in case they matter later:

`TimerIsRunning()` was `timer_->running()`. It is false outside the loop and also false while timing is paused, which is how it caught pause-of-paused.

The `skipped()` early return was for `ScopedPauseTiming`. Its destructor calls `ResumeTiming()`, so a benchmark that calls `SkipWithError()` inside a `ScopedPauseTiming` scope trips the assert on the way out. That is pre-existing behaviour, not something this PR should change.

Both are behaviour changes, so they are out. I can file the double-pause case as its own issue if it is worth fixing.

Tests are back to the four out-of-loop cases, plus one check that a run which is not diagnosed reports a real and CPU time below a second. Release and debug both pass.

devtejasx:
@LebedevRI @dmah42 The changes-requested review here predates the revert. The patch is back to only moving the assertion to the caller — no `skipped()` early return, no `TimerIsRunning()`, no added control flow at any call site, so none of the behavioural changes you objected to are still in it.

Could one of you take another look when you have time?

LebedevRI:
Err, gtest, wth???
```
87: D:\a\benchmark\benchmark\test\diagnostics_gtest.cc(73): error: Death test: RunOne("BM_pause_before_loop")
87:     Result: failed to die.
87:  Error msg:
87: [  DEATH   ] 
87: 
87: [  FAILED  ] Diagnostics.PauseOutsideOfTheLoopIsDiagnosed (9 ms)
87: [ RUN      ] Diagnostics.ResumeOutsideOfTheLoopIsDiagnosed
87: Running main() from gmock_main.cc
87: D:\a\benchmark\benchmark\test\diagnostics_gtest.cc(80): error: Death test: RunOne("BM_resume_before_loop")
87:     Result: failed to die.
87:  Error msg:
87: [  DEATH   ] 
87: 
87: [  FAILED  ] Diagnostics.ResumeOutsideOfTheLoopIsDiagnosed (8 ms)
```

LebedevRI:
So why would `assert()` do nothing in those two particular configurations (Release w/ Shared Libs),
but work fine in other instances (Release w/ Static libs, all Debug variants)? VS is weird.

devtejasx:
@LebedevRI It's the export on `State`. In a shared build the test sees `class __declspec(dllimport) State`, and MSVC won't expand `PauseTiming()` at the call site there. It calls the copy exported from benchmark.dll instead, which was built with the library's NDEBUG, so the assert is gone. Static builds inline it, and a Debug DLL still has its asserts, so only Release + shared fails.

On MSVC 19.latest at /O2 /MD the call compiles to `rex_jmp __imp_?PauseTiming@State@@QEAAXXZ`. The same wrapper without the assert does get inlined, so it's specifically the body with `assert()` that it declines.

Marking both wrappers `BENCHMARK_ALWAYS_INLINE`, like `begin()`/`end()`, fixes it: `_wassert` ends up in the caller at /O2 and /O2 /Ob1, and with NDEBUG only the tail call to `PauseTimingImpl()` is left. Pushed that, plus a merge of main.

test/diagnostics_gtest.cc
dmah42, line 1: this is a _weird_ name for this test. shouldn't these tests be part of an existing googletest instead?

dmah42 (changes requested):

LebedevRI:
> @LebedevRI It's the export on `State`. In a shared build the test sees `class __declspec(dllimport) State`, and MSVC won't expand `PauseTiming()` at the call site there. It calls the copy exported from benchmark.dll instead, which was built with the library's NDEBUG, so the assert is gone. Static builds inline it, and a Debug DLL still has its asserts, so only Release + shared fails.

OMGWTAF...

> On MSVC 19.latest at /O2 /MD the call compiles to `rex_jmp __imp_?PauseTiming@State@@QEAAXXZ`. The same wrapper without the assert does get inlined, so it's specifically the body with `assert()` that it declines.
> 
> Marking both wrappers `BENCHMARK_ALWAYS_INLINE`, like `begin()`/`end()`, fixes it: `_wassert` ends up in the caller at /O2 and /O2 /Ob1, and with NDEBUG only the tail call to `PauseTimingImpl()` is left. Pushed that, plus a merge of main.

LebedevRI (approved):
LG, thank you!

(@dmah42 up to you)

dmah42:
it's a nit, but i really don't like the addition of a new test called "diagnostic".

LebedevRI (changes requested):
/\

devtejasx:
@dmah42 Moved the tests into the existing `benchmark_gtest.cc` (suites `TimingDeathTest` / `TimingTest`) and deleted `diagnostics_gtest.cc`. Nothing else changed apart from a merge of main.

@LebedevRI that's the change your last review pointed at.

test/benchmark_gtest.cc
```diff
@@ -164,6 +164,92 @@ TEST(AddCustomContext, DuplicateKey) {
   global_context = nullptr;
 }
 
+// PauseTiming() and ResumeTiming() assert when called outside the benchmark
+// loop (#2235).
```
dmah42: drop the PR reference. it's not useful information.

docs/user_guide.md
```diff
@@ -1347,6 +1347,12 @@ BENCHMARK(BM_SetInsert_With_Timer_Control)->Ranges({{1<<10, 8<<10}, {128, 512}})
 ```
 <!-- {% endraw %} -->
 
+Both calls are only valid inside the benchmark loop. Calling them outside it
+stops a timer that was never started, which adds an absolute clock reading to
+the reported time instead of a duration. This is asserted, and the assertion is
+compiled into the benchmark rather than into the library, so it fires whenever
+the benchmark itself is built without `NDEBUG`.
```
LebedevRI: I'm not sure if i've commented on this before, but the docs are user-facing,
all of this reads slightly off. Here's how i'd word it:
```suggestion
Note that you can only pause the timing if the timer is actually running,
and likewise, you can only resume the timing only if the timer is paused.
Such invalid pause/resume calls are diagnosed if the
user code is built without `NDEBUG`.
```

LebedevRI (approved):

LebedevRI (approved):

dmah42 (approved):

Issue #2263, opened 2026-07-23
[BUG] Fail to build with `-Werror` on macOS.


divVerent:
**Describe the bug**
Compile warnings on macOS. Breaks build of BoringSSL because that one uses `-Werror` in its own `cxxopts`.

**System**
Which OS, compiler, and compiler version are you using:
  - OS: macOS 26.5.2
  - Compiler and version: Apple clang version 21.0.0 (clang-2100.1.1.101); Target: arm64-apple-darwin25.5.0; Thread model: posix

**To reproduce**
1. sync to commit `8b66b54f7e1bf6b25390dca1dea3f18a40e607f9`
2. `bazel build --cxxopt=-Werror :all` (or even just `bazel build :all`)
3. See error (or multiple screen pages of warning spam)

**Actual behavior**
```
INFO: Analyzed 3 targets (0 packages loaded, 627 targets configured).
ERROR: /private/tmp/benchmark/BUILD.bazel:41:11: Compiling src/benchmark_runner.cc failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target //:benchmark) external/rules_cc++cc_configure_extension+local_config_cc/cc_wrapper.sh -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object ... (remaining 43 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from src/benchmark_runner.cc:15:
In file included from src/benchmark_runner.h:23:
In file included from src/perf_counters.h:30:
src/mutex.h:79:40: error: mutex 'mut_' is still held at the end of function [-Werror,-Wthread-safety-analysis]
   79 |   void lock() ACQUIRE() { mut_.lock(); }
      |                                        ^
src/mutex.h:79:32: note: mutex acquired here
   79 |   void lock() ACQUIRE() { mut_.lock(); }
      |                                ^
src/mutex.h:80:34: error: releasing mutex 'mut_' that was not held [-Werror,-Wthread-safety-analysis]
   80 |   void unlock() RELEASE() { mut_.unlock(); }
      |                                  ^
2 errors generated.
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.202s, Critical Path: 1.00s
INFO: 14 processes: 1 action cache hit, 9 internal, 5 darwin-sandbox.
[...]

PR #2275, opened 2026-08-01
Detect thread safety attribute support in mutex.h (#2263)

devtejasx:
The thread safety annotations were only enabled when the surrounding build system defined HAVE_THREAD_SAFETY_ATTRIBUTES, but nothing in the tree ever defines it. Build systems that turn on -Wthread-safety by default -- Bazel does on macOS -- therefore analyzed Mutex/MutexLock with every annotation erased. Where the standard library annotates std::mutex itself (libc++), that makes the analysis conclude that Mutex::lock() leaks the lock and that Mutex::unlock() releases a lock that was never held, which is a hard error under -Werror:

  src/mutex.h:79:40: error: mutex 'mut_' is still held at the end of function
  src/mutex.h:80:34: error: releasing mutex 'mut_' that was not held

Enable the annotations whenever the compiler actually supports them, probed with __has_attribute(capability), so the wrappers are analyzed with the contract they are meant to have. An explicit HAVE_THREAD_SAFETY_ATTRIBUTES from the build system still takes precedence, and
BENCHMARK_DISABLE_THREAD_SAFETY_ATTRIBUTES forces the previous no-op behaviour back.

Verified with clang 21 + libc++ (reproduces before, clean after), clang 21 + libstdc++, gcc 15 and MSVC; the full library builds clean under -Wall -Wextra -Wthread-safety -Werror with clang.

---

responsibility for it.


src/mutex.h
```diff
@@ -8,6 +8,22 @@
 
 // Enable thread safety attributes only with clang.
 // The attributes can be safely erased when compiling with other compilers.
+//
+// Detect support instead of relying on the build system to opt in: build
```
dmah42: but nothing has been changed in the build system so this is not "instead of" is it?

devtejasx:
Correct — nothing in the build system changes, and the wording was wrong. Reworded to state the situation plainly rather than describe it as a replacement:

`HAVE_THREAD_SAFETY_ATTRIBUTES` is left to whoever compiles the library, and nothing in this tree defines it. That is fine as long as `-Wthread-safety` is off. Bazel turns it on by default, and there the wrappers get analysed with their own annotations erased: `Mutex::lock()` is then seen acquiring `mut_` and never releasing it, which is an error wherever the standard library annotates `std::mutex` itself (libc++, hence macOS in #2263).

src/mutex.h
```diff
@@ -8,6 +8,24 @@
 
 // Enable thread safety attributes only with clang.
 // The attributes can be safely erased when compiling with other compilers.
+//
+// HAVE_THREAD_SAFETY_ATTRIBUTES is left to whoever compiles the library, and
+// no build in this tree defines it. That is fine as long as -Wthread-safety is
+// off, but Bazel turns it on by default, and there the wrappers below get
```
dmah42: i'm a bit confused. if Bazel turns it on, then the macro HAVE_THREAD_SAFETY_ATTRIBUTES should be enabled. or is this trying to suggest that Bazel turns it on but doesn't enable the macro?

either way, i think this whole comment should be focused on why we need to do this check here (and not, say, in the build system: if we know Bazel enables it then why not set the cc_defines in the build script?)

src/mutex.h
```diff
@@ -8,6 +8,24 @@
...
+// standard library annotates std::mutex itself (libc++). Probe for the
+// attributes so this cannot happen. An explicit HAVE_THREAD_SAFETY_ATTRIBUTES
+// still wins, and BENCHMARK_DISABLE_THREAD_SAFETY_ATTRIBUTES forces the
+// attributes off.
+#if !defined(HAVE_THREAD_SAFETY_ATTRIBUTES) &&              \
+    !defined(BENCHMARK_DISABLE_THREAD_SAFETY_ATTRIBUTES) && \
+    defined(__has_attribute)
+#if __has_attribute(capability)
+#define HAVE_THREAD_SAFETY_ATTRIBUTES
+#endif
```
LebedevRI: (I think something like this overall diff should be done.)

Is there any compiler other than clang that supports TSA anyway?
Can we just check for clang, and perhaps do a version check?

LebedevRI: (to be more specific, i strongly suspect that `__has_attribute(capability)` is bogus/unsufficient,
that only covers `CAPABILITY()`, what about others? Can we just assume they are implied by it?)

devtejasx:
Both addressed.

@LebedevRI — dropped `__has_attribute(capability)`; you're right that probing one attribute says nothing about the other twenty, and clang is the only compiler that implements any of them. It is now just:

```c++
#if !defined(HAVE_THREAD_SAFETY_ATTRIBUTES) && defined(__clang__)
#define HAVE_THREAD_SAFETY_ATTRIBUTES
#endif
```

No version check: the attributes used here have all been in clang since 3.x, well below anything this project supports.

@dmah42 — on the confusion: yes, Bazel turns on the *warning* without defining the *macro*. Its default C++ toolchain adds `-Wthread-safety` whenever the compiler accepts it ([rules_cc](https://github.com/bazelbuild/rules_cc/blob/edc6a8d7227d76d798ac4041efc8fdacee9d5fd7/cc/private/toolchain/unix_cc_configure.bzl#L663)), and nothing tells it about `HAVE_THREAD_SAFETY_ATTRIBUTES`. So the wrappers get analysed with their annotations erased, which is only an error where the standard library annotates `std::mutex` itself — libc++, hence macOS and not Linux.

On setting it from `BUILD.bazel` instead: that would fix the Bazel build, but the same `BUILD.bazel` also has to serve gcc, which rejects the attributes, so it would need the compiler condition anyway. Since it is a property of the compiler rather than of the build, keeping the decision next to the attributes means every build system gets it right without knowing about the macro. The comment now says that instead of what it said before.

Verified against clang 21 + libc++ (reproduces before, clean after), clang 21 + libstdc++, gcc 15 and MSVC; the library still builds clean under `-Wall -Wextra -Wthread-safety -Werror`.

dmah42:
i would have thought https://bazel.build/docs/configurable-attributes would be the right way to solve it for compiler.

devtejasx:
That works for the Bazel build, and if it is what you want I will send it — `@rules_cc//cc/compiler:clang` gives the `config_setting`, and it would be a `select()` on `local_defines`.

The reason I did not: `-Wthread-safety` is not Bazel-specific. Any consumer can turn it on through any build system, and CMake has the same hole today — `cmake -DCMAKE_CXX_FLAGS=-Wthread-safety` with clang reproduces #2263 with no Bazel involved. Fixing it per build file means fixing it once per build system and keeping each one in step with the compiler check, whereas whether the annotations are valid is a property of the compiler alone. That is also what @LebedevRI suggested above ("Can we just check for clang").

So the trade-off as I see it: one line in the header that every build system inherits, versus a `select()` that fixes the reported case and leaves the others. Happy to switch — or to add the `select()` on top, if you would rather Bazel state it explicitly as well. Which would you prefer?

dmah42 (dismissed):

dmah42:
i'm good with this. @LebedevRI ?

LebedevRI:
I don't quite follow. Why is this being done on the buildsystem level? What about CMake?
I was basically going to stamp the diff that was present during my last review, i.e. https://github.com/google/benchmark/commit/c28327407a7bf4199f6ffb219c1dbdd8e93f4dd4,
modulo nits, i.e. just unconditionally enable it for clang, but do that in a header, not build system...

LebedevRI (changes requested):

devtejasx:
@LebedevRI Fair point, the select() only fixed Bazel. I'd moved it there after dmah42's configurable-attributes suggestion. It's now back to the header check you reviewed:

    #if !defined(HAVE_THREAD_SAFETY_ATTRIBUTES) && defined(__clang__)

in `mutex.h`, with `BUILD.bazel` untouched, so CMake is covered too.

@dmah42 this replaces the version you approved. Are you OK with the header check?

dmah42 (approved):

LebedevRI (approved):

PR #2276, opened 2026-08-01
Exclude setup/teardown from the MemoryManager measurement window (#2149)

devtejasx:
RunMemoryManager() opened the MemoryManager window before allocating the ThreadManager and running the per-benchmark Setup(), and closed it after Teardown(). Everything the library and the fixture allocate around the benchmark was therefore attributed to the benchmark itself, so an exact allocation count was impossible to obtain.

Move Start()/Stop() to bracket only RunInThread(), which is the same region the timers cover. Setup()/Teardown() already run outside the timed region, so this makes the memory window consistent with the time window.

Add memory_manager_ordering_gtest, which registers a MemoryManager that records whether the window is open and asserts the fixture callbacks never observe it as open. It fails on the previous ordering ("Teardown ran inside the MemoryManager Start/Stop window") and passes with this change.

---

responsibility for it.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -462,17 +462,17 @@ void BenchmarkRunner::RunWarmUp() {
 
 MemoryManager::Result BenchmarkRunner::RunMemoryManager(
     IterationCount memory_iterations) {
-  memory_manager->Start();
   std::unique_ptr<internal::ThreadManager> manager;
   manager.reset(new internal::ThreadManager(1));
   b.Setup();
+  memory_manager->Start();
   RunInThread(&b, memory_iterations, 0, manager.get(),
               perf_counters_measurement_ptr,
               /*profiler_manager=*/nullptr);
-  manager.reset();
-  b.Teardown();
   MemoryManager::Result memory_result;
   memory_manager->Stop(memory_result);
+  manager.reset();
+  b.Teardown();
   memory_result.memory_iterations = memory_iterations;
   return memory_result;
 }
--- a/test/memory_manager_ordering_gtest.cc
+++ b/test/memory_manager_ordering_gtest.cc
@@ -1,0 +1,74 @@
+// Setup()/Teardown() must run outside the MemoryManager Start()/Stop()
+// window, as they already do outside the timed region. See #2149.
+
+#include <vector>
+
+#include "benchmark/benchmark.h"
+#include "gtest/gtest.h"
+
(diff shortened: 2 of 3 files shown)


src/benchmark_runner.cc
```diff
@@ -462,17 +462,23 @@ void BenchmarkRunner::RunWarmUp() {
...
     IterationCount memory_iterations) {
-  memory_manager->Start();
+  // Set up the thread manager and run the user's Setup() outside the
+  // measured window. These happen outside the timed region too, so the
+  // library's own ThreadManager allocation and any per-benchmark fixture
+  // allocations must not be attributed to the benchmark's memory usage.
   std::unique_ptr<internal::ThreadManager> manager;
   manager.reset(new internal::ThreadManager(1));
   b.Setup();
+  MemoryManager::Result memory_result;
```
dmah42: why have you moved the declaration of the result?

src/benchmark_runner.cc
```diff
@@ -462,17 +462,23 @@ void BenchmarkRunner::RunWarmUp() {
 
 MemoryManager::Result BenchmarkRunner::RunMemoryManager(
     IterationCount memory_iterations) {
-  memory_manager->Start();
+  // Set up the thread manager and run the user's Setup() outside the
```
dmah42: is this a useful comment or just the AI spewing the decisions its made?

test/memory_manager_ordering_gtest.cc
dmah42, line 1: is this test actually testing something useful? discuss.

devtejasx:
All three addressed; the runner change is now a four-line move.

- **The `memory_result` declaration is back where it was.** I had moved it on the issue author's suggestion, in case `Result` ever allocates. Nothing in the fix needs it, so it was noise.
- **Comment dropped.**
- **On the test.** It registers a `MemoryManager` that flips a flag in `Start()`/`Stop()`, and a benchmark whose `Setup()`/`Teardown()` assert that flag is false. Against main it fails with `Teardown ran inside the MemoryManager Start/Stop window`; with this change it passes. What it pins is the ordering the PR is about: nothing else in the suite observes when the measurement window opens relative to `Setup()`, so moving `Start()` back above `b.Setup()` in a later refactor would go unnoticed. If you would rather not carry it, I'll drop it and leave just the runner change.

dmah42 (approved):

dmah42:
thanks

PR #2277, opened 2026-08-01
Give ProfilerManager access to the State being profiled (#2234)

devtejasx:
Custom profilers that open a named region need the benchmark name, but the ProfilerManager hooks receive no context at all.

Adding an overload of AfterSetupStart()/BeforeTeardownStop() that takes a State is not source compatible: the new overload hides the old one in every existing subclass. Instead, add a protected ProfilerManager::GetState() accessor that returns the State of the run being profiled. Existing subclasses keep compiling untouched.

The accessor returns a pointer, not a reference, because there is no meaningful answer outside the hooks. State sets the pointer immediately before invoking a hook and clears it as soon as the hook returns, so a caller outside AfterSetupStart()/BeforeTeardownStop() observes nullptr rather than a stale State. The test asserts both: the benchmark name is readable inside the hook, and the accessor is null outside it.

---

responsibility for it.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1604,6 +1604,11 @@ If set, the `ProfilerManager::AfterSetupStart` and
 end of a separate benchmark run to allow user code to collect and report
 user-provided profile metrics.
 
+From within either hook, `ProfilerManager::GetState()` returns a pointer to
+the `benchmark::State` of the run being profiled, which gives access to run
+information such as the benchmark name (e.g. for registering named profiling
+regions). Outside of the two hooks it returns `nullptr`.
+
 Output collected from this profiling run must be reported separately.
 
 <a name="using-register-benchmark" />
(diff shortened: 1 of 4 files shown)


devtejasx:
Closed without a comment, so I can't tell what to change. Was it the approach — a protected `GetState()` accessor with the `State` reachable only inside the two hooks — or the feature itself?

If it is the approach, the alternative that avoids both problems raised on the earlier attempts is a differently *named* hook rather than an overload: `AfterSetupStart(const State&)` hid the existing `AfterSetupStart()` in subclasses, which is what made #2249 source-incompatible, but a new name with a default implementation that forwards to the old one does not. Happy to send that instead.

If the feature isn't wanted, that's a fine answer too — #2234 can be closed and I'll stop pushing on it.

dmah42:
there's no motivation for the feature, as far as i can tell. noone is asking for it, the issue doesn't explain _why_ someone would want to do it.

PR #2278, opened 2026-08-01
Fix clang-tidy findings in src/ (#1925)

devtejasx:
Works through part of the collated clang-tidy list, restricted to changes that cannot alter behaviour:

- misc-const-correctness: declare locals const where they are never reassigned, spelling the qualifier first to match the surrounding code.
- readability-uppercase-literal-suffix: uppercase integer literal suffixes.
- readability-simplify-boolean-expr: distribute the negation in IsTruthyFlagValue() instead of negating a disjunction, and drop a manual counting loop in favour of the value it was recomputing.
- misc-use-anonymous-namespace: move the file-local console reporter helpers IgnoreColorPrint() and FormatTime() into an anonymous namespace rather than marking them static.

No functional change. Release and debug builds are clean and the test suite passes.

---

responsibility for it.

### DIFF
--- a/src/benchmark_runner.cc
+++ b/src/benchmark_runner.cc
@@ -105,7 +105,7 @@ BenchmarkReporter::Run CreateRunReport(
   report.repetition_index = repetition_index;
   report.repetitions = repeats;
 
-  if (report.skipped == 0u) {
+  if (report.skipped == 0U) {
     if (b.use_manual_time()) {
       report.real_accumulated_time = results.manual_time_used;
     } else {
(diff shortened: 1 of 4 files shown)


dmah42:
did you consider why these checks aren't enabled?

devtejasx:
I read the question as: the config globs them in, but that is not the same as the project wanting them applied. `.clang-tidy` does list `misc-*` and `readability-*`, and clang-tidy 20 reports six `misc-const-correctness` hits in `complexity.cc` alone — but `WarningsAsErrors` is empty and nothing runs it in CI, so the glob is a starting point for #1925 rather than a mandate.

So I have cut this down to the three findings that are not a matter of taste:

- `0u` to `0U`,
- two negated disjunctions distributed (`RunInThread()` and `IsTruthyFlagValue()`),
- the two file-local console reporter helpers moved from `static` into an anonymous namespace.

The `misc-const-correctness` batch, which was most of the previous diff, is gone. If what is left is still more churn than you want against #1925, say so and I'll close it.

dmah42:
so maybe we should start by running clang-tidy on CI (or even on pre-commit) before we add more fixes.  fwiw, i wouldn't care about u vs U. the anonymous namespace fix is nice.

devtejasx:
Thanks for the feedback. It sounds like agreeing on clang tidy enforcement should come first, so I will close this PR for now. Once the project decides which checks to enforce, I can send focused cleanup patches that match that polic

Issue #1754, opened 2024-02-09
[BUG] With NDEBUG, early-returning without SkipWithError leads to indefinite retries


adrianroos:
**Describe the bug**
With NDEBUG, early-returning without SkipWithError leads to indefinite retries

**System**
Which OS, compiler, and compiler version are you using:
  - OS: ```6.5.13-1rodete2-amd64 (building for Android cuttlefish)```
  - Compiler and version:  clang --version: ```Android (11368308, +pgo, +bolt, +lto, +mlgo, based on r510928) clang version 18.0.0 (https://android.googlesource.com/toolchain/llvm-project 477610d4d0d988e69dbc3fae4fe86bff3f07f2b5)```

**To reproduce**
1. Run https://cs.android.com/android/_/android/platform/hardware/interfaces/+/f850de6732287b3c5539a56751d21a4a1b170f42:vibrator/bench/benchmark.cpp;bpv=1;bpt=0;drc=e29acd8194c6df2fc81613cebdb2dc4507ec4c1b
2. Benchmark prints the context, then hangs

Note how the benchmark functions early return before reaching the measure loop. Also note that Android compiles with NDEBUG set.

Minimal repro case (**compile with `-DNDEBUG=1`**)
```
static void BM_Foobar(benchmark::State& state) {
  return;
}
// Register the function as a benchmark
BENCHMARK(BM_Foobar);
```

**Expected behavior**

Prints an error & skips / aborts, instead of retrying indefinitely.

**Root cause**

https://github.com/google/benchmark/blob/b04cec1bf90c3d8e47739bb3271607a18d8b5106/src/benchmark_runner.cc#L133 does not trigger an abort if NDEBUG is set; instead, the benchmark function gets retried indefinitely. 

Verbose logs:
```c++
-- LOG(2): Running VibratorBench_V1_0/on for 1
-- LOG(2): Ran in 0/0
-- LOG(2): Running VibratorBench_V1_0/on for 1
-- LOG(2): Ran in 0/0
-- LOG(2): Running VibratorBench_V1_0/on for 1
-- LOG(2): Ran in 0/0
-- LOG(2): Running VibratorBench_V1_0/on for 1
-- LOG(2): Ran in 0/0
-- LOG(2): Running VibratorBench_V1_0/on for 1
-- LOG(2): Ran in 0/0
-- LOG(2): Running VibratorBench_V1_0/on for 1
-- LOG(2): Ran in 0/0
[...]

Issue #1504, opened 2022-10-25
[BUG] RangeMultiplier(1) creates infinite loop


tgaldes:
**Describe the bug**
BENCHMARK(myfunction)->RangeMultiplier(1)->Range(1, 2);
creates an infinite loop 

**System**
redhat 8, g++ 8.5

**Expected behavior**
I would expect here that RangeMultiplier(1)->Range(1, 2) yields the same behavior as ->Range(1, 2)
just increment by 1 each time

dominichamon:
do you not get "src/benchmark_register.cc:347: RangeMultiplier: Check `multiplier > 1' failed. "?

tgaldes:
No.

Tyler

dominichamon:
ah i see.. the `CHECK` is removed in non-debug builds.

we don't have a great way to do error propagation in release builds, but the tl;dr for this issue is: it's working as intended that it isn't supported.

PR #2280, opened 2026-08-01
Diagnose a non-growing RangeMultiplier in release builds (#1504)

devtejasx:
RangeMultiplier(1) hangs the binary. The guards that reject it -- the `multiplier > 1` check in RangeMultiplier() and `mult >= 2` in AddRange() -- are BM_CHECKs, which compile away under NDEBUG, so in a release build the multiplier is accepted and AddPowers() then multiplies by one forever looking for a value that reaches `hi`. Debug builds abort with a clear message; release builds spin with no output at all.

Add BM_CHECK_RUNTIME (the name suggested in #1754), which is the existing BM_CHECK without the NDEBUG opt-out, and define BM_CHECK to it in debug builds so there is only one implementation. Use it for the two range multiplier preconditions.

These run once per BENCHMARK() registration, never inside the measured loop, so the concern about paying for checks in the critical path does not apply here. Nothing changes for debug builds, and release builds now abort with

  RangeMultiplier: Check `multiplier > 1' failed.

instead of hanging.

Covered by two new cases in benchmark_gtest that swap in an abort handler and assert both preconditions fire; the RangeMultiplier() one exercises the library as built, i.e. with NDEBUG.

### DIFF
--- a/src/benchmark_register.cc
+++ b/src/benchmark_register.cc
@@ -370,7 +370,9 @@ Benchmark* Benchmark::Teardown(const callback_function& teardown) {
 }
 
 Benchmark* Benchmark::RangeMultiplier(int multiplier) {
-  BM_CHECK(multiplier > 1);
+  // Checked in every build configuration: a multiplier of 1 or less makes the
+  // range generation in AddRange() unable to terminate.
+  BM_CHECK_RUNTIME(multiplier > 1);
   range_multiplier_ = multiplier;
   return this;
 }
--- a/src/benchmark_register.h
+++ b/src/benchmark_register.h
@@ -18,7 +18,9 @@ typename std::vector<T>::iterator AddPowers(std::vector<T>* dst, T lo, T hi,
                                             int mult) {
   BM_CHECK_GE(lo, 0);
   BM_CHECK_GE(hi, lo);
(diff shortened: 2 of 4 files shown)


LebedevRI:
This spam is starting to get rude.

dmah42:
we're not going to accept changes like this without discussion and a demonstration of an understanding of the existing reasons for how the codebase is today.  also failing to acknowledge AGENTS.md shows a contempt for the project that is unhelpful. Given the sheer number of PRs you're opening, @devtejasx , i'm starting to get to the point of finding a way to ban you.

Consider your approach carefully.

devtejasx:
bro I am solving multiple issue and then pull them together . have you not seem time been last PR and these time PR .Its more than week.If you want, I can solve 1 issue in 2 day- if these approach is fine let me know

dmah42:
my issue is not the time. my issue is the lack of discussion or understanding about why things might be the way things are before throwing a PR at us.

just because we _can_ write code doesn't mean we _should_ write code. AI has enabled code changes to not be a bottleneck, and that's great, but writing code was never the most important thing. Understanding and discussing continues to be.

devtejasx:
ok bro, got it from next time we will discuss and then i will open PR . Is these fine for you

PR #2290, opened 2026-08-22
Ignore skipped benchmark when computing geomean

Krzmbrzl:
Skipped benchmarks have their real and CPU times set to zero. If they are part of the data for which the geomean is calculated, these zeros cause the computed geomean to be zero. Hence, a single skipped benchmark will effectively disable the computation of a geomean.

This seems unfortunate as in many cases there are lots of other benchmarks that still ran normally and for which one might still be interested in the overall geomean change.

### DIFF
--- a/tools/gbench/report.py
+++ b/tools/gbench/report.py
@@ -191,6 +191,8 @@ def calculate_geomean(json):
     for benchmark in json["benchmarks"]:
         if "run_type" in benchmark and benchmark["run_type"] == "aggregate":
             continue
+        if "skip_message" in benchmark:
+            continue
         times.append(
             [
                 get_timedelta_field_as_seconds(benchmark, "real_time"),


tools/gbench/report.py
LebedevRI: This should also handle skipped and errored-out benchmarks, i would imagine.

tools/gbench/report.py
```diff
@@ -191,6 +191,8 @@ def calculate_geomean(json):
     for benchmark in json["benchmarks"]:
         if "run_type" in benchmark and benchmark["run_type"] == "aggregate":
             continue
+        if "skip_message" in benchmark:
```
LebedevRI: Does this cover errored-out benchmarks?

LebedevRI (commented):
Nice one.
Could you please add test coverage? (see `tools/compare.py`)

tools/gbench/report.py
Krzmbrzl: It doesn't - I'll add coverage for it as well though

Krzmbrzl:
I guess something could be done about comparing benchmarks where only one version of it was skipped or errored. However, I don't think that a timing comparison in this case is of any use and to report anything else seems like it would require a bit more work. For the time being benchmarks are excluded if they were skipped or they errored in one of the compared runs. That seems reasonable to me.

LebedevRI (dismissed):
Ok, test seems to somewhat cover the behavior. I'll take this.

LebedevRI (approved):

LebedevRI:
@Krzmbrzl thank you!

Issue #2282, opened 2026-08-06
[BUG] GCC optimizes away the write with `benchmark::DoNotOptimize`


SaiCharanMarrivada:
GCC16.1 optimizes away the write when we use `benchmark::DoNotOptimize`.  

## Minimal reproducible example
```cpp
void f() {
    int *a = new int[1 << 16];
    for (int i = 0; i < 1 << 16; i++) {
        benchmark::DoNotOptimize(a[i] = 10); // this is optimized away by the compiler
    }
    delete[] a;
}
```

Compiler explorer(CE) link: https://godbolt.org/z/j538536ss. The CE link contains a possible fix for the issue.

LebedevRI:
@SaiCharanMarrivada could you please file an issue with GCC?
They should just provide a builtin that would do the right thing,
or at worst expand to the correct ASM sequence.

SaiCharanMarrivada:
@LebedevRI Sure I will raise it. What do you think about the fix given in the compiler explorer link?

ramanayakakumara01-ship-it:
void f() {
    int *a = new int[1 << 16];
    for (int i = 0; i < 1 << 16; i++) {
        benchmark::DoNotOptimize(a[i] = 10); // this is optimized away by the compiler
    }
    delete[] a;
}

PR #2291, opened 2026-08-24
Support DoNotOptimize on types with const members (#1997)

### LINKED ISSUE #1997: [BUG] DoNotOptimize on values with const members again

devtejasx:
Fixes #1997.

`DoNotOptimize(Tp&)` and `DoNotOptimize(Tp&&)` bind the argument as an asm
**output** operand. A class with a `const` data member has nothing to store into,
so both compilers reject it:

```
gcc   15.2.0  : error: read-only reference 'value' used as 'asm' output
clang 20.1.8  : error: invalid lvalue in asm output
```

## The change

One trait, `internal::IsValidAsmOutput`, plus four overloads — two for the clang
branch and two for gcc — that route non-writable types to an **input** operand:

```cpp
asm volatile("" : : "m"(value) : "memory");
```

The barrier is not weakened by that. The `"memory"` clobber already forces the
object into memory and forbids the compiler assuming it unchanged; the only thing
given up is the claim that the asm *writes* it, which was never true for these
types anyway.

Notes on the trait:
- `is_move_assignable` is included deliberately. Move-only types (e.g. a struct
  with a `unique_ptr` member) are perfectly valid asm outputs and must not be
  demoted to inputs.
- Arrays are never assignable but *are* valid outputs, so `is_array` is checked
  outright.
- A class with a reference member is also a valid output but is not
  distinguishable here; it is conservatively treated as an input.

## On the size of it

This is deliberately the small version. On the earlier #2254 the objection was
"why is this such a big problem that we need such complex code" — that patch added
8 overloads. This adds **4**, and collapses the fallbacks to a single `m`-input
overload per value category, which is what the reply to that review proposed.

## Verification

Reproducer (a struct with a `const` member, passed as lvalue, const lvalue and
rvalue, alongside `int` / `vector` / `string` / array regression guards):

| | pristine `main` @ 267a111 | with this change |
|---|---|---|
| gcc 15.2.0 | fails, 2 errors | compiles clean |
| clang 20.1.8 | fails, 2 errors | compiles clean |

- Header stays C++11-clean — the README promises C++11 for *use*, C++17 only to
  build, so no `if constexpr` in a public header.
- `objdump -d` shows **byte-identical object code** for `int`, `vector`, `string`,
  array and rvalue uses on both compilers, so existing call sites are unaffected.
- `const` lvalues still route to the existing deprecated const-ref overload and
  still emit the deprecation warning.

---

*AI disclosure, per AGENTS.md:* an AI coding assistant was used while producing
this change — to locate the failing overloads, draft the trait and overloads, and
run the before/after compiler checks above. I reviewed the result, chose the trait
membership (including keeping `is_move_assignable`) and the input-operand
approach, and verified the compiler output myself. I understand the change and can
argue it on review.


LebedevRI:
I still have the same comment i posted in https://github.com/google/benchmark/issues/2282#issuecomment-5208446677

> could you please file an issue with GCC?
> They should just provide a builtin that would do the right thing,
> or at worst expand to the correct ASM sequence.

devtejasx:
Happy to raise it with them — I'd just like to file the right bug, because I think these are two separate problems.

#2282 is a miscompile: the write is elided. That's the one a proper builtin would fix.

This PR is a different failure — both compilers reject the code outright:

```
gcc   15.2.0 : error: read-only reference 'value' used as 'asm' output
clang 20.1.8 : error: invalid lvalue in asm output
```

`DoNotOptimize(Tp&)` binds its argument as an asm *output*, and a class with a `const` data member has nothing to store into. A new GCC builtin wouldn't fix that for any compiler already released, and clang rejects it as well, so it isn't GCC-specific.

Two questions, then:

1. Do you want the report scoped to the elision in #2282, or should the const-member rejection go with it?
2. Is this PR blocked on that report, or can it land as the workaround for currently-shipping compilers while the builtin is pursued upstream?

dmah42:
i think this change looks reasonable. @LebedevRI do you have concerns with the PR as it stands, even if it's a workaround for a bug in the compiler?

LebedevRI (commented):
Question: can these be
```diff
- asm volatile("" : : "m"(value) : "memory");
+ asm volatile("" : : "r,m"(value) : "memory");
```
(perhaps the other way around for gcc, as suggested by other cases)

devtejasx:
@LebedevRI Yes. The fallback now uses "r,m" for clang and "m,r" for gcc, same as the output overloads. A small const-member struct now stays in a register; with plain "m" both compilers spilled it to the stack.

It compiles on gcc 5–15 and clang 3.9–20 at -O0 and -O2 in C++11, including large and non-trivially-copyable types, so gcc didn't need a size split.

include/benchmark/utils.h
```diff
@@ -88,35 +121,54 @@ inline BENCHMARK_ALWAYS_INLINE
...
+    DoNotOptimize(Tp& value) {
+  asm volatile("" : : "m,r"(value) : "memory");
+}
+
+template <class Tp>
+inline BENCHMARK_ALWAYS_INLINE
+    typename std::enable_if<!internal::IsValidAsmOutput<Tp>::value>::type
+    DoNotOptimize(Tp&& value) {
+  asm volatile("" : : "m,r"(value) : "memory");
+}
```
LebedevRI, line 171: When must it be `m,r` instead of `r,m`? For which code and compiler?

include/benchmark/utils.h
```diff
@@ -60,13 +76,30 @@ inline BENCHMARK_ALWAYS_INLINE void DoNotOptimize(Tp& value) {
...
+    DoNotOptimize(Tp& value) {
+  asm volatile("" : : "r,m"(value) : "memory");
+}
+
+template <class Tp>
+inline BENCHMARK_ALWAYS_INLINE
+    typename std::enable_if<!internal::IsValidAsmOutput<Tp>::value>::type
+    DoNotOptimize(Tp&& value) {
+  asm volatile("" : : "r,m"(value) : "memory");
+}
```
LebedevRI, line 102: Do we need both overloads here?

LebedevRI (commented):
Some more thoughts.
https://godbolt.org/z/45co8P814

PR #2293, opened 2026-09-03
Add opt-in cross-thread statistics for multithreaded runs

### LINKED ISSUE #1902: [FR] Report cross-thread statistics for multithreaded runs

Frank20021:
## Summary
- Implements the design posted on #1902: opt-in mean/median/stddev/cv **across threads** of one multithreaded run.
- Default summed (then `Finish`) counter row is unchanged. Extra `RT_Aggregate` rows use a `thread_` prefix so they do not collide with repetition `_mean` / `_median` / `_stddev` / `_cv`.
- Enable with `->ReportThreadStatistics()` or `--benchmark_report_thread_statistics=true`. Per-thread samples are captured in `RunInThread` before `Increment`; `ComputeStats` still only sees iteration runs.

## Test plan
- [x] `thread_statistics_test` (unequal `{1,2,3}` counters, `kAvgThreads`, feature off, `threads:1`)
- [x] `user_counters_threads_test`
- [x] `repetitions_test --benchmark_repetitions=3`

Fixes #1902

AI usage: I used Cursor as a coding assistant to locate the runner/reporter paths and draft this change. I reviewed the capture/`Finish`/`ComputeStats` wiring and the tests myself.


include/benchmark/benchmark_api.h
```diff
@@ -180,6 +181,8 @@ class BENCHMARK_EXPORT Benchmark {
   BigOFunc* complexity_lambda_;
   std::vector<internal::Statistics> statistics_;
   std::vector<int> thread_counts_;
+  bool report_thread_statistics_specified_;
+  bool report_thread_statistics_;
```
dmah42, line 185: do we need both? if the default is false does it matter if it's specified or not?

src/benchmark_register.cc
```diff
@@ -426,6 +428,12 @@ Benchmark* Benchmark::DisplayAggregatesOnly(bool value) {
   return this;
 }
 
+Benchmark* Benchmark::ReportThreadStatistics(bool value) {
+  report_thread_statistics_specified_ = true;
```
dmah42, line 432: how does this interact with other per benchmark settings? do we need to make sure that the user hasn't set incompatible flags?

src/benchmark_runner.cc
```diff
@@ -139,6 +141,106 @@ BenchmarkReporter::Run CreateRunReport(
...
+  std::vector<UserCounters> finished;
+  finished.reserve(n);
+  std::vector<double> real_times;
+  std::vector<double> cpu_times;
+  real_times.reserve(n);
+  cpu_times.reserve(n);
+
+  const IterationCount thread_iters = thread_results.front().iterations;
+
+  for (const auto& tr : thread_results) {
```
dmah42, line 162: this looks like code we have elsewhere. is there any opportunity to share the code instead of duplicating it?

src/benchmark_runner.h
```diff
@@ -87,6 +89,7 @@ class BenchmarkRunner {
   bool warmup_done;
   const int repeats;
   const bool has_explicit_iteration_count;
+  const bool report_thread_statistics;
```
dmah42, line 92: do we need this to be a member?

test/thread_statistics_test.cc
```diff
@@ -0,0 +1,230 @@
+
+#undef NDEBUG
```
dmah42, line 2: why?

dmah42 (changes requested):

PR #2294, opened 2026-09-07
Report the complexity aggregates in the family's time unit

lenamonj:
`ComputeBigO` never set `big_o.time_unit`, so the BigO row kept the `Run()` default of nanoseconds and its coefficient was rendered in ns whatever unit the family declared. It now carries the family's unit, as the RMS row already did.

The new `BM_Complexity_O_N_ms` family declares milliseconds: on main its BigO row reports `"time_unit": "ns"`. The test fails on main and passes here.


src/complexity.cc
```diff
@@ -229,6 +229,8 @@ std::vector<BenchmarkReporter::Run> ComputeBigO(
   big_o.cpu_accumulated_time = result_cpu.coef;
   big_o.report_big_o = true;
   big_o.complexity = result_cpu.complexity;
+  // The coefficient is a time; reporters render it in the family's unit.
```
dmah42: unnecessary comment

test/complexity_test.cc
```diff
@@ -20,13 +20,17 @@ namespace {
 int AddComplexityTest(const std::string& test_name,
                       const std::string& big_o_test_name,
                       const std::string& rms_test_name,
-                      const std::string& big_o, int family_index) {
+                      const std::string& big_o, int family_index,
+                      const std::string& time_unit = "ns",
```
dmah42: why are there defaults here?

dmah42 (changes requested):

lenamonj:
Removed the comment and the defaults; each family passes its unit and CSV pattern explicitly.

LebedevRI (commented):
CSV reporter is deprecated and should not be used.

lenamonj:
Dropped the CSV reporter change and its test pattern; the PR is now the one-line time_unit fix with the JSON check.

test/complexity_test.cc
```diff
@@ -265,7 +267,34 @@ const std::string complexity_capture_name =
 
 ADD_COMPLEXITY_CASES(complexity_capture_name, complexity_capture_name + "_BigO",
                      complexity_capture_name + "_RMS", "N",
-                     /*family_index=*/9);
+                     /*family_index=*/9, "ns");
+// ========================================================================= //
+// ------------------- Testing BigO with a declared unit ------------------- //
```
LebedevRI, line 272:
```suggestion

// ------------------- Testing BigO with a declared unit ------------------- //
```

LebedevRI (commented):
Ah, hmm, when initially reading the change, somehow i missed that it was also 'fixing' JSON.
I thought it was only making CSV do the same as JSON reporter.

After sleeping on this, i do think this makes sense.

I do have a suggestion though: is there a test that breaks
if you intentionally scale the value of the "bigo" field in JSON reporter
by `x1000`? There should be one, we should have a test
to ensure that the reported bigo really does have the correct magnitude
as implied by the unit.

lenamonj:
There is now: scaling both coefficients by 1000 in `json_reporter.cc` fails `complexity_benchmark`, because the millisecond family's `real_coefficient`, about `1.0004e-03` ms per N, has to match `1\.[0-9]+e-03`; at 6169934 all 87 tests passed under that scaling. `cpu_coefficient` stays `%float` because it is measured CPU time that varies from run to run.

LebedevRI (approved):

dmah42 (approved):

PR #2297, opened 2026-09-10
Py tooling: fix Windows terminal color output and handle color flags in compare.py (fixes #642)

### LINKED ISSUE #642: On Windows there is formating error when using compare.py
So I am trying to compare my benchmark but it is really hard because the output contains
 garbage characters. 

python.exe D:\_Dev\Playground\gBenchmarkVulkan/bazel-gbenchmarkvulkan/external/gbenchmark/tools/compare.py filters D:\_Dev\Playground\gBenchmarkVulkan/benchmark.json NoExcept WithExcept
Comparing NoExcept to WithExcept (from D:\_Dev\Playground\gBenchmarkVulkan/benchmark.json)
Benchmark                                                    Time             CPU      Time Old      Time New       CPU Old       CPU New
-----------------------------------------------------------------------------------------------------------------------------------------
[92mvkInstanceCreationUnique[NoExcept vs. WithExcept][0m[97m         +0.0067[0m[97m         +0.0250[0m     107057091     107769236[0m      56818182      58238636

[benchmark.txt](https://github.com/google/benchmark/files/2203784/benchmark.txt)

jdymitarai:
﻿### Summary
This PR fixes issue #642 where `compare.py` produced literal ANSI escape characters (`^[[92m`, `^[[0m`, etc.) on Windows consoles, and ensures that color settings (`--color`, `--no-color`, `--benchmark_color=...`) are properly recognized regardless of argument position.

### Root Cause
1. **Windows Console VT Processing**: On Windows, ANSI escape sequences (`\033[...]`) require `ENABLE_VIRTUAL_TERMINAL_PROCESSING` (`0x0004`) to be set on the console output buffer. Without this flag, `cmd.exe` and standard console hosts print literal escape sequences.
2. **Auto-detection & Redirects**: `compare.py` previously defaulted `color=True` unconditionally, even when stdout was not a TTY (such as redirected to a file or piped to another tool), or when the standard `NO_COLOR` environment variable was set.
3. **Argparse Remainder Flag Leak**: Because subparsers in `compare.py` collect trailing options into `benchmark_options` (`nargs=argparse.REMAINDER`), placing `--no-color` or `--benchmark_color=false` after the subcommand or arguments caused the flag to be captured as benchmark runner options instead of setting `args.color`, while triggering a spurious warning when inputs were JSON files.
4. **Test import issue**: In `tools/gbench/report.py` line 1492, `import util` caused `ModuleNotFoundError` when run via `python -m unittest`.

### Changes
- **Windows VT Processing**: Enable `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on Windows console handles when available, and disable color formatting if VT mode is not supported by the console.
- **Color Auto-Detection**: Auto-detect TTY output (`sys.stdout.isatty()`) and respect `NO_COLOR` standard (https://no-color.org/).
- **Flag Extraction**: Allow both `--color` and `--no-color` on root parser, and extract `--color`, `--no-color`, and `--benchmark_color=false|true|auto` from trailing `benchmark_options`.
- **Suppress False Warning**: Avoid printing "WARNING: passing --benchmark_color=... has no effect since both inputs are JSON" in `check_inputs`.
- **Test Fix & Unit Tests**: Fix `from gbench import util` in `report.py`, and add unit tests in `compare.py` covering color flag parsing, option resolution, and `NO_COLOR` handling.
- Added contributor details to `AUTHORS` and `CONTRIBUTORS`.

### Verification
- All 35 Python unit tests in `tools/` pass:
  ```bash
  python -m unittest discover -s tools -p "*.py"
  ```
- C++ library builds cleanly with CMake on MSVC:
  ```bash
  cmake --build build --config Release --target benchmark
  ```

---
*Disclosure: AI assistance (LLM) was used in formulating this patch and tests.*


tools/compare.py
```diff
@@ -73,10 +146,17 @@ def create_parser():
         "Internally, all the actual runs are still used, e.g. for U test.",
     )
 
-    parser.add_argument(
+    color_group = parser.add_mutually_exclusive_group()
+    color_group.add_argument(
```
dmah42, line 150: isn't the opposite of --no-color just --color? why do we need both flags?

jdymitarai:
Previously, \compare.py\ only had \--no-color\ (with \default=True\), which meant color could only be turned off, and could never be explicitly forced on (e.g., when piping output into \less -R\ or in CI runners where stdout is not detected as an interactive TTY).

We added \--color\ to allow explicitly forcing color output, while retaining \--no-color\ for 100% backwards compatibility with existing user scripts and CLI habits. When neither flag is specified (\default=None\), \compare.py\ auto-detects whether stdout is an interactive TTY and whether VT processing is supported (and honors the standard \NO_COLOR\ environment variable).

dmah42:
> Previously, \compare.py\ only had --no-color\ (with \default=True), which meant color could only be turned off, and could never be explicitly forced on (e.g., when piping output into \less -R\ or in CI runners where stdout is not detected as an interactive TTY).
> 
> We added --color\ to allow explicitly forcing color output, while retaining --no-color\ for 100% backwards compatibility with existing user scripts and CLI habits. When neither flag is specified (\default=None), \compare.py\ auto-detects whether stdout is an interactive TTY and whether VT processing is supported (and honors the standard \NO_COLOR\ environment variable).

that explains what it does, but not why. why do we need to force color on?

LebedevRI:
(This also probably needs LLM disclosure notice.)

jdymitarai:
Thanks for the feedback!

### Why force color on?
When compare.py runs in non-interactive environments—such as CI pipelines (GitHub Actions, GitLab CI, Buildkite) or when piping through pagers/utilities (compare.py ... | less -R or compare.py ... | tee log.txt)—sys.stdout.isatty() evaluates to False. 

Without a flag to explicitly force color on, the output will unconditionally strip colors in those environments, even though modern CI dashboards and pagers like less -R fully support rendering ANSI color codes. Allowing users to pass --color mirrors standard CLI behavior across tooling (e.g., git --color=always, grep --color=always, pytest --color=yes, ls --color=always) so that colored diffs can be preserved in CI logs and terminal pipelines.

---
*Disclosure: AI assistance (LLM) was used in formulating this patch and tests.*

tools/compare.py
```diff
@@ -41,6 +41,8 @@ def check_inputs(in1, in2, flags):
         # When both sides are JSON the only supported flag is
         # --benchmark_filter=
         for flag in util.remove_benchmark_flags("--benchmark_filter=", flags):
+            if flag.startswith("--benchmark_color="):
```
dmah42, line 44: why is this here?

dmah42:
i guess i don't have a particularly strong opinion. it's a lot of code to add for a fairly niche use-case.

LebedevRI:
I have a question. If this change itself does not enable the coloring,
and you have to change the callers (ci etc) to pass extra CLI argument to benefit,
can you not instead change CI etc to globally enable coloring in the terminal?
Why must this be specifically done in this tooling?

I don't really know much about windows terminal,
but i'd guess there is a way to do that with a simple terminal command?

jdymitarai:
Hi @LebedevRI, thanks for asking! Let me clarify both points:

### 1. Does this enable coloring automatically?
**Yes, it does enable coloring automatically out-of-the-box without requiring any extra CLI arguments.**

For any interactive terminal session (Linux, macOS, and Windows):
- \compare.py\ auto-detects that stdout is a TTY (\sys.stdout.isatty() == True\).
- On Windows, the reason issue #642 occurred is that the Windows Console host (\conhost.exe\ / \cmd.exe\) has \ENABLE_VIRTUAL_TERMINAL_PROCESSING\ (\

tools/compare.py
LebedevRI: +1. I do not get why `--color` needs to exist. This is the default already.

LebedevRI (commented):
To make this less contentious, i'd propose
1. Drop `--color`. The default is already always-on with `--no-color` opt-out.
2. Drop `--benchmark_color=` handling.
3. Keep `enable_virtual_terminal_processing()`
4. Adjust `should_use_color()` into a legality check - should auto-disable colors when coloring is not possible.

jdymitarai:
Hi @LebedevRI,

Thank you for the constructive suggestion! I have updated the PR following your 4 points:

1. **Dropped \--color\**: Kept only the original \--no-color\ flag (with \default=True\).
2. **Dropped \--benchmark_color=\ handling**: Removed all internal option interception.
3. **Kept \enable_virtual_terminal_processing()\**: Enables VT mode on the Windows console handle so ANSI sequences render natively on Windows.
4. **Adjusted \should_use_color()\ into a legality check**:
   - Auto-disables color when \NO_COLOR\ is present in \os.environ\.
   - Auto-disables color when \sys.stdout\ is not an interactive TTY (or dumb terminal).
   - On Windows, auto-disables color if \enable_virtual_terminal_processing()\ fails.

The PR diff is now much smaller and focused strictly on Windows VT console support and the legality check. All tests pass.

tools/compare.py
LebedevRI:
```suggestion
        default=should_use_color(),
```

tools/compare.py
```diff
@@ -534,6 +575,31 @@ def test_benchmarksfiltered_with_remainder_after_doubleminus(self):
         self.assertEqual(parsed.filter_contender[0], "e")
         self.assertEqual(parsed.benchmark_options[0], "g")
 
+    def test_benchmarks_no_color_flag(self):
+        parsed = self.parser.parse_args(
+            ["--no-color", "benchmarks", self.testInput0, self.testInput1]
+        )
+        self.assertFalse(parsed.color)
+
+    def test_benchmarks_color_default_true(self):
```
LebedevRI: I think we can't test what the default is.

tools/compare.py
```diff
@@ -534,6 +575,31 @@ def test_benchmarksfiltered_with_remainder_after_doubleminus(self):
...
+    def test_should_use_color_no_color_env(self):
+        orig = os.environ.get("NO_COLOR")
+        try:
+            os.environ["NO_COLOR"] = "1"
+            self.assertFalse(should_use_color())
+        finally:
+            if orig is None:
+                os.environ.pop("NO_COLOR", None)
+            else:
+                os.environ["NO_COLOR"] = orig
```
LebedevRI: This test does not look hermetic.
Can `unittest.mock.patch` be used here?

jdymitarai:
Thanks @LebedevRI for the review!

I have applied all three suggestions in commit \57ffcd9\:
1. Set \default=should_use_color()\ directly on the \--no-color\ flag in \create_parser()\, eliminating redundant post-parse color checking in \main()\.
2. Removed \	est_benchmarks_color_default_true\ to avoid testing an environment-dependent dynamic default.
3. Updated \TestShouldUseColor\ to be hermetic using \unittest.mock.patch\ (testing both \NO_COLOR\ environment variable suppression via \@patch.dict\ and non-interactive TTY via \@patch('sys.stdout.isatty')\).

All 16 unit tests pass.

tools/compare.py
```diff
@@ -534,6 +574,22 @@ def test_benchmarksfiltered_with_remainder_after_doubleminus(self):
...
+
+class TestShouldUseColor(unittest.TestCase):
+    @patch.dict(os.environ, {"NO_COLOR": "1"})
+    def test_should_use_color_no_color_env(self):
+        self.assertFalse(should_use_color())
+
+    @patch.dict(os.environ, {}, clear=True)
+    @patch("sys.stdout.isatty", return_value=False)
+    def test_should_use_color_not_a_tty(self, mock_isatty):
+        self.assertFalse(should_use_color())
```
LebedevRI: I'd think this should test the actual result of parsing, not `should_use_color()`.

tools/gbench/report.py
```diff
@@ -1489,7 +1489,7 @@ def load_result():
         cls.json = load_result()
 
     def test_json_diff_report_pretty_printing(self):
-        import util
+        from gbench import util
```
LebedevRI: Hm, is this to allow running tests from prebuilt package? 
I'm mainly worried that it would no longer be possible to run tests in source checkout.

jdymitarai:
Hi @LebedevRI, thanks again for the thorough feedback!

I've pushed commit \25fb093\ addressing both points:

1. **Test actual parsed result**:
   - Replaced \TestShouldUseColor\ with test methods directly inside \TestParser\ (\	est_benchmarks_color_default_no_color_env\, \	est_benchmarks_color_default_not_a_tty\, \	est_benchmarks_color_default_tty\).
   - These directly verify \parsed.color\ on \parser.parse_args(...)\ with hermetic mocks for \NO_COLOR\, \sys.stdout.isatty()\, and Windows VT console mode.

2. **Source checkout vs package/discovery imports**:
   - In \	ools/gbench/report.py\, updated the import to:
     \\\python
     try:
         from gbench import util
     except ImportError:
         import util
     \\\
   - This ensures running directly from source checkout (\python tools/gbench/report.py\ or \cd tools/gbench && python report.py\) works without error, while also working seamlessly when discovered from the root repository or installed as a package.

tools/gbench/report.py
```diff
@@ -1489,7 +1489,10 @@ def load_result():
         cls.json = load_result()
 
     def test_json_diff_report_pretty_printing(self):
-        import util
+        try:
+            from gbench import util
+        except ImportError:
+            import util
```
LebedevRI: As far as i can tell, from source checkout, any of the following currently works:
```
cd googlebenchmark && python tools/gbench/report.py 
cd googlebenchmark/tools && python gbench/report.py 
cd googlebenchmark/tools/gbench && python report.py 
```
So these tests can already be run.
I feel like this change is unrelated to the rest of the diff,
with which i think i'm okay with. Can this hunk be dropped?

tools/compare.py
```diff
@@ -56,6 +57,45 @@ def check_inputs(in1, in2, flags):
         sys.exit(1)
 
 
+def enable_virtual_terminal_processing():
```
LebedevRI, line 60: I don't suppose python has this functionality in it's standard library?

tools/compare.py
```diff
@@ -56,6 +57,45 @@ def check_inputs(in1, in2, flags):
...
+                return bool(
+                    kernel32.SetConsoleMode(handle, mode.value | 0x0004)
+                )
+            return False
+        except Exception:
+            return False
+    return True
+
+
+def should_use_color():
```
LebedevRI, line 84: I don't suppose python has this functionality in it's standard library?

LebedevRI (commented):
Now this looks reasonable.

tools/gbench/report.py
jdymitarai: Dropped the 
eport.py hunk in 16efad6 so this PR stays strictly focused on compare.py.

tools/compare.py
jdymitarai: Python's standard library does not provide a high-level API (such as os.enable_vt()) for toggling Windows console modes. The canonical, zero-external-dependency way to do it in pure Python is via Python's built-in ctypes standard library module calling Win32 GetConsoleMode/SetConsoleMode with ENABLE_VIRTUAL_TERMINAL_PROCESSING (0x0004). (Some packages rely on os.system('') which had an undocumented side effect of enabling VT in MSVCRT on Windows 10, but ctypes is the clean, explicit, and direct standard library approach without spawning a subshell).

LebedevRI (approved):

LebedevRI:
@jdymitarai thank you.

PR #2302, opened 2026-09-21
Report the CPU affinity mask size in the run context (#1812)

### LINKED ISSUE #1812: [BUG] Core Affinity doesn't seem to work (Or Reporter wrong?)
**Describe the bug**
I would like to run the benchmark on a particular CPU core. The [docs]( says:

```
2. Set the benchmark program's task affinity to a fixed cpu.  For example:
   ```sh
   taskset -c 0 ./mybenchmark
```
However, when I run the `basic_test` app, I see the following:
```
build git:(main) taskset -c 0 ./test/basic_test
2024-07-11T11:29:31+02:00
Running ./test/basic_test
Run on (24 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x12)
  L1 Instruction 32 KiB (x12)
  L2 Unified 1280 KiB (x12)
  L3 Unified 25600 KiB (x1)
Load Average: 5.07, 4.92, 2.97
-----------------------------------------------------------------------------------------
Benchmark                                               Time             CPU   Iterations
-----------------------------------------------------------------------------------------
BM_empty                                            0.327 ns        0.325 ns   2148747493
BM_empty/threads:24                                 0.296 ns        0.328 ns   2130524928
...
```
 I think it might be reporting here that is wrong. Looking at `top`, I can verify that only core 0 is used. The code in `reporter`[https://github.com/google/benchmark/blob/main/src/reporter.cc#L49C29-L49C37) seems to use a static number of cores.  

**System**
  - OS: ubuntu 22.04 
  - Compiler and version: gcc 11.4

**To reproduce**
1. sync to commit : `ea71a14891474943fc1f34d359f9e0e82476ffe1`
2. cmake: ` cmake -D BENCHMARK_DOWNLOAD_DEPENDENCIES=1 -S . -B build`
3. make: `cmake --build build/ -j20`
4. `taskset -c 0 ./build/test/basic_test `

**Expected behavior**
I would expect the test to run only on core 0 and the output of the test be:
`Run on (1 X 4700 MHz CPU s)` instead of `Run on (24 X 4700 MHz CPU s)`

renvins:
`Run on (24 X 4700 MHz CPU s)` reports the machine CPU's count, so a benchmark restricted with `taskset -c 0` gives no indication that the restriction took effect.

So as discussed in the related issue, we decided to add a new field to `CPUInfo`, `num_cpus_in_affinity_mask` (0 when it cannot be determined, e.g. macOS), using the same path used in `ThreadAffinityGuard`.

Now the console reporter appends `, N in affinity mask` only when N < num_cpus, so the output is unchanged when a restriction is not applied. On the other hand, the JSON reporter emits the field as well, whenever it is known, next to `num_cpus`.

I ran all the tests and everything passes as expected.

Fixes #1812


renvins:
@dmah42 tagging for visibility :)

